PR #31, opened 2016-06-23
Very early / simple module system

SimonJF:
This pull request implements a very simple module system for single files. As an example, it is now possible to write:

```
module A {
  var a = 0;
  var b = 0;

  module B {
    fun foo() {
      print("World!")
    }
    var a = 0;
  }

  fun foo() {
    print("Hello!");
    # Here, we've got A.B.foo in scope, but B.foo isn't getting renamed
    # to A.B.foo.
    B:::foo()
  }

}

fun bar() { print("Worldddd") }

open A
foo()
```

Which will print 

```
[simon@guadeloupe scraps]$ links moduletest.links
Hello!
World!
() : ()
```

This is all implemented using a desugaring pass "desugarModules", which eliminates Module and Open.


SimonJF:
Found a bug, closing this for now.

Issue #58, opened 2016-09-02
Add missing modules test file


jgbm:

SimonJF:
@dhil which is the file that was missing? modules.tests and the modules directory seem to be added. I checked out a clean version of sessions, and (modules) tests passed all fine:

```
[simon@guadeloupe links]$ ./run-tests 
tests/annotations.tests
tests/bindings.tests
1 failure
tests/booleans.tests
tests/collections.tests
1 failure
tests/conditionals.tests
tests/continuations.tests
1 failure
tests/database.tests
tests/functions.tests
tests/lexing.tests
tests/library.tests
tests/messages.tests
tests/modules.tests
tests/numbers.tests
tests/patterns.tests
3 failures
tests/polymorphism.tests
tests/records.tests
tests/regex.tests
tests/sessions.tests
tests/strings.tests
tests/tuples.tests
tests/variants.tests
tests/xml.tests
6 total failures
307 total successes
```

dhil:
It would appear to be the test "module chasing":
```
$ ./test-harness tests/modules.tests 
 SUCCESS: Modules hide inner definitions
 SUCCESS: Basic qualified binding resolution
 SUCCESS: Inner module qualified binding resolution
 SUCCESS: Open allows basic unqualified binding access
 SUCCESS: Module nesting allows partially-qualified names to be used for resolution
 SUCCESS: Open allows partially-qualified names to be used for resolution
 SUCCESS: Open still allows fully-qualified names to be used
 SUCCESS: Opened module does not shadow bindings after opening
 SUCCESS: Opened module shadows previous bindings after opening
 SUCCESS: Cyclic dependencies outlawed
Traceback (most recent call last):
  File "./test-harness", line 100, in <module>
    main()
  File "./test-harness", line 96, in main
    evaluate(name, code, **opts)
  File "./test-harness", line 64, in evaluate
    proc = Popen((code, extern_args), stdout=PIPE, stderr=PIPE, env=env)
  File "/usr/lib/python2.7/subprocess.py", line 711, in __init__
    errread, errwrite)
  File "/usr/lib/python2.7/subprocess.py", line 1343, in _execute_child
    raise child_exception
OSError: [Errno 2] No such file or directory
```

fehrenbach:
I still get this when running `./test-harness tests/modules.tests`
```
!FAILURE: Module chasing
  error: Module chasing
      (Unexpected return code: expected `0'; got `1')
  error: Module chasing
[...]

Issue #25, opened 2016-01-28
*** Error: Query.Eval.DbEvaluationError("Error projecting from record")


fehrenbach:
I have some code that triggers this (in sessions), but used to work (in cb0e214).
I tried bisecting and I think things start breaking with the closure conversion project at around 74cb2f4.

The exact error message keeps changing (bad pattern matches, bad assertions).
There are a couple of bug fix commits 13d7ffc, 59d863f, that mention queries, but don't fix this one.

There is probably a smaller test case than this, but this is what I got. Expected result `[(name="b"), (name="d")] : [(name:String)]`.

``` sql
create table xml (
  id int primary key,
  parent int,
  name text,
  pre int,
  post int
  );

insert into xml (id, parent, name, pre, post) values
  (0, -1, '#doc', 0, 13),
  (1, 0, 'a', 1, 12),
  (2, 1, 'b', 2, 5),
  (3, 2, 'c', 3, 4),
  (4, 1, 'd', 6, 11),
  (5, 4, 'e', 7, 8),
  (6, 4, 'f', 9, 10);
```

```
var db = database "stefan";

var xml = table ("xml") with
    (id : Int,
     parent : Int,
     name : String,
     pre : Int,
     post : Int) from db;

typename Axis = [| Self
                 | Child
                 | Descendant
                 | DescendantOrSelf
                 | Following
                 | FollowingSibling
                 | Rev:Axis
                 |];

typename Path = [| Axis:Axis
                 | Seq:(Path, Path)
                 | Name:String
                 | Filter:Path
                 |];

typename Node = (id:Int,name:String,parent:Int,post:Int,pre:Int);

fun axis(ax) {
  switch (ax) {
    case Self -> fun (s, t) { s.id == t.id }
    case Child -> fun (s, t) { s.id == t.parent }
    case Descendant -> fun (s, t) { s.pre < t.pre && t.post < s.post }
    case DescendantOrSelf -> fun (s, t) { s.pre <= t.pre && t.post <= s.post }
    case Following -> fun (s, t) { s.post < t.pre }
    case FollowingSibling -> fun (s, t) { s.post < t.pre && s.parent == t.parent }
    case Rev(ax) -> var rev = axis(ax); fun (s, t) { rev(t, s) }
  }
}

sig path : (Path) ~> (Node, Node) -> Bool
fun path(p: Path) {
  switch (p) {
    case Seq(p, q) ->
      var p = path(p);
      var q = path(q);
      fun (s, u) {
[...]

PR #59, opened 2016-09-02
Build and test on Travis

fehrenbach:


fehrenbach:
@SimonJF could you have a look at this? I think I screwed something up with the merge in test-harness. In the build log (https://travis-ci.org/links-lang/links/builds/157149352) it says
```
 SUCCESS: Cyclic dependencies outlawed
Traceback (most recent call last):
  File "./test-harness", line 109, in <module>
    main()
  File "./test-harness", line 101, in main
    evaluate(name, code, **opts)
  File "./test-harness", line 64, in evaluate
    proc = Popen((code, extern_args), stdout=PIPE, stderr=PIPE, env=env)
  File "/opt/python/2.7.10/lib/python2.7/subprocess.py", line 710, in __init__
    errread, errwrite)
  File "/opt/python/2.7.10/lib/python2.7/subprocess.py", line 1335, in _execute_child
    raise child_exception
OSError: [Errno 2] No such file or directory
```
Any idea of why that would happen, and what should have happened instead?

fehrenbach:
Also, there's a test for #25, so the build will be broken until we (=Sam) fix it.

SimonJF:
@fehrenbach, seems that this is a result of #58. It should be fixed in ba1bf64ada441071c174b9f0b54c574713e5b2e9 if you get time to check (I haven't merged yet).

Issue #63, opened 2016-09-15
Introducing git and code conventions


jstolarek:
I would like to suggest introducing some working conventions for Links project:
- **Clean source code from trailing whitespaces.** These have no semantic meaning for humans or compilers but have meaning for source control systems like git. I would be radical here and suggest that once all major branches are merged and the Gorgie release is made we should just remove all trailing whitespace in one go. Once we've done that we could install a git hook that prevents pushing trailing whitespaces to the repository to make sure they are not re-introduced.
  
  If you want Emacs to highlight trailing whitespaces add these two lines to your `.emacs` file:

```
(setq-default show-trailing-whitespace t)
(setq-default indicate-empty-lines t)
```
- **Stop using merge commits.** Looking at the git history I see that the current convention is to use merge commits. I think this is a Bad Thing as it totally obscures commit history. Firstly, I don't think it is really possible to follow a history that looks like this:

![1](https://cloud.githubusercontent.com/assets/1597921/18555816/23dc67f6-7b61-11e6-8c75-e3d92b2b7157.png)

  Secondly, it is not possible to figure out what commits do. Take for example commit 5786aa191b30478254f017be8629c076c2ba83b1. It introduces a lot of changes in the source code but the commit message gives me (or any of you in six months) no clue as to what these changes are about. It only mentions that some branch was merged but that's not really useful - merged branches usually get deleted and following what work was done on a deleted branch is non trivial (at least to me). So I would propose to stop using merge commits and start using `git rebase` + `git merge --ff-only`.
# TODO
- [ ] set "rebase merging" strategy as the default. This has to be done in repository settings to which I do not have access. In this way "rebase merging" will be the only way to merge pull requests. To change this go to Settings -> Options -> Merge button and untick "Allow merge commits" and "Allow rebase merging".

jamescheney:
These both sound like reasonable things to me.  I think we have generally been using whatever the default git/github behavior is out of ignoramce (at least on my part) - using git well seems to require more conscious effort than svn.  I think the main thing needed is to make instructions explaining how to do the right thing (not just mentioning the correct commands that someone who knows what they are doing would use), and stick them on the wiki somewhere.

jstolarek:
>  I think the main thing needed is to make instructions explaining how to do the right thing

Certainly. I can prepare such instructions. I just wanted to be sure that people would be willing to follow them.

fehrenbach:
[...]

PR #69, opened 2016-09-27
Sessions+shredding

fehrenbach:
This adds back shredding (behind a flag).

### DIFF
--- a/basicsettings.ml
+++ b/basicsettings.ml
@@ -91,9 +91,6 @@ let optimise = Settings.add_bool("optimise", false, `User)
 (* Compile & cache whole program, closures, and HTML *)
 let cache_whole_program = Settings.add_bool("cache_whole_program", false, `User)
 
-(* Inlining with-clauses *)
-let inline_with = Settings.add_bool("inline_with", true, `User)
-
 (* Use keys in shredding *)
 let use_keys_in_shredding = Settings.add_bool("use_keys_in_shredding", true, `User)
 
--- a/basicsettings.ml
+++ b/basicsettings.ml
@@ -107,11 +107,5 @@ let fast_stitch = Settings.add_bool("fast_stitch", true, `User)
 (* experimental: faster version of execute_select *)
 let fast_execute_select = Settings.add_bool("fast_execute_select", true, `User)
 
-(* experimental: alternative, two-pass stitching code *)
-let deforest_stitching = Settings.add_bool("deforest_stitching",true,`User)
-
-(* Generic flag for A/B testing *)
-let feature_test = Settings.add_bool("feature_test",false,`User)
-
 (* Paths to look for .links files in chasing pass *)
 let links_file_paths = Settings.add_string("links_file_paths", "", `User)
--- a/basicsettings.ml
+++ b/basicsettings.ml
@@ -94,11 +94,5 @@ let cache_whole_program = Settings.add_bool("cache_whole_program", false, `User)
 (* Use keys in shredding *)
 let use_keys_in_shredding = Settings.add_bool("use_keys_in_shredding", true, `User)
 
-(* experimental: faster stitching code *)
-let fast_stitch = Settings.add_bool("fast_stitch", true, `User)
-
-(* experimental: faster version of execute_select *)
-let fast_execute_select = Settings.add_bool("fast_execute_select", true, `User)
-
 (* Paths to look for .links files in chasing pass *)
 let links_file_paths = Settings.add_string("links_file_paths", "", `User)
--- a/basicsettings.ml
+++ b/basicsettings.ml
@@ -37,21 +37,7 @@ let welcome_note = Settings.add_string ("welcome_note",
  | |   | | , \\| |   /  \\  \\\n\
  | |___| | |\\ \\ | |\\ \\ _\\  \\\n\
  |_____|_|_| \\__|_| \\_|____/\n\
-Welcome to Links with session types", `System)
-
-(* (* Alternative banner designs *)
-"  _     _ __    _ _  __  ___
- / |   | |  \\  |   |/ / / ._\\
- | |   | | , \\ | |   /  \\  \\
- | |___| | |\\ \\  | |\\ \\__\\  \\
- |_____._._| \\___|_| \\______/",
-"                                         _
-  _     _ __    _ _  _ ___        <>    |_>
- / |   | |  \\  | | |'// ._\\      //     |  
- | |   | | , \\ | |  / \\  \\      //      |
- | |___| | |\\ \\  | | \\_\\  \\  ,-//      _|_
- |_____._._| \\___|_|\\_____/  \\_/      <_|_>"
-*)
+Welcome to Links with session types and shredding", `System)
 
 (** Allow impure top-level definitions *)
 let allow_impure_defs = Settings.add_bool("allow_impure_defs", false, `User)
@@ -67,6 +53,12 @@ struct
 
   let hide_database_info = Settings.add_bool("js_hide_database_info", true, `System)
 end
+
+module Shredding = struct
+  let relax_query_type_constraint = Settings.add_bool("relax_query_type_constraint", false, `User)
+  let shredding = Settings.add_bool("shredding", false, `User)
+end
+
 
 (** Caveat: don't [Open basicsettings] because the above module
    conflicts with the Js module from js.ml*)
@@ -98,6 +90,28 @@ let optimise = Settings.add_bool("optimise", false, `User)
 
 (* Compile & cache whole program, closures, and HTML *)
 let cache_whole_program = Settings.add_bool("cache_whole_program", false, `User)
+
+(* Inlining with-clauses *)
+let inline_with = Settings.add_bool("inline_with", true, `User)
+
+(* Use keys in shredding *)
+let use_keys_in_shredding = Settings.add_bool("use_keys_in_shredding", true, `User)
+
+
+(* experimental: faster unflattening code *)
+let fast_unflatten = Settings.add_bool("fast_unflatten", true, `User)
+
+(* experimental: faster stitching code *)
+let fast_stitch = Settings.add_bool("fast_stitch", true, `User)
+
+(* experimental: faster version of execute_select *)
+let fast_execute_select = Settings.add_bool("fast_execute_select", true, `User)
+
+(* experimental: alternative, two-pass stitching code *)
+let deforest_stitching = Settings.add_bool("deforest_stitching",true,`User)
+
+(* Generic flag for A/B testing *)
+let feature_test = Settings.add_bool("feature_test",false,`User)
 
 (* Paths to look for .links files in chasing pass *)
 let links_file_paths = Settings.add_string("links_file_paths", "", `User)
--- a/basicsettings.ml
+++ b/basicsettings.ml
@@ -94,10 +94,6 @@ let cache_whole_program = Settings.add_bool("cache_whole_program", false, `User)
 (* Use keys in shredding *)
 let use_keys_in_shredding = Settings.add_bool("use_keys_in_shredding", true, `User)
 
-
-(* experimental: faster unflattening code *)
-let fast_unflatten = Settings.add_bool("fast_unflatten", true, `User)
-
 (* experimental: faster stitching code *)
 let fast_stitch = Settings.add_bool("fast_stitch", true, `User)
 
--- a/database.ml
+++ b/database.ml
@@ -104,69 +104,6 @@ let result_signature field_types result =
     in build rs []
 ;;
 
-(*experimental *)
- let _build_record_array (rs: (string * (Types.datatype * int)) list) (row:string array) = 
-    let rec build rs l = 
-      match rs with
-      | [] -> l
-      | (name,(t,i))::rs' -> 
-	  build rs' ((name,value_of_db_string (Array.get row i) t)::l)
-      | _ -> assert false
-    in build rs []
-;;
-
-
-
-(*
-Old version, retained in case of regression *)
-(*
-
-let execute_select_old
-    (field_types:(string * Types.datatype) list) (query:string) (db : database)
-    : Value.t =
-
- (* BUG: Lists can be too big for List.map; need to be careful about recursion *)
-
-  let result = db#exec query in
-
-    match result#status with
-      | `QueryError msg -> 
-        raise(Runtime_error("An error occurred executing the query " ^ query ^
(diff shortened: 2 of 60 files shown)


fehrenbach:
I don't understand why github thinks this can't be rebased. I rebased parts of another branch on top of sessions and cherry picked a couple other commits and then manually fixed things that broke. It should just be a fast-forward merge.

When clicking the "Rebase and merge" button, does github want to do the rebase again?

Anyways, since #63 is not the law yet, can we just merge this? I don't want to do it again.

jstolarek:
Can it be the case that github wants to merge it with the default branch (read: master, not sessions)?

> Anyways, since #63 is not the law yet, can we just merge this? I don't want to do it again.
:( *If* we want to stop using merge commits we have to start doing it.

But wait a second. If you have done rebase locally and all is well, and at the same time github is doing something silly, why not just ff-merge locally and push changes? We can only rely on github to review the patches and then push merges manually.

jstolarek:
By the way, I guess this is the point where I'd like to discuss commit squashing :-) Do we need all these commits? From a quick glance it seems that at least some of them are intermediate steps that are only partially working. My view is that commits should be self-contained entities, each providing at least one functionality (ideally: exactly one self-contained functionality). I am aware that cleaning commit history before merging is extra work and might be perceived as a burden. Again, I claim this time investment is something that pays off in the future when we need to inspect history for things like finding bugs.

I wonder how others feel about this. Perhaps we should discuss this at a standup meeting on Friday?

basicsettings.ml
```diff
@@ -37,21 +37,7 @@ let welcome_note = Settings.add_string ("welcome_note",
...
- | |___| | |\\ \\  | |\\ \\__\\  \\
- |_____._._| \\___|_| \\______/",
-"                                         _
-  _     _ __    _ _  _ ___        <>    |_>
- / |   | |  \\  | | |'// ._\\      //     |  
- | |   | | , \\ | |  / \\  \\      //      |
- | |___| | |\\ \\  | | \\_\\  \\  ,-//      _|_
- |_____._._| \\___|_|\\_____/  \\_/      <_|_>"
-*)
+Welcome to Links with session types and shredding", `System)
```
jstolarek: I just had a thought that perhaps instead of modifying Links' welcome message we should have a `--version` switch that displays information about features of the current build.

fehrenbach:
> :( *If* we want to stop using merge commits we have to start doing it at some point...

With respect, I'm not sure *we* want that to the same extent as *you*. :)

As for commit squashing: Most of these commits are by other people than me (Sam and James), cherry-picked into sessions (some time ago). I did not do any of the actual implementation of shredding. I'm not comfortable rewriting them all into one. Even if they were okay with it, I'm not going to try to isolate patches with "exactly one self-contained functionality" in other people's years old code. This was more than enough work already.

There are some of my own WIP commits in there. My only excuse for them is that I'm lazy (and don't quite understand why some of them show up here at all, when they shouldn't (and no, this is not a PR against master)) and that *I* don't believe that the time investment to go through them and rewrite is worth it in the end. We don't look at history all that often.

> I wonder how others feel about this. Perhaps we should discuss this at a standup meeting on Friday?

I feel like the next Links meeting would be more appropriate. I'm afraid the discussion might take a while.

fehrenbach:
> github is doing something silly

Yeah. It fast forwards perfectly fine locally and it's not a PR against master.

jstolarek:
> With respect, I'm not sure we want that to the same extent as you. :)

Yes, I realize that I am the one pushing for changes others might not want or consider worthwhile bothering. My opinion is that these changes would have on overall beneficial effect on the project at the expense of a bit of extra work. But if others don't agree then I am not going to force anyone and will simply drop the matter. The point is that working conventions only make sense if all people working on the code adhere to them.

Regarding squashing: frankly speaking I am also having difficulties understanding the commit history of this branch. Have you merged sessions branch into yours and then rebased on top of sessions? Frankly speaking I don't know git well enough to understand what exactly happens in such a situation.

jamescheney:
I think it would be good to develop.  However, I don't think it will be practical to apply these to previous branches/developments (going back in some cases several years), particularly since one of the conventions I'd like to establish is that we merge "experimental" branches with useful code in them as soon as possible, rather than allowing them to live a long (long long) time after active development / research on them has finished.  

I think the main thing to check regarding merging shredding+sessions into sessions is that any problematic behavior is confined to the shredding code (which is disabled by default).

In this case (and in other cases of long-lived branches) would "squash and merge" be so bad?

basicsettings.ml
```diff
@@ -99,5 +91,27 @@ let optimise = Settings.add_bool("optimise", false, `User)
 (* Compile & cache whole program, closures, and HTML *)
 let cache_whole_program = Settings.add_bool("cache_whole_program", false, `User)
```
jamescheney: Many of these flags were only useful in experimental evaluation/performance profiling, it may make sense to remove them and specialize the underlying shredding code with sensible default values.

database.ml
```diff
@@ -65,66 +65,197 @@ let execute_insert_returning (table_name, field_names, vss, returning) db =
...
+let execute_select_result
+    (field_types:(string * Types.datatype) list) (query:string) (db: database)  =
+  let result = (db#exec query) in
+    (match result#status with
+       | `QueryOk -> 
+           result,
+	   result_signature field_types result
+       | `QueryError msg -> raise (Runtime_error ("An error occurred executing the query " ^ query ^ ": " ^ msg)))
+
+(* Experimental code to build result using getvalue *)
```
jamescheney: This appears to not be used anywhere now.

database.ml
```diff
@@ -65,66 +65,197 @@ let execute_insert_returning (table_name, field_names, vss, returning) db =
...
+  in do_map 0 []  
+    ;;
+    
+let build_result_fun ((result:Value.dbvalue),rs) = 
+  `List (result#map (fun row ->
+                     `Record (build_record rs row))
+	   )
+    ;;
+
+(* experimental *)
```
jamescheney: This appears not to be used anymore.

database.ml
```diff
@@ -65,66 +65,197 @@ let execute_insert_returning (table_name, field_names, vss, returning) db =
...
+      match rs with
+      | [] -> l
+      | (name,(t,i))::rs' -> 
+	  build rs' ((name,value_of_db_string (row i) t)::l)
+      | _ -> assert false
+    in build rs []
+;;
+
+(*experimental *)
+ let build_record_array (rs: (string * (Types.datatype * int)) list) (row:string array) =
```
jamescheney: This appears not to be used anywhere anymore.

database.ml
```diff
@@ -65,66 +65,197 @@ let execute_insert_returning (table_name, field_names, vss, returning) db =
...
+      | [] -> l
+      | (name,(t,i))::rs' -> 
+	  build rs' ((name,value_of_db_string (Array.get row i) t)::l)
+      | _ -> assert false
+    in build rs []
+;;
+
+
+
+(*
```
jamescheney: We can probably get rid of this now.

database.ml
```diff
@@ -65,66 +65,197 @@ let execute_insert_returning (table_name, field_names, vss, returning) db =
...
+    : Value.t =
+  let result,rs = execute_select_result field_types query db in 
+  build_result (result,rs)
+;;
+    
+
+let execute_select  
+  (field_types:(string * Types.datatype) list) (query:string) (db : database)
+    : Value.t =
+  if Settings.get_value Basicsettings.fast_execute_select
```
jamescheney: We can probably hardwire this to the "true" branch, since there is nothing shredding-specific about this

fehrenbach:
Okay children, gather 'round, it is story time. Once upon a time there was a branch called shredding. It branched off of then master in August 2012.

Back then there was another branch called sand. Sand is a place--the home of the computer science department in Tübingen. The sand branch was the home of Alexander Ulrich's work to use the pathfinder query optimizer to compile queries in Links, and thereby support nested queries.

Starting August 2012 Sam Lindley implemented query shredding in Links. Shredding was an alternative approach to the work in the sand branch, by then already mostly untouched for a year. Shredding allows nested queries, but not aggregations, but does not rely on pathfinder.

Another year passes. Enter James Cheney.

James writes some more benchmarks for both shredding and sand and fixes some performance problems in the shredding branch.

In the meantime, master has moved on to follow the research interests of the ABCD project. There is some work on session types. The sessions branch is created.

The year is 2014. Enter Stefan Fehrenbach.

Pathfinder and the sand branch, but not Sand the place in Tübingen, have long since been abandoned. The only way to install pathfinder at the point is to install database-supported Haskell from Hackage, find the foreign dependencies and extract the pathfinder code from there. At that point it does not even build with recent versions of libXML2.

Database-supported Haskell moves away from pathfinder to a new approach called query flattening.

I decide to revive the shredding branch, because I need records in query results for where-provenance and I thought it would be less effort than implementing it myself, even though I do not yet need nested results. I dig through the history to find the start of the shredding implementation. As said before, it branched off of master in 2012. Even master has since become a bit stale, with almost all of the work happening in the sessions branch.

Merging seemed to be to difficult. Thus I started a new branch from sessions, cherry-picking commits from shredding. I managed to compile and run examples and benchmarks.

I used the revived shredding branch as the basis for work on where- and why-provenance. 

My revived shredding branch also accumulated all sorts of minor features and also tried to follow the sessions branch closely. This was a somewhat misguided attempt to convince Sam to just use my branch as the new master. (Make merging easy by following development and have cool new features as the incentive.) This branch always lived in my fork of the Links repo and is the default branch there (not the ancient master!, or sessions).

The year is 2016. Enter Jan Stolarek.

Pace of Links development picked up. We are actually considering a release.

I have given up on the misguided notion of my shredding branch becoming the new master. Over the last couple of months I have slowly been moving features over to sessions (new and improved emacs mode, ignoring tests, database tests, travis, ...). These things were fairly minor, written by me, and easily extracted.

The bulk of remaining differences are due to shredding. The merge was a lot of work. It is ancient code that was not written by me. Both me and the Skye project might need nested queries. I want to get it merged, not let it rot again.

Since the new doctrine is no merge commits, I rebased parts of my shredding branch (the one that was sessions with shredding cherry-picked on top) on top of current sessions plus a bit of other stuff. I think there's a merge commit in there somewhere, because I failed at git.

This is the story of this pull request. Some of the history is second hand, corrections welcome. The morale of the story is: don't let your topic branches rot. Especially when the topic branch in question is already the merge of a previously abandoned topic branch.

query.ml
```diff
@@ -29,10 +29,15 @@ struct
   type query = comprehension list
 end
```
jamescheney: It would be desirable to split query.ml into two modules, one that does shredding and the other one the standard/traditional one...

fehrenbach:
> In this case (and in other cases of long-lived branches) would "squash and merge" be so bad?

The shredding code itself, that is Sam's code, has a number of different commits that show some structure of the implementation.
James's changes are mainly performance related, as far as I can tell.
Then there's some code that I touched during the merge.

If we squash it we lose all of this information.
In an ideal world, we probably would rewrite into four patchsets (original impl (Sam), perf work (James), introduced by merge (me), later stuff building on top (also me)), but I don't think it's that simple.

For other long-lived branches it might be easier. Especially if they have a single author. Then git blame will still tell you who to bug about it.

As you said, shredding is behind a flag, and all (automatic) tests pass (except the one that does not work in sessions either).

jamescheney:
Thanks, Stefan.  Yes, this is definitely largely My Fault.  I'm looking over the pull request to add code review comments, but these are probably mostly things that I should do (perhaps before we merge).  In particular, I'm not entirely comfortable with the idea of replacing the existing query.ml with the shredding one, because I don't think it's been tested with database drivers other than postgresql (even in the flat case).  I think it would be better to create a separate module for that stuff.  There is also some extra-experimental code from Sam in there about giving queries list semantics.

I also wonder if you were using the same set of flags to run the where/lineage queries as I was.  Many of them were introduced solely to experiment with various things that seemed like they might be optimizations, to see if they really helped or not.  These flags are off by default, totally undocumented, of course, and not mentioned at all in the shredding paper...  so, on the plus side, enabling them may be an easy way of speeding up the shredded queries from the PPDP paper.  (The links config file I used for the experiments is:  

se_cache=false
debug=true
database_driver=postgresql
database_args=...

inline_with=true
use_keys_in_shredding=true
fast_unflatten=true
fast_stitch=true
fast_execute_select=true
deforest_stitching=true
feature_test=true

relax_query_type_constraint=true
shredding=true

)

How to proceed: I know you have other more interesting stuff to do, so how about if I have a look at it over the next few days and try to factor query.ml into the "standard" one and the "shredding" one and rewire so that the standard one is called by default.  That way, cleaning up (or not) the shredding version can be done independently and any problems with that should be confined and not bother anyone else.  If that turns out to be hard, we discuss further, otherwise merge.

PS also we were still using Subversion at the beginning of the story, I think, so about half of the commits were translated over from there.

jamescheney:
Oops, I take it back: all of the settings that should be enabled for shredding to be fast are enabled by default in basicsettings.ml.  Most of them have no effect if shredding is disabled, only fast_execute_select is relevant to flat queries.  Darn, no free performance win!

query.ml
jamescheney: Done in 9598801

jamescheney:
I fixed the problem with factorial.links, the code in the shredding version of evalir.ml was stale and for some reason this problem only showed up when I split the queryshredding code off in its own file.  I'll remove the flags and specialize / eliminate dead code later today.

basicsettings.ml
jamescheney: Done

database.ml
jamescheney: Removed

database.ml
jamescheney: Removed

database.ml
jamescheney: Removed

database.ml
jamescheney: Removed

database.ml
jamescheney: Done

jamescheney:
I think this is ready to merge.  There are some more things I'd like to do to decrease duplication of code between the flat and nested query engines (flagged as TODOs in query.ml and queryshredding.ml) but these will take time I don't have this week.

jamescheney:
As discussed just now, comments on this pull request are solicited until Friday, October 14.  If there are no objections by then, it will be merged.

jamescheney:
There have been no objections, so merging.  I believe we discussed "create a merge commit" as the least-bad option for this long-running branch.

Issue #43, opened 2016-07-28
Type variable names are too localised


slindley:
Type variable names are local to individual types rather than whole error messages. The current implementation threads a renaming environment through the pretty-printer. It may be easier to do something more imperative.
# TODO
- [x] figure out why gripers are called several times even though only one error message is printed. **Answer:** this is caused by unification that tries to unify types and backtracks in case of failure. Sam says that unification throws exceptions containing error messages, these exceptions get caught and backtracking is done. Only the last one is printed. This is probably not what unification should be doing. Error messages should be generated lazily. See #85 
- [x] all gripers should have their code refactored: calls to `code`/`show_type` should be assigned to let-bindings. This ensures proper evaluation order. Otherwise, subsequent calls to `^` are evaluated right-to-left and if the calls are inlined then variables are generated backwards. **UPDATE** This does not seem to work as intended. Variables names are generated in reverse order anyway. The probable reason is that they are not generated lazily enough. Needs fixing.
- [x] figure out why `hide_fresh_type_vars` is enabled by default. This does not seem to me like a useful thing to do. Perhaps we should reconsider?
- [x] Documentation: describe each new function, explain motivation for refactoring gripers (proper evaluation order), describe mechanism of name generation
- [ ] Fix the bug where effect type variable is not printed in the error message.

jstolarek:
Pasting Sam's more detailed instructions on how to demonstrate the problem:

Let's suppose we try to write a compose function with type:

```
((a) -e-> b, (b) -e-> c) -e-> (a) -e-> c
```

If we get the composition the right way round:

```
links> sig h : ((a) -e-> b, (b) -e-> c) -e-> (a) -e-> c
...... fun h(f, g)(x) {g(f(x))};
```
then it works fine:
```
h = fun : ((a) -b-> c, (c) -b-> d) -b-> (a) -b-> d
```
If we get the order of the functions wrong
```
links> sig h : ((a) -e-> b, (b) -e-> c) -e-> (a) -e-> c
...... fun h(f, g)(x) {f(g(x))};
```
then we get the following error message:
```
<stdin>:2: Type error: The function
    `g'
has type
    `(_) -> _'
while the arguments passed to it have types
    `_'
and the currently allowed effects are
    `|_'.
In expression: g(x).
```
This is actually exhibiting a slight variant of the problem. Because each type variable occurs only once in each type, the pretty-printer decides to display them all as `_`. We can disable this behaviour by switching off the `hide_fresh_type_vars` setting:

[...]

PR #73, opened 2016-09-28
Global type variable names in error messages

jstolarek:
See #43. Don't merge yet.

### DIFF
--- a/types.ml
+++ b/types.ml
@@ -66,11 +66,9 @@ struct
       deriving (Eq, Show)
   let make name arity =
     let id = Utility.gensym ~prefix:"abstype:" () in
-      { id    = id ;
-        name  = name ;
-        arity = arity }
-  let arity { arity = arity } = arity
-  let name  { name  = name  } = name
+      { id; name; arity }
+  let arity { arity; _ } = arity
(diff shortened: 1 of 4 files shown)


tests/shredding/config
```diff
@@ -0,0 +1,7 @@
+database_driver=postgresql
```
fehrenbach: You accidentally added the shredding config.

types.ml
```diff
@@ -1749,7 +1753,7 @@ struct
                                           && is_present (FieldEnv.find v fields))
                         true values in
 
-       let arrow =
+       let arrow () =
```
jstolarek: One of them but unfortunately there's still more :-/

jamescheney:
Is this now ready to be merged?

jstolarek:
> Is this now ready to be merged?

There is one more thing that remains to be done: always displaying type variable and effect names when printing error messages. This should be simple to do but I need to consult @slindley about one implementation detail.

Alternatively, we can merge this in - because it essentially does what it should - and then merge the mentioned change separately.

jstolarek:
Merged as 3e19badbb65df83a9fd15bdfbffb85334745719b

PR #77, opened 2016-10-01
Conditional linking for the oasis-based build system

Drup:
Things work in the following way: With the `Object` section, oasis builds a `.cm[ox]` that encapsulate all the side effects of a module. When a flag is enabled, we add this cmo to the dependencies of `links.cm[ox]`. The code is very commented and should be read commit-by-commit.

Doing all this with libraries is much more painful, because we need to add `linkall` and be careful about the order of dependencies, ocamlbuild is not very good at such delicate things.

Beware, there is a bit of confusion, since there is a binary and a library called `links`. I put them in different directories to avoid confusions. The library should probably be renamed.

I took the occasion to move some files around, at it made things a bit more convenient. In particular, database implementation are isolated on purpose, in order for dependencies not to get confused. I also added a directive to know the available drivers, for testing purposes ... I tested for postgresql, sqlite3 in byte and native mode.

@gildor478 Apparently, `use_obj_foo` is not yet available. This would make this much easier. You might also be interested by the scheme for conditional linking, as we discussed it several time.

Sidenote: you should get rid of camlp4, it makes the compilation time unbearable....


gildor478:
@Drup 
Actually I started working on the conditional linking (for Executable, Libraries are a different problem) and using object. I was stopped by:
https://github.com/ocaml/ocamlbuild/issues/106

At one point, it will be done, but it will probably not be available in OASIS 0.4.8 (maybe 0.4.9).

Drup:
@gildor478 Yes, I figured that out when I saw the bug report. As you can see, the proposed solution works quite well.

gildor478:
@Drup would you be kind enough to extract 2 patches from this PR, one showing the diff for the use_obj and one for the conditional deps and open a bug againt OASIS for that ? This is just to keep track of this when I will resume working on the topic.

Drup:
@gildor478 I'm not sure I understand. You mean you want the patch if, hypothetically, I had `use_obj` at my disposal (and the same if I had conditional deps ?)

gildor478:
I just want to have an example how to link with object and conditionally. 

Re-reading your PR I suppose, for object this is:
+    (** We plug in dependencies manually. See comment on {!use}.
 +        We tag the cm[xo] in order to force the link order.
 +    *)
 +    dep ["file:"^main-.-"cmx"] [file-.-"cmx"] ;
 +    dep ["file:"^main-.-"cmo"] [file-.-"cmo"] ;

And the example for conditional linking:
+(** When a flag is enabled, we tag the main file with a dependency towards
 +    the internal corresponding library.
 +*)
 +let link_conditionally env (flag, file) =
 +  let _lib = lib_of_flag flag in
 +  if oasis_flag env flag then begin
 +    (* We need conditional linking, so only the binaries need the
 +       conditional link, not the source file.
 +    *)
 +    tag_file (main-.-"native") [pkg flag ];
 +    tag_file (main-.-"byte")   [pkg flag ];
 +
 +    (** We plug in dependencies manually. See comment on {!use}.
 +        We tag the cm[xo] in order to force the link order.
 +    *)
 +    dep ["file:"^main-.-"cmx"] [file-.-"cmx"] ;
 +    dep ["file:"^main-.-"cmo"] [file-.-"cmo"] ;
 +  end
 +
 +(** The list of flag that have associated libraries
 +    that we want conditionally linked. *)
 +let flags = [
 +  "postgresql", "lib/postgres/pg_database" ;
 +  "mysql", "lib/mysql/mysql_database" ;
 +  "sqlite", "lib/lite/lite_database" ;
 +  "sqlite3", "lib/lite3/lite3_database" ;
 +  "monetdb", "lib/monetdb/m5_database" ;
 +]

Drup:
Yes, that's it. I think the content of the `myocamlbuild.ml` should be pretty much self explanatory for you. By adding the dependency between `cm[xo]`, ocamlbuild properly add it to the command line when doing the final link.

One thing I had to be careful is that the ocamlfind flag `package(foo)` is not conditional (as opposed to `pkg_foo`, which oasis used to use). So, if we put everything in the main directory, this doesn't work well. I'm of the opinion that oasis should be much more careful with tagging (currently it work by tagging entire directories, this cause issues quite often).

myocamlbuild.ml
```diff
@@ -1,44 +1,67 @@
...
-     else
-       Command.execute (mv "links.byte" "links")
-  | _ -> ()
- *)
- )
+    (* Run the normal build process *)
+    dispatch_default hook ;
+
+    let env_filename = Pathname.basename BaseEnvLight.default_filename in
+    let env = BaseEnvLight.load ~filename:env_filename ~allow_empty:true () in
```
jstolarek: From my understanding of documentation it suffices to say:

```
let env = BaseEnvLight.load () in
```
The `env_filename` that you construct is identical to the default value of `filename` that is assumed if no argument is passed. `allow_empty` seems to be irrelevant here. And indeed this shorter version seems to work. Am I missing something?

jstolarek:
Oh, and one thing that I have to undo with this patch is moving of all source files into their own source directory. I agree this is definitely a good thing to do, but it has to wait. There are several large patches to be merged in the near future and I don't want to cause unnecessary merge conflicts. So the source files will get their own directory but not just yet.

myocamlbuild.ml
Drup: I remember a subtlety about that, but I don't remember which exactly. You can try. :)

myocamlbuild.ml
```diff
@@ -1,44 +1,67 @@
 (* OASIS_START *)
-(* DO NOT EDIT (digest: d41d8cd98f00b204e9800998ecf8427e) *)
```
Drup: Yes, oasis let it, but it's useless and will be replaced anyway.

Drup:
> One issue I've ran into is that make clean does not remove setup.data. The result is that this erroneously persists things across different runs of configure. For example if I say configure --enable-postgresql and then run configure again, the PostgreSQL backend will be enabled,

That's the intended behavior of `configure`. You can use `make distclean` to also removes the result of `configure`.

Drup:
> Oh, and one thing that I have to undo with this patch is moving of all source files into their own source directory. [...] There are several large patches to be merged

Honestly, I would just fix said patches. git is clever about these kinds of thing, it should just be a matter of rebasing. :)

In particular, putting the database files in the same directory as the rest *will* cause issues (as I explained in a previous comment).

jstolarek:
> Honestly, I would just fix said patches. git is clever about these kinds of thing, it should just be a matter of rebasing. :)

I would do the same, but it is not me working on these patches :-)

jamescheney:
Thanks very much Gabriel for the contribution! However, as Jan says, changing the directory structure while other changes are in progress is potentially problematic.  

I think shredding+sessions-2 is ready to merge - there are some minor things I'd like to improve about that but nothing urgent (or that I'll have time for in the next few days).

I don't know the status of other branches that are supposedly going to be merged soon.  I think we should discuss next week.  We are am **trying** to decrease the number of long-running branches.

jstolarek:
Gabriel, I have a few more questions. Previous `Makefile` had some extra options like:

```
ifdef PROF
OCAMLOPT := ocamlopt -p -inline 0
endif

OCAMLYACC := ocamlyacc -v

OCAMLFLAGS=-dtypes -w Ae-44-45 -g -cclib -lunix
```
I am now trying to recover these setting in Oasis:
1. Profiling. I noticed I can `./configure --enable-profile` to have `-p` passed to `ocamlopt`. But how do I pass `-inline 0`? I could use `NativeOpt` to pass extra arguments to `ocamlopt` from Oasis but I can't find any way to test whether `profile` flag has been enabled. I can define my own flag but then I have to pass two flags to `configure` which seems silly.
2. ocamlyacc. I see no way of passing extra options here :-/
3. Ocamlflags: I suppose I can just pass them to `NativeOpt` and `ByteOpt`, but this means lots of duplication (I have to repeat options twice in two section - library and executable). I'd rather just write them once. I can probably tell `ocamlbuild` to pass these options to `ocamlc`. Not sure if this is the best solution.

Again, I can probably devise hacks around these but I feel they will be ugly.

Issue #76, opened 2016-09-30
Decide which database backends we support


jstolarek:
At the moment there are five database backends in Links source code:
- PostgreSQL
- MySQL
- SQLite
- SQLite3
- MonetDB

Per today's discussion we should probably do some cleanup:
- Why are there two SQLite backends? Do we really need both?
- Do all backends support all the features of Links? James suggested that SQLite(3) might not support all the operations required for query shredding to work. Is this indeed the case? If so, how do we handle situations when query shredding is used with SQLite backend? Do we emit an error message or just crash unexpectedly?
- It was suggested that one of the backends is enabled by default in the Links distribution that is published to OPAM. Stefan suggested SQLite as it requires no extra setup and allows people to easily try out Links. James suggested PostgreSQL as the one we currently support the most.
- Is anyone using MonetDB backend?
# TODO
- [x] delete SQLite 2 backend if we don't plan on supporting it
- [x] see whether MonetDB backend works
- [ ] decide on a default backend and enable it in `_oasis` file
- [x] make sure that we get reasonable error messages when trying to use shredding with database engines that don't support all the features required by shredding to work.

jamescheney:
>     Why are there two SQLite backends? Do we really need both?

The first one is for SQLite 2, which is incompatible (different OCaml interfacing libraries).  I don't think there is a reason to maintain the SQLite (2) version.


>     Do all backends support all the features of Links? James suggested that SQLite(3) might not support all the operations required for query shredding to work. Is this indeed the case? If so, how do we handle situations when query shredding is used with SQLite backend? Do we emit an error message or just crash unexpectedly?

Shredding has only been tested with Postgres, and I believe it relies on features that Postgres supports and MySQL/Sqlite may not.  I intend to add a run-time error if we try to use shredding with a query that refers to a non-Postgres DB (it would be more work to catch this statically, similarly to how we currently don't statically check that a query refers to tables all from the same DB.)

>     It was suggested that one of the backends is enabled by default in the Links distribution that is published to OPAM. Stefan suggested SQLite as it requires no extra setup and allows people to easily try out Links. James suggested PostgreSQL as the one we currently support the most.

I wasn't arguing that PostgreSQL would be a better defauly than SQLite - the latter has the advantage that all you need is the SQLite library, the database is just stored in a plain file and you don't have to install + configure a database.  This would be enough for a lot of toy applications.

>     Is anyone using MonetDB backend?

[...]

Issue #62, opened 2016-09-15
Replacing OCamlMakefile with ocamlbuild?


jstolarek:
Currently used build method - OCamlMakefile - seems to have two drawbacks:
- it puts all the compilation artefacts in the same directory as the source files. I looked at the documentation and it seems that OCamlMakefile can't do better
- file dependencies need to be figured out manually and listed explicitly in the `Makefile`

I was wondering if it would make sense to move from OCamlMake to `ocamlbuild`? As a quick and dirty attempt I managed to compile links with:

```
ocamlbuild -use-ocamlfind -pkgs 'bigarray,num,str,deriving.syntax,deriving.syntax.classes,deriving.runtime,lwt,lwt.syntax,lwt.unix' -syntax camlp4o "links.native"
```

This is definitely not sufficient as it does not seem to compile database backends. It would require a bit more investigation to figure out how to deal with that. Do others think this is worth doing? The benefits would be clean source tree (all `.o`, `.annot`, `.cmi` and `.cmx` files would land in `_build` directory) and there would be no need to manually figure out module dependencies.

See #78 for an uo-to-date list of TODOs.

fehrenbach:
Last time I checked, ocamlbuild did not do parallel compilation.

jstolarek:
AFAIK it still doesn't. Would this be a blocker?

fehrenbach:
I guess not. If you can figure out the database stuff, go for it :)

(My understanding is that the database backends contain side-effecting code at the toplevel with which they register themselves at runtime. It should be enough to convince ocamlbuild that they need to be built and linked into the final executable.)

jstolarek:
Another option that looks promising is [Oasis](https://github.com/ocaml/oasis). It is built on top of ocamlbuild. At a first glance it seems that with Oasis it might be easier to set up all the complex build rules required for building database backends.

yallop:
[topkg](http://erratique.ch/software/topkg/doc/Topkg.html) is also worth a look.  Several packages are currently switching over to topkg from _oasis ([e.g.](https://github.com/mirage/ocaml-git/pull/158)).  Package descriptions are generally very succinct, and it's well integrated with other parts of the ecosystem (ocamlbuild, ocamlfind, OPAM, etc.)

[I hope you don't mind my jumping in.  I think I wrote some of the current OCamlMakefile-based setup for Links, so I feel a certain responsibility here.]

jstolarek:
> I hope you don't mind my jumping in.

[...]

Issue #61, opened 2016-09-15
Figure out how to do a release


jstolarek:
There are several files in the repo related to preparing a release tarball: `checkfiles` and `make_release` scripts as well as `MANIFEST` and `COMANIFEST` data files. These are woefully out of date. `MANIFEST` and `COMANIFEST` mention files that no longer exist and omit files that exist. Files generated by `checkfiles` are not properly cleaned by `make clean` and are not ignored by git. These are all easy things to fix but the questions are whether the upcoming release will be made using these scripts and whether it is worth to invest time in fixing this? Perhaps it would make sense to make a release using OPAM?

Since this is about making a release I am assigning this to Gorgie milestone, but if anyone more knowledgeable than me thinks this is incorrect please remove the milestone.
# TODO
- [x] remove old scripts and data files from the repo: `checkfiles`, `MANIFEST`, `COMANIFEST`, `make_release`
- [x] remove `RELEASE-CHECKLIST`, but make sure relevant things are placed on the wiki
- [x] update contents of `INSTALL` file. Perhaps point to wiki?
- [x] create OPAM description file
- [x] verify that the package installs correctly with all the required data files (most importantly, `prelude.links`, but also user documentation)
- [x] check how to enable/disable database backends during installation with opam
- [x] check what happens when libraries required by database backends are installed in the system. Is the package automatically reinstalled with the new backend enabled?
- [ ] make sure the installation process and release process are described. Release should be described on the wiki [here](https://github.com/links-lang/links/wiki/Release-checklist). Installation should be described in `INSTALL` file (or perhaps on the wiki?) Also, the database backend wiki page needs to be updated.

jstolarek:
Oh, and I forgot that `make_release` assumes that the project is stored in Subversion, not git. This would also need to be updated.

jstolarek:
As part of making an official release we should:

  * [ ] assign version number to Links
  * [ ] Decide on a license and put LICENSE file in the repo

fehrenbach:
2: It's GPL, see https://github.com/links-lang/links/blob/master/COPYING

jstolarek:
Thanks. Marked as done. (1) and (3) remain.

jstolarek:
Note to self: assuming we use opam to build and distribute a release I need to figure out how to enable particular database backends. Is it possible to pass extra flags to opam for conditional compilation based on oasis? Should we assume a default backend (postgres?) and if user wants a different one they need to build from source?

[...]

PR #78, opened 2016-10-03
Oasis build system

jstolarek:
Here's a preliminary patch replacing OCamlMake build system with Oasis (#62). Big thanks go to Gabriel Radanne for his help. Please review this patch but don't merge yet.

What we get with Oasis:
- no need to specify dependencies manually. All the modules still need to be listed but this can be done in any order (I suggest sticking to alphabetic order)
- all build artefacts are placed in `_build` directory and no longer clutter the directory with source files
- Oasis conveniently generates `configure` and `Makefile`, so now we build things like `./configure --enable-postgresql --enable-tests --enable-docs && make && make docs && make test`
- it also seems that we can maintain parallel builds. This can be controlled with `THREADS` environment variable. By default parallel build is on.

What changes in our workflow:
- As developers we need to have `oasis` package installed. Oasis generates several files (`setup.ml`, `configure`, `Makefile`). These files are stored inside the repo, which means that end users should not need oasis to install Links. At first I was worried about stroging automatically generated files in the repository but I believe these files will hardly ever change. I think adding new flags and sections to `_oasis` file will change `configure` script and that's it. From now on we use `configure`/`make` routine.

Questions
- I have removed removed `deriving` directory along with executable inside it. Things seem to magically work nevertheless. Can someone more knowledgeable than me verify that this is indeed OK?
# TODO
- [ ] parametrize test script on the generated executable name. This solves the problem that oasis generates file named `links.native` rather than `links`. This can be done in `_oasis` file with `Command: testing_script $links_executable`, where `$links_executable` is executable section name
- [x] I was unable to recover some setting present in the original `Makefile`. One that worries me the most is `OCAMLFLAGS=-dtypes -w Ae-44-45 -g -cclib -lunix`. I will keep working on this. **EDIT:** fix this in `_tags` file - see comments below
- [ ] Each database backend needs to be in its separate directory. We should think what this directory should be. I will strongly insist on putting all source file in a separate directory once all large branches (like application server) are merged. I suggest to put all source files in `src` and backends on `src/drivers/`. At the moment they are in `lib`, per Gabriel's patch. I guess it is best to take a look at other OCaml projects and see whether they tend to store their library source files in `src` or `lib`
- [ ] All of this needs to be documented on the wiki. I will write the documentation once I get an 'OK' to merge this change.
- [ ] Documentation sections of `_oasis` file need tweaking before we make a release. We need to specify which files should be installed and where. Related to #61.
- [ ] We should decide on a default backend and enable it. Currently all backends are disabled. Related to #76.


jstolarek:
I think I should elaborate a bit more on this one:

> I was unable to recover some setting present in the original `Makefile`. One that worries me the most is `OCAMLFLAGS=-dtypes -w Ae-44-45 -g -cclib -lunix`. I will keep working on this.

Actually, Oasis has a simple way of passing extra arguments to the compiler. One either uses `ByteOpt` or `NativeOpt` setting, depending on whether one wants to build bytecode version or native version. The problem is that these options would have to be duplicated. Slightly annoying but not a big deal.

Also, which version do we want to build by default: byte or native? I see no easy way to switch between them other than changing entry in the `_oasis` file. Sadly, Oasis does not allow field `CompiledObject` to be conditional depending on a flag.

Drup:
> We should decide on a default backend and enable it. Currently all backends are disabled.

You don't need to: just add an OR dependency in the opam file. In any cases, that's better left for the opam description.

> One that worries me the most is OCAMLFLAGS=-dtypes -w Ae-44-45 -g -cclib -lunix.

Add to `_tags` the following line:
```
true: annot, warn(Ae-44-45)
```
annot is the new name of dtypes. `-g` is added by default by ocamlbuild and I'm pretty sure you don't need the rest (not even sure why it was there to begin with). You might want `bin_annot` there too (for merlin).

jstolarek:
> Add to `_tags` the following line:

Won't `_tags` be overwritten by oasis?

Drup:
> Won't _tags be overwritten by oasis?

No, just give oasis his section and it'll be happy. [Example](https://github.com/ocsigen/tyxml/blob/master/_tags).

jamescheney:
For comparison, here is a branch that uses OMake instead of ocamlbuild.  Conditional building works much the same way as the existing OCamlMakefile approach, and does not require reorganizing the directory.

https://github.com/links-lang/links/tree/jrc-omake-test

jamescheney:
Closing without merging, as just discussed.

PR #87, opened 2016-11-07
Issue 27

fehrenbach:
This (somewhat arbitrarily) decides that '@' is not allowed in record fields and uses it as the separator for nested field names in queries. '@' seems to be allowed in SQL column names for at least PostgreSQL.


jamescheney:
This looks reasonable to me, but we should make a note of this assumption and may need to adjust if some other DBMS doesn't allow @.

PR #88, opened 2016-11-10
Use multi-row insert, clean up some cruft, and silence warnings

fehrenbach:


pg_database.ml
```diff
@@ -201,7 +188,7 @@ let get_pg_database_by_string args =
           process.  This has to be done by acquiring UID, finding corresponding
           entry in passwd table and reading user's login name. *)
        let user = if user = ""
-                  then (Unix.getpwuid (Unix.getuid ())).pw_name
+                  then let open Unix in (Unix.getpwuid (Unix.getuid ())).pw_name
```
jamescheney: I'm not sure what the "let open Unix in" part adds to this.  Couldn't we leave out "Unix." prefixes in the rest of the expression?

PR #91, opened 2016-11-18
Fix #89 and remove outdated database backends

### LINKED ISSUE #89: *_database.ml out of date (except pg_database.ml)
The shredding code changed the dbvalue parent class / interface and only Pg_database.dbresult has been updated to implement the additional methods.  Either the missing methods need to be implemented for the other database classes or the shredding code needs to be specialized to only work for pg_database.

I propose to retire the lite_database.ml and m5_database.ml files and update the other two.

Making this a blocker for release since it stops people from using databases other than postgres.

jamescheney:
This fixes issue 89 by adding the missing methods to the dbresult class and its derivatives in pg_database, lite3_database and mysql_database.  It also removes the lite_database and m5_database modules since the OCaml libraries they depend on do not appear to be maintained (at least as OPAM packages).


jamescheney:
It would also probably be a good idea to write tests that exercise the sqlite3 and mysql functionality.

jamescheney:
I have also added a reduced version of the xpath test that is failing due to bug #25.  This may be helpful to @slindley in investigating that problem.

jstolarek:
Ugh.. I was hoping that comment above will show on da19c0d commit.

jamescheney:
As mentioned on #92, I think we should merge this first, then #92 (after fixing any problems due to the overlapping commits) since this fixes some problems that aren't fixed yet in #92.  Alternatively, merging this into #92 first and dropping this request would be fine.

jamescheney:
Merging now, #92 may need to be updated after that.

jstolarek:
It definitely will need a rebase. I'll do that tomorrow after dev meeting.

Issue #72, opened 2016-09-28
"links" executable name collides with already existing software


jstolarek:
This is a bit silly but was anyone ever been bothered by the fact that "Links" name - and thus `links` executable as well - collides with [already existing web browser](http://links.twibright.com/)? As part of my local setup I renamed Links the Programming Language executable to `linksc` to avoid name clash. Given that we plan to make a release, perhaps it would be a Good Thing to rename the executable to `linksc` (or something else) to avoid problems on systems that also have the Links web browser installed?
# TODO
- [ ] make any kind of decision.

jstolarek:
Once we decide on executable name we should make sure that this is indeed what is generated and installed in the system. See #61 and #62.

LeifW:
I was naming it "links-lang" locally because of that. Though I imagine less
and less using that browser these days.

On Sep 28, 2016 10:37 AM, "Jan Stolarek" <notifications@github.com> wrote:

This is a bit silly but has anyone ever been bothered by the fact that
"Links" name - and thus links executable as well - collides with already
existing web browser <http://links.twibright.com/>? As part of my local
setup I renamed Links the Programming Language executable to linksc to
avoid name clash. Given that we plan to make a release, perhaps it would be
a Good Thing to rename the executable to linksc (or something else) to
avoid problems on systems that also have the Links web browser installed?

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/links-lang/links/issues/72>, or mute the thread

jstolarek:
> I was naming it "links-lang" locally because of that.

I'm glad to hear that it's not only me being affected by this.

Perhaps `linksi` would be a better name, since it is really an interpreter. Let's save `linksc` until we have a compiler.

jstolarek:
@slindley decided on the name "linx".

PR #92, opened 2016-11-23
Use OPAM for making a release

### LINKED ISSUE #76: Decide which database backends we support

jstolarek:
This addresses #61 by using OPAM for making a release. There are three separate commits here.

First drops two unused database backends. This fixes #76 but conflicts with one of the commits in #91. I have decided to incorporate changes made by @jamescheney in #91 because this was a natural thing to do for my patch.

Second commit changes Makefile to automatically check for presence of database backends in the system. This is required for OPAM but is also a sensible change on its own, thus a separate commit. 

Third commit is a proper fix for #61 that drops outdated release scripts and adds an `opam` file.

One thing worth pointing out here are optional dependencies of the Links package. No database backend is enabled by default but Links package will recompile if any of the backends is installed in the system. With the change in the `Makefile` we no longer need to define environmental variables.  Potential downsides are:

1) this relies on shell command which is not part of Make standard (or so I read).  It works with GNU make on Linux.  James, could you test whether this works on Mac?

2) if a given backend package is installed the corresponding backend will always be enabled and compiled.  I hope this is not a problem.

## TODO

  * [x] make sure that prelude is installed in `lib` directory and that links knows how to find it.


jamescheney:
I tried doing "make" and got an error (the error a problem with lite3_database.ml being out of data that is fixed by #91).  I suggest we merge #91 before this.

It seems to be including the sqllite3, mysql and postgresql libraries automatically, as desired, on a Mac.  Does this test the thing you were uncertain about, or did you want me to test something else too?

jstolarek:
> I suggest we merge #91 before this.

Agreed.

>  Does this test the thing you were uncertain about?

Yes. I was uncertain whether the `shell` command will be available in Mac version of `make`.

jamescheney:
The standard Mac version seems to be GNU make (I have v3.81).  It is also possible to install/use a more recent version if needed.  But seems to be fine as is.

fehrenbach:
Not sure whether the branch/PR ownership etiquette allows this, but I fixed the travis build for you.

jstolarek:
Thanks!

jstolarek:
Did we agree how to point our Links installation to the Prelude location? There was a suggestion for a wrapper script. This should be simple. Alternatively, OPAM allows also to display a custom text message after installation is finished, so we can also reproduce current `make install` behaviour and tell user to define `LINKS_LIB` variable. There was also a suggestion to search pre-defined directories for a prelude file but this would be a larger implementation effort and probably out of scope of this PR.

jamescheney:
I think we agreed to do whatever you think is easiest/best, after looking at what "make install" does.  The shell script idea would address #72 at the same time.  

It probably wouldn't hurt to print a message explaining intended usage (plain "links" with config file or LINKS_LIB, vs. shell script) on installation.  Say something like "You can run Links interactively using the default configuration using the 'runlinks' script.  To connect to a database or run as a web application, consult the installation instruction/README."

basicsettings.ml
```diff
@@ -24,14 +24,35 @@ let concurrent_server = Settings.add_bool ("concurrent_server", true, `System)
...
 (** Name of the file containing the prelude code. *)
-let prelude_file = 
+let prelude_file =
   let prelude_dir = match Utility.getenv "LINKS_LIB" with
-      None -> Filename.dirname Sys.executable_name
+    (* If user defined LINKS_LIB then it takes the highest priority *)
     | Some path -> path
+    | None ->
+        (* If LINKS_LIB is not defined then we search in current directory *)
+        let current_dir = Filename.dirname Sys.executable_name in
```
fehrenbach: `current_dir` sounds like current working directory. This is not what is intended here. The idea is to get the path to the links executable (wherever it is) and find the prelude in the same folder.

(I just found this https://github.com/dbuenzli/logs/issues/14 which suggests that that would not work if links is bytecompiled. We might want to check/fix that.)

basicsettings.ml
```diff
@@ -24,14 +24,35 @@ let concurrent_server = Settings.add_bool ("concurrent_server", true, `System)
...
+        if Sys.file_exists (Filename.concat current_dir "prelude.links") then
+          current_dir
+        else
+          try
+            (* If all else failed we search for OPAM installation of Links and
+               use a prelude that it provides *)
+            let opam_lib_dir =
+              input_line (Unix.open_process_in "opam config var lib 2>/dev/null") in
```
fehrenbach: We might want to use `opam config var links:lib`. It prints `"#undefined"` for me (I don't have links installed through opam). `opam config var ocamlbuild:lib` prints `"/home/stefan/.opam/system/lib/ocamlbuild"` for me. There's probably nobody who has different lib dirs from the defaults, but at least we save ourselves the concat .. "links" line.

(Not sure whether we need to do something in the opam file to populate the var links:lib thing.)

jstolarek:
One point to make is that in order to make an OPAM release we need to have a tarball available for download. Probably the best way of doing that will be making a tag on github. More details for the curious: https://opam.ocaml.org/doc/Packaging.html#GettingafullOPAMpackage

basicsettings.ml
jstolarek: I'll change the variable name to something more accurate.

> I just found this dbuenzli/logs#14 which suggests that that would not work if links is bytecompiled.

Up until two days ago I used to byte-compile links and this code worked (it was there before my changes). So that would be surprising...

fehrenbach: Yeah, that was me being stupid. We of course don't care what the executable name is, we just need the path to it.

basicsettings.ml
jstolarek: Using `links:lib` is a good idea - I've updated the code. Thanks!

jamescheney:
> James, could you please test whether the SQLite3 backend works correctly? I want to be sure that everything works as it should after a rebase.

It builds successfully for me (with mysql, postgresql and sqlite3 enabled).  I haven't tested that database access for sqlite3 or mysql work; I've created an issue to remind us to do that at some point (not necessarily now).

> >    The shell script idea would address #72 at the same time.

> I'm sorry, I don't follow. The idea behind #72 is not to have executable called links - how does an extra script address that?

I thought we had discussed the possibility of keeping the "links" executable, and adding a script or scripts that runs links, as one possible resolution to #72.  If the script(s) have a hardwired path to our links executable, then the existence of links-the-web-browser binary will not break use of links-the-language, and vice versa (assuming that links-the-web-browser-binary gets installed to /usr/bin or somewhere earlier in the PATH than the links/bin path.)

jstolarek:
>  (assuming that links-the-web-browser-binary gets installed to /usr/bin or somewhere earlier in the PATH than the links/bin path.)

I believe this is a false assumption: OPAM prepends it `bin` directory to `PATH` rather than appending it. So I think that solution won't work for us.

dhil:
Maybe we can use the `lib` section in `.install`? C.f. https://opam.ocaml.org/doc/1.1/Packaging.html#Manuallyinstallingbinariesorlibaries

jstolarek:
> Maybe we can use the lib section in .install to fix the path for prelude.links?

The problem is that we still need to pre-compile the prelude and `.install` files only seem to support copying pre-existing files. We could make prelude pre-compilation a part of `make`, but I don't think this is worth the effort.

This is now merged.

PR #93, opened 2016-11-29
Newer version of modules system

SimonJF:
This PR adds the newer version of the modules system as described in the Wiki. The name resolution system is an implementation of Neron et al.'s scope graph formalism.

This should be less ad-hoc and cleaner than the previous system.

Additionally, since it's a rather major change, the module system is hidden behind a flag "-m" (or, setting the setting flag modules to true).

Still unsupported is partially-qualified imports from external files -- they all have to be fully-qualified at the moment. It's probable that we'd be looking at some sort of tradeoff here (such as explicit import ordering, or explicit declaration of all imported modules) if we want this, but I don't think it's crucial for the release.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -89,6 +89,8 @@ SOURCES = $(OPC)                                \
           parse.mli parse.ml                    \
           sugarTraversals.mli  sugarTraversals.ml       \
 					moduleUtils.mli moduleUtils.ml \
+					uniquify.mli uniquify.ml \
+					scopeGraph.mli scopeGraph.ml \
 					chaser.mli chaser.ml \
 					desugarModules.mli desugarModules.ml \
           desugarDatatypes.mli desugarDatatypes.ml      \
--- a/basicsettings.ml
+++ b/basicsettings.ml
@@ -88,6 +88,9 @@ let allow_stale_cache = Settings.add_bool("allow_stale_cache", false, `System)
 (* Optimization pass? *)
 let optimise = Settings.add_bool("optimise", false, `User)
 
+(* Allow modules? *)
+let modules = Settings.add_bool("modules", false, `User)
+
 (* Compile & cache whole program, closures, and HTML *)
 let cache_whole_program = Settings.add_bool("cache_whole_program", false, `User)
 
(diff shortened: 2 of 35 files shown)


SimonJF:
There's one more change I need to make (namely, not invoking the chaser for FQ names, only invoking for imports). Patch incoming, please don't merge until then. Thanks!

jamescheney:
As discussed, this is ready to merge.  

However, this pull request can't be "rebased and merged" due to conflicts.  Shall we squash it into one commit, or does someone want to manually rebase it?

jstolarek:
> this pull request can't be "rebased and merged" due to conflicts. Shall we squash it into one commit, or does someone want to manually rebase it?

But squashing into one commit won't make the conflicts go away. Conflicts will still be there, no matter whether the branch is merged or rebased.

jamescheney:
Well, the green button above remains green and claims that everything will be fine if we "squash and merge".  (Continuing to follow my "monkey randomly bashing at the keyboard" approach to github.)  If you choose "Rebase and merge" it says "This branch cannot be rebased due to conflicts.
Rebasing the commits of this branch on top of the base branch cannot be performed automatically due to conflicts encountered while reapplying the individual commits from the head branch. "

Which I thought meant that there are conflicts in intermediate states but not between the initial and final state if they are squashed into one commit.

SimonJF:
This is strange; I rebased (and fixed conflicts) before opening the PR.

My assumption was that it was going to be squashed, since there are many intermediate commits in there.

jstolarek:
That's weird - the green button tells me that there are no conflicts. I guess this is why I never use github for merging anything :-/

SimonJF:
The rebase went through fine once I fixed the error. Is there anything stopping us from doing a squash and merge?

jamescheney:
The "squash and merge" button is enabled, but after pressing it, it asks to confirm the commit message (which is all of the above commit messages concatenated).  Simon, maybe it would be better if you summarize the squashed commit?

jamescheney:
Looking at the affected files, my guess would be that it's something silly like conflicting changes to the Makefile, or something.

jstolarek:
Will the rebase&merge be happening now? If so, I'll wait with my rebase of #73.

PR #96, opened 2016-12-02
Stay in shell!

dhil:
Changed the default behaviour of 'display_fatal' such that it no longer exits the shell when running in interactive mode.


jstolarek:
I will merge this next week unless there are any objections.

PR #99, opened 2016-12-07
More Efficient Module System

SimonJF:
The scope graph paper was really, really nice -- but the resolution algorithm contained therein didn't seem to be efficient enough (even after I threw all the dynamic programming I could at it).

This PR implements a smaller, simpler version of the module system. Unlike my first attempt, I actually know what's going on in this one. Unlike the previous version, module desugaring is very quick this time.

I'm hoping that this will be the last *major* revision to this, certainly before the release.

I have tested this by porting the monolithic agario implementation to a modular one, which now works. You can see this here:
https://github.com/links-lang/links-examples/commit/51ff00dba6b7da5566869c59a2a908dec75c68da


jamescheney:
I think we agreed to merge this.  Merge as is, or squash?  If squash, Simon, can you do that and summarize the commits?

SimonJF:
I'm going to squash and merge, I think. I'll do that now.

PR #101, opened 2016-12-15
First attempt at in-Links web server

jgbm:
Work on Links web server.

Still works on simple examples.

The commit history is useless, and should be squashed.  I was of the impression that GitHub would give a way to do this easily. :)


SimonJF:
Awesome! FYI GH allows "squash and merge" so it'll be squashed into a single commit when it gets merged into sessions, thankfully no need to do it manually.

What are the main differences in running? Is there a separate executable or is it a flag?

jgbm:
@SimonJF Check examples/webserver/buttons.links, for example.  In all likelihood, you shouldn't need to set flags or anything.  You will probably need to know that the default port is 8080 (which I will document in the hypothetical wiki pages).

SimonJF:
Ah, I see, thanks!

jamescheney:
One thing that needs to be done is to make the CI test pass.  This may be easy, the build currently fails with this message:

> ocamlfind: Package `cohttp' not found

lib.ml
```diff
@@ -22,6 +22,7 @@ let alias_env : Types.tycon_environment =
 let datatype = DesugarDatatypes.read ~aliases:alias_env
 
 let cgi_parameters = ref []
```
jamescheney: I gather these are among the global references that need to be made thread-local.

lib.ml
```diff
@@ -1518,6 +1507,14 @@ let env : (string * (located_primitive * Types.datatype * pure)) list = [
...
                   `Record [])),
      datatype "(Socket) ~> ()",
+     IMPURE);
+    "unsafeAddRoute",
+    (`PFun (fun _ -> assert false),
+     datatype "(String,(String) ~> a) ~> ()",
+     IMPURE);
+    "servePages",
+    (`PFun (fun _ -> assert false),
+     datatype "() ~> ()",
```
jamescheney: Given that servePages does not return, could its result type be "a" instead?

lib.mli
```diff
@@ -33,10 +33,14 @@ val primitive_stub_by_code : Var.var -> Value.t
 val primitive_name : Var.var -> string
 val primitive_location : string -> Sugartypes.location
 val primitive_arity : string -> int option
-val cgi_parameters : (string * string) list ref 
+val cgi_parameters : (string * string) list ref
```
jamescheney: Another global that should be thread-local?

lib.mli
```diff
@@ -33,10 +33,14 @@ val primitive_stub_by_code : Var.var -> Value.t
...
 
+val cohttp_server_response : (string * string) list -> string -> (Cohttp.Response.t * Cohttp_lwt_body.t) Lwt.t
 val print_http_response : (string * string) list -> string -> unit
 
 val primitive_names : string list
 
 val prim_appln : Env.String.name -> Ir.value list -> Ir.tail_computation
+
+val cgi_parameters : (string * string) list ref
```
jamescheney: Is there a reason "val cgi_parameters" is declared twice?

prelude.links
```diff
@@ -1166,3 +1166,7 @@ fun assertEq (actual, expected) server {
       sysexit(2);
   }
 }
+
+####
+sig addRoute : (String, (String) ~> Page) ~> ()
```
jamescheney: This "addRoute" / "startServer" stuff seems awfully imperative. What happens if I call startServer twice?  What happens if I call addRoute after startServer?  Given that Links is supposedly functional, is there an obstacle I'm not seeing to defining a "functional" startServer that takes a list of pairs (String, String ~> Page) and starts the server using the list as the route table?  Or is there some benefit to the imperative API that I'm not seeing?

jstolarek:
I think we should have the two new dependencies listed as such in the `opam` file.

lib.mli
jgbm: Let's call that a merge error. :)

lib.ml
jgbm: Yes, I think they're ```cgi_parameters```, ```cookies```, ```http_response_headers````, ```http_response_code```

.travis.yml
```diff
@@ -11,7 +11,7 @@ before_install:
   - eval `opam config env`
   - psql -c 'create database links;' -U postgres
 install:
-  - opam install -y deriving lwt postgresql sqlite3 mysql
+  - opam install -y deriving lwt postgresql sqlite3 mysql conduit cohttp
```
jstolarek: Stefan, is `conduit` necessary here? It _seems_ to me that this is a transitive dependency, ie. `conduit` is required by `cohttp`, but is not directly used by Links. @jgbm, is that correct? If so than just installing `cohttp` will be fine.

fehrenbach: I don't know. That's what the INSTALL file says.

jstolarek: Indeed. @jgbm, can you clarify?

slindley:
This branch breaks the old CGI mode. Most examples work initially, but subsequent interaction with the server results in the following error:

    "Assert_failure proc.ml:92:14" 

For instance, buttons.links displays a form, but gives this error when I click on a button.

slindley:
Apparently global variables disappear from the environment when the client calls the server. For instance, try changing the beginning of `progress.links` to

    var one = 1;
    
    fun compute(count, total) server {
     if (count < total) {
      showProgress(count, total);
      compute(count+one, total)
     } else "done counting to " ^^ intToString(total)
    }

This is particularly a problem with some of the database examples where we use global variables to store the names of database tables.

slindley:
Bizarrely, if you enter `7937` or larger into the textbox in `examples/webserver/progress.links` then the following error arises on the `7937`-th client -> server call:

    Error handling ((headers
      ((accept "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8")
       (accept-encoding "gzip, deflate") (accept-language "en-US,en;q=0.5")
       (connection keep-alive) (content-length 53)
       (content-type application/x-www-form-urlencoded) (host localhost:8080)
       (referer http://localhost:8080/)
       (user-agent
        "Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:50.0) Gecko/20100101 Firefox/50.0")))
     (meth POST) (resource /) (version HTTP_1_1) (encoding (Fixed 53))): (Invalid_argument "base64     decode gave error: AfYNAvENAAADgT7yDQAAA4A")

I don't know if this has anything to do with the appserver code because in CGI mode I've never tried a number that large before.

slindley:
The 7937 problem is already present in the CGI version.

Issue #132, opened 2017-01-23
Instantiation error: Arity mismatch


dhil:
After merging with `sessions` branch, the following example causes an `Instantiate.ArityMismatch` error to occur in the `effects+sessions` branch:
```
fun state(m) {
  handle(m) {
    case Return(x) -> fun(s) { x }
    case Get(k)      -> fun(s) { k(s)(s) }
    case Put(p,k)    -> fun(s) { k(())(p) }
  }
}

fun runState(m, s0) { 
  var f = state(m); 
  f(s0)
}
```
See stack trace:
```
links> fun runState(m, s0) { var f = state(m); f(s0) };
Arity mismatch in type application
Expression: `TAppl ((`TAbstr ({contents = [(7041, (`Unl , `Any ), `Type ...);
                               (7042, (`Unl , `Any ), `Row ...);
                               (7043, (`Unl , `Any ), `Type ...)]},
                   (`TAppl ((`Var "state", ...),
                             [`Type `MetaTypeVar ...; `Row ({}, ..., false);
                              `Type `MetaTypeVar ...; `Presence `Var ...;
                              `Presence `Var ...]), ...)), ...),
         [`Type `MetaTypeVar ...; `Row ({}, ..., false);
          `Type `MetaTypeVar ...; `Presence `Var ...; `Presence `Var ...])
Type: (() {Get:a,Put:(a) {}-> ()|b}~> c) {Get{d},Put{e}|b}~> (a) {Get{d},Put{e}|b}~> c
Raised at file "transformSugar.ml", line 121, characters 6-37
Called from file "transformSugar.ml", line 554, characters 24-36
Called from file "transformSugar.ml", line 339, characters 26-34
Called from file "transformSugar.ml", line 554, characters 24-36
Called from file "transformSugar.ml", line 345, characters 26-34
Called from file "transformSugar.ml", line 554, characters 24-36
Called from file "transformSugar.ml", line 350, characters 26-34
Called from file "transformSugar.ml", line 554, characters 24-36
Called from file "transformSugar.ml", line 339, characters 26-34
Called from file "transformSugar.ml", line 554, characters 24-36
Called from file "transformSugar.ml", line 345, characters 26-34
Called from file "transformSugar.ml", line 554, characters 24-36
Called from file "transformSugar.ml", line 315, characters 26-34
Called from file "transformSugar.ml", line 554, characters 24-36
Called from file "transformSugar.ml", line 607, characters 24-32
[...]

PR #133, opened 2017-01-23
Fix issue #132

dhil:
As it turns out this is an issue that has been reintroduced. It was initially resolved back in August 1, 2015 by commenting out FixTypeAbstractions in `frontend.ml`.


dhil:
Approved by myself.

PR #137, opened 2017-01-26
Fix #130.

### LINKED ISSUE #130: Make thread-specific global variables thread-local
There are a collection of global variables, primarily in lib.ml, that track bits of state specific to individual web requests.  This include things like the CGI parameters, cookies, response code, and so forth.  

In the current implementation of the app server, these are still global.  This means that concurrent web requests can cause races, overwrite each others cookies, and perform similarly nonsensical operations.

Lwt has a mechanism for thread-local state, using keys.  This mechanism is used in proc.ml to provide thread_ids.  It should be used for all the other thread-specific state as well.

SimonJF:
This PR fixes #130, ensuring that request-specific parameters (for example, cookies and HTTP response codes) are handled in a way specific to that request.

Before, these parameters were held as global variables in lib.ml, meaning that this data would be shared across different requests in webserver mode. Now, when initiating the request (Webif.serve_request for CGI; Webserver.callback for the webserver), a Lib.requestData record is constructed, which is passed as a parameter to the functions in EvalIR.

A side-effect of this patch is that each function in EvalIR now needs to take a Lib.requestData record as a parameter (since evaluating some primitive functions will need to get / set request-specific data). There may be a nicer way to do this which would require fewer changes? 

Tested with loginFlow in both webserver and CGI modes, fixing the wrong behaviour displayed pre-patch.

### DIFF
--- a/lib.ml
+++ b/lib.ml
@@ -18,13 +18,19 @@ let alias_env : Types.tycon_environment =
 
 let datatype = DesugarDatatypes.read ~aliases:alias_env
 
-let cgi_parameters = ref []
-let cookies = ref []
-(*  http_response_headers: this is state for the webif interface. I hope we can
-    find a better way for library functions to communicate with the web
-    interface. *)
-let http_response_headers = ref []
-let http_response_code = ref 200
+type requestData = {
+  cgi_parameters : (string * string) list ref;
+  cookies : (string * string) list ref;
+  http_response_headers : (string * string) list ref;
+  http_response_code : int ref
+}
+
+let empty_request_data () = {
+    cgi_parameters = ref [];
+    cookies = ref [];
+    http_response_headers = ref [];
+    http_response_code = ref 200
+  }
 
 (*
   assumption:
@@ -56,7 +62,7 @@ and row_values db = function
 
 type primitive =
 [ Value.t
-| `PFun of Value.t list -> Value.t ]
+| `PFun of Value.t list -> requestData -> Value.t ]
 
 type pure = PURE | IMPURE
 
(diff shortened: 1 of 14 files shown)


lib.ml
```diff
@@ -18,13 +18,19 @@ let alias_env : Types.tycon_environment =
...
 let datatype = DesugarDatatypes.read ~aliases:alias_env
 
-let cgi_parameters = ref []
-let cookies = ref []
-(*  http_response_headers: this is state for the webif interface. I hope we can
-    find a better way for library functions to communicate with the web
-    interface. *)
-let http_response_headers = ref []
-let http_response_code = ref 200
+type requestData = {
```
jamescheney: Nitpick: it might be more consistent to call this "request_data"

lib.ml
```diff
@@ -18,13 +18,19 @@ let alias_env : Types.tycon_environment =
...
+  cgi_parameters : (string * string) list ref;
+  cookies : (string * string) list ref;
+  http_response_headers : (string * string) list ref;
+  http_response_code : int ref
+}
+
+let empty_request_data () = {
```
jamescheney: Nitpick: This generates a new (empty) request_data, rather than defining a unique empty structure, so maybe call it "new_request_data"?

lib.ml
```diff
@@ -56,7 +62,7 @@ and row_values db = function
 
 type primitive =
 [ Value.t
-| `PFun of Value.t list -> Value.t ]
+| `PFun of Value.t list -> requestData -> Value.t ]
```
jamescheney: Nitpick: adding the request_data argument as the first argument might preserve more of the subsequent code (not a big deal though)

lib.ml
```diff
@@ -97,47 +103,70 @@ let string_to_xml : Value.t -> Value.t = function
...
+  (`PFun (fun args _ ->
       match args with
         | [c] -> (box_float (fn (unbox_float c)))
         | _ -> assert false),
    datatype "(Float) -> Float",
   pure)
 
 let p1 fn =
-  `PFun (fun args ->
+  `PFun (fun args _ ->
```
jamescheney: You could define p1D etc. first and then define p1, p2, p3 as special cases e.g. p1 fn = p1d (fun _ x -> fn)

SimonJF:
Thanks very much for the code review! Your comments are very helpful. I'm midway through adding them (and some other changes suggested by @slindley) but hopefully this should be good to go tomorrow.

SimonJF:
CI should be going now. I've addressed @jamescheney's comments, as well as incorporated `request_data` into `Value.env` as requested by @slindley.

Due to the introduction of a cyclic dep between `Lib` and `Value`, I've factored out the `request_data` structure (and appropriate accessor / mutators) into `requestData.ml{i}`.

If all's OK after, mind if this is squashed into one commit? The current commit history isn't hugely useful.

jamescheney:
Squashing to one commit seems reasonable to me.

SimonJF:
Thanks!

PR #141, opened 2017-02-02
Compatibility with 4.02.x compilers

dhil:
This patch lowers the dependency from OCaml 4.04 to OCaml 4.02.x (at least, I haven't tested with earlier compilers). The motivation for this patch is that the `multicore-compiler` branch depends on the Multicore OCaml compiler which is based upon the stock OCaml 4.02.2 compiler.

In this patch I simply define some utility functions which are otherwise available in OCaml 4.03+.

I have updated the `opam` file accordingly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ before_install:
   - sudo apt-get -qq update
   - sudo apt-get install -y ocaml-nox ocaml-native-compilers camlp4-extra opam
   - opam init -a
-  - opam switch 4.02.2
+  - opam switch 4.04
   - eval `opam config env`
   - psql -c 'create database links;' -U postgres
 install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,8 @@ before_install:
   - eval `opam config env`
   - psql -c 'create database links;' -U postgres
 install:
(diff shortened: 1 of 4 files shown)


SimonJF:
You might want to update the Travis file, too, which will keep us honest :)

dhil:
@SimonJF which file is the *Travis file*?

SimonJF:
@dhil: https://github.com/links-lang/links/blob/master/.travis.yml.

Now we're compatible with 4.02.3 again, you might actually get away with deleting the "opam switch" altogether, since I think the stock Ubuntu compiler from the PPA is 4.02.3.

dhil:
@SimonJF thanks!

dhil:
@fehrenbach would you happen to know why Travis cannot find `cohttp` although it has been installed?

fehrenbach:
I don't. My best guess at the moment is that it is something stupid with installing multiple things at the same time. Maybe opam builds lwt while cohttp is being configured, thus cohttp does not compile its optional lwt backend, or something? I'd try putting opam install -y cohttp on its own line, but I don't really expect it to help.

fehrenbach:
Wait, no, that doesn't make any sense. We already use cohttp.lwt in the master branch. Sorry. I have no idea.

dhil:
I have given up on the Travis configuration. It remains a mystery to me why it fails to fetch the dependencies.

PR #146, opened 2017-03-03
Distribution merge phase 2

SimonJF:
This commit adds everything except the distributed communication itself,
which will need to wait until it's completely threadsafe.

**Major changes**

*Client IDs*

Each client is associated with a unique ID when it connects to the
server. This ties in with the location system described below.

*Location system*

We introduce a new abstract type, Location. This encapsulates whether
something takes place either at the server, or at a particular client.
This comes in handy when thinking about spawning processes in realpages
mode.

Whereas before the type for `addRoute` was:

```sig addRoute: (String ~> Page) ~> ()```

we now have

```sig addRoute: ((String, Location) ~> Page) ~> ()```.

`Location`s can be used in the new `spawnAt` and `newAP` functions, for
example:

```
sig spawnSomethingAtClient : (Location) ~e~> Pid(e)
fun spawnSomethingAtClient(clientLoc) {
  spawnAt(clientLoc, { fun () -> () })
}

fun mainPage(_, clientLoc) {
  var pid = spawnSomethingAtClient(clientLoc);
  <#></#>
}

fun main() {
  addRoute("/", mainPage);
  servePages();
}
```

The location of the current place at which something is running can be
retrieved using `here`:

```
here : () ~> Location
```

*JSONisation*

JSON has been split into two parts: resolveJsonState and json.ml, and
json.ml has a new mli file.

The rationale for this is to separate the pure JSON functions and the
side-effecting JSON functions -- now, json.ml is pure, and thus doesn't
depend on proc.ml any more (hurrah!). This makes it a lot easier to, for
example, encode and decode websocket messages.

resolveJsonState handles the side-effecting operations, for example
popping all initial messages from a realpages-imitated mailbox prior to
initial delivery.

A major consequence of this is that there is no longer a need to include
a reference (PID) to a process spawned with `spawnClient` (or `spawnAt`
with a client location) -- this is done automatically from the list of
client-spawned processes.

*Clientside representation of sessions*

The clientside implementation of session-typed communication has been
rewritten to match the server implementation of session-typed
communication. This helps immensely when doing distribution.

*Representation / serialisation of access points / PIDs*
Access points and PIDs now also record whether they were created on the
server, or created on a client. Both are correctly serialised into, and
deserialised from, JSON.


SimonJF:
Beyond heartbroken to say it, but this merge got botched and reintroduced a ton of regressions. Closing for now.

PR #148, opened 2017-03-07
Pretty print values

fehrenbach:
Please test.

Travis failure is due to breaking of long lines in small terminals. Maybe line breaks should be disabled for tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ before_install:
   - eval `opam config env`
   - psql -c 'create database links;' -U postgres
 install:
-  - opam install -y deriving lwt postgresql sqlite3 mysql cgi base64 cohttp
+  - opam install -y deriving lwt postgresql sqlite3 mysql cgi base64 cohttp ANSITerminal
 script:
   - make -j2 nc
   - make tests
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 
 OCAMLMAKEFILE = ./OCamlMakefile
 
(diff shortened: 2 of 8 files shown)


jamescheney:
This looks helpful.  However, I'm a little wary of enabling indentation/pretty-printing for XML by default - in HTML, whitespace between "structural" tags is not significant but in XML, it can be.  

Would it make sense to enable this by default only when using the REPL, but disable in web mode / appserver mode?

fehrenbach:
Yes.

On:
- REPL

Off:
- Appserver
- CGI

Unsure:
- run file (links FILE) & exit
- run expression (links -e EXPR) & exit

I guess for the last two we could try to detect whether stdout is connected to a terminal and pretty print, or not, and not pretty print. Or just default to off.

dhil:
@fehrenbach do you still want this PR merged? If so would you kindly bring this PR in sync with the master branch?

jamescheney:
I think this was waiting on progress on some kind of testing framework, that we could use to ensure that the XML printing that happens in web mode is not affected by this change.

fehrenbach:
Ideally I would have liked for someone besides me to test this, but nobody came forward. I think I'll just merge it in. If it does break anything we can fix it, or revert.

jamescheney:
Did you mean to close (unmerged) this rather than merging/squashing?

jamescheney:
No, wait, I forgot that this was the old pull request, never mind.

Issue #153, opened 2017-03-28
Better default behavior for appserver?


jamescheney:
The appserver is great because it means we can set multiple entry points into an application. This is just what you want for complex applications, but for simple ones (e.g. the vast majority of our toy examples) it necessitates some "boilerplate" at the end of the Links program that looks like this:

```
fun mainPage(_) {
  page
    ... display the actual main page ...
}

fun main() {
  addRoute("",mainPage);
  servePages()
}

main()
```

This means in particular that the naive exercise of copying an "example program" off the Links demo programs page and running "links" on it yields unhelpful results along the lines of:
```
(0, fun, fun) : Page
```

Since I need to teach some people how to do simple things with Links tomorrow, this boilerplate saddens me.

One way this could be nicer is if it were possible to call addRoute/servePages from the toplevel, i.e. inline the call to main().  Another even nicer thing would be if running `linx` (possibly with some to-be-defined flag) on a program whose final action is to construct a page were to result in starting the appserver with that page as the main page.  Is there an obvious obstacle to this solution?

jgbm:
Some discussion resulted in suggesting that we could have a couple of convenience functions, such as:

| Function | Description |
|---|----|
| ```serveThis(page)``` | Starts the server, serving a single page |
| ```serveThese(pageList)``` | Starts the server with a list of routes and pages (or page-generating functions) |
|---|---|

These should be mostly one-liners in terms of the existing functions.

jamescheney:
yup, will do.

jamescheney:
Closed by #156.

Issue #152, opened 2017-03-13
rename built "links" executable?


jamescheney:
We currently build the "links" executable and rename it to "linx" during OPAM package installation only.  I suggest we change the executable name built by the makefile to "linx" for uniformity.

jstolarek:
I guess this is a duplicate of #72 :-) I'm in favour of renaming the executable but I'm not sure if other's agree. @slindley ?

jamescheney:
#72 was closed as a result of adding a line that renames "links" to "linx" when it is installed by OPAM.  This means that you have to do something different if you are using the OPAM package vs. if you are building from source.  My proposal is that we should change the name of the executable to "linx" when it is built rather than building it as "links" and then renaming.  I thought this is what was intended by "renaming" in #72.

jstolarek:
> My proposal is that we should change the name of the executable to "linx" when it is built rather than building it as "links" and then renaming.

That was my original proposal as well.

jamescheney:
We discussed this and the outcome was to keep the Makefile as is but add a symlink "linx" -> "links" in whatever location "links" is built.  (Perhaps also do this as part of make install.)  I will do this.

PR #156, opened 2017-04-26
Fixing #152 and #153

jamescheney:


jamescheney:
That sounds reasonable to me.

SimonJF:
Thanks!

PR #158, opened 2017-05-07
Fix #106.

### LINKED ISSUE #106: Error locations in other modules not reported correctly
Currently the chaser is using dummy_position when inserting modules from other files, which loses error location information. Instead, we (I) should generalise "position" to include a file name.

SimonJF:
(Sorry it took so long! A really embarrassing issue with a trivial fix!
I was thinking it would take a major refactoring; instead all that was
needed was the appropriate calls to be made to resolvePositions :))


dhil:
Looks good! This change is uncontroversial, but also unusual since it contributes more deletions than additions.

PR #164, opened 2017-05-10
Fix #139.

### LINKED ISSUE #139: JS Preamble not included when following links in realpages
This might be my lack of understanding of the post-realpages semantics (indeed, this sort of thing wasn't allowed before, AFAICT?) but it seems that when clicking a link, that the JS library and preamble isn't included.

```
fun sayHi() { debug("Hi!") }

fun page2() {
  page
    <html>
      <body>
        <div l:onload="{sayHi()}"></div>
        <h1>Hello!</h1>
      </body>
    </html>
}

fun mainPage(_) {
  page
    <html>
      <body>
        <a l:href="{page2()}">next</a>
      </body>
    </html>
}

fun main() {
  addRoute("/", mainPage);
  servePages()
}

main()
```

My understanding of what should happen: mainPage is generated on the server, user clicks "next" link, page2 is generated on the server, delivered to the client, "Hi!" appears in the debug console.

Currently: the main page is loaded with the preamble, but after clicking the "next" link to load page2, the page2 is just the plain HTML (without preamble).

I volunteer to fix, but I'd just like to make sure that it's A) allowed behaviour in the first place and B) an actual bug.

SimonJF:
We need to ensure that generate_real_client_page is also called for ServerConts, as well as invoked for EvalMain.

This patch adds a callback taking a Value of type Page, and passing it to generate_real_client_page after render_cont is invoked.

### DIFF
--- a/webif.ml
+++ b/webif.ml
@@ -149,7 +149,7 @@ struct
     let json_state = ResolveJsonState.add_ap_information client_id json_state in
     ResolveJsonState.add_process_information client_id json_state
 
-  let perform_request valenv run render_cont req =
+  let perform_request valenv run render_cont render_servercont_cont req =
     let req_data = Value.request_data valenv in
     let client_id = RequestData.get_client_id req_data in
     let client_id_str = ClientID.to_string client_id in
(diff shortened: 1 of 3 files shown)


SimonJF:
@elordin Does this work for you?

elordin:
From what I can see, it looks like it works.

SimonJF:
Jolly good! I'll wait for someone to have a look and then we can merge it in.

Issue #150, opened 2017-03-11
spawnWait is not implemented on the server


SimonJF:
There is no reason why `spawnWait` should only be a client function -- it makes semantic sense on the server, too. In fact, having it only as a client function leads to much confusion.

It shouldn't be a big job to write the server version of this.

PR #167, opened 2017-05-12
Implement spawnWait on the server

SimonJF:
This should address #150.


SimonJF:
The CI is failing because of an issue with the polymorphic Church numerals test, but I think this was broken before this patch.

slindley:
Yes. I've fixed that now.

PR #168, opened 2017-05-14
Fixes exception propogation

### LINKED ISSUE #97: Exceptions are rethrown twice
With PR #96 the Links shell no longer exits when an I/O error such as "file not found" occurs. However, when a I/O error occurs it gets rethrown twice. This is not a problem with the PR, but rather it exposes a cosmetic problem with the current code infrastructure. There are three identical error handlers, each sitting in a different layer in the code. It would be better to lift error handling such as "file not found" to the toplevel.

dhil:
Try this patch @slindley to see whether it fixes your error messages. It is a quick hack. I have basically removed some exception handlers, relying on the existence of a top-level "catch-them-all" handler. For presentational stuff this fix ought to be adequate, however in the long term I would like to fine-grain the exception handling such that every layer throws and handles a particular set of exceptions. 

Fixes #97

### DIFF
--- a/links.ml
+++ b/links.ml
@@ -68,7 +68,7 @@ let process_program ?(printer=print_value) (valenv, nenv, tyenv) (program, t) =
   lazy (process_program ~printer (valenv, nenv, tyenv) (program, t)) <|measure_as|> "process_program"
 
 (** Read Links source code, then optimise and run it. *)
-let evaluate ?(_handle_errors=Errors.display_fatal) parse (_, nenv, tyenv as envs) =
+let evaluate ?(handle_errors=Errors.display_fatal) parse (_, nenv, tyenv as envs) =
   let evaluate_inner x =
     let (program, t), (nenv', tyenv') = parse (nenv, tyenv) x in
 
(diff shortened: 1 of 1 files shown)


SimonJF:
I haven't yet got a minimal test case where exceptions are thrown twice. What do you mean by an "IO Exception"?

If you've got one knocking around so that I can make "variations on a theme", that'd be great!

Doing
```
links> hd([]);
*** Fatal error : hd() of empty list
links> 
```
doesn't quit the toplevel, as expected, before or after the patch.


However, I have uncovered a bug where an exception will still exit the toplevel, if an exception occurs in a spawned thread: 

```
links> spawn { haveMail() };
Fatal error: exception (Failure "The haveMail function is not implemented on the server yet")
```
This fails before and after this patch. Not sure whether this should be a separate issue.

dhil:
@SimonJF 

Try 
```
links> @load "this-file-does-not-exist.links"
```
or an evaluation error
```
links> switch (Foo) { }
```

These expressions used to generate exceptions which were being printed three times each.

SimonJF:
Works for me!

Issue #122, opened 2016-12-24
XML support is incomplete


slindley:
For instance:
   * Ill-formed XML can sometimes be accepted
   * CDATA literals are not accepted.
   * XML namespaces are not handled.
   * Non-ASCII characters are not handled.

elordin:
- Non-single-word element names

elordin:
There is actually a built-in
```TypeScript
xmlToVariant : (Xml) ~> mu b . [[|Attr:(String, String)|Node:(String, b)|Text:String|]]
```
though I could not find one the other way round

If there were a `variantToXml`, 
> - ability to construct/deconstruct XML
> - add / change the children (+ attributes) of a node.

could easily be implemented

PR #169, opened 2017-05-19
Add makeXml function to create arbitrary XML elements

elordin:
see #122

### DIFF
--- a/prelude.links
+++ b/prelude.links
@@ -456,6 +456,23 @@ fun asList(t) server {
   for (x <-- t) [x]
 }
 
+# Create an arbitrary XML element
+# @param {String} name Name of the element ([a-z]+)
+# @param {[(String, String)]} attributes List of attribute-value-pairs
+# @param {Xml} children Child XML nodes
+sig makeXml : (String, [(String, String)], Xml) ~> Xml
+fun makeXml (name, attributes, children) {
+  var attributes = fold_left(fun (prev, curr) {
+    var (attr, value) = curr;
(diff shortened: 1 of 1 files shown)


prelude.links
```diff
@@ -456,6 +456,23 @@ fun asList(t) server {
...
 
+# Create an arbitrary XML element
+# @param {String} name Name of the element ([a-z]+)
+# @param {[(String, String)]} attributes List of attribute-value-pairs
+# @param {Xml} children Child XML nodes
+sig makeXml : (String, [(String, String)], Xml) ~> Xml
+fun makeXml (name, attributes, children) {
+  var attributes = fold_left(fun (prev, curr) {
+    var (attr, value) = curr;
+    prev ^^ " " ^^ attr ^^ "=\"" ^^ value ^^ "\""
```
jamescheney: This has the potential for injection attacks if attributes containing quotes are passed in.  For example one could inject arbitrary XML content inside an element constructed using makeXML by providing an attribute with a value such as 

`xyz\"/> ... arbitrary XML content ...<foo key=\"`

which would give different results from what would happen if we were able to construct XML directly rather than having to go through the parser.

So it should only be used on strings that have been sanitized/checked (parseXML has the same problem, of course...)

jamescheney:
Happy to merge, if you could add a line pointing out that the string arguments aren't "escaped" and it is the caller's responsibility to ensure that any tainted input is sanitized before being passed to `makeXml`

prelude.links
elordin: Resolved in 3e16580

jamescheney:
Great, thanks.

PR #171, opened 2017-05-21
References

dhil:
This PR adds support for ML-style references to Links. The motivation is simply that both @SimonJF and I have expressed interests in having references in Links. In particular, I am interested in using references in the compiler benchmarks.

I have opted for an OCaml-like semantics for references. The implementation is rather simple, it adds a new abstract type `Ref(a::Type)` and three new library functions:

* `ref : (a) ~> Ref(a)` for creating references
* `deref : (Ref(a)) ~> a` for dereferencing
* `refmodify : (Ref(a),a) ~> ()` which assigns the provided value to the provided reference cell. In `prelude.links` I have defined a binary operator `:=` which is an alias of `refmodify`. 

The reference type can be thought of as `typename Ref(a::Type) = (contents:a)`. That is a reference is an one element tuple. Since the `Ref` type is abstract it is not possible to construct a reference by hand or pattern match on a reference.

Both the client and server side support references. In order to avoid running references on the database the three reference operations above all cause the effect context to have the `wild` effect. An alternative design would be to introduce another built-in effect, say, `ref` in order to have fine-grained tracking of references. However, down the line once the effect handlers (with operation polymorphism) implementation has been merged into the master branch I would like to have an effect type `Ref` with constructors `New`, `Modify`, and `Deref` whose default interpretation is in terms of raw reference cells, thereby allowing the programmer to supply its own local interpretation of references.

This is a early stage implementation of references; there are many design considerations yet to consider. Some questions are:

* What is the semantics of reference serialisation? Is it even meaningful?
* What is the semantics of sending a reference to another process?
* Should the `Ref` type be exposed in other to permit pattern-matching on references as in OCaml?

I do not think this PR is ready to be merged yet. I have opened this PR mainly such that we can discuss and converge on a design.

### DIFF
--- a/defaultAliases.ml
+++ b/defaultAliases.ml
@@ -15,5 +15,6 @@ let alias_env : Types.tycon_environment =
       "AP"      , `Abstract Types.access_point;
       "EndBang" , `Alias ([], Types.make_endbang_type);
       "Socket"  , `Abstract Types.socket;
-      "Location", `Abstract Types.spawn_location
+      "Location", `Abstract Types.spawn_location;
+      "Ref"     , `Abstract Types.reference_cell;
     ]
--- a/evalir.ml
+++ b/evalir.ml
@@ -3,7 +3,6 @@ open Ir
 open Lwt
 open Utility
 open Proc
-open ProcessTypes
 
 let lookup_fun = Tables.lookup Tables.fun_defs
 let find_fun = Tables.find Tables.fun_defs
--- a/examples/index.html
+++ b/examples/index.html
@@ -22,6 +22,10 @@
     <a href="progress.links">progress bar</a>
     (<a href="../examplessrc/progress.links">source</a>)
   </li>
+  <li>
+    <a href="refcells.links">reference cells</a>
+    (<a href="../examplessrc/refcells.links">source</a>)
+  </li>
   </ul>
   The following examples appeared in earlier drafts of the same paper
   <ul>
@@ -58,7 +62,7 @@
     (<a href="../examplessrc/validate.links">source</a>)
   </li>
   </ul>
-  
+
   Other examples
   <ul>
   <li>
--- a/examples/refcells.links
+++ b/examples/refcells.links
@@ -1,0 +1,50 @@
+sig += : (Ref(Int), Int) ~> ()
+op r += v client {
+  r := (deref(r) + v)
+}
+
+sig incr : (Ref(Int)) ~> ()
+fun incr(i) client {
+    i += 1
+}
+
+sig decr : (Ref(Int)) ~> ()
+fun decr(i) client {
+    i += (-1)
+}
+
+sig worker : () {hear:[|Decrement|Increment|]|_}~> ()
+fun worker() client {
+    var i = ref(0);
+    fun loop() {
+      switch(recv()) {
+        case Decrement -> decr(i)
+        case Increment -> incr(i)
+      };
+      var n = deref(i);
+      ignore(domSetPropertyFromRef(getNodeById("display"), "value", intToString(n)));
+      loop()
+    }
+    loop()
+}
+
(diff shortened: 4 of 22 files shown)


jamescheney:
I think this is a relatively large can of worms to open, unless the ways references are used are carefully restricted.  

There's a lot of work on transparent handling of persistent / shared data structures in distributed settings (i.e. I remember reading papers about this in graduate school).  This line of work was largely abandoned because of the high cost of ensuring consistency across different sites and difficulty of choosing a single policy that works well for a range of applications.

If the purpose of having references is to allow for efficient algorithms based on mutable state, but references are also expected to work "transparently" between the client and server, the overhead of monitoring and notifying other aliases of changes to a refernence could nullify the desired advantage.  (e.g. if the server creates a reference and it "leaks" to n clients, each of which modifies the reference concurrently, who wins? and what locking mechanism ensures some notion of "consistency"?)

In principle, transparently portable references are already "definable" in Links using processes (as in the pi-calculus): 

* "ref(s)" creates a process that accepts Get() and Set(s) messages
* when a Get() message is received the state is returned
* when a Set(s) message is received the state is updated

This would build on existing mechanisms for distribution rather than requiring new ones, or if locking/mutual exclusion is needed then that could be added to the interface.  It might also be good to have several flavors corresponding to different usage patterns, rather than having to decide on one fixed implementation strategy.

I think a proposal to create reference types that are server-side only or client/side only would be OK, though - but this might not handle the cases you or Simon had in mind.  So it would be good to understand what uses of references are most important (server only? client only? migration?).

SimonJF:
The reason I was thinking about references is that I'll soon be supervising an intern who will be implementing a declarative frontend library, possibly inspired by my (with Denuziere and Granicz) PADL'15 paper.

I need to give more thought to the best way of implementing this; i.e. whether it can be implemented in Links itself, or whether it'll need modifications to the various libraries. Nonetheless, we needed (F#) references, which were (IIRC) implemented on the client. For this use case, I'll only need client-side references.

My opinion is that references should be treated like processes: delivered upon page generation, but not serialised thereafter (other than via reference IDs, similar to PIDs). I think this is what the current PR seems to be proposing.

Nonetheless, adding mutability breaks actor abstractions. Care needs to be taken here. I think having an additional reference effect, which isn't serialisable, might help.

dhil:
The particular applications I have mind are process/thread schedulers which I would like to implement on top of (potentially) mutable queue structures.

> There's a lot of work on transparent handling of persistent / shared data structures in distributed settings [..] This line of work was largely abandoned because of the high cost of ensuring consistency across different sites and difficulty of choosing a single policy that works well for a range of applications.

I would like to keep the semantics simple and avoid all the issues that distribution entails.

jamescheney:
We discussed this and decided it can be merged if it's modified so that references are server-side only (for now), cannot be serialized, and activated by a flag (which is unset by default).

dhil:
I will bring this branch in sync shortly. But I want to merge this after my effect handlers extension has been merged.

dhil:
Pending PR #227.

jamescheney:
Is this planned for incoporation in Dalry?  If so we should tag it appropriately.

dhil:
> Is this planned for incoporation in Dalry? If so we should tag it appropriately.

Done!

dhil:
No longer on the critical path for Dalry. I will fix this after ICFP.

jamescheney:
Given that #240 shows how to accommodate arbitrary OCaml-native data structures/first order APIs (albeit with some work to be done to align with the module system and make the types work), is it worth thinking about doing references the same way, instead of adding to lib.ml?

dhil:
Yes, it makes sense to reconsider this PR. I think I'll merge references into my private branch for the experimental evaluation of handlers that I am doing instead.

PR #177, opened 2017-05-29
Use deps from opam file for CI build

fehrenbach:
I forgot to update the opam file with dependencies. That should not happen. This would avoid the issue entirely and we only need to maintain one list of dependencies. Let's see whether that works...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,8 @@ before_install:
   - opam upgrade -y
   - psql -c 'create database links;' -U postgres
 install:
-  - opam install -y --deps-only
+  - opam pin add links-dev . -n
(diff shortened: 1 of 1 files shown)


SimonJF:
Oooh! This would be nice. Let's see whether it works, or whether more sacrifices to the Travis gods are needed.

SimonJF:
Nice.

Can the pin be removed after the install, or would this break things? Would having the pin there already mean that subsequent installs error?

fehrenbach:
I have no idea!
Note: I use a different name (links-dev) and `install --deps-only`, so hopefully it would not break things.
I'll add an unpin thing, just in case..

PR #178, opened 2017-05-29
don't merge

fehrenbach:
does this actually use *this* opam file as opposed to whatever the released links has as dependencies/whatever is in the cache?


fehrenbach:
That seems to have worked.

PR #189, opened 2017-06-06
Improve JavaScript printing

elordin:
- Remove the debug popup
- Replace debug, print, and dump implementations to use `console`
- Add `show` implementation


SimonJF:
Having just tested, I'm in favour of this; the patch works; and the code is nice.

I say we run it by people at the Links meeting tomorrow but other than that I'm happy to merge.

PR #197, opened 2017-06-08
Add native readline functionality, removing the need for rlwrap (Fixes #51)

### LINKED ISSUE #51: Native readline wrapping / history / tab completion without rlwrap
rlwrap does the job, but it would be nice to have native readline functionality

SimonJF:
Uses the linenoise package and hooks everything together (which is a bit
of a faff with the Lexer).

This works (to my satisfaction) to be the same as rlwrap's version,
including histories, parsing of partial inputs, and handling of
interrupts such as ctrl-c and ctrl-d.

Other people might want to test to see if it is satisfactory for them
too.

(should be squashed due to initial commit)


SimonJF:
@elordin could you please let me know whether this works on the Windows subsystem thingy?

(Thanks! I appreciate you being my "does my code screw over the Windows users?" test subject!)

dhil:
I have encountered some problems with this patch:

1. `Ctrl-D` to end input does not work. Although, if the input buffer is empty, then `Ctrl-D` correctly exits the REPL. (Incidentally `rlwrap` does not support the former behaviour either).
2. Command history is not preserved between instances.

I find item 2 especially annoying as I often jump in and out of the REPL.

SimonJF:
Thanks for testing! I'll have a look.

dhil:
Thanks. 

Just for clarification: It is only preservation of command history that I think is worth fixing.

The `Ctrl+D` behaviour would make for a more unixy experience, but since `rlwrap` doesn't support it I suggest we do not bother supporting it either.

dhil:
I will look into this soon. Note to self: test whether `CTRL+R` works.

dhil:
Command history search (CTRL+R) does not work. Also, if I use rlwrap the Links repl starts panicking on every keystroke, e.g.
```
 _     _ __   _ _  __  ___
/ |   | |  \ | | |/ / / ._\
| |   | | , \| |   /  \  \
| |___| | |\ \ | |\ \ _\  \
|_____|_|_| \__|_| \_|____/
Welcome to Links version 0.6.1 (Gorgie)
links> ^[[*** Parse error: <stdin>:1
Unexpected character : 
  
[   ^
links> A^[[A^[[*** Parse error: <stdin>:1
Unexpected character : 
  A
[    ^
```

If we were to accept this PR in its current state then I think it should at least be possible to disable this "native-readline" feature via a flag or config (such that I can keep using rlwrap).

SimonJF:
I will add the ability to default back to the "no-readline" functionality, and will try my best to not have it interfere with rlwrap (which is rather worrying).

I'm going to throw in the towel on Ctrl-R though; it's not exported by Linenoise, and I have spent too long on this issue already to implement that functionality from scratch.

dhil:
Question: In light of our discussion, will anyone be using the proposed built-in readline feature? Or will everyone just keep running Links wrapped in rlwrap?

SimonJF:
I don't use Ctrl-R, and newcomers to Links probably won't, so a nonzero number>

SimonJF:
I couldn't replicate the parse error bug. Nonetheless, the most recent patch adds `native_readline`, which you can set to `false` if you want the old behaviour.

jamescheney:
Is this ready to go now? @dhil, does the flag option resolve your concern?

jamescheney:
Is the command line flag (to disable this and allow rlwrap) intended to happen before we make the new release?

jamescheney:
I have added the flag.  

@dhil, @slindley can you please check that "rlwrap linx -r" does what you expect / resolve your reviews?  I tried it and Ctrl-R seems to work.

@SimonJF, can you resolve the conflicts with the master branch that stop this from being resolved, and merge once @dhil and @slindley approve?  It looks like a fairly straightforward conflict due to your settings and #227 adding settings in the same place, but I don't want to screw it up.

SimonJF:
Superseded. See https://github.com/links-lang/links/pull/252

Issue #46, opened 2016-07-28
Make Xml a user-defined type constructor


slindley:
We could define Xml as a variant type.

PR #202, opened 2017-06-15
Improve XML support

elordin:
- Add XML construction from variant
- Improve `makeXml`
- Add `appendChild`
- Add `setAttribute`
- Add `prependChild`
- Add `replaceChildren`

See #122

### DIFF
--- a/lib.ml
+++ b/lib.ml
@@ -680,8 +680,26 @@ let env : (string * (located_primitive * Types.datatype * pure)) list = [
                                            | (`XML x) -> Value.value_of_xmlitem x
                                            | _ -> failwith "non-XML passed to xmlToVariant") xs)
                     | _ -> failwith "non-XML passed to xmlToVariant")),
-   datatype "(Xml) ~> mu n.[ [|Text:String | Attr:(String, String) | Node:(String, n) |] ]",
+  datatype "(Xml) ~> mu n.[ [|Text:String | Attr:(String, String) | Node:(String, n) |] ]",
+  IMPURE);
+
+  "xmlItemToVariant",
+  (`Server (p1 (fun v ->
+      match v with
+        | (`XML x) -> Value.value_of_xmlitem x
+        | _ -> failwith "non-XML passed to xmlItemToVariant")),
+    datatype "(XmlItem) ~> mu n. [|Text:String | Attr:(String, String) | Node:(String, [ n ]) |]",
+    IMPURE);
+
+  "variantToXml",
+  (`Server (p1 (fun v -> `List (List.map (fun i -> `XML (i)) (Value.xml_of_variants v)))),
+   datatype "(mu n.[ [|Text:String | Attr:(String, String) | Node:(String, n) |] ]) ~> Xml",
    IMPURE);
+
+  "variantToXmlItem",
+  (`Server (p1 (fun v -> `XML (Value.xmlitem_of_variant v))),
+    datatype "(mu n. [| Text: String | Attr: (String, String) | Node: (String, [ n ]) |]) ~> XmlItem",
+    IMPURE);
 
   "getTagName",
   (p1 (fun v ->
--- a/prelude.links
+++ b/prelude.links
@@ -463,18 +463,74 @@ fun asList(t) server {
 # @param {[(String, String)]} attributes List of attribute-value-pairs
 # @param {Xml} children Child XML nodes
 # @returns {Xml}
-sig makeXml : (String, [(String, String)], Xml) ~> Xml
-fun makeXml (name, attributes, children) {
-  var attributes = fold_left(fun (prev, curr) {
-    var (attr, value) = curr;
-    prev ^^ " " ^^ attr ^^ "=\"" ^^ value ^^ "\""
-  }, "", attributes);
-
-  var startTag = "<" ^^ name ^^ attributes ^^ ">";
-  var endTag = "</" ^^ name ^^ ">";
-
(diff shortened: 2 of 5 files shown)


prelude.links
```diff
@@ -463,19 +463,75 @@ fun asList(t) server {
...
 # @returns {Xml}
-sig makeXml : (String, [(String, String)], Xml) ~> Xml
-fun makeXml (name, attributes, children) {
-  var attributes = fold_left(fun (prev, curr) {
-    var (attr, value) = curr;
-    prev ^^ " " ^^ attr ^^ "=\"" ^^ value ^^ "\""
-  }, "", attributes);
+sig makeXml : (String, [(String, String)], Xml) ~> XmlItem
+fun makeXml (name, attrs, children) {
+  hd(variantToXml([ Node(name, map(fun (attr) { var (n, v) = attr; Attr(n, v) }, attrs) ++ xmlToVariant(children)) ]))
```
jamescheney: the mapped function could be simplified: `fun ((n,v)) { Attr(n, v) }`

prelude.links
```diff
@@ -463,19 +463,75 @@ fun asList(t) server {
...
+}
+
+# Appends a XML forest to an XmlItems children
+# @param {XmlItem} item
+# @param {Xml} child
+# @returns {XmlItem}
+sig appendChild : (XmlItem, Xml) ~> XmlItem
```
jamescheney: parenthetical note: if we defined XMLItem as the variant type then we could avoid the dynamic check by constraining the first argument type to be `[|Node:(String,Xml)|]`...

prelude.links
```diff
@@ -463,19 +463,75 @@ fun asList(t) server {
...
+# Replaces all children of an XmlItem
+# @param {XmlItem} item
+# @param {Xml} newChildren
+# @returns {XmlItem} item with new children
+sig replaceChildren : (XmlItem, Xml) ~> XmlItem
+fun replaceChildren (item, newChildren) {
+  switch (xmlToVariant([ item ])) {
+    case [ Node(n, c) ] -> hd(variantToXml([ Node(n, xmlToVariant(newChildren)) ]))
+    case _ -> error("Cannot replace children of " ^^ show(item))
+  }
```
jamescheney: These operations seem fine, but are they needed given that we already have direct update operations to the DOM tree?

tests/xml.tests
```diff
@@ -95,3 +95,8 @@ stdout : foo<a/><b>bar</b>fubar : Xml
 Labelization bug
 (escape cc in <form l:action="{exit (cc)}" />) ++ <p/>
 stdout : @.* : \[XmlItem\]
+
+XML-Variant-Conversion
+variantToXml(xmlToVariant(<div class="i-am-a-div"><ul><li/><li></li><li>A</li></ul></div>))
+stdout : <div class="i-am-a-div"><ul><li/><li/><li>A</li></ul></div> : Xml
+
```
jamescheney: Tests for the new append/setAttribute/replace operations?

lib.ml
```diff
@@ -683,6 +683,11 @@ let env : (string * (located_primitive * Types.datatype * pure)) list = [
    datatype "(Xml) ~> mu n.[ [|Text:String | Attr:(String, String) | Node:(String, n) |] ]",
    IMPURE);
 
+  "variantToXml",
+  (`Server (p1 (fun v -> `List (List.map (fun i -> `XML (i)) (Value.xml_of_variants v)))),
+   datatype "(mu n.[ [|Text:String | Attr:(String, String) | Node:(String, n) |] ]) ~> Xml",
```
jamescheney: Not a huge fan of having the attributes floating around in the XML element list, as opposed to attached to the Node (e.g. as a list of attribute/value paires, or better a finite map) but this is fine for now.
Oh... I see that this is just reflecting what Links does internally

value.ml
```diff
@@ -764,3 +764,16 @@ and value_of_xmlitem =
     | Text s -> `Variant ("Text", box_string s)
     | Attr (name, value) -> `Variant ("Attr", `Record [("1", box_string name); ("2", box_string value)])
     | Node (name, children) -> `Variant ("Node", `Record [("1", box_string name); ("2", value_of_xml children)])
+
+let rec xml_of_variants vs = match vs with
+  | (`List variant_items) -> List.map xmlitem_of_variant variant_items
+  | _ -> failwith "Cannot construct xml from variants"
+and xmlitem_of_variant =
```
jamescheney: exposing the xmlitem version of this as a Links-callable function would make it possible to avoid some oercions to/from singleton lists in the library calls...

prelude.links
elordin: Resolved in `3a5bbef`

tests/xml.tests
elordin: Resolved in `3a5bbef`

value.ml
elordin: Resolved in `3a5bbef`

lib.ml
elordin:
```
typename XmlItem = [| Text: String | Node: (String, Map(String, String), [ XmlItem ]) |]
```
would be a better alternative, but since that will most likely require changes in a few more place, I'd say it is out of scope for this PR.

jamescheney: OK with me (cf #46)

elordin:
I just realized there is a name clash between the builtin `replaceChildren` and the added one in `prelude.links`. 
Since I fixed that alongside adding some other things I'm going to close this PR and open a new one once those changes are finalized.

PR #204, opened 2017-06-15
Explicit "event" variable in jslib

SimonJF:
This is useful for two reasons. Firstly, window.event isn't supported in
Firefox, so this should add support back for some applications.
Secondly, this should fix errors with "getTime" trying to access an
undefined event, for some reason.

### DIFF
--- a/lib/js/jslib.js
+++ b/lib/js/jslib.js
@@ -1,3 +1,6 @@
+// Support events in firefox, which doesn't support window.event.
+var event = undefined;
+
 // set up optimised setZeroTimeout
 (function() {
 	var timeouts = [];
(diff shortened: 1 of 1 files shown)


dhil:
Which browsers have you tested this fix on?

SimonJF:
Chrome and Firefox. It might be worth someone testing it on Edge, too.

lib/js/jslib.js
```diff
@@ -10,9 +13,10 @@
 		window.postMessage(messageName, "*");
 	}
 
-	function handleMessage(event) {
-		if (event.source == window && event.data == messageName) {
-			event.stopPropagation();
+	function handleMessage(_event) {
+		if (_event.source == window && _event.data == messageName) {
+      event = _event;
```
dhil: Have you configured your editor to insert tabs rather than spaces?

SimonJF: Oops! No, I've always been a spaces guy. Will fix tomorrow.

SimonJF:
@elordin could you please check whether this works on Edge?

SimonJF:
Thanks for testing, everyone!

PR #212, opened 2017-06-24
Add int case for float unboxing (Fix #206)

### LINKED ISSUE #206: Soundness error due to JS'es treatment of round floats
```
[simon@bodygold ~]$ node
> - 1.0
-1
> -1.0
-1
> -(1.0)
-1
> -1.1
-1.1
```

Now, say we have a client program which JSONises `-1.0`. It will get parsed in and treated as an *integer* -- and we then lose type soundness (not great).

The only way I can see of getting around this (save for carrying around type information, which we won't want to do) is boxing floats on the client. Although the prospect does fill me with terror, given the performance hit that would be incurred.

SimonJF:
As per #206, floats can be incorrectly boxed as integers upon deserialisation from JSON. 

This patch adds an ``` `Int ``` case in `value.ml`'s `unbox_float` which rectifies this, and should be a cheap way to fix the issue.

### DIFF
--- a/json.ml
+++ b/json.ml
@@ -62,7 +62,7 @@ let jsonize_location : Ir.location -> string = function
   | `Native  -> "native"
   | `Unknown -> "unknown"
 
-let rec jsonize_value : Value.t -> json_string =
+let rec jsonize_value' : Value.t -> json_string =
   function
   | `PrimitiveFunction _
   | `Continuation _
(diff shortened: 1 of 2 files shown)


SimonJF:
I intended it to be permanent, since the JSON otherwise only gets printed on initial delivery and via RPCs, but I can separate this out into a separate PR if you like.

jamescheney:
No, that's fine, just curious (I have the bad habit of making debugging-message changes and then forgetting to remove them...)

Issue #188, opened 2017-06-06
Fix and document FFI


SimonJF:
The "alien" function will make interfacing with external libraries a lot nicer.

It currently doesn't work because of the requirement to have "extras.js" loaded. Other than that, I doubt it will be too difficult (at least for first-order functions).

It will be nice to integrate this with the module system, too, to have:

```
alien javascript module "test.js" Test {
  setTitle : (String) ~> ()
}

Test.setTitle "foo"
```

I will do this today.

SimonJF:
This is underway. Progress so far is alien module desugaring, and some grungy fixes to the backend. We're no longer getting "what on earth is this!?" errors in irtojs anymore, but I'm yet to do a bit of engineering to ensure correct interaction with the CPS compiler (essentially needing to generate `LINKS.kify` stubs).

There's still some work to do here though, and it's certainly not ready yet. Thanks to some temporary hacks on a fork, Jake's progress isn't blocked on this anymore. That said, I need it soon. It won't be ready before Dalry, most probably, but soon thereafter.

SimonJF:
https://github.com/links-lang/links/wiki/JavaScript-FFI

This has been merged and documented.

PR #213, opened 2017-06-25
Basic FFI (Address #188)

SimonJF:
This PR proposes a simple "foreign function interface" -- that is, a way to call native JavaScript functions from Links itself.

Concretely, say we have the following file called `test.js` in `lib/js`:

```
function _setTitle(s) {
  document.title = s;
}

function _alertBox(s) {
  alert(s);
}

var setTitle = LINKS.kify(_setTitle);
var alertBox = LINKS.kify(_alertBox);
```

(note, we adopt the same convention as in `jslib.js` -- we define a direct-style function prefixed with an underscore, and CPSify it using `kify`).

Now, consider `examples/ffi.links`. We can do the following:

```
alien javascript module "test.js" Test {
  setTitle : (String) ~%~> ();
  alertBox : (String) ~%~> ();
}

fun mainPage(_, _) {
  var _ = spawnClient {
    Test.setTitle("Hello!");
    Test.alertBox("Hola!")
  };

  page
    <html><body><h1>Hi!</h1></body></html>
}

fun main() {
  addRoute("/", mainPage);
  servePages()
}

main()
```

The definition `alien javascript module "test.js" Test` defines a module referring to the native JS file `test.js`, defining two functions `setTitle` and `alertBox`, and assigning them the appropriate types. These can then be used to set the title and display an alert box.

This works by changing some of the internals to fix problems with the old `alien` definitions which weren't extending the environment properly, but also due to a new desguaring pass defined in `desugarAlienModules.ml`. This is an incredibly simple pass:

```
alien javascript module "test.js" Test {
  setTitle : (String) ~> ()
  alertBox : (String) ~> ()
}
```

desugars to

```
module Test {
   alien javascript "test.js" setTitle : (String) ~> ()
   alien javascript "test.js" alertBox : (String) ~> ()
}
```

`alien` definitions can be used on their own, without modules. Note that plain `alien` definitions now also take an additional parameter detailing the associated file.

Comments welcome!

*Please squash when merging!*


elordin:
The `.js` files are required to be in `lib/js` ?

I think it makes more sense to give the absolute URL which will be  `<script src="` here `"></script>`:
Those files will usually be project-specific. Using `lib/js` is esentially forcing the user to maintain all of those globally.
Giving the URL instead allows maintaining the `js`-files in the project directory and serving them with `serverStatic` + keeping the option of globally sticking them in whatever is used as `jslibdir`

elordin:
Also, #188 says:
> Fix and **document** FFI

;)

elordin:
With the FFI and the whole convention of underscoring the direct-style functions and `kify`ing them:
What about constants? And Records containing functions?
e.g.
```JavaScript
const a = 12;
const foo = {
  bar: 5,
  baz: (a, b) => a + b,
};
```
```
alien javascript module "foo.js" Foo {
  a: Int
  foo: (
    bar: Int,
    baz: (Int, Int) ~> Int
  )
}
```

And do `alien` statements not require semicolons (inside and after)? If so, what is the reasoning behind what statements require them and which don't?

SimonJF:
@elordin:

I think your comment about the location of the JS files is a good one, and I'll implement it.

Wrt documentation, I fully intend to do so, as I have done for every larger feature I've written so far. I'll do it when the PR is finalised.

Wrt the "reverse CPS", I haven't thought about this yet (I haven't needed it). What would need to happen right now is for the functions to essentially be written in CPS style (i.e. kify'ed in the record). 

```
const a = 12;
const foo = {
  bar: 5,
  baz: LINKS.kify((a, b) => a + b),
};
```

The spirit of this at the moment is for wrapper code to be written, which in turn calls the "real" JS code.

dhil:
What is the motivation for the sugar `alien ... module ...`? Why not only support its desugared form?

SimonJF:
It makes sense to me to have each FFI file namespaced. It's only a small amount of code, and IMO it's a lot neater than having the duplicated information on each line.

On 29/08/17 17:26, Daniel Hillerström wrote:

What is the motivation for the sugar alien ... module ...? Why not only support its desugared form?

dhil:
To me, with the sugar it looks as if Links has two kinds of module systems. 

In order to avoid the name duplication, I would rather have
```
alien javascript "foo.js" {
  setTitle : (String) ~> ()
  alertBox : (String) ~> ()
}
```
and then have it desugared into
```
alien javascript "foo.js" setTitle : (String) ~> ()
alien javascript "foo.js" alertBox : (String) ~> ()
```
This way the FFI would be agnostic to the module system.

SimonJF:
I disagree somewhat -- I think the namespacing makes a lot of sense, since the functions are actually defined in a different file. If people didn't want them namespaced then the standard syntax is still there, albeit a bit more verbose. OTOH, with your suggestion, perhaps we could do

module Foo {
  alien javascript "foo.js" {
    setTitle : (String) ~> ()
    alertBox : (String) ~> ()
  }
}


Which is a bit more verbose but also a bit more general. I wouldn't be too averse to this, yet I still slightly prefer the original (although that might just be because I'm biased :) ).

Does anyone else have any strong opinions?

On 29/08/17 18:18, Daniel Hillerström wrote:

To me, with the sugar it looks as if Links has two kinds of module systems.

In order to avoid the name duplication, I would rather have

alien javascript "foo.js" {
  setTitle : (String) ~> ()
  alertBox : (String) ~> ()
}


and then have it desugared into

alien javascript "foo.js" setTitle : (String) ~> ()
alien javascript "foo.js" alertBox : (String) ~> ()


This way the FFI would be agnostic to the module system.

elordin:
I too consider namespacing a good idea. Being able to mix `alien` imports with Links-defined code in a module would be good though:
```
module Foo {
  alien javascript "foo.js" {
    setTitle : (String) ~> (); 
    alertBox : (String) ~> (); 
  }
  
  fun bar () { ... }
  var baz = 5;
}
```

because 
```
module MyCoolLib { ... }
alien javascript module "myCoolLib.js" MyCoolLibJS { ... }
```
looks weird.

SimonJF:
The more I think about this, the more I agree. More general is better! Thanks for the great suggestions!

I will go with

```
alien javascript "foo.js" {
  ...
}
```

which, by convention, will often but not always inhabit a module.

jamescheney:
+1 to orthogonality.  It might also be interesting to have something similar for OCaml-native FFIs (via dynlink libraries) and it would be nice to have a syntactically similar mechanism.  (But I think we still have work to do on hwo that should work before merging #240).

SimonJF:
...that was an entertaining merge :)

OK, so I have implemented @elordin's suggestion about allowing JS files in arbitrary locations (great suggestion!) and decoupled `alien` blocks from `module` blocks as per @dhil's (also great!) suggestion.

Let's see what the CI says, but other than that, please do take a look and see what you think :)

SimonJF:
I also seem to have messed up the Git merge (which also seems to reflect the 20k+ lines of algebraic effect code). Not quite sure what to do here, does someone with better git-fu than me have an idea? Maybe opening a fresh PR?

SimonJF:
I will open a fresh PR without the 22k+ additions which I accidentally pulled into this one.

PR #214, opened 2017-06-25
Add ability to have a "standard library"

SimonJF:
All this essentially does is adds a distinguished directory which is (by
default) included in the module path. The key idea is that we should be
able to import things from the standard library when we need them,
rather than having things in the prelude which are there all the time.

(also, have a basic binary tree library as a freebie...)


SimonJF:
(See PR #213)

Issue #192, opened 2017-06-08
Effect handlers extension


dhil:

SimonJF:
(Just as an aside, I find it hilarious that you've written two MSc theses and two papers on effect handlers in Links, opened an issue called "Effect handlers extension", and assigned *me!* to it!)

dhil:
> [...] and assigned me! to it!

I'll do all the heavy lifting, then you can take all the credit for resolving this "issue".

PR #216, opened 2017-06-27
Extend XML support; Clean up JS

elordin:
* XML 🔁  Variant type conversions
* Various XML util methods
* Name-spaced XML (currently by variant-type only, not via manual input + parser)
* Changed JS XML-representation from `type Xml = [ string, string, Object?, Array<Xml>? ]` to `{ type: string, tagName?: string, attrs?: Object, children?: Array<Xml>, text?: string }` (makes the code much more legible and profiling shows no clear performance benefit for either)
* Remove the YAHOO JS library
* Some general clean up in `jslib.js`


SimonJF:
Would it be possible for you to talk us through the main changes at the Links meeting on Monday? 

There's a lot in this PR at the moment (including breaking changes) and it would be good to get an overview of everything. In fact, I would probably suggest (if possible) breaking these into smaller commits, (removing the YAHOO.JS library being a single commit, for example -- by the way, thanks so much for doing that!!!)

elordin:
> Would it be possible for you to talk us through the main changes at the Links meeting on Monday? 

Absolutey

> (including breaking changes)

shouldn't be anything breaking

> I would probably suggest (if possible) breaking these into smaller commits

I started that but since I changed the JSON format of `Xml` to add namespace support, I had to touch all methods that use it in the `jslib` which are quite a few. 
So separating that from purely js-improvements is a lot of work.

examples/citations.links
```diff
@@ -31,11 +31,11 @@ fun header(title) {
 sig suggest : (String, String) ~> ()
 fun suggest(authorPrefix, paperPrefix) client {
   var (authorList, paperList) = completions(authorPrefix, paperPrefix);
-  replaceChildren(
+  domReplaceChildren(
```
SimonJF: Eventually we'll want to namespace these properly using some kind of "virtual module", I think. But this namespacing-by-prefixing is fine for now.

irtojs.ml
```diff
@@ -761,16 +761,14 @@ let rec generate_toplevel_bindings : Value.env -> Json.json_state -> venv -> Ir.
       state, venv, xs, f -<- g
 
 let script_tag body =
-  "<script type='text/javascript'><!--\n" ^ body ^ "\n--> </script>\n"
+  "<script type='text/javascript'><!-- 'use strict';\n" ^ body ^ "\n--> </script>\n"
```
SimonJF: :+1:

irtojs.ml
```diff
@@ -761,16 +761,14 @@ let rec generate_toplevel_bindings : Value.env -> Json.json_state -> venv -> Ir.
...
 
 let make_boiler_page ?(cgi_env=[]) ?(onload="") ?(body="") ?(html="") ?(head="") defs =
   let in_tag tag str = "<" ^ tag ^ ">\n" ^ str ^ "\n</" ^ tag ^ ">" in
   let debug_flag onoff = "\n    <script type='text/javascript'>var DEBUGGING=" ^
     string_of_bool onoff ^ ";</script>"
   in
-  let extLibs = ext_script_tag "regex.js"^"
-  "            ^ext_script_tag "yahoo/yahoo.js"^"
```
SimonJF: :heart: :heart: :heart:

lib.ml
```diff
@@ -671,7 +678,23 @@ let env : (string * (located_primitive * Types.datatype * pure)) list = [
   (`Client, datatype "(DomNode) ~> Bool",
   PURE);
 
-  (* Section: Accessors for XML *)
+  (* Create XML element from string, attributes and childre *)
```
SimonJF: children

lib.ml
```diff
@@ -680,35 +703,67 @@ let env : (string * (located_primitive * Types.datatype * pure)) list = [
                                            | (`XML x) -> Value.value_of_xmlitem x
                                            | _ -> failwith "non-XML passed to xmlToVariant") xs)
                     | _ -> failwith "non-XML passed to xmlToVariant")),
-   datatype "(Xml) ~> mu n.[ [|Text:String | Attr:(String, String) | Node:(String, n) |] ]",
+  datatype "(Xml) ~> mu n.[ [|Text:String | Attr:(String, String) | Node:(String, n) | NsAttr: (String, String, String) | NsNode: (String, String, n) |] ]",
+  IMPURE);
+
+  "xmlItemToVariant",
+  (`Server (p1 (fun v ->
```
SimonJF: Is there any reason that this only makes sense on the server?

I could imagine wanting to create XML on the client, without having to do a client call. If this is outside of the scope of the current PR then fair enough, but it would be good to have an issue requesting a client implementation as well.

lib.ml
```diff
@@ -1214,6 +1270,12 @@ let env : (string * (located_primitive * Types.datatype * pure)) list = [
     datatype ("(String) ~> Int "),
     PURE));
 
+  ("strcontains",
```
SimonJF: Seems like the convention is camelCase. Could this be changed to `strContains`?

lib/js/jslib.js
```diff
@@ -1761,31 +1524,26 @@ function _swapNodes(x, y) {
...
-//replaceChildren : xml -> domRef -> ()
-function _replaceChildren(xml, parent) {
+/**
+ * domReplaceChildren : xml -> domRef -> ()
+ * @param {Xml} xml
+ * @param {DomNode} parent
+ * @returns {()}
+ */
+function _domReplaceChildren(xml, parent) {
   newNodes = LINKS.XmlToDomNodes(xml);
```
SimonJF: In strict mode this must be `const newNodes` -- I got an error running the distributed chatserver as it is (but fixed it by adding `const` here)

lib/js/jslib.js
```diff
@@ -1761,31 +1524,26 @@ function _swapNodes(x, y) {
...
-//    parent.removeChild(parent.firstChild);
-//  }
+  parent.innerHTML = '';
 
-  for(i = 0; i < newNodes.length; i++)
-    _domAppendChildRef(newNodes[i], parent);
-
-  return {}
+  newNodes.forEach(function (node) { _domAppendChildRef(node, parent) });
+  return { };
```
SimonJF: `return CONSTANTS.UNIT`?

lib/js/jslib.js
elordin: Not necessarily `const`, `var` or `let` would do as well, but yeah, it needs to be declared.
That is a bit of a problem with JS. Unless the line is actually executed you don't know whether it is broken. So I might have overlooked some other instances of undeclared variables that were not executed yet. =/

SimonJF: Sure, you're right, my comment was over-specific on the type of binder needed.

Not catching them all straight away is OK -- but ideally we want the examples to work.

lib.ml
elordin: "convention" is disputable (see e.g. `fold_left`) but since the convention for JS is camelCase, that would be a good way to go

lib/js/jslib.js
elordin: Fixed in b9cade1

lib/js/jslib.js
elordin: Fixed in b9cade1

lib.ml
elordin: Fixed in b9cade1

lib.ml
elordin: Fixed on b9cade1

lib.ml
elordin: Fixed in b9cade1

elordin:
- `draggable.links` was broken node-equality due to incomplete `isUnit`
- `mandelbrot.links` threw errors since I threw one when `getNodeById` returned `null`
- `crop.links` was broken because I (partially accidentally) removed the `onXpage` events. I still find them a bit dodgy but I re-added them.

All three are fixed in b9cade1

SimonJF:
I think you mentioned over Slack that there are a few other things you wanted to clean up prior to me looking at this again. Let me know when it's good to go :)

dhil:
We should consider whether we merge this PR before or after PR #227. As merging either breaks the other.

dhil:
Since this PR removes the YAHOO library, I wonder how the following program behaves
```
fun err() { error("I'll throw; who'll catch?") }
fun mainPage(_,_) {
   page
   <html>
    <body>
     <form l:onsubmit="{err()}">
      <input type="submit"/>
     </form>
    </body>
   </html>
}
fun main() {
  addRoute("/", mainPage);
  servePages()
}
```
If I am not mistaken the YAHOO library handled errors in events.

jamescheney:
Since we agreed that #227 / #192 were on the critical path for v0.7 Dalry, and had not made such a decision for this PR, I suggest that we merge #227 and release the next version first, then consider this.  (Hopefully, modifying this PR to track the changes in #227 will be less work than changing #227 to match this one would be...)

Sorry @elordin !

elordin:
Not a problem. 

After a running a number of tools on `jslib.js` I found a number of additional fun things, e.g. calls to undefined functions. So I have some time to fix those as well.

I'm also trying out a few things to 
- statically check for such errors (mostly via `tsc` and `eslint`/`tslint`
- reduce the amount of global variables
- modularise it better
- reduce the amount of underscore-prefixes
- remove the need to kify everything

I probably will have something to report on that in the coming week.

From what I see in the @dhil s PR, the biggest change is that the list functions have been moved to the top of the file. So integrating those changes with what I have done and am doing isn't too much effort (relatively).

elordin:
@dhil Regarding that error example: 
It might be a matter of personal preference, but I would consider it undesired behaviour if the Yahoo lib were to catch that error. When I call `error` in Links, I would expect that error to occur. Having a lib catch that somewhere internally would mess with my expectations.

dhil:
> [...] but I would consider it undesired behaviour if the Yahoo lib were to catch that error.

I should have been clearer. What I meant is that the Yahoo library propagates errors raised during event handling (different execution context), ultimately re-raising them at the top level. I just wonder how error propagation works now.

elordin:
Closing this until for re-open once, as discussed, #227 is merged.

dhil:
This can be revisited now.

PR #218, opened 2017-07-05
Fix haveMail

SimonJF:
haveMail was operating on an incorrect PID. This fixes it (and un-breaks the games...)


elordin:
see also #216

SimonJF:
Ah, nice one!

I promise I will look at that PR in more detail soon, just swamped with other stuff at the moment...

PR #221, opened 2017-07-08
Remove nocrypto dep, add ANSITerminal dep to opam

SimonJF:


fehrenbach:
Did I not add ANSITerminal to the opam file? I'm confused.

SimonJF:
I did a clean install of Links the other day; seems it wasn't picking up that dep... I thought you did as well, so I was pretty baffled, but maybe you added it to one place and it needed to be another. Who knows!

PR #222, opened 2017-07-08
Treat carriage returns as whitespace

### LINKED ISSUE #220: DOS-style newlines not treated as whitespace
# Input

```
fun foo(x) {x+1}
foo(42)
```
where the newline after `}` is a DOS CR+LF newline.  (i.e., the result of `unix2dox` of the above program)

# Desired output

```
43 : Int
```

# Actual output

```
Unexpected character : 
  fun	foo(x) {x+1}
                   ^
```

SimonJF:
This should fix #220


jamescheney:
Looks good, how about adding a test case?

SimonJF:
Added :)

jamescheney:
Thanks!

PR #223, opened 2017-07-08
Emulate DOM reference serialisation (fix #207)

### LINKED ISSUE #207: Calls to client-side functions from server side
we call a client-side (side-effecting) function such as `replaceNode` from a function marked as `server`, the right thing does not appear to happen.  For example changing line 6 of [this](https://github.com/links-lang/links-tutorial/blob/master/3_form_client/form_client.links) to make handleForm run on the client, seems to lead to the replaceNode call silently failing.  I can see in the web dev console that there are two trampolined calls from the server to the client (to perform `getNodeById` and then `replaceNode`) and it looks like the http response in each case is appropriate, but nothing actually happens.  

Leaving the location unspecified, or specifying `client`, works.  But I would expect this to either work if `server` is specified or to get an error message indicating that client-side functions cannot be called from server-side contexts.

SimonJF:
This assigns a unique reference to serialised DOM nodes and stores them in
a dictionary, allowing them to be serialised and deserialised correctly.

This should fix #207.

### DIFF
--- a/json.ml
+++ b/json.ml
@@ -80,6 +80,7 @@ let rec jsonize_value' : Value.t -> json_string =
         ", \"environment\":" ^ s in
     "{\"func\":\"" ^ Js.var_name_var f ^ "\"," ^
     " \"location\":\"" ^ location ^ "\"" ^ env_string ^ "}"
+  | `ClientDomRef i -> "{\"_domRefKey\":" ^ (string_of_int i) ^ "}"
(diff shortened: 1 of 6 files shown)


jamescheney:
Looks reasonable to me, and seems to work on the example I ran into.  Thanks!

jamescheney:
I think we agreed to merge this but we should make a note of the fact that it may be fragile in the presence of multiple clients.  (Should that be a new issue?)

PR #224, opened 2017-07-08
Made addRoute take a single parameter again (Fix #199)

### LINKED ISSUE #199: addRoute API
In Links 0.6, addRoute had this type:

```
sig addRoute : (String, (String) ~> Page) ~> () 
```

but recently (4c358dc) it changed:

```
sig addRoute : (String, (String, Location) ~> Page) ~> ()
```

Since the shortcuts such as `serveThis` and `serveThese` haven't been introduced to a numbered Links version yet, this means that any code that we write for 0.6 has to change for 0.7 and vice versa.  This complicates various things such as the tutorial.

Proposed solution 1: Revert `addRoute` to the previous type, introduce a new function `addRouteWithLocation` with the new type.

Proposed solution 2: keep addRoute as is, add a new function `addSimpleRoute` with the new type

Proposed solution 3: define a RequestHandler type that is (for now) a variant with cases for simple handlers:
```
type RouteHandler = [|SimpleHandler: (String) ~> Page| Handler : (String,Location) ~> Page|]
sig addRoute : (String,RouteHandler) ~> ()
```

Proposed solution 4: something something row polymorphism
```
sig addRoute : (String,(path:String,location:Location) ~> ()
```

I lean toward solution 3, even though it is a breaking change w.r.t 0.6, since it seems more future-proof if we decide we want to change anything else about route handlers.

(in that case `serveThis` and `serveThese` should be modified so as to use the `RouteHandler` type...)

SimonJF:
As per the discussion in issue #199, revises addRoute to take a single parameter again.

Adds:
  - addRoute : (String, (String) ~> Page) ~> ()
  - addSimpleRouteHandler : (String, (String) ~> Page) ~> ()
  - addLocatedRouteHandler : (String, (String, Location) ~> Page) ~> ()

addRoute and addSimpleRouteHandler have the same behaviour.

addLocatedRouteHandler allows the page render function to be passed an
explicit location, for use in things like spawnAt.

### DIFF
--- a/examples/distribution/chatserver/chatServer.links
+++ b/examples/distribution/chatserver/chatServer.links
@@ -111,7 +111,7 @@ fun main() {
   var (clientAP:AP(ChatServer)) = new();
   var loopPid = spawn { serverLoop(Topic("Hello, ABCD!"), [], []) };
   var _ = spawn { acceptor(clientAP, loopPid) };
-  addRoute("/", fun (_, _) { ChatClient.mainPage(clientAP) } );
+  addRoute("/", fun(_) { ChatClient.mainPage(clientAP) } );
   addStaticRoute("/css", "css", [("css", "text/css")]);
   serveWebsockets();
   servePages()
--- a/examples/distribution/clientProcessOnServer.links
+++ b/examples/distribution/clientProcessOnServer.links
@@ -26,7 +26,7 @@ fun mainPage() {
 }
 
 fun main() {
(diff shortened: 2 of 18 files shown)


jamescheney:
I think we agreed that this was ready; any objection to merging?

dhil:
> I think we agreed that this was ready; any objection to merging?

The sooner, the better. I am implementing (and tidying up existing) web examples with PR #227. It would be good to have this interface available.

jamescheney:
Done!

PR #232, opened 2017-07-18
Add explicit error handling for routes.

jgbm:
Adds an error hander per route, called should the route throw an `EvaluationError` or `Wrong` exception, and updates the existing route adding functions to use a default error handler, providing an informative error message.

### DIFF
--- a/prelude.links
+++ b/prelude.links
@@ -1203,8 +1203,13 @@ sig addRoute : (String, (String) ~> Page) ~> ()
 fun addRoute(path, f) { addLocatedRouteHandler(path, fun(p, _){f(p)}) }
 
 sig defaultErrorPage : (String, String, Location) ~> Page
-fun defaultErrorPage(path, error_string, location) {
-  bodyP(<html><body><h1>Wrong</h1></body></html>)
+fun defaultErrorPage(_, error_string, _) {
+  bodyP(<html>
+        <body>
+        <h1>Wrong</h1>
+        <p>Links has encountered an internal error; the error message is <em>{stringToXml(error_string)}</em></p>
+        </body>
+        </html>)
 }
 
 sig addLocatedRouteHandler : (String, (String, Location) ~> Page) ~> ()
--- a/prelude.links
+++ b/prelude.links
@@ -1197,13 +1197,21 @@ fun assertEq (actual, expected) server {
 #### APP SERVER ####
 
 sig addSimpleRouteHandler : (String, (String) ~> Page) ~> ()
-fun addSimpleRouteHandler(path,f) { unsafeAddRoute(path, fun(p, _){f(p)}) }
+fun addSimpleRouteHandler(path,f) { addRoute(path,f) }
 
 sig addRoute : (String, (String) ~> Page) ~> ()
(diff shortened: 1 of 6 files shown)


prelude.links
```diff
@@ -1197,13 +1197,21 @@ fun assertEq (actual, expected) server {
...
+
+sig defaultErrorPage : (String, String, Location) ~> Page
+fun defaultErrorPage(path, error_string, location) {
+  bodyP(<html><body><h1>Wrong</h1></body></html>)
+}
 
 sig addLocatedRouteHandler : (String, (String, Location) ~> Page) ~> ()
-fun addLocatedRouteHandler(path,f) { unsafeAddRoute(path, fun(p, l){f(p, l)}) }
+fun addLocatedRouteHandler(path,f) { addLocatedRouteWithErrors(path, fun(p, l){f(p, l)}, defaultErrorPage) }
+
```
jamescheney: It's a little unfortunate that the only way to add a non-default (located) error handler is to use the located route handler interface...

prelude.links
```diff
@@ -1197,13 +1197,21 @@ fun assertEq (actual, expected) server {
...
+sig addLocatedRouteWithErrors : (String, (String, Location) ~> Page, (String, String, Location) ~> Page) ~> ()
```
jamescheney: Out of curiosity, is the Location passed into the normal result handler ever different from the one passed to the error handler?

prelude.links
```diff
@@ -1197,13 +1197,21 @@ fun assertEq (actual, expected) server {
...
 
 sig addSimpleRouteHandler : (String, (String) ~> Page) ~> ()
-fun addSimpleRouteHandler(path,f) { unsafeAddRoute(path, fun(p, _){f(p)}) }
+fun addSimpleRouteHandler(path,f) { addRoute(path,f) }
 
 sig addRoute : (String, (String) ~> Page) ~> ()
-fun addRoute(path, f) { unsafeAddRoute(path, fun(p, _){f(p)}) }
+fun addRoute(path, f) { addLocatedRouteHandler(path, fun(p, _){f(p)}) }
+
+sig defaultErrorPage : (String, String, Location) ~> Page
```
jamescheney: What about this slightly more useful (if less efficient) definition:

```
fun defaultErrorPage(_,error_string,_) {
  page
    <html>
    <body>
    <h1>Links internal error</h1>
    <em>{stringToXml(error_string)}</em>
    </body>
    </html>
}
```

jgbm: It turns out you can't use `page` sugar in the Prelude... but otherwise that seems fine (if somewhat less in the _spirit_ of Links, I would have thought....)

prelude.links
jgbm: It occurred to me when writing this that the number of ways to add routes was proliferating rapidly, and we might want to avoid a quadratic number of addXRouteWithY functions.  But I didn't have a good alternative approach in mind.

prelude.links
jgbm: Not that I can think of.

prelude.links
jamescheney: Fair enough, I don't have a better idea either.

PR #237, opened 2017-08-24
Disable dependency build caching on Travis

fehrenbach:


fehrenbach:
This doesn't solve the postgresql problem, but we might want to merge this anyway.

Currently Travis caches the ~/.opam directory. As a result builds are faster, but they may also rely on libraries that are not explicitly mentioned as dependencies, because they happen to be installed already. Also, versions are a bit unpredictable.

PR #238, opened 2017-08-28
dynload databases & Ocaml std libs

Yuren-Zhong:


jamescheney:
Thanks!

The Travis CI test appears to be failing because

```
cmnd=./links --config=tests/database/config --database=/home/harvey/dynlinx/pg_db/pg_database.cmxs tests/database/factorials.links
```
contains an absolute path to the pg_database shared library, which should be changed to a relative path (or to an absolute path that is valid in the VM created by Travis).  This seems like it may be easy to fix.

Makefile.config
```diff
@@ -0,0 +1,10 @@
+# Prefix for installation - default is /usr/local
+LINKS_PREFIX=/home/harvey/linx
```
jamescheney: This file should be removed - we only incldye the sample config file.

dhil:
Your branch appears to be out of date as this PR appears to rollback previous PRs. Please synchronise with the master branch.

dynload.ml
```diff
@@ -0,0 +1,56 @@
+
+(* dynamically load plugins.*)
+
+module StrM = Map.Make(String)
+module StrS = Set.Make(String) 
+module StrMM = MoreLabels.Map.Make(String)
+module StrMS  = MoreLabels.Set.Make(String)
+
+
```
jamescheney: Add a comment explaining that this is to reference all OCaml standard libraries so that they get included, hence we may need to modify this if the standard library changes.  (The OCaml -linkall flag might do this automagically instead, we should check.)

evalir.ml
```diff
@@ -423,11 +423,11 @@ struct
       unblock out2;
       apply_cont cont env (`Record [])
     (* end of session stuff *)
-    | `PrimitiveFunction ("unsafeAddRoute", _), [pathv; handler; error_handler] ->
```
jamescheney: This looks like it undoes a change that was made to Links since you created your branch, we should be careful not to undo such changes.

basicsettings.ml
```diff
@@ -153,10 +149,3 @@ let print_colors = Settings.add_bool ("print_colors", false, `User)
 
 (* Base URL for websocket connections *)
 let websocket_url = Settings.add_string("websocket_url", "/ws/", `User)
-
```
jamescheney: This looks like it undoes work done since you created your branch.

basicsettings.ml
```diff
@@ -24,36 +24,32 @@ let concurrent_server = Settings.add_bool ("concurrent_server", true, `System)
 (** Set this to [true] to print types when printing results. *)
 let printing_types = Settings.add_bool ("printing_types", true, `User)
 
-(* Looks for a given file, either in the current directory or in the Links opam path *)
```
jamescheney: I'm not sure of the reason for this change, is it undoing reorganization that was done concurrently with your branch?

json.ml
```diff
@@ -80,7 +80,6 @@ let rec jsonize_value' : Value.t -> json_string =
         ", \"environment\":" ^ s in
     "{\"func\":\"" ^ Js.var_name_var f ^ "\"," ^
     " \"location\":\"" ^ location ^ "\"" ^ env_string ^ "}"
-  | `ClientDomRef i -> "{\"_domRefKey\":" ^ (string_of_int i) ^ "}"
```
jamescheney: This looks like it undoes a recently merged fix.

PR #240, opened 2017-08-28
Dynamic loading database and Ocaml std lib

Yuren-Zhong:


lib.ml
```diff
@@ -1580,7 +1570,7 @@ let env : (string * (located_primitive * Types.datatype * pure)) list = [
      IMPURE);
     "unsafeAddRoute",
     (`PFun (fun _ -> assert false),
-     datatype "(String, (String, Location) ~> a, (String, String, Location) ~> a) ~> ()",
```
jamescheney: This looks like another regression (#232).

run-tests
```diff
@@ -74,7 +74,7 @@ run_database_testsuite () {
           t=${line[1]}
           expect_broken=1
         fi
-        cmnd="./links --config=$1/config $1/$t.links"
+        cmnd="./links --config=$1/config --database=/plugins/pg_db/pg_database.cmxs $1/$t.links"
```
jamescheney: This doesn't look right - it seems to me like it will look for the absolute path /plugins/pg_db/pg_database.cmxs rather than relative to the current path.  removing the initial "/" may fix that.

links.ml
```diff
@@ -532,6 +538,14 @@ let whole_program_caching_main () =
   in
    Webif.serve_request envs prelude filename
 
+let construct_plugin_name x = 
+  "/home/harvey/links/plugins/ocamllib/ocaml_" ^ x ^ ".cmxs"
```
jamescheney: Here is another absolute path dependency that needs to be removed.  Maybe use LIB_DIR as the starting point?

plugins/lite3_db/Makefile
```diff
@@ -0,0 +1,24 @@
+
+PATH := $(PATH):/home/harvey/links/deriving
```
jamescheney: Absolute path dependency, maybe this should be ../../deriving or something?

plugins/lite3_db/Makefile
```diff
@@ -0,0 +1,24 @@
...
+ifneq ($(SQLITE3_LIBDIR),)
+   DB_AUXLIBS += $(SQLITE3_LIBDIR)
+   DB_CLIBS   += sqlite3
+   DB_LIBS    += sqlite3
+endif
+
+OCAMLLDFLAGS = -shared
+
+LIBS    = $(DB_LIBS) 
+INCDIRS = $(DB_AUXLIBS) $(EXTRA_INCDIRS) /home/harvey/dynlinx/links
```
jamescheney: Another absolute path, there are lots in these Makefiles.  We may need to think of a better way to set this up so that the appropriate Links interfaces are available when compiling the plugins.

Yuren-Zhong: Maybe I should use LINK_PREFIX in the Makefile.config?

Yuren-Zhong: Since the plugins are compiled independently, it needs to know where to find necessary .cmx files.

dhil: The `LINKS_PREFIX` variable is only available during the compilation of Links. Rather, I think we would want a variable called something like `LINKS_LD_LIBRARY_PATH` (analogous to `OCAML_LD_LIBRARY_PATH`) which would follow the typical `PATH` conventions, i.e. list of directory names separated by colons.

plugins/lite3_db/Makefile
dhil: I think it should be `PATH := $(PATH):deriving`, the deriving preprocessor is in a direct subdirectory called "deriving" relative to the Makefile.

dhil:
I wonder what happens if one were to dynamically load a module whose name clashes with one used by the Links compiler, e.g. the `Utility` module? 

I suppose it might work fine for plain modules, but for functorised modules it may not.

jamescheney:
@dhil I think dynamically loaded code is never visible to already-loaded code.  So loading another module with the same name as an already-loaded one is not a problem, it is only visible to the already-loaded code.  (Just think what would have to happen to the already-loaded executable code in memory for anything else to happen.)  Instead, for the dynamically loaded code to affect the existing code, it is necessary for the loaded code to execute something that has visible side effects, such as adding data/function pointers to a reference or list ref (which is what happens here, in *_database.ml and other places).

jamescheney:
@Yuren-Zhong Thanks for making this branch / pull request and being responsive to comments/change requests.  I believe today or tomorrow is your last day here and we are probably not in a position to merge it into the main system yet, but having it as a branch that is close to being integrated with the master branch is helpful.  

In any case this is really cool!  I hope you had an enjoyable visit and I hope it will be OK if we can ask you a few questions later.

Yuren-Zhong:
@jamescheney I'm glad that it can be useful. It's been great staying here and I'm very happy to answer questions about what I did.

dhil:
Why don't we merge this into a new branch, say `dynamic-loading`, on this repository for now?

jamescheney:
Sounds good to me - I'll give it a try.

jamescheney:
OK, I tried changing the target branch to a new branch "dynamic-loading" from the UI, and strange things happened - lots of recent commits from "master" appeared and I'm not sure that's what we want, so maybe it would be better to merge this PR into that branch from the command line, or by creating a new PR.

jamescheney:
I want to try one mroe thing to see if I can make the tests run, then I'm giving up and merging.

jamescheney:
Hmm, OK, it builds again but can't find the pg_database shared lib, probably because it's not being built.  Probably means we need to build it in travis.yml. Which probably means it's time to merge this in as a branch and fix it, not keep making piecemeal commits through the web UI.

PR #246, opened 2017-09-06
Fix mysterious postgresql 4.1.0 build failure

fehrenbach:
Turns out we were trying to recompile the OPAM-installed version of postgresql. I still don't know why it started breaking now, but we shouldn't do that. See here for the full story: https://github.com/mmottl/postgresql-ocaml/issues/21

This PR assumes that we will only ever use the OPAM-installed version. This way we can use the regular mechanism for linking with OPAM-installed dependencies. We don't require the whole CLIBS, EXTRA_LIBDIR, and stuff anymore, I think.

I only tested the postgresql backend.

Someone who knows anything about Makefiles should check whether this makes sense.
Anyone who cares about MySQL/SQLite should check them.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,31 +7,21 @@ export OCAMLFLAGS=-syntax camlp4o
 
 PATH := $(PATH):deriving
 
-POSTGRESQL_LIBDIR=$(shell ocamlfind query postgresql)
-SQLITE3_LIBDIR=$(shell ocamlfind query sqlite3)
-MYSQL_LIBDIR=$(shell ocamlfind query mysql)
-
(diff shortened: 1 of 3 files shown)


dhil:
Good job @fehrenbach -- thank you!

jamescheney:
It would be good to merge this before making the new version (Dalry/0.7).  Is there something preventing this from being merged?

jamescheney:
Oh wait, I'm supposed to review it - sorry, that happened while I was at ICFP.

jamescheney:
I will merge this tomorrow unless anyone objects (or merges first).

PR #248, opened 2017-09-15
Extend XML support; Clean up JS

elordin:
- [x] XML <-> Variant type conversions
- [x] Various XML util methods
- [x] Name-spaced XML (currently by variant-type only, not via manual input + parser)
- [x] Remove the YAHOO library
- [x] Some general clean up in `jslib`
- [x] Add strict mode in `jslib`
- [x] Update js compiler to use strict mode


elordin:
Tested all (nodb) examples + chatserver. All work as intended.

elordin:
CI fails with
```
!FAILURE: XML Construction
  error: XML Construction
      (Unexpected stdout: expected `<foo id="bar" class="baz"><div/></foo> : XmlItem'; got `<foo id="bar"class="baz"><div/></foo> : XmlItem')
```
(`/tests/xml.tests`)

Test-case is
```
XML Construction 
makeXml("foo", [ ("id", "bar"), ("class", "baz") ], <div/>)
stdout : <foo id="bar" class="baz"><div/></foo> : XmlItem  
```
which, when I execute it manually, produces the correct output.

Can someone else please double check this ?!

fehrenbach:
That's probably my fault. I'll have a look.

elordin:
Don't think it's your fault. 
The error also occurs when running the tests locally.
But somehow the space between attributes is only missing in this specific test-case and only when running as a test-case.
Other tests are fine and executing that code manually also produces the correct output.

fehrenbach:
Okay, so this works on master.
In your branch it misses the space when `print_pretty` is set to true.

```
stefan@icarus:~/src/links xml-support$ ./links --config=pretty.config -e 'makeXml("foo", [ ("id", "bar"), ("class", "baz") ], <div/>)'
<foo id="bar" class="baz"><div/></foo> : XmlItem
stefan@icarus:~/src/links xml-support$ ./links --config=nopretty.config -e 'makeXml("foo", [ ("id", "bar"), ("class", "baz") ], <div/>)'<foo id="bar"class="baz"><div/></foo> : XmlItem
```

I don't have to time right now to dig through >6000 changed lines. If you still have a more fine-grained history around, you might want to try to bisect this.

There seems to be a newline missing, too. This suggests string_of_value is not working as intended. Another way is to force string_of_value use by using `show`:

```
stefan@icarus:~/src/links xml-support$ ./links --config=nopretty.config -e 'show(makeXml("foo", [ ("id", "bar"), ("class", "baz") ], <div/>))'
"<foo id="bar"class="baz"><div/></foo>" : String
stefan@icarus:~/src/links xml-support$ ./links --config=pretty.config -e 'show(makeXml("foo", [ ("id", "bar"), ("class", "baz") ], <div/>))'
"<foo id="bar"class="baz"><div/></foo>" : String
```

This also works in master, though...

fehrenbach:
One idea: do you call string_of_value on `Attr`s and concatenate?

dhil:
@elordin do the examples run with `enable_handlers = true`?

jamescheney:
If you test locally, make sure you delete prelude.links.cache.  Otherwise,
Links might happily load the old version of the prelude and use the oldv
ersion of makeXml, shadowing the buggy version.

jamescheney:
Hmm... seems to be a difference in how `XmlItem` vs. `Xml`s (== `[XmlItem]`s) are printed.

Also Links correctly parses XML without spaces between closing `"`s and the following attribute name.  Is that legal XML?

```
$ ./links -w
 _     _ __   _ _  __  ___
/ |   | |  \ | | |/ / / ._\
| |   | | , \| |   /  \  \
| |___| | |\ \ | |\ \ _\  \
|_____|_|_| \__|_| \_|____/
Welcome to Links version 0.7 (Dalry)
links> makeXml("foo", [ ("id", "bar"), ("class", "baz") ], <div/>);
<foo id="bar"class="baz"><div/></foo> : XmlItem

links> <foo id="bar"class="baz"><div/></foo>;
<foo id="bar" class="baz"><div/></foo> : Xml

links> [makeXml("foo", [ ("id", "bar"), ("class", "baz") ], <div/>)];
<foo id="bar" class="baz"><div/></foo> : [XmlItem]
```

elordin:
> Also Links correctly parses XML without spaces between closing "s and the following attribute name. Is that legal XML?

No, it is not.

What exactly is the `-w` flag for? Since without it there is no missing space between attributes.

jamescheney:
`-w` puts Links into web mode, which (along with some other complicated conditions) suffices to disable pretty-printing:

https://github.com/links-lang/links/blob/c6b76f55deb4110027dff837bf8a52c18e5002c2/links.ml#L18

elordin:
Just checked on `master`. Same behaviour.

So I'll just add the test case with `ignore` for now and defer fixing that as a bug.

With that this PR is ready for merging.

jamescheney:
The immediate reason for the regression is probably that changing the return type of `makeXml` from `Xml` (in prelude.links) to `XmlItem` (in lib.ml) unmasked a bug in XML pretty printing, but so far I can't see what the bug is.  Happy with merging.

dhil:
Can you confirm that you have tested the web examples with `enable_handlers=true`?

elordin:
> Web examples ~"chatserver" and~ "progress" are broken when enable_handlers=true.

FTFY. Progress-bar example works, chatserver still as broken as it is on `master`

dhil:
@elordin please synchronise the branch `xml-support` with `master`. Then I will try some handlers examples with your changes.

elordin:
Just for the record:
```
Failed to execute 'atob' on 'Window': The string to be decoded is not correctly encoded.
```
has, in all cases that I have seen so far, been a `500 Internal Server` error. So there should be an error on the servers `stderr` (which is the jsonparse error in this case because I only updated the way XML is encoded in one direction, m.c.)

PR #252, opened 2017-09-15
Native Readline, Round 2

SimonJF:
Again, I messed up a merge, so here is the (hopefully satisfactory) linenoise patch so that newcomers won't need rlwrap...


basicsettings.ml
```diff
@@ -256,8 +256,26 @@ module Unify = struct
...
 
-(* Standard library path *)
-let stdlib_path = Settings.add_string ("stdlib_path", "", `User)
+  (* Standard library path *)
+  let stdlib_path = Settings.add_string ("stdlib_path", "", `User)
+end
+
+module Readline = struct
+  (* Path for readline history file *)
+  let readline_history_setting = Settings.add_string("readline_history_path", "", `User)
```
dhil: This option may be unnecessary. It seems to be of little value. On the other hand, I don't mind it either.

PR #257, opened 2017-09-19
Fix #253

### LINKED ISSUE #253: Some impure programs never yield
Spawning a divergent computation like `fun f() { f() } f()` causes the spawned process to never give up control (i.e. yielding), meaning that a process ID is never returned
```
links> spawn { fun f() { f() } f() };
<loops indefinitely>
```
Because the divergent computation is not pure one would expect it to (eventually) yield
```
links> fun f() { f() };
f = fun : () ~> _::Any
```
Running an effectful function like `random` during computation causes the right behaviour
```
links> spawn { fun f() { ignore(random()); f() } f() };
Pid Server (srvPid_0) : Process ({ wild|_ })
```
Note the signature of the function is the same as before
```
links> fun f() { ignore(random()); f() };
f = fun : () ~> _::Any
```

SimonJF:
This should fix issue #253.

Turns out that there is only one place that apply doesn't call apply_cont, and that's when evaluating a computation.

All this patch does is to add a yield in this specific case.


dhil:
Immediate thought: Does this fix affect the semantics of pure computations? A pure computation never yields.

PR #258, opened 2017-09-19
Design things

elordin:
- [ ] ~Add the updated links-lang frontpage so the example server no longer goes "Nope" without `/examples`~
- [x] The example-page now loads the examples as sub-page, keeping the example-list
- [x] Source is opened in new a new window/tab

### DIFF
--- a/examples/css/doc.css
+++ b/examples/css/doc.css
@@ -1,0 +1,234 @@
+body {
+  margin: 0;
+  padding: 0;
+  font-family: 'Segoe UI', Tahoma, Geneva, Verdana, sans-serif;
+}
+
+a {
+  text-decoration: none;
+  color: #091F3F;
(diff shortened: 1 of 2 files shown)


SimonJF:
I think this is gorgeous, well done!

One minor quibble -- the examples appearing under the "The following examples appeared in earlier drafts of the same paper" heading on the examples page seem to show the source instead of the examples. Does that happen for you as well?

SimonJF:
Sorry, you're right of course! I hadn't noticed that they were DB ones...

elordin:
While editing those files I really should also remove the `./` from link-hrefs. They really don't belong there and lead to inconsistent behaviour.

elordin:
As per @slindley s request, the redesigned frontpage is now part of the [github.io page](https://links-lang.github.io) so examples returns **Nope** again.

PR #261, opened 2017-09-20
Tail call optimise jslib

### LINKED ISSUE #228: Tail call optimise jslib
Pull request #216 makes the Links client side code base compliant with JavaScript [strict mode](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode). As a result our code base gets to benefit from tail call optimisation (TCO) in Chrome (TCO is not supported by Firefox or Microsoft Edge; I think it is supported by default by Safari) which ought to improve the performance of Links client side code.

Now, it would be reasonable to expect that we get TCO for free once we turn on strict mode as the Links JavaScript compiler generates code in continuation passing style. However being JavaScript reasonable is unreasonable. JavaScript employs [an unusual definition of tail position](http://2ality.com/2015/06/tail-call-optimization.html#pitfall-solo-function-calls-are-never-in-tail-position). As an example consider this code
```javascript
function foo() { bar(); }
```
One might be inclined to think that `bar` is invoked in tail position. But it is not, because it is semantically equivalent to 
```javascript
function foo() { bar(); return; }
```
which in turn is equivalent to
```javascript
function foo() { bar(); return undefined; }
```
Thus `bar` must be marked by a `return` statement to be in tail position, i.e.
```javascript
function foo() { return bar(); }
```
This is unfortunate, because the vast majority of code in jslib has no explicit return statements. Therefore, I suggest to amend jslib such that we can take advantage of TCO in the (near) future.

dhil:
This PR inserts explicit return statements in `jslib.js` to expose opportunities for JavaScript engines to tail call optimise.

Fixes #228.

### DIFF
--- a/lib/js/jslib.js
+++ b/lib/js/jslib.js
@@ -6,7 +6,7 @@
 var lsNil = null;
 
 function _lsNilF() {
-	return null;
+    return null;
 }
 var lsNilF = _lsNilF;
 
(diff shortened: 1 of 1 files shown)


elordin:
I would fix the indents in the morning, if you don't mind, then I'm happy to merge.

SimonJF:
Is this going in before the release? I want to do the release by midday.

dhil:
I don't want this in the 0.7 release. I intend to merge after the release today. It will be part of an eventual 0.7.1 release.

SimonJF:
Righto, sounds good. Thanks for letting me know!

elordin:
Unless anyone is opposed, I'm going to merge this tomorrow with the note-to-self that there is still a lot of inconsistency in `jslib`.

Issue #249, opened 2017-09-15
Login flow example produces Type error


elordin:
When running the _login flow using sendSuspend_ example, upon clicking the "logout" link, the server registers a `Type error unboxing list: 0`

(Example works though, it's just a print on the server stdout/stderr)

SimonJF:
Weird. I'll take a look at this.

PR #262, opened 2017-09-20
Fix Issue #249

SimonJF:
Well, that was an adventure.


elordin:
Quality programming right there !

Issue #81, opened 2016-10-12
Move source files to their separate directory


jstolarek:
I suggest that we move all the source files into a separate `src` directory to keep the project structure clean.

There was a question yesterday whether this is something that should be done as part of the release or perhaps sometime later. I strongly believe this should be done before the stable release. Once we have an official release we don't want stable and development branches to diverge too quickly. Otherwise backporting bugfixes from development branch to stable branch will be problematic.

jamescheney:
We determined that this should be done after the appserver, modules and any bugfix merges are done but before creating the release.

jamescheney:
Delaying to Dalry.

dhil:
I think we should delay this until whether we decide whether to go with `jbuilder` as our build system, because `jbuilder` will naturally enforce/impose some directory structure.

jamescheney:
Fine with me.  I haven't had time to look at jbuilder yet.

jamescheney:
Anyone know what the next version name after Dalry is?  Or should we tag this "0.7.x"?  I think it would be good to do as a minor version bump since it should not affect the actual behavior of programs, just the build setup.

dhil:
I am happy to go with Dalry 0.7.2. We can reserve the next major version for Simon's sessions work :-)

jamescheney:
Now done by #265

Issue #79, opened 2016-10-03
Create .merlin file


jstolarek:
It might make our like easier if we configure Merlin - see @Drup's hints in #78 and tutorials: [here](https://github.com/the-lambda-church/merlin/wiki/project-configuration) and [here](https://opam.ocaml.org/blog/turn-your-editor-into-an-ocaml-ide/).

jstolarek:
Sam says in the past there were problems with setting up Merlin to work with deriving.

Drup:
Indeed. But `ppx_deriving` works fine, so time to switch :3

jstolarek:
Unassigning myself - not working actively on this in foreseeable future.

jamescheney:
This will be superseded by merging #265, since jbuilder supports generating merlin files.

jamescheney:
Now done by #265

PR #265, opened 2017-09-21
Switch build system to jbuilder

### LINKED ISSUE #62: Replacing OCamlMakefile with ocamlbuild?

dhil:
This pull request ports the Links build process to jbuilder from OCamlMakefile. It splits Links into five packages of which two are mandatory (in order to run Links) and the rest are optional. The two mandatory packages are

- `links-core` which exposes the Links source as a library.
- `links` which contains the executable.

The executable `links` depends on `links-core`. The optional packages are

- `links-postgresql`, a dynamic loadable database driver for PostgreSQL.
- `links-mysql`, a dynamic loadable database driver for MySQL.
- `links-sqlite3`, a dynamic loadable database driver for SQLite3.

Each of the optional database drivers depends on `links-core`. Currently, available database drivers are discovered and loaded automatically.

Resolves #62, #79, #81, #284, #302.


jamescheney:
Questions:

1. The travis tests fail because the postgresql driver isn't being built when Links is built.  I don't see anything in the jbuild file or Makefile that tells jbuilder to compile and link the pg_database.ml file if package postgresql is installed, but I don't see anything that tells it not to compile / link pg_database.ml either.  I guess it is analyzing the source code / dependencies and deciding that pg_database is not needed because nothing references it?  Could we fix this to insert an artificial dependency on each db driver file if the corresponding library is installed?

2.  Is there a convention for building / running unit tests (e.g. building a separate executable that runs a bunch of tests written using ounit)?

dhil:
1. jbuilder does not compile `pg_database.ml` because no module depends on. I have been looking into whether it is possible to dynamically select at build time whether to include postgres, SQL lite, or MySQL bindings. But I haven't found a satisfying solution yet. 

I actually think the right solution is to have a configure step in which one selects which database drivers to include. Then they could be static dependencies in the jbuilder file.

It should be easy to bring it up to date. I can try tomorrow.

2. Yes, unit tests would be a separate executable.

The "jbuilder way" would be to have three directories: `bin`, `lib`, and `test`, where `lib` contains the bulk of the source code, `bin` contains the source code for the main Links executable which will depend on `lib`, and `test` contains the unit test executable which will also depend on `lib`.

jamescheney:
Ok re 2.  Re 1, would that be like the good old days when you had to edit
Makefile.config to say what database drivers were present? I guess we'd
have to enable postgres by default to run the DB tests?

Maybe it'd be better to build the database drivers as separate libraries
and load them dynamically ala what Harvey did over the summer - you could
then (if understand correctly) define packages links-postgresql,
links-mysql etc to install the different drivers.

dhil:
Regarding dynamic loading: yes! That would be good. It would solve the issue.

Regarding the configure step: I was thinking about having an actual OCaml script called "configure" which generates the jbuild files.

jamescheney:
OK (see #284).  So maybe a sensible path would be:

- [ ] Define a config script that looks at the environment to determine what DB libraries are available, and constructs an approproaite jbuild file.
- [ ] Once #240 is done, get rid of the configure script and instead we provide several OPAM packages that provide the appropriate driver libraries.

(I still think OMake would be able to handle this, maybe I should make a competing pull request for discussion.  It looks like jbuilder may be superior in terms of integration with opam though.)

frank-emrich:
Is this supposed to build as it currently is? I wanted to base my experiments for dynamic loading of DB drivers on this.

 I tried make clean && make, but the links.opam file only contains "opam", which seems odd to me.

dhil:
> Is this supposed to build as it currently is? 

It should. Make sure you have a clean directory though. Then `make nc` ought to build Links.

frank-emrich:
I have now been able to look at the modularization for database drivers in greater detail.
(It turns out that my previous problems with building Daniel's branch were caused by some weirdness resulting from the combination of git, symbolic links and the Virtualbox files system... In essence, git thought it had created a symbolic link from "links.opam" to file "opam", but instead the filesystem just created a file named "links.opam" containing the word "opam". To git, however, it must have looked like everything was checked out correctly. So the build failed although git said that everything was checked out cleanly. That's one day of my life I won't get back...)

I see two options for modularizing the build process:
1. We create opam packages links-postgres, links-sqlite, etc, each of them containing a specific driver and each of them being able to be installed individually via opam.
The package links-postgres would then depend on the packages "links" and "postgresql".
At runtime, when creating a connection using the driver name "postgres", we then use ocamlfind (not the command line tool, but the ocaml library providing its functionality) to ask for the path of package links-postgres and load it.
I think this is what @jamescheney  described above.

2. I've played around with jbuilder and I've seen that it allows to specify certain libraries as optional.
Here, libraries are one level below packages w.r.t. modularization. In particular, one package (like "links") can internally be build from multiple libraries. Marking such a library as optional means that it is only built and installed if all of its dependencies are present.
We could then specify a library links.postgresql which would be part of the overall package "links" on opam. This library would only be built and installed if the package postgresql is present on the system.
This would have the advantage that there would only be a single "links" package on opam and all the magic would happen behind the scences. An additional advantage would be that building the package/project would be quite easy, because jbuilder does the detection about which database libraries to build and which not without any additional scripting, etc required from our side.

Note that right now, postgresql (and other DBs) are stated as optional dependencies of the "links" package. This has the following, very pleasant effect: Suppose a user has neither links nor postgresql installed on their system and issues "opam install links". Now, links is automatically built without postgresql support. If the user afterwards uses "opam install postgresql", the dependency management is clever enough to rebuild links after postgresql was installed. JBuilder would then recompile links, this time with support for postgresql.

I personally like the second option much better than the first one. It is much less "hackish" since it would utilize the mechanisms for optional dependencies offered by the build system, instead of some tricks using ocamlfind and dynamic linking at runtime.

jamescheney:
I think option 2 is essentially what we have now, only using Makefile hacks to conditionally compile support for Postgres or other DB drivers if present.  I like the way this works too and if there's a straightforward way of getting this behavior with jbuilder then let's just do that.

What I don't like about this (that potentially justifies dynamic linking) is that we would need to create a new Links version in the future whenever we want to support another database.  Right now the number of DB drivers available for ocaml is small and finite, but this might change.  So in the longer term something like option 1 would be preferable.  But I think we were only considering doing it now because it seemed just as much work as 2; if 2 is easier then let's just do that.  

If we eventually decide we do want dynamic linking for DB drivers too (#282) then I think this would be compatible with option 2: the optional links.postgresql library would then not necessarily require recompiling Links, and we could also add new such libraries (or allow users to create their own) without recompiling either.

frank-emrich:
I see your point about having to release a new version of Links every time we add a new DB driver. As a compromise, we could separate the drivers from the links package, i.e. release links and links-dbsupport on opam individually. The links-dbsupport package could comprise all database drivers and internally use what I proposed as option 2, i.e. conditional compilation of the drivers for the DBs found on the system. It would then suffice to release a new version of links-dbsupport every time we add/modify a driver and only this package would have to be recompiled, not the "base-package" links itself.

I didn't realize that the automatic re-compilation when installing a database was already happening. Then my porposed solution no. 2 is indeed just what we have right now, only integrated with jbuilder.

Implementing solution 1 isn't much additional effort since I implemented most of the more tricky parts already during my experimenting (in particular, determining the path where an opam package named links-myfancydatabase is installed locally and loading it dynamically when establishing a database connection). If you prefer, we could go entirely for dynamic linking and ditch the conditional compilation.
If we don't want to rely on opam only for installing database drivers, we could additionally provide a command-line argument to provide the paths where to load database drivers from. This part seems to be what Harvey did in his branch and I could re-use it.

jamescheney:
It looks like the tests need to be modified to ensure the postgresql driver is loaded, right now they fail with a message saying that postgresql driver is not loaded.  

I personally would prefer for the system to find these automatically the first time a driver is needed (or to be able to force loading using a command line option / config parameter) rather than always have to write `dload(links-postgres.cmxs")` in the code.  I think Frank said he has already experimented with that way of doing things.  Is that easy to accommodate or should we merge this and improve it later?

frank-emrich:
Yes, what I did was mostly finding out how to ask FindLib (the underlying library used by the ocamlfind tool) to query for the directory where an opam package (like links-postgresql) is installed. I already have code which does exactly that once a database connection is requested and there is no corresponding driver registered, yet.

What my current code does not handle is the problem of loading dependencies of the Links database drivers, that only worked with OCamlMake, not with jbuilder. I think Daniel proposed to create a small stub that knows about the dependencies, loads them first and only then loads the actual links database driver. Unless @dhil has already created code for that, I could do that and see if it works as expected.

dhil:
> I personally would prefer for the system to find these automatically the first time a driver is needed (or to be able to force loading using a command line option / config parameter) rather than always have to write dload(links-postgres.cmxs") in the code. I think Frank said he has already experimented with that way of doing things. Is that easy to accommodate or should we merge this and improve it later?

I agree, there should **not** be a Links function for loading code. What I have implemented at the moment, though, is a _directive_ `@dload` to load plugins. This can only be used in the REPL. A command line interface for `@dload` remains to be implemented.

>  I think Daniel proposed to create a small stub that knows about the dependencies, loads them first and only then loads the actual links database driver. Unless @dhil has already created code for that, I could do that and see if it works as expected.

Yes, I have thought about it, but I not yet implemented anything along those lines. Feel free to give it a go.

jamescheney:
Ah, sorry, I hadn't read the code carefully.  That makes a lot of sense.  

I think the simplest thing (enabling us to close / merge this) would be to provide a command line option / config parameter for db driver loading, e.g. 

```
linx --load_db_driver=postgres
```
or
```
#config file
load_db_driver=postgres
```
and hard-wire the dynamic loading needed for that.  This would be a little ugly (and still require some cmdline/config file changes, but not code changes) but if the fully-automated approach Frank outlines above turns out to be trickier than expected I don't want us to keep this PR open indefinitely.

dhil:
I think "hard-wiring" postgres (or mysql or sqlite3) as a special loadable plugin is fine for now. To stay both backward and forward compatible, we should probably make the dynamic load implicit. By that I mean having Links looking up whether the appropriate drivers are installed and then load them during start up. Once we have a proper interface for loading plugins we can disable this behaviour.

jamescheney:
This seems to be more or less done now.  I suggest that we merge #306 and any remaining bugfix PRs and then finalize this, then merge this and make sure everything still fits together, before creating the next version.  Does that make sense?

dhil:
Yes. I agree. Let's merge this last.

dhil:
If Travis is happy, then this PR is ready for squash'n'merge. @rudihorn it would be good if you would try out a small unit test sample.

dhil:
In light of janestreet/jbuilder#188 I wonder whether we should defer merging this PR until jbuilder (or whatever its new name will be) hits release 1.0.

jamescheney:
Ha, indeed.  I even had a copy of that back in the day.  

We could release "now" and fix with 0.7.3 along with the ocaml 4.0.6 stuff, or push this back to 0.7.3.  Aren't race conditions fun.

dhil:
OK. This PR is ready to go. I suggest we merge in on Monday.

jamescheney:
Sounds like a plan, does this mean @SimonJF's session exceptions will be merged after this or is that expected to happen sooner?

dhil:
I expect @SimonJF's PR to be merged in on Thursday.

dhil:
I have merged in @SimonJF's exceptional sessions. There are no remaining blockers on this PR. So we can merge this after you (@SimonJF and @jamescheney) have reviewed it.

SimonJF:
I'm happy to review; please note however that my review won't be an expert one since I have little experience with build systems (but this is a good opportunity to learn!).

Makefile
```diff
@@ -1,190 +1,25 @@
--include ./Makefile.config
+.PHONY: nc native clean tests install uninstall clean
+.DEFAULT_GOAL: nc
 
-OCAMLMAKEFILE = ./OCamlMakefile
+nc:
+	jbuilder build -j 4 @install
+	@echo "#!/bin/sh" > links
+	@echo "LINKS_LIB=\"$(shell pwd)/_build/default/bin\" LINKS_LD_LIBRARY_PATH=\"$(shell pwd)/_build/default/pg-driver\" $(shell pwd)/_build/default/bin/links.exe \"\$$@\"" >> links
```
SimonJF: This means that the JS path will be set incorrectly, unless Links is installed.

bin/dyn_db_hack.ml
```diff
@@ -0,0 +1,55 @@
+(* This module automatically discovers and loads database
```
SimonJF: Does this really need to be called `dyn_db_hack`? It may well be a hack, and that's fine (although it doesn't look too egregious to me) but I'd prefer a different filename.

dhil: Fixed (modulo the renaming of `dyn_db_hack.ml`).

It is a hack. Once a principal solution emerges this hack should be removed.

SimonJF:
I still can't get the JS library found if I use a custom configuration file, even when there are no JS library settings enabled. It works if I don't use a custom config file.

Can we look at this together again tomorrow?

SimonJF:
As requested, here is my config file:

```
debug=true
show_compiled_ir=false
use_cache=false
make_cache=false
modules=true
jsliburl=/lib/
base_url=/links/
show_json=false
enable_handlers=totally
database_driver=postgresql
database_args=:5432:simon:
js_pretty_print=true
session_exceptions=uncool
```

The same result happens when deleting jsliburl and base_url:

```
debug=true
show_compiled_ir=false
use_cache=false
make_cache=false
modules=true
show_json=false
enable_handlers=totally
database_driver=postgresql
database_args=:5432:simon:
js_pretty_print=true
session_exceptions=uncool
```

dhil:
@SimonJF what directory is Links looking for jslib.js in?

dhil:
@SimonJF, I am guessing it is failing on your end because you haven't specified `jslibdir`.

dhil:
@jamescheney the variable `$LINKS_LD_LIBRARY_PATH` need only be set when running the development version of `links`. It should all be backwards compatible. We can easily add a new CLI arg or setting, but I would prefer not to pollute that space any further, and leave the design space wide open for the future.

jamescheney:
OK, great.

dhil:
For good measure I better add the reason why `$LINKS_LD_LIBRARY_PATH` needs to be set for the development version of Links.

The resolution mechanism in used by the database binding loader is simple: it using `findlib` to query the installed libraries looking for `links-postgresql`, `links-mysql`, and `links-sqlite3`. How to load each of them is built in to the loader. This works well when Links is running in release mode, however, this is inconvenient when developing `links` as the database bindings might be modified which would only be loaded after during a full `opam install links-<database-bindings>`. Instead, it is possible to instruct the resolution mechanism to search non-standard paths via `$LINKS_LD_LIBRARY_PATH` (it is named after a corresponding variable used by the OCaml compiler).

SimonJF:
I can confirm that it works if I set `jslibdir`.

Given that it all works now, and is a vast simplification over what we had before, I'm eager to merge.
Are you happy with it being merged now, @dhil?

SimonJF:
Merged. Congrats and thanks, @dhil!

jamescheney:
Woo hoo!

PR #273, opened 2017-09-23
Add history manipulation to the examples page

elordin:
- When viewing on desktop, history is updated for the currently viewed example
- Loading an URL with query parameters loads the example in the frame
- On mobile it redirects to the example page

### DIFF
--- a/examples/index.html
+++ b/examples/index.html
@@ -7,6 +7,7 @@
     <link rel="stylesheet" type="text/css" href="/examples/css/doc.css" />
     <script>
       window.addEventListener('load', function () {
+        const examples = 'examples/';
         const frame = document.getElementById('exampleFrame');
         const nav = document.querySelector('nav');
 
@@ -15,23 +16,24 @@
         }
 
         function replaceExample (url) {
+          const exampleUrl = '/' + examples + url;
           if (isMobile()) {
-            location = url;
+            location = exampleUrl;
           } else {
-            frame.data = url;
+            frame.data = exampleUrl;
           }
         }
 
         // Set initial example
         if (location.hash) {
-          replaceExample('/' + location.hash.slice(1));
+          replaceExample(location.hash.slice(1));
         }
 
         // Intercept navigation
         window.addEventListener('hashchange', (e) => {
           e.preventDefault();
           if (location.hash) {
-            replaceExample('/' + location.hash.slice(1));
+            replaceExample(location.hash.slice(1));
           }
           history.replaceState({ }, '', location.hash);
         });
@@ -44,7 +46,7 @@
             a.addEventListener('click', function (e) {
               if (!isMobile()) {
                 e.preventDefault();
-                location.hash = a.getAttribute('href').slice(1);
+                location.hash = a.getAttribute('href').slice(1).replace(examples, "");
               }
             });
           }
--- a/examples/index.html
+++ b/examples/index.html
@@ -2,24 +2,53 @@
 <html>
   <head>
     <meta charset="utf-8">
-    <meta name=viewport content="width=device-width, initial-scale=1">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
     <title>Links examples</title>
     <link rel="stylesheet" type="text/css" href="/examples/css/doc.css" />
     <script>
       window.addEventListener('load', function () {
-        const iframe = document.getElementById('exampleFrame');
+        const frame = document.getElementById('exampleFrame');
         const nav = document.querySelector('nav');
-        document.querySelectorAll('nav a.frame-link').forEach(function (a) {
+
+        function isMobile () {
+          return !(nav.getBoundingClientRect().width < document.body.getBoundingClientRect().width / 2);
+        }
+
+        function replaceExample (url) {
+          if (isMobile()) {
+            location = url;
+          } else {
+            frame.data = url;
+          }
+        }
+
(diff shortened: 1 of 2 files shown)


examples/index.html
```diff
@@ -9,12 +9,30 @@
       window.addEventListener('load', function () {
         const iframe = document.getElementById('exampleFrame');
         const nav = document.querySelector('nav');
+
+        function isMobile () {
+          return !(nav.getBoundingClientRect().width < document.body.getBoundingClientRect().width / 2);
+        }
+
+        // Initial iframe src based on URL search parameter
+        const queryMatch = location.search.match(/example=[^&]+/g);
```
dhil: Could we potentially match on anchors? Say `www.links-lang.org/examples#dictSuggest.links`.

dhil:
(shouldn't viewport on line 5 be in enclosed in quotes?)

elordin:
> (shouldn't viewport on line 5 be in enclosed in quotes?)

Actually not necessarily. But yeah, for consistency it should.

I miss that all the time, because I have written that line only once in my life and copy-paste that from there ever since.

examples/index.html
elordin: for obivous reasons it can't just be the name of the Links file but must be the full path, but yeah, it now uses `location.hash`

dhil: I see. It is not behaving as I expected. Moving back and forth in the browser history do not update the displayed content. I suppose this is due to the `iframe`. Maybe what I really want is to rid of the `iframe`, and use a method similar to YouTube and Twitter for rendering.

dhil: `examples#/examples/dictionary/dictSuggest.links` do not read as nice as `examples#dictionary/dictSuggest.links`. So if we cannot get rid of the `/examples/` then I probably do not want to use anchors.

elordin: Twitter, YouTube etc. have the advantage of knowing what their content looks like.
Since our examples are pretty much arbitrary we have to use an `iframe` (unless we make an AJAX request, parse the HTML and construct the DOM tree locally, but really... no!).

It is possible to emulate some of what you want with `onpopstate` etc. but that will take a little longer than 5 minutes ;)

Regarding the slash: We can of course trim/append the leading slash but then we would make very implicit assumtions about the `href` attribute.

An alternative is to just set the URL to the example page. Disadvantage here is, that when you reload the page/share the URL you will end up on the page with _only_ the example and no side-navigation etc.

jamescheney:
I don't understand why this is tagged with milestone 0.7.1.  I tagged the two "critical" problems with OPAM / config file finding only, and intended that we make a new bugfix release as soon as those are fixed (which now seems to be the case).  Anything else should be under 0.7.2 or later because I want to have a usable 0.7.x release as soon as possible.

examples/index.html
dhil: > Regarding the slash: We can of course trim/append the leading slash but then we would make very implicit assumtions about the href attribute.

I think it is OK make such assumptions as this is a thing intrinsic to the examples page. Personally, I dislike urls with query parameters because they leak implementation details. The reason I like the anchor is that `links-lang.org/examples#progress.links` reads like record projection :-). The examples page is essential one big record.

> It is possible to emulate some of what you want with onpopstate etc. but that will take a little longer than 5 minutes ;)

We should really have navigation on the examples page working properly.

> Twitter, YouTube etc. have the advantage of knowing what their content looks like.
Since our examples are pretty much arbitrary we have to use an iframe (unless we make an AJAX request, parse the HTML and construct the DOM tree locally, but really... no!).

Thanks for the clarification.

dhil:
@jamescheney free feel to untag it. But navigation on the new examples page is broken.

examples/index.html
```diff
@@ -2,19 +2,36 @@
...
+
+        function isMobile () {
+          return !(nav.getBoundingClientRect().width < document.body.getBoundingClientRect().width / 2);
+        }
+
+        // Initial iframe src based on URL target
+        if (location.hash) {
+          const exampleUrl = location.hash.slice(1);
+          if (isMobile()) {
+            console.info('mobile');
```
elordin: No complaint about the blatant print-debugging here? ;)

examples/index.html
elordin: Turns out `iframe` and `history` don't go well together.

Iframe-loading adds an entry to the browser history (which is not editable).
Adding a hash (`loaction.hash = `, `replaceState`, `pushState`) adds an entry to the browser history.
So when we update both there are two new entries in the browser history for the single user-action. 
When the user then starts navigating via back and forward there every other state has hash and iframe out of sync.

dhil: > Adding a hash (loaction.hash =, replaceState, pushState) adds an entry to the browser history.

I take this would not be a problem with query parameters?

elordin: unfortunately no. Only real difference is that with hash I have the benefit of the `onhashchange` event to work with. Otherwise they are pretty much the same.

dhil:
Works like a charm! Thanks!

I removed the 'examples/' prefix from the url right of the anchor. Now it reads nicely.

jamescheney:
OK, I understand why you wanted to get this into 0.7.1.  Is this now ready for merging so that @SimonJF  can run through releasing 0.7.1?

elordin:
Tested in: Firefox, Opera, Vivaldi, Edge (and IE with graceful degradation).

PR #283, opened 2017-09-28
Add crypt hash and verification functions

elordin:
Whenever there is a login or something to build, password really shouldn't end up in the DB in plain text !

This PR adds
- [x] `crypt : (String) ~> String` which hashes a password so it can be safely stored in the database
- [x] `verify : (String, String) ~> Bool` which verifies that a password matches a hash

Please especially check whether there are changes missing for the added dependency.

(Also: What is the correct OCaml way to get rid of the warning `implicit elimination of optional arguments ?count, ?seed` ?)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 
 OCAMLMAKEFILE = ./OCamlMakefile
 
-PACKS=str deriving unix cgi base64 ANSITerminal linenoise cohttp lwt websocket websocket-lwt.cohttp
+PACKS=str deriving unix cgi base64 ANSITerminal linenoise cohttp lwt websocket websocket-lwt.cohttp safepass
 export OCAMLFLAGS=-syntax camlp4o
 
 PATH := $(PATH):deriving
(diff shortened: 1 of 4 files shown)


dhil:
What is the interface of the function yielding this warning? And how do you call it?

elordin:
See [the docs](
http://ocaml-safepass.forge.ocamlcore.org/apidoc/Bcrypt.html#VALhash)
```OCaml
val hash : ?count:int -> ?seed:string -> string -> hash
```
called with just the last string-argument (so `hash str`; [real code](https://github.com/links-lang/links/blob/07c4b73b3fcea35f409bb00bd41c3c65831478ee/lib.ml#L1665) looks a bit different for... reasons)

dhil:
The warning is caused by the composition. In such cases you should apply the labelled arguments explicitly. You can always use their default values, i.e.
```
... -<- (hash ?count:None ?seed:None) -<- ....
```

jamescheney:
Looks OK to me, modulo the fact that we should have a better way to add OCaml libraries/functions (#285).

PR #286, opened 2017-09-29
Base url setting

SimonJF:


SimonJF:
Incredibly irritatingly, this is still not quite enough to get Links examples properly working on the `frege` server due to the way it is set up -- requests are given relative to the initial `/links` prefix, but the JS is actually given as an absolute URL!

But this is useful anyway for a server with a non-bonkers setup (it's similar to something like Jekyll).

SimonJF:
There we go, that should do it.

PR #294, opened 2017-11-01
Fixed Issue 35 (use linked lists in JS)

frank-emrich:
One thing I'm not terribly happy about is the fact that I currently represent Nil as [] (i.e., the empty Javascript Array). This is mostly due to the fact that it's short and easy to parse. If we used something different, like a dedicated string or the empty object {}, it would be kind of hacky to parse this to the empty list in jsonparse.mly.

In addition, I could have squashed a few commits. In particular, what I did in 
bed80f3 was later mostly reverted. Do we pay a lot of attentition to squashing such useless commits?

### DIFF
--- a/lib/js/jslib.js
+++ b/lib/js/jslib.js
@@ -5,13 +5,12 @@ const LINKEDLIST =  new function(){
   
   return Object.freeze({
 
-    //Constant representing the Nil Object
-    //the JS constructor goes out of scope after this, so we can construct just one
-    //This allows us to see a difference between undefined/null and the empty list
-    Nil: "nil",//;new NilConstructor(),
+    //Using [] here for easier parsing (if we used e.g. a string, JSON parsing would become ambiguous)
+    Nil: [],
+
+    //Use this to check if something is Nil (using == or === may fail, since we have e.g. that  [] === [] is false in JavaScript)
+    isNil: function(n) {return n !== undefined && Array.isArray(n) && n.length === 0},
   
-    //Not a JS constructor, but a function calling it 
-    //(i.e., it behaves like one excepts when calling "Cons(x,xs)")
     Cons: function(x,xs) {DEBUG.assert(LINKEDLIST.isLinkedList(xs), "Second argument must be linked list"); return {_head: x, _tail: xs}},
 
 
@@ -20,7 +19,7 @@ const LINKEDLIST =  new function(){
      * but this would return false for objects created for deserialized objects
      */
     isLinkedList: function(v) {
-      return v === Nil 
+      return LINKEDLIST.isNil(v)
         || v.hasOwnProperty("_head") && v.hasOwnProperty("_tail");
         //could test list-property recursively, but seems overly expensive
         //&& LINKEDLIST.isLinkedList(v._tail) 
@@ -28,7 +27,7 @@ const LINKEDLIST =  new function(){
     },
 
     forEach: function(list, f) {
-      while(list != Nil) {
+      while(!LINKEDLIST.isNil(list)) {
         var cur = list._head;
         f(cur);
         list = list._tail;
@@ -38,7 +37,7 @@ const LINKEDLIST =  new function(){
     toArray: function(list) {
       var res = [];
       var cur = list;
-      while(cur !== Nil) {
+      while(!LINKEDLIST.isNil(cur)) {
         res.push(cur._head);
         cur = cur._tail;
       }
@@ -62,7 +61,7 @@ const LINKEDLIST =  new function(){
 
     take: function (n, xs) {
       var arr = [ ];
-      while (xs !== Nil && n > 0) {
+      while (!LINKEDLIST.isNil(xs) && n > 0) {
         arr.push(xs._head);
         xs = xs._tail;
         --n;
@@ -71,7 +70,7 @@ const LINKEDLIST =  new function(){
     },
 
     drop: function(n, xs) {
-      while (xs !== Nil && n > 0) {
+      while (!LINKEDLIST.isNil(xs) && n > 0) {
         xs = xs._tail;
         --n;
       }
@@ -80,7 +79,7 @@ const LINKEDLIST =  new function(){
 
     length: function (xs) {
       var out = 0;
-      while (xs !== Nil) {
+      while (!LINKEDLIST.isNil(xs)) {
         out += 1;
         xs = xs._tail;
       }
@@ -89,20 +88,20 @@ const LINKEDLIST =  new function(){
 
     head: function(v) {
       DEBUG.assert(TYPES.isLinkedList(v), "Not a linked list"); 
-      return v === Nil ? _error('head') : v._head; 
+      return LINKEDLIST.isNil(v) ? _error('head') : v._head; 
       }, 
 
     tail: function(v) {
       DEBUG.assert(TYPES.isLinkedList(v), "Not a linked list"); 
-      return v === Nil ? _error('tail') : v._tail; 
+      return LINKEDLIST.isNil(v) ? _error('tail') : v._tail; 
       },
 
     last: function(xs) {
-      if (xs === Nil) {
+      if (LINKEDLIST.isNil(xs)) {
         return _error('last');
       }
       var out = xs._head;
-      while (xs !== Nil) {
+      while (!LINKEDLIST.isNil(xs)) {
         out = xs._head;
         xs = xs._tail;
       }
@@ -111,7 +110,7 @@ const LINKEDLIST =  new function(){
 
       revAppend: function(xs, ys) {
         var out = ys;
-        while (xs !== Nil) {
+        while (!LINKEDLIST.isNil(xs)) {
           out = LINKEDLIST.Cons(head(xs), out);
           xs = LINKEDLIST.tail(xs);
         }
@@ -120,14 +119,14 @@ const LINKEDLIST =  new function(){
 
     append: function(xs, ys) {
       DEBUG.assert(TYPES.isLinkedList(xs) && TYPES.isLinkedList(ys), "Not linked lists"); 
-      if (xs === Nil) {
+      if (LINKEDLIST.isNil(xs)) {
         return ys;
       }
       const origxs = xs;
       const rootEl = _Cons(xs._head, Nil);
       let curr = rootEl;
       xs = xs._tail;
-      while (xs !== Nil) {
+      while (!LINKEDLIST.isNil(xs)) {
         curr._tail = _Cons(xs._head, Nil);
         xs = xs._tail;
         curr = curr._tail;
@@ -138,7 +137,7 @@ const LINKEDLIST =  new function(){
 
     at: function(xs, i) {
       var out;
-      while (xs !== Nil && i >= 0) {
+      while (!LINKEDLIST.isNil(xs) && i >= 0) {
         out = xs._head;
         xs = xs._tail;
         --i;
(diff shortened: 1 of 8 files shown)


jamescheney:
We normally do "squash and merge" or, if necessary, "rebase and merge" (or manually rebasing for those with sufficient git-fu, unlike me).   See:

https://github.com/links-lang/links/wiki/Working-conventions

dhil:
> One thing I'm not terribly happy about is the fact that I currently represent Nil as [] (i.e., the empty Javascript Array).

Why not use `null`? Using `[]` to denote the empty list is somewhat troublesome as `[] === []` and `[] == []` are both `false`, whereas `null === null` and `null == null` are both `true` in JavaScript.

Also using `[]` as nil means that (source level) homogeneous lists would be represented by heterogeneous lists at runtime which may lead to all kinds of strangeness if we are not careful.

frank-emrich:
@jamescheney I tested all the examples now and fixed the additional bugs I found. I also tested with --enable-handlers as suggested by Daniel. I found out that one example wasnt't working in the current master when --enable-handers is set, and created a separate issue for that.

@dhil null was previously used for Nil. I thought that may not be ideal since we then have to parse null to the empty list in jsonparse.mly, but then I realized that currently null is parsed to the empty record & there's a comment saying that raising an error may be better at this point. So I guess just sticking with null is okay, I changed it back to null now.

jamescheney:
This looks reasonable to me.  Maybe @slindley should look at it when he gets back too?  I seem to recall that the original issue claimed that using linked lists would avoid some bad performance associated with arrays, e.g. sorting becoming quadratic.  Has this been checked?

frank-emrich:
In general, calling tail() on a list involved copying the whole remainder in the old JS code. This is not the case any longer. But I don't know of any particular example I could use to see if there is a measurable performance benefit resulting from my changes.

Some of the runtime machinery in jslib.js still uses JS Arrays, but only for internal purposes. I am cautiously optimistic that I eliminated all occurences where such runtime stuff could "leak" into Links programs. This involved adapting some of the code that deals with XML representations of DOM nodes, but that should work now.

jamescheney:
It looks like this is ready to merge.  I'll do that Monday if no one objects.

lib/js/jslib.js
```diff
@@ -1,187 +1,255 @@
...
-  while (xs !== null && i >= 0) {
-    out = xs._head;
-    xs = xs._tail;
-    --i;
-  }
-  // FIXME should check i / index out of bounds here
-  return out;
-}
-const lsAt = _lsAt;
+    take: function (n, xs) {
```
dhil: 2) With the new (uniform) list representation I think this would be defined in the prelude rather as a built-in function. To the best of my knowledge the Links compiler does not depend on `take`.

lib/js/jslib.js
```diff
@@ -1,187 +1,255 @@
...
+        --n;
+      }
+      return LINKEDLIST.lsFromArray(arr);
+    },
 
-function _lsEmpty (xs) {
-	return xs === null;
-}
-const lsEmpty = _lsEmpty;
+    drop: function(n, xs) {
```
dhil: Replace `take` with `drop` in 2)

lib/js/jslib.js
```diff
@@ -1,187 +1,255 @@
...
-  var arr = [ ];
-  while (xs !== null && ys !== null) {
-    arr.push({ "1": xs._head, "2": ys._head }); // { ctor:"_Tuple2", _0:x, _1:y }
-    xs = xs._tail;
-    ys = ys._tail;
-  }
-  return _lsFromArray(arr);
-}
-const lsZip = _lsZip;
+    length: function (xs) {
```
dhil: As with 2) I think this can be user-defined now.

lib/js/jslib.js
```diff
@@ -1,187 +1,255 @@
...
-function _lsMinimum(xs) {
-  var currentMin = _lsHead(xs);
-  while (xs !== null) {
-    if (xs._head < currentMin) currentMin = xs._head;
-    xs = xs._tail;
-  }
-  return currentMin;
-}
-const lsMinimum = _lsMinimum;
+    last: function(xs) {
```
dhil: I think this can be user-defined as well.

lib/js/jslib.js
```diff
@@ -1,187 +1,255 @@
...
+      while (!LINKEDLIST.isNil(xs) && i >= 0) {
+        out = xs._head;
+        xs = xs._tail;
+        --i;
+      }
+      // FIXME should check i / index out of bounds here
+      return out;
+    },
+
+    empty: function(xs) {
```
dhil: Same as 2)

lib/js/jslib.js
```diff
@@ -1,187 +1,255 @@
...
+      while (!LINKEDLIST.isNil(xs)) {
+        curr._tail = _Cons(xs._head, Nil);
+        xs = xs._tail;
+        curr = curr._tail;
+      }
+      curr._tail = ys;
+      return rootEl;
+    },
+
+    at: function(xs, i) {
```
dhil: Same as 2)

lib/js/jslib.js
```diff
@@ -1,187 +1,255 @@
...
+      }
+      // FIXME should check i / index out of bounds here
+      return out;
+    },
+
+    empty: function(xs) {
+      return LINKEDLIST.isNil(xs);
+    },
+
+    zip: function(xs, ys) {
```
dhil: Same as 2)

lib/js/jslib.js
```diff
@@ -1,187 +1,255 @@
...
+      var lst = Nil;
+      if (a <= b) {
+        do {
+          lst = LINKEDLIST.Cons(b, lst);
+        } while (b-- > a);
+      }
+      return lst;
+    },
+
+    replicate: function(n, item) {
```
dhil: Same as 2)

lib/js/jslib.js
```diff
@@ -1,187 +1,255 @@
...
+      var arr = [ ];
+      while (!LINKEDLIST.isNil(xs) && !LINKEDLIST.isNil(ys)) {
+        arr.push({ "1": xs._head, "2": ys._head }); // { ctor:"_Tuple2", _0:x, _1:y }
+        xs = xs._tail;
+        ys = ys._tail;
+      }
+      return LINKEDLIST.lsFromArray(arr);
+    },
+
+    range:  function(a, b) {
```
dhil: Same like a specialised version of `intRange` which we have special sugar for in the source syntax, i.e. `[1...42]`. I wonder whether `range` is used anywhere?

lib/js/jslib.js
```diff
@@ -1,187 +1,255 @@
...
+      var out = Nil;
+      while (n > 0) {
+        out = LINKEDLIST.Cons(item, out);
+        --n;
+      }
+      return out;
+    },
+
+    // faster than _replicate? not really?
+    repeat: function(n, x) {
```
dhil: Same as 2)

lib/js/jslib.js
```diff
@@ -1,187 +1,255 @@
...
+      return LINKEDLIST.lsFromArray(arr);
+    },
+
+    // FIXME: max and min rely on '<' and '>', which
+  // may not do the right thing for non-primitive types
+  // (of course, we really want something like type classes
+  // in order to be able to handle this kind of situation
+  // more robustly)
+
+    min: function(xs) {
```
dhil: Same as 2)

lib/js/jslib.js
```diff
@@ -1,187 +1,255 @@
...
+      var currentMin = LINKEDLIST.head(xs);
+      while (!LINKEDLIST.isNil(xs)) {
+        if (xs._head < currentMin) currentMin = xs._head;
+        xs = xs._tail;
+      }
+      return currentMin;
+    },
+
+    max: function(xs) {
```
dhil: Same as 2)

lib/js/jslib.js
```diff
@@ -1269,6 +1359,12 @@ const LINKS = new (function() {
 
       return true;
     }
+    if (TYPES.isLinkedList(l) && l != null &&
+        TYPES.isLinkedList(r) && r != null) {
+      return LINKEDLIST.equal(l,r)
+
+    return true;
```
dhil: Dead code, this return statement can never be reached.

lib/js/jslib.js
```diff
@@ -1,187 +1,255 @@
...
-var lsNil = null;
+    //Using [] here for easier parsing (if we used e.g. a string, JSON parsing would become ambiguous)
+    Nil: null,
 
-function _lsNilF () { return null; }
-const lsNilF = _lsNilF;
+    //Use this to check if something is Nil (using == or === may fail, since we have e.g. that  [] === [] is false in JavaScript)
+    isNil: function(n) {return n === null},
+  
+    Cons: function(x,xs) {DEBUG.assert(LINKEDLIST.isLinkedList(xs), "Second argument must be linked list"); return {_head: x, _tail: xs}},
```
dhil: **Nitpick** why is `Cons` capitalised?

lib/js/jslib.js
```diff
@@ -1,187 +1,255 @@
 'use strict';
 
 /* LIST MANIPULATING FUNCTIONS */
+const LINKEDLIST =  new function(){
+  
+  return Object.freeze({
 
-var lsNil = null;
+    //Using [] here for easier parsing (if we used e.g. a string, JSON parsing would become ambiguous)
+    Nil: null,
```
dhil: **Nitpick** why is `Nil` capitalised?

lib/js/jslib.js
```diff
@@ -1,187 +1,255 @@
 'use strict';
 
 /* LIST MANIPULATING FUNCTIONS */
+const LINKEDLIST =  new function(){
```
dhil: Please avoid using `new function` as it suffers from potential security and performance issues. It is somewhat similar to be using `eval` to dynamically create and execute arbitrary code. Rather use 
```
const LINKEDLIST = (function() {
   return ....;
}());
```

lib/js/jslib.js
frank-emrich: It was like that before, I guess the original author wanted to mimick the "capitalize data constructors" rule found in languages like Haskell?

lib/js/jslib.js
frank-emrich: same as for Nil

jamescheney:
I'll defer merging until the requested changes are addressed.

lib/js/jslib.js
frank-emrich: I haven't found any occurences

frank-emrich:
I have pushed the requested changes

lib/js/jslib.js
slindley: I'm not sure if this is related, but some functions were capitalised at some point in order to ensure that there could never be a clash with user defined functions (which must always begin with lower case letters). There is a long outstanding bug with pattern matching compilation whereby one can witness strange behaviour by redefining hd and tl (collections.tests: Case patterns (with redefined hd)). It would make sense to address these name hygiene problems more systematically.

lib/js/jslib.js
dhil: Interesting, maybe it is in fact dead code.

jamescheney:
Is this now ready to go?  Let's discuss tomorrow.

frank-emrich:
The current status now fixes all of the issues raised by Daniel except the removal of the hard-coded list functions. I've created #298 for this.
This PR should now be ready to merge.

jamescheney:
@dhil, is there anything left in your code review that should be fixed at this point (i.e. not deferred to #298)

jamescheney:
Manually dismissing @dhil's review, will merge at the end of today.

jamescheney:
Merged.  Whew!  Thanks @frank-emrich !

PR #300, opened 2017-12-07
Exception Handling with Session Types

SimonJF:
This PR adds session exception handling, as described in the draft paper, "Session Types without Tiers" (http://simonjf.com/drafts/chomp-draft-nov17.pdf)

The key idea is to add three new constructs:

- `raise` raises an exception
- `try L as x in M otherwise N` evaluates `L`, binding the result to `x` in `M` if successful, and evaluating `N` if not
- `cancel M` allows consumption of a linear session channel endpoint

The system works by a notion of "channel cancellation" -- that is, channels that are either discarded explicitly through a use of `cancel` or implicitly through being in the aborted continuation when an exception is raised, are marked as cancelled. When their peer endpoint attempts to perform an action that wouldn't succeed (for example, receiving from an empty buffer, where the peer endpoint is offline), then an exception is raised.

As an example, consider:

```
fun go() {
  try {
    var t = fork (fun (s) { cancel(s) });
    var (res, _) = receive(t);
    res
  } as (x) in {
    "result: " ^^ intToString(x)
  } otherwise {
    "exception"
  }
}

go()
``` 

The send endpoint `s` is cancelled. Thus, receiving from its peer would block forever. However, with this extension, an exception would be raised and "exception" would be printed.

The extension is implemented both on the client and on the server.

The extension is off by default, but can be enabled by using the `--session-exceptions` flag, or setting `session_exceptions=true` in the config file.


fehrenbach:
Some good commit messages in here. 👍 
(No, I don't have anything constructive to add...)

jamescheney:
Looks like there are now some conflicts w.r.t. the recently merged PRs.  Also it seems that there are many ancient commits in this PR for some reason.  @SimonJF can you take a look and either resolve or re-make this PR?

SimonJF:
Will do.

PR #303, opened 2017-12-09
Fix issue 293

jamescheney:
Trivial changes to make routine REPL responses go to stdout not stderr.


jamescheney:
To get one of the builds to work I had to add an option to allow unauthenticated packages when installing.  I'm not sure it's necessary, may just have been a transient failure in the package install system.  Is this a bad idea?

fehrenbach:
I restarted the build and it looks like it was a transient failure. No idea why.

PR #304, opened 2017-12-10
Fix 280

SimonJF:
Ensure that `generalise` is called so that FTVs in FFI declarations are universally


SimonJF:
WTF is with this additional nonsense in all of my PRs!? Closing.

PR #305, opened 2017-12-10
Fix #280

SimonJF:
Ensure FFI annotations are generalised, so FTVs are universally quantified.

PR #306, opened 2017-12-10
Session Exceptions

SimonJF:
This PR adds session exception handling, as described in the draft paper, "Session Types without Tiers" (http://simonjf.com/drafts/chomp-draft-nov17.pdf)

The key idea is to add three new constructs:

  * `raise` raises an exception
  * `try L as x in M otherwise N` evaluates `L`, binding the result to `x` in `M` if successful, and evaluating `N` if not
  * `cancel M` allows consumption of a linear session channel endpoint

The system works by a notion of "channel cancellation" -- that is, channels that are either discarded explicitly through a use of cancel or implicitly through being in the aborted continuation when an exception is raised, are marked as cancelled. When their peer endpoint attempts to perform an action that wouldn't succeed (for example, receiving from an empty buffer, where the peer endpoint is offline), then an exception is raised.

As an example, consider:

```
fun go() {
  try {
    var t = fork (fun (s) { cancel(s) });
    var (res, _) = receive(t);
    res
  } as (x) in {
    "result: " ^^ intToString(x)
  } otherwise {
    "exception"
  }
}

go()
```

The send endpoint s is cancelled. Thus, receiving from its peer would block forever. However, with this extension, an exception would be raised and "exception" would be printed.

The extension is implemented both on the client and on the server.

The extension is off by default, but can be enabled by using the `--session-exceptions` flag, or setting `session_exceptions=true` in the config file.

### DIFF
--- a/sugartypes.ml
+++ b/sugartypes.ml
@@ -267,7 +267,7 @@ and phrasenode = [
 (* | `Fork             of binder * phrase *)
 | `CP               of cp_phrase
 | `TryInOtherwise   of (phrase * pattern * phrase * phrase * Types.datatype option)
-| `Raise (* TEMPORARY, FOR TESTING ONLY *)
+| `Raise
 ]
 and phrase = phrasenode * position
 and bindingnode = [


jamescheney:
I looked over the code and don't see any obvious problems but am far from an expert on what's going on.  @SimonJF, should someone else review this at this point or is it more or less ready to go?

Also, it would still be helpful to add to the wiki page on "Distribution", even if just to copy the above description into a section.  Pull requests are a non-obvious place to look for this kind of information.

SimonJF:
It would be good to get @dhil's input to sanity check it, but I'm happy with it going in at this point.

More than happy to add a wiki page (probably a separate one to distribution; while session exceptions do work in the distributed setting, explicit cancellation allows them to be used on a single VM too).

dhil:
I suggest we merge #305 and #307 first. I will review this PR after I have finalised those.

channelVarUtils.ml
```diff
@@ -0,0 +1,151 @@
+open Value
+open Utility
+open Pervasives (* PIPES *)
+
+(* TODO: Maybe it would be nice to have some kind of visitors for the IR?
```
dhil: We have visitors for the IR. The visitor type and base implementation are in `ir.{ml,mli}`.

evalir.ml
```diff
@@ -549,12 +593,12 @@ struct
...
         match value env db, value env name, value env keys, (TypeUtils.concrete_type readtype) with
           | `Database (db, params), name, keys, `Record row ->
-	      let unboxed_keys =
-		List.map
-		  (fun key ->
-		    List.map Value.unbox_string (Value.unbox_list key))
-		  (Value.unbox_list keys)
-	      in
+        let unboxed_keys =
+    List.map
```
dhil: Fix whitespace here

evalir.ml
```diff
@@ -570,33 +614,33 @@ struct
            | None -> computation env cont e
            | Some (db, p) ->
                begin
-		 if db#driver_name() <> "postgresql"
-		 then raise (Errors.Runtime_error "Only PostgreSQL database driver supports shredding");
-		 let get_fields t =
+     if db#driver_name() <> "postgresql"
```
dhil: Fix whitespace

evalir.ml
```diff
@@ -570,33 +614,33 @@ struct
...
                    | _ -> assert false
-		 in
+     in
                  let execute_shredded_raw (q, t) =
-		   Database.execute_select_result (get_fields t) q db, t in
-		 let raw_results =
-		   Queryshredding.Shred.pmap execute_shredded_raw p in
-		 let mapped_results =
-		   Queryshredding.Shred.pmap Queryshredding.Stitch.build_stitch_map raw_results in
+       Database.execute_select_result (get_fields t) q db, t in
```
dhil: Fix whitespace

evalir.ml
```diff
@@ -606,7 +650,7 @@ struct
                    []
                in
                apply_cont cont env (Database.execute_select fields q db)
-	 end
+   end
```
dhil: Fix whitespace

ir.mli
```diff
@@ -84,7 +84,8 @@ and special =
   | `Select of (name * value)
   | `Choice of (value * (binder * computation) name_map)
   | `Handle of handler
-  | `DoOperation of (name * value list * Types.datatype) ]
+  | `DoOperation of (name * value list * Types.datatype)
+  ]
```
dhil: This should not be dangling on its own line.

lib/js/jslib.js
```diff
@@ -3131,46 +3245,65 @@ function prepareDelegatedChannels(chans) {
...
 }
 
-function _send(v, c) {
-  var sendEP = c._sessEP1;
-  if (sendEP in _buffers) {
+function send(v, c, cancellation_thunk, kappa) {
+  const peer_ep = c._sessEP1;
+  const local_ep = c._sessEP2;
+
+  /* New semantics: don't raise exn on send
```
dhil: Remove this "code"?

value.ml
```diff
@@ -649,15 +670,39 @@ module Eff_Handler_Continuation = struct
...
+                   continuation_thunk = continuation_thunk
+                 }) else (Trap continuation_thunk)
              | None -> handle ((User_defined h, pk) :: k') k
              end
           | (identity, pk) :: k -> handle ((identity, pk) :: k') k
-          | [] -> E.error (Printf.sprintf "no suitable handler for operation %s has been installed." opname)
+          | [] ->
+              begin
+              if opname = session_exception_operation then
+                (* TODO: Is this correct??? *)
```
dhil: What are you trying to do here?

SimonJF:
@dhil -- thanks for the review! Due to a git cockup on my part, I need to close this PR and reopen it. But I'll address your comments before doing so.

PR #307, opened 2017-12-10
Fix #149

### LINKED ISSUE #149: Inner functions cannot have type signatures
For example, we get a parse error with the following:

```
fun foo() {
  sig bar : () -> String
  fun bar() {
    "Hello!"
  }
  bar()
}
foo()
```

SimonJF:
Seems that the simple parser fix seems to work.


jamescheney:
Looks reasonable/mergeable to me.  Any reason not to go ahead?

dhil:
I will review (and hopefully merge) this later today.

PR #311, opened 2017-12-12
* reverse backticks so that they equal pipe operations

### LINKED ISSUE #288: Proposal: Replace <| |> as infix "backtick" operators in OCaml source
`|>` and `<|` are common "pipe" idioms for reverse function application. Heavily used in F# and widely gaining adoption elsewhere (Elixir, for example), these operations allow one to chain operations in "pipelines" as so:

```
let example xs =
  List.map (fun x -> if Char.is_upper x then [x] else []) xs
  |> List.concat
```
instead of, say

```
let example xs =
  List.concat (List.map (fun x -> if Char.is_upper x then [x] else []) xs)
```

It's a nice way of writing code. Currently, however, these are shadowed as "backtick" notation for infix operators in `utility.ml`, in order to do things such as the following:

```
  let valenv, v = lazy (Eval.run_program valenv program) <|measure_as|> "run_program" in
  lazy (printer t v) <|measure_as|> "print";
  valenv, v
```

This means that one must explicitly do `let open Pervasives in` before a pipeline to shadow the backtick notation.

I propose we change this, so that we can use the idiomatic pipeline constructs. I don't really mind what we change them to, but I'd really like to have pipeline notation without having to explicitly import Pervasives every time.

jamescheney:
Fixes #288, I think; objections?

### DIFF
--- a/links.ml
+++ b/links.ml
@@ -85,12 +85,12 @@ let process_program ?(printer=print_value) (valenv, nenv, tyenv) (program, t) ex
   let (globals, _) = program in
   Webserver.init (valenv, nenv, tyenv) globals external_files;
 
-  let valenv, v = lazy (Eval.run_program valenv program) <|measure_as|> "run_program" in
-  lazy (printer t v) <|measure_as|> "print";
+  let valenv, v = lazy (Eval.run_program valenv program) |>measure_as<| "run_program" in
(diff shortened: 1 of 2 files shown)


SimonJF:
Strong accept from me.

fehrenbach:
LGTM

PR #313, opened 2017-12-14
Removed CGI & caching code (Issue 297)

### LINKED ISSUE #297: Remove caching and CGI (and other unused/legacy dependencies)
Caching appears no longer to be used, and can/should be unwound.

Likewise, old web mode and related dependencies can/should be removed.

frank-emrich:
The only thing that remains are a few parameters named cgi* being passed around (for example in webif, webserver), but it seems like they are still used.

Resolves #297.


SimonJF:
I would like to review this. I'll review this after the ABCD meeting taking Mon & Tues next week, unless there's a pressing need for it to be merged before then.

frank-emrich:
no, there's absolutely no urgency about this :)

PR #316, opened 2017-12-27
Revised jbuilder setup

dhil:
This PR merges the `links-core` package into the `links` package such that the latter package now exposes both the binary and the Links source (modulo `links.ml`) as a library.


dhil:
The main Links source is still contained within the subdirectory named `core`. I would like a better name for this directory. We can easily rename this later without affecting the (new) package `links`.

PR #318, opened 2018-01-22
Cohttp_lwt_body --> Cohttp_lwt.Body

SimonJF:
Should fix CI?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,11 +12,6 @@ before_install:
   - eval `opam config env`
   - opam install -y postgresql
   - psql -c 'create database links;' -U postgres
-after_failure:
-  - echo "in after_failure"
-  - cat /home/travis/.opam/4.04.0/build/links.0.7.1/*.env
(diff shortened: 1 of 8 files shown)


fehrenbach:
I'm still confused by the pp, it says 
```
ocamlc core/webserver.{cmo,cmt} (exit 2)
```
but reports the error as coming from core/webserver.pp.ml

But otherwise it looks like it compiles core/debug without problems and before it errors.

I'm out of ideas.

fehrenbach:
Eh? So the error message for ambiguity in an unrelated module is "Debug.print is not a value"? Is that something one is expected to know? WTF?! Does Lwt or something have a nested Debug module that one of the `open`s brings into scope and we end up looking up stuff in there instead of our global Debug module?

fehrenbach:
Actually, the last almost makes sense, but no warning? WTF?

SimonJF:
Yeah, bloody bonkers isn't it. Ah well. Everything's working again now!

This should mean that I FINALLY make the release tomorrow morning -- squashing and merging for now though.

PR #324, opened 2018-02-12
Ensure channels created on server are correctly sent to client (Fix #323)

### LINKED ISSUE #323: Channels created on server cause an error when used on the client
Realpages semantics mean that a function is run on the server when generating a page. If that function makes a channel, and the channel is delivered as part of the webpage, then an error will be thrown when the client attempts to use it.

```
fun mainPage(s) {
  spawnClient { var (res, _) = receive(s); print(res))
  page
    <html><h1>Hi!</h1></html>
}
addRoute("/", fun (_) { var s = fork(fun(t) { send("hello!", t)}); mainPage(s) } )
```

Will fail, for example. This means that `fork` is fairly useless in the web setting.

SimonJF:
This patch inspects all channels sent to the client, includes the buffers in the initial JSON dump, and updates the client locations.

This should fix #323.

### DIFF
--- a/core/evalir.ml
+++ b/core/evalir.ml
@@ -99,10 +99,11 @@ struct
          Some (Settings.get_value Basicsettings.websocket_url) else
          None in
      let st = List.fold_left
-       (fun st_acc arg -> ResolveJsonState.add_val_event_handlers arg st_acc)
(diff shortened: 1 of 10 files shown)


SimonJF:
Would anyone have any objections to me merging this?

PR #326, opened 2018-02-26
Macro-expand try-in-otherwise

SimonJF:
Currently, `try-in-otherwise` handlers may not contain linear variables in their continuations, ensuring the type-safe translation into linear effect handlers. However, this is quite restrictive.

This patch remedies this by inlining a simple macro-translation:

```
 [| try L as x in M otherwise N |]
    =
  switch (try [| L |] as x in Just(x) otherwise Nothing) {
     case Just(x) -> [| M |]
     case Nothing -> [| N |]
  }
```

Consequently, without touching the typechecker which still implements the more restrictive rule, we can now correctly and safely type `cancel10.links`.


core/frontend.ml
```diff
@@ -54,6 +53,7 @@ struct
       let _program = CheckXmlQuasiquotes.checker#program program in
       let () = DesugarSessionExceptions.settings_check program in
       ((( ExperimentalExtensions.check#program
+       ->- DesugarSessionExceptions.sanitise_linear_handlers
```
dhil: This should **technically** only be run whenever `session_exceptions` is enabled.

dhil:
Please rebase this PR on top of the master branch.

dhil:
Ping @SimonJF.

SimonJF:
Thanks for the ping. I intend to revive and fix up this PR on Friday.

Issue #124, opened 2016-12-24
The progress bar example fails when the entered number is 7937


slindley:
An error is thrown at 7937.

    Error handling ((headers
      ((accept "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8")
       (accept-encoding "gzip, deflate") (accept-language "en-US,en;q=0.5")
       (connection keep-alive) (content-length 53)
       (content-type application/x-www-form-urlencoded) (host localhost:8080)
       (referer http://localhost:8080/examples/progress.links)
       (user-agent
        "Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:50.0) Gecko/20100101 Firefox/50.0")))
     (meth POST) (resource /examples/progress.links) (version HTTP_1_1)
     (encoding (Fixed 53))): Stream.Failure

slindley:
I've tracked the problem down to the `Dump` feature of the `deriving` library. Changing the serialiser to use `Pickle` instead fixes this problem. Unfortunately `Pickle` leads to an error with the formlet examples (and possibly others).

yallop:
Do you have a standalone test case for the `Dump` problem?

slindley:
No. All I have at the moment is the progress bar example... but at least the error still occurs if I tweak it to start counting from 7936.

dhil:
On the `effects+sessions` branch this error is thrown whenever the value is greater than 127. I suspect the problem may be due to new continuation layout.
```
Error handling ((headers
  ((accept "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8")
   (accept-encoding "gzip, deflate") (accept-language "en-US,en;q=0.5")
   (connection keep-alive) (content-length 53)
   (content-type application/x-www-form-urlencoded) (host localhost:10088)
   (referer http://localhost:10088/examples/progress.links)
   (user-agent
    "Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:50.0) Gecko/20100101 Firefox/50.0")))
 (meth POST) (resource /examples/progress.links) (version HTTP_1_1)
 (encoding (Fixed 53))): Stream.Failure
```

dhil:
I confirmed my suspicion, using the fact that **generalised continuation** of this program has at most length 1 I hardcoded the compression/serialisation of this continuation:
```
type compressed_continuation = compressed_delim_continuation
and compressed_delim_continuation = (Ir.var * compressed_env) list
[...]

Issue #293, opened 2017-10-26
Links interactive shell outputting to stderr when it should be stdout


NickWu007:
Hi,

I was playing with Links and trying to make an online interactive shell for Links, and I realised when I input something like:

`var a = 2;`

I would get `a = 2 : Int` which is expected, but it is in `stderr` instead of `stdout`.

This is a bit strange to me since this message is not an error, more like a helpful reminder, which technically should be through `stdout`. If I type in some gibberish then the message would make sense to be in `stderr`.

Let me know if this makes sense.
Nick

NickWu007:
This would be a better illustration. Red background means it is coming from `stderr` and light grey is `stdout`.

<img width="781" alt="screen shot 2017-10-27 at 10 42 21" src="https://user-images.githubusercontent.com/17936144/32098180-c9cdfd86-bb03-11e7-89d0-f5a036b24b7f.png">

dhil:
I am not sure what the rationale (if any) is behind this decision. But the behaviour is caused by `prerr_endline` on line 294 in `links.ml` (c.f. https://github.com/links-lang/links/blob/master/links.ml#L294). 

If you want to change this now, I suggest you fork the repository and patch it yourself. Looking at the source, maybe line 267 will have to be changed as well (c.f. https://github.com/links-lang/links/blob/master/links.ml#L267)

NickWu007:
Hi Daniel,

Thanks for the explanation. I think it would make sense to change the lines you mentioned in the master branch as well.

Nick

jamescheney:
should be fixed by #303

PR #332, opened 2018-04-16
Migrate from deriving to ppx_deriving, allow building with 4.06.0 

### LINKED ISSUE #30: Replacing "deriving" with "ppx_deriving"
Links does not compile with the latest OCaml compiler (4.03), because the package "deriving" is not compatible with 4.03 (actually I think it is one of the dependencies of "deriving" that's causing the issue). 

The package "deriving" appears to be obsolete, and for future projects with Links it may very well be worthwhile replacing "deriving" with "ppx_deriving". My own project would benefit from this change as well as the OCaml Multicore backend will be released for 4.03+.

According to Sam switching from "deriving" to "ppx_deriving" is not an entirely mechanical process. Furthermore, the proposed change is likely to be thorough, and since it may benefit Links in general James suggested that we implement this change on the de facto master-branch "sessions". 

The question remains: _Which parts of Links are sensitive to this change?_

### LINKED ISSUE #299: OCaml 4.06 compatibility
Work is needed to make Links (and dependent libraries) compatible with OCaml 4.06  @dhil can give more detailed examples.

frank-emrich:
The follwoing contributions are in this PR:
C1) I migrated all uses of the old deriving preprocessor to the ppx_deriving and translated all customizations accordingly. I've replaced the Dump/Pickle serialisation of deriving by ppx_deriving_yojson (fixes #30)
C2) I wanted to see whether ocaml 4.06.0 improves on the problem with warnings explained below. Therefore, I've established compability with ocaml 4.06.0. To this end, I've patched the cgi library (which seems to be abandoned) and included the two files directly into our code. Since we want to get rid of cgi anyway (see #293, #313), this should be an acceptable interim soolution. I also fixed a minor problem that occurred when compiling with 4.06.0 (fixes #299)

Problems:
P1) If the build fails on your machine, this is probably caused by some bugs in jbuilder and ppx_deriving that were only fixed recently. I did "opam upgrade" after running into these problems, so I'm not sure which version exactly of these two fixed the issue. To be on the safe side, we could just require the current version of both packages in our opam file.

P2) Previously, deriving generated modules implementing the desired functionality, for example a module Show_foo if some type foo was annoted with deriving (Show). ppx_deriving generated functions instead, in the case of [@@deriving show] these would be show_foo : foo -> string and pp_foo : Format.formatter -> foo -> unit.
The former function behaves like expected and uses the latter function, which implements the actual formatting.

One problem arising from this is that while unused modules are okay, unused functions trigger warnings by the compiler. Thus, we are now seeing dozens of warnings about unused generated functions. This is not simply due to superfluos deriving annotations. For example, the pp_foo function above may be required by some other pp_ function, but the show_foo function may not be used at all, thus yielding a warning that cannot easily be fixed.

The reason why I switched to 4.06.0 is that it allows warnings to be disabled on a  more fine-grained level than previous versions. This is already used by ppx_deriving, which generates for example some [@ocaml.warning "-39"] attributes, which disable warnings about unneeded "rec". However, the fact that we still see so many warnings shows that they are not generated in all places where they should be. There seem to be some github  issues in the repo of ppx_deriving about this.

Since having a big amount of warnings is undesirable, I suggest we implement one of the following solutions:
S1) We could help fixing the actual problems in ppx_deriving. This seems to be non-trivial since ppx_deriving uses the meta-programming facilities of OCaml and the code isn't exactly understandable at first sight.
S2) We could add the necessary attibutes to disable the warnings in our own code. Either for individual types/functions, or by just disabling the two types of warnings that we currently see on a per-module basis.

I briefly tested whether the change from Dump/Pickle to Yojson caused any problems when serialising continuations  but you're more than welcome to check if your favorite Links feature still works as expected.


core/jbuild
```diff
@@ -10,6 +10,6 @@
   (wrapped false)
   (modes (native))
   (flags (:standard -safe-string -dtypes -w Ae-44-45-60 -g -cclib -lunix -thread))
-  (libraries (str deriving unix safepass cgi base64 ANSITerminal linenoise
+  (libraries (str deriving yojson ppx_deriving_yojson.runtime unix safepass base64 ANSITerminal linenoise
               cohttp lwt websocket websocket-lwt.cohttp))
-  (preprocess (action (run deriving/deriving ${<})))))
+  (preprocess (pps (ppx_deriving.std ppx_deriving_yojson)))))
```
dhil: Brilliant

links.opam
```diff
@@ -56,7 +56,8 @@ remove: [
 depends: [
   "jbuilder" {build}
   "deriving" {build}
```
dhil: I think `deriving` should be removed as a build dependency.

frank-emrich:
@slindley I have verified that using yojson fixes #124 and the other examples work as well, in particular the formlet one. I've tested it with and without handlers being enabled

links.opam
```diff
@@ -7,7 +7,7 @@ bug-reports: "https://github.com/links-lang/links/issues"
 license: "GPL-2"
 
 available: [
-  ocaml-version >= "4.04.0" & ocaml-version <= "4.06.0"
+  ocaml-version >= "4.04.0" & ocaml-version <= "4.06.1"
```
dhil: We should be able to remove the upper bound now, i.e. I think it is save to have `ocaml-version >= "4.04.0`" now.

links.opam
```diff
@@ -60,7 +60,7 @@ depends: [
   "base64"
   "linenoise"
   "ANSITerminal"
-  "lwt"
+  "lwt" {>= "3.1.0"}
```
dhil: Why do you need to constrain LWT?? Do we have some other dependency that depends on 3.1.0?

core/cgi.ml
```diff
@@ -0,0 +1,328 @@
+(*
```
dhil: It seems `cgi.ml` has been readded. I thought we had gotten rid of it.

frank-emrich: See P2) in my PR description:  tldr: The removal of cgi isn't merged in yet, so I patched cgi as an immediate solution to work with 4.06.0

links.opam
frank-emrich: done!

frank-emrich:
I have now solved the problem with warnings. In some cases, some deriving show/eq/yojson annotations were indeed unnecessary. In those 5 modules where some warnings simply could not be fixed, I added module-wide annotations to suppress the following two types of warnings:
1) unused functions
2) unnecessary rec flags in let 

From my side, this is now ready for merging

dhil:
Cheers. I will have a look at this on Tuesday, when I am back in Edinburgh.

core/utility.ml
```diff
@@ -33,9 +32,11 @@ let ( <| ) f arg = f arg
 
 (** {0 Maps and sets} *)
 module type OrderedShow = sig
-  type t
+  type t (**[@@deriving show]**)
   val compare : t -> t -> int
-  module Show_t : Deriving_Show.Show with type a = t
+  val show : t -> string
+  val pp : Format.formatter -> t -> unit
```
dhil: Remove this.

core/utility.ml
```diff
@@ -33,9 +32,11 @@ let ( <| ) f arg = f arg
 
 (** {0 Maps and sets} *)
 module type OrderedShow = sig
-  type t
+  type t (**[@@deriving show]**)
```
dhil: Uncomment this.

core/types.ml
```diff
@@ -571,17 +570,17 @@ let is_sessionable_row = is_sessionable_row IntSet.empty
...
 (* useful for debugging: types tend to be too big to read *)
 (*
-module Show_datatype = Show_unprintable (struct type a = datatype end)
-module Show_field_spec = Show_unprintable (struct type a = field_spec end)
-module Show_field_spec_map = Show_unprintable (struct type a = field_spec_map end)
-module Show_row_var = Show_unprintable (struct type a = row_var end)
-module Show_row = Show_unprintable (struct type a = row end)
-module Show_meta_type_var = Show_unprintable (struct type a = meta_type_var end)
-module Show_meta_row_var = Show_unprintable (struct type a = meta_row_var end)
+let pp_datatype = fun _ _ -> ()
```
dhil: Should print ".."

PR #346, opened 2018-06-15
More lazy unification errors

### LINKED ISSUE #85: Error messages generated too eagerly
When working on #43 I noticed that function generating error messages (a.k.a. gripers in `typeSugar` module) are called several times even though only one error message gets printed. @slindley says this is caused by unification doing backtracking: error messages are generated and thrown as exceptions, they get caught and another attempt at unification is made. It would be much better to generate error messages lazily, ie. only when they are actually going to get printed.

jstolarek:
This patch makes error message generation more lazy. What works:

 * `fun_apply` griper is called only once when reporting incorrect function type. Thus error message is only generated once, when we know it will be displayed to the user

What doesn't:

  * if I comment out first two lines of `build_tyvar_names` in `types.ml`, which are the lines responsible for resetting variable name counter, I can easily see that some variable names are being generated even though they are not displayed to the user. This can be witnessed by variable names in error message starting from letters like `k` and not `a`. Is this expected behaviour? Are we generating some variable names for internal use only? If not, then this patch requires further work to determine why this happens. I suspect this might be taking place outside of `typeSugar.ml` module.

This is intended to fix #85.

### DIFF
--- a/core/typeSugar.ml
+++ b/core/typeSugar.ml
@@ -1628,7 +1628,7 @@ let rec close_pattern_type : pattern list -> Types.datatype -> Types.datatype = 
        (* TODO: expand applications? *)
       | `Application _ -> t
 
-type unifyResult = UnifySuccess | UnifyFailure of (Unify.error * SourceCode.pos)
+type unify_result = UnifySuccess | UnifyFailure of (Unify.error * SourceCode.pos)
 
 let raise_unify ~(handle:Gripers.griper) ~pos error t1 t2 =
   begin
--- a/core/typeSugar.ml
+++ b/core/typeSugar.ml
@@ -1649,8 +1649,7 @@ let unifyOrRaise ~(handle:Gripers.griper) ~pos (_, ltype1 as lt1, (_, rtype1 as 
   | UnifyFailure (err, pos) -> raiseUnify ~handle ~pos err lt1 rt1
   end
 
-(* Expects both unify arguments to succesfully unify.  Raises a unify exception
-   if at least one unification fails *)
+(* Expects at least one pair of arguments to succesfully unify *)
 let unifyOr ~(handle:Gripers.griper) ~pos ((_, ltype1), (_, rtype1))
                                           ((_, ltype2) as lt2, ((_, rtype2) as rt2)) =
   begin
--- a/core/typeSugar.ml
+++ b/core/typeSugar.ml
@@ -1630,7 +1630,7 @@ let rec close_pattern_type : pattern list -> Types.datatype -> Types.datatype = 
 
 type unifyResult = UnifySuccess | UnifyFailure of (Unify.error * SourceCode.pos)
 
-let raiseUnify ~(handle:Gripers.griper) ~pos error t1 t2 =
+let raise_unify ~(handle:Gripers.griper) ~pos error t1 t2 =
   begin
     match error with
     | `Msg s -> Debug.print ("Unification error: "^s)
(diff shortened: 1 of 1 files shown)


core/typeSugar.ml
```diff
@@ -1628,18 +1628,41 @@ let rec close_pattern_type : pattern list -> Types.datatype -> Types.datatype =
...
-        begin
-          match error with
-            | `Msg s ->
-                Debug.print ("Unification error: "^s)
-            | _ -> ()
-        end;
-        handle ~pos ~t1 ~t2 ~error
+type unifyResult = UnifySuccess | UnifyFailure of (Unify.error * SourceCode.pos)
+
+let raiseUnify ~(handle:Gripers.griper) ~pos error t1 t2 =
```
dhil: **Disclaimer** the following is nitpick. Throughout the code base we predominantly use shake case for function names and type names, although, some modules do use camel case. In spirit of converging to a consistent coding style throughout the code base, I suppose `raiseUnify` should be `raise_unify`.

core/typeSugar.ml
```diff
@@ -1628,18 +1628,41 @@ let rec close_pattern_type : pattern list -> Types.datatype -> Types.datatype =
...
+    | `Msg s -> Debug.print ("Unification error: "^s)
+    | _ -> ()
+  end;
+  handle ~pos ~t1 ~t2 ~error
+
+let unify ~pos unifyTys  =
+  try Utils.unify unifyTys; UnifySuccess
+  with Unify.Failure error -> UnifyFailure (error, pos)
+
+let unifyOrRaise ~(handle:Gripers.griper) ~pos (_, ltype1 as lt1, (_, rtype1 as rt1)) =
```
dhil: Same as my previous comment.

core/typeSugar.ml
```diff
@@ -1628,18 +1628,41 @@ let rec close_pattern_type : pattern list -> Types.datatype -> Types.datatype =
...
+  begin
+  match unify ~pos (ltype1, rtype1) with
+  | UnifySuccess -> ()
+  | UnifyFailure (err, pos) -> raiseUnify ~handle ~pos err lt1 rt1
+  end
+
+(* Expects at least one pair of arguments to succesfully unify *)
+let unifyOr ~(handle:Gripers.griper) ~pos ((_, ltype1), (_, rtype1))
```
dhil: Same as my previous comment.

core/typeSugar.ml
slindley: I think you mean "snake case". But yes you're right, compound variable names should never use camel case. This is the standard convention for OCaml source code. The conventions for modules and type constructors are a little less clear, but we tend to use camel case for those.

@jstolarek please change the variable and type names to use underscores.

dhil: @slindley yes, I meant to type "snake case"...

jstolarek: > for modules and type constructors (...) we tend to use camel case.
> please change the variable and type names to use underscores.

So, do you want the type constructor `unifyResult` to remain in camel case or do you want it to become `unify_result`?

slindley: it should be unify_result - as should any other compound name that begins with a lower case letter

core/typeSugar.ml
```diff
@@ -1630,7 +1630,7 @@ let rec close_pattern_type : pattern list -> Types.datatype -> Types.datatype =
 
 type unifyResult = UnifySuccess | UnifyFailure of (Unify.error * SourceCode.pos)
```
dhil: Yes, types follow the same scheme as terms, so it should really be `unify_result`

jstolarek: Done now.

PR #348, opened 2018-06-20
Change the semantics of print

dhil:
This PR changes the semantics of the built-in function `print` such that it no longer emits a newline character. A new function `println : (String) ~> ()` has been added to the prelude which prints a string followed by a newline character.

In addition, the lexer now recognises `\b` as an escape character.

Finally, I have patched all the examples in the `examples/` directory to conform with the new semantics of `print`.


examples/handlers/aop.links
```diff
@@ -6,7 +6,7 @@ sig runState :
         (Comp ({Get:(()) {}-> s,Put:(s) {}-> ()|e}, a)) ->
   (s) -> Comp ({Get{_},         Put{_}         |e}, (a, s))
 fun runState(m)(s)() {
-  (open handle(m) {
```
frank-emrich: it looks like you committed some unrelated other changes here

examples/handlers/aop2.links
```diff
@@ -1,12 +1,12 @@
 typename Zero = [||];
 
-typename Comp(e::Row,a) = (() ~e~> a);
```
frank-emrich: same in this file

examples/handlers/number_games.links
```diff
@@ -0,0 +1,65 @@
+# Classic guess a number game
+typename Comp(e::Eff,a) = () ~e~> a;
+typename Answer = [|Low|Correct|High|];
+
```
frank-emrich: added new example here?

examples/sessions/Elvina/IRC_Test.links
```diff
@@ -117,20 +117,20 @@ fun connectToIRCServer() {
...
     # End of tests.
     writeToSocket("QUIT\n", rs);
-    print("QUIT");
+    println("QUIT");
     closeSocket(rs);
     ()
 }
 
-# Writes LinksCalc: and given expression to socket; prints it.
+# Writes LinksCalc: and given expression to socket; printlns it.
```
frank-emrich: your regex should have been something like print[^a-zA-Z] to prevent breaking comments containing "print"

examples/handlers/aop.links
dhil: Yes, I fixed a broken example too.

examples/handlers/number_games.links
dhil: Apparently, yes... This is the example from my ProWeb talk.

examples/sessions/Elvina/IRC_Test.links
dhil: Fixed.

Issue #345, opened 2018-06-15
Inconsistent behaviour of Regex in REPL and Server


Corneliusbusch:
The behaviour of a function using regular expressions is inconsistent when used on the server and the Links REPL.

The following minimal example is supposed to print "False" to the Console after the button has been clicked as it checks whether the string of "-2" is a positive integer. 
However, in the Browser it prints true, whereas when run in the Links REPL, the function correctly rejects "-2" as a positive integer. 

The minimal example: 
```
sig isPosInt: (String) -> Bool
fun isPosInt(n){
    (n =~ /^[0-9]+$/)
}

sig main_page : (_) ~> Page
fun main_page(_) {
    page
    <html>
        <div id="content">
           <button l:onclick="{if( isPosInt("-2") ) print("True") else print("False") }">
                 Is -2 a positive Integer? 
           </button>
        </div>
    </html>
}

sig main: () ~> ()
fun main() {
    addRoute("/", main_page);
    servePages()
}

main()
```

When called in the Links REPL, the correct result is returned. 
```
links> sig isPosInt: (String) -> Bool
...... fun isPosInt(n){
......     (n =~ /^[0-9]+$/)
...... };
isPosInt = fun : (String) -> Bool
links> isPosInt("-2");
false : Bool
```

jamescheney:
This seems likely to be due to differences between client (JS) and server (OCaml/PCRE) implementations of regexes.  Are you sure the regex operation is happening on the server and not the client?  What happens if you put the "server" keyword in "isPosInt"?  Not that you necessarily want to be calling the server just to do regex matching.

Corneliusbusch:
If I add the "server" keyword in "isPosInt", it evaluates correctly and prints "False". However, if I add the "client" keyword, it prints "True". So, I guess, the inconsistency is between REPL and client?

dhil:
The compilation of regular expressions to the client is completely broken. The regex `/^2$/` seemingly gets compiled as a list object 
```
{"regex": 
    {"_label": "Seq", 
     "_value": {"_head": 
                       {"_label": "StartAnchor", "_value": {}}, 
                     "_tail": 
                       {"_head": 
                          {"_label": "Simply", "_value": "2"}, 
                        "_tail": 
                            {"_head": 
                                {"_label": "EndAnchor", "_value": {}}, 
                             "_tail": null}}}}}}
```
The whole object is being fed into `_tilde` as `regex`
```
function _tilde(s, regex) {
[...]

PR #356, opened 2018-06-28
Fixing #345: Inconsistent Regex behaviour

Corneliusbusch:
The problem has been solved by implementing a custom regex compiler as the built in implementation returned the empty string.


lib/js/jslib.js
```diff
@@ -1404,9 +1404,66 @@ const LINKS = new (function() {
...
+    case "StartAnchor":{
+      return "^";
+    }
+    case "Repeat":{
+      let op = undefined;
+      switch(expr._value[1]._label){
+        case "Plus": op = "+"; break;
+        case "Star": op = "*"; break;
+        case "Question": op = "?"; break;
+        default: throw("Error: Unrecognised Operation");
```
dhil: Make the error message a little bit more descriptive, e.g. `"Unrecognised operation '" + expr._value[1]._label + "'"`. Perhaps use `_error` rather than `throw`.

lib/js/jslib.js
```diff
@@ -1404,9 +1404,66 @@ const LINKS = new (function() {
...
+    case "Group":{
+      return "(" + _regexCompile(expr._value) + ")";
+    }
+    case "Quote":{
+      return "\\" + _regexCompile(expr._value);
+    }
+    case "Replace":{
+      throw("Error: \"Replace\" not implemented");
+    }
+    default: throw("Unrecognised Tag in Regex Expression");
```
dhil: Perhaps `"Unrecognised label '" + expr._label + "'"`

lib/js/jslib.js
```diff
@@ -1404,9 +1404,66 @@ const LINKS = new (function() {
...
+  }
+  return null;
```
dhil: Maybe add a comment here `// Can never be reached`. And use `_error` rather than `throw`.

dhil:
This PR resolves #345 .

lib/js/jslib.js
Corneliusbusch: Fixed.

lib/js/jslib.js
Corneliusbusch: Fixed.

lib/js/jslib.js
Corneliusbusch: Fixed.

fehrenbach:
Does this PR make this file obsolete? lib/js/regex.js
Should we delete that?

dhil:
Ah, good point! Now I understand why regex compilation did not work on the client. The `regex.js` monkey patches `Regex.compile`, however it seemingly wasn't included (properly) in the generated page code.

I suppose we can get rid of it...

Issue #360, opened 2018-07-03
Incomplete closing of recursive patterns


dhil:
During type inference of a `switch`-expression, the type checker is at some point going to attempt to close the inferred pattern type. The implementation of `close_pattern_type` appears to be incomplete as the following example triggers an assertion failure
```
typename Term = [|Var:Int|Lam:(Int, Term)|];
switch(Var(0)) {
   case (_ : Term) -> 0
}

*** Error: "Assert_failure core/typeSugar.ml:1638:32"
```

The code of interest in `typeSugar.ml` is
```
| `MetaTypeVar point ->
          begin
            match Unionfind.find point with
              | `Body t -> cpt pats t
              | `Var _ -> t
              | `Recursive _ -> assert false
          end
```

I wonder what it means to "close a recursive type variable"? How is it any different from "closing an ordinary type variable"?

PR #361, opened 2018-07-03
Resolves links-lang/links#360

dhil:
Fixes issue #360. 

I picture two potential fixes, either return `t` as in the case for `Var` or unroll the recursive type once, and keep track of the recursive variables already encountered to prevent non-termination. I opted for the latter approach, which may prove be a somewhat conservative option as I have been unable to produce an example that requires unrolling.

### DIFF
--- a/core/typeSugar.ml
+++ b/core/typeSugar.ml
@@ -1438,10 +1438,13 @@ let type_binary_op ctxt =
    If there are no _ or variable patterns at a variant type, then that
    variant will be closed.
 *)
-let rec close_pattern_type : pattern list -> Types.datatype -> Types.datatype = fun pats t ->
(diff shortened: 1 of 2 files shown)


dhil:
The following is a simple example that unrolls the recursion to expose another recursive type variable, I am still not sure whether it is strictly necessary to unroll for correctness...

```
switch (Foo) { 
  case (_ : (mu a . [|Foo|Bar:a|Baz:(mu b. b)|_|])) -> 0 
}
```

dhil:
Apparently whether to unroll deeply depends on the lexicographic ordering of constructor names. If we alpha-convert the previous program, such that we swap the `A` and `C` constructors, then the type checker only unrolls a single level for each recursive type variable
```
switch (A) { 
  case (_ : (mu a . [|A|B:a|C:(mu b. b)|_|])) -> 0 
}
```

PR #362, opened 2018-07-08
Removes safepass

dhil:
A recent update to safepass (verison 3.0.0) has added a new optional argument to the API for `hash`, which causes a new warning during compilation of Links
```
      ocamlc core/.links.objs/lib.{cmo,cmt}
File "core/lib.ml", line 1610, characters 66-102:
Warning 48: implicit elimination of optional argument ?variant
    ocamlopt core/.links.objs/lib.{cmx,o}
File "core/lib.ml", line 1610, characters 66-102:
Warning 48: implicit elimination of optional argument ?variant
```
Setting a default value for `variant` would require everybody to upgrade their `safepass` package, however, as there appear to be no uses of the built-in `crypt` and `verify` I thought we might as well remove the dependency altogether.

Arguably, functionality a la `safepass` should be provided via a foreign function interface rather than being built-in.

### DIFF
--- a/core/jbuild
+++ b/core/jbuild
@@ -10,5 +10,5 @@
   (wrapped false)
   (modes (native))
   (flags (:standard -safe-string -dtypes -w Ae-44-45-60 -g -cclib -lunix -thread))
-  (libraries (str yojson ppx_deriving_yojson.runtime unix safepass base64 ANSITerminal linenoise cohttp lwt websocket websocket-lwt.cohttp findlib))
+  (libraries (str yojson ppx_deriving_yojson.runtime unix base64 ANSITerminal linenoise cohttp lwt websocket websocket-lwt.cohttp findlib))
   (preprocess (pps (ppx_deriving.std ppx_deriving_yojson)))))
--- a/core/lib.ml
+++ b/core/lib.ml
@@ -1603,18 +1603,6 @@ let env : (string * (located_primitive * Types.datatype * pure)) list = [
     (`PFun (fun _ -> assert false),
     datatype "() ~> ()",
     IMPURE);
-
-    (* Crypt API *)
-
-    "crypt",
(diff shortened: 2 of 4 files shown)


frank-emrich:
you need to remove safepass from the jbuild files as well

fehrenbach:
This was one of the previous intern's work, wasn't it? Presumably they added it for a reason. I'm not sure a warning is reason enough to remove it, especially since the proposed alternative FFI is currently not viable, ignoring the warning is easy, and fixing the warning is only one `opam whatnot` + a lower bound in the opam file away.

dhil:
The API is unused. I'll remind you that we have a "no warnings" in master branch policy.

fehrenbach:
1. Am I confused? Isn't it exposed as a built-in function. How could you possibly know whether it is used by someone?
2. Do we have a "no warnings" policy? How come we don't enforce it?
3. There are other ways:
  a) We can just pass the argument and require a newer safepass. We have restricted dependency version ranges before (and even currently do for lwt).
  b) We disable warnings 44, 45, and 60 already.

dhil:
1. No Links source program makes use of it
2. Yes. We do enforce it on an ad-hoc basis.
3. a) true but that is suboptimal, like our current restriction of the OCaml compiler. I would not compare the specific case of lwt, as it is an integral part of the Links code base, safepass is not. b) we shouldn't need to silence 44, 45, or 60. I am aware that we currently silence 32 and 39, and we should be able to get rid of the 32-warnings, but we might need to keep silencing 39 because one of the PPX extensions `let rec`s non-recursive functions.

dhil:
Why do you want this bloat in the code base? Useful in what sense? Shall I take it as you volunteer to maintain compatibility with `safepass` and thereby the restriction in the OPAM file in Links?

PR #365, opened 2018-07-10
Allow pure/tame toplevel expressions and bindings

### LINKED ISSUE #364: No tame expressions allowed at toplevel
Consider the following code defining a tame function and calling it at the toplevel:
```
sig f : (Int) {}-> Int
fun f(x) {x}
links> f(2);
```
This yields the following error:
```
Type error: The function
    `f'
has type
    `(Int) {}-> Int'
    `Int'
    `wild'
In expression: f(2).
```

frank-emrich:
This fixes #364.

The idea is to change the row denoting the expected effects to an empty row with a flexible row variable each time before type-checking a toplevel expression or binding. In the process, that row will be unified to whatever effects the toplevel expression or binding actually has.

After that, we try to unify the effect row with either an empty, closed row or a closed row containing only the wild effect. At the point of that check, we do not know the particular shape of e.g. the offending expression, which means it didn't make sense to re-use any of the existing gripers. Instead, I created two new ones. The resulting error message looks like this:

../my_examples/toplevel_impure.links:28: Type error: This toplevel binding has unhandled effects.
In expression: var z = f3().

### DIFF
--- a/core/typeSugar.ml
+++ b/core/typeSugar.ml
@@ -1722,7 +1722,6 @@ let check_toplevel_effects_expr tyenv expr =
   check_toplevel_effects tyenv pos griper
 
 
-
 (** check for duplicate names in a list of pattern *)
 let check_for_duplicate_names : Sugartypes.position -> pattern list -> string list = fun pos ps ->
   let add name binder binderss =
@@ -2101,8 +2100,6 @@ let usage_compat =
 
 let usages_cases bs =
   usage_compat (List.map (fun (_, (_, _, m)) -> m) bs)
-
-
 
 let rec type_check : context -> phrase -> phrase * Types.datatype * usagemap =
   fun context (expr, pos) ->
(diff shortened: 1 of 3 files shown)


fehrenbach:
Sounds like a good plan.

How hard would it be to add the offending effects to the error message?

dhil:
@fehrenbach it should be easy, and that will indeed be one of my review requests!

frank-emrich:
I didn't implement that immediately because it seemed to require some hacks but I've now found a nice way of printing the offending effects :)

dhil:
Ping @frank-emrich.

frank-emrich:
It was my impression that this is by design and we should close this PR then

dhil:
It is by design, which makes sense for compilation, but makes less sense for live demoing in the REPL. You can reopen the PR if you want to get in.

PR #369, opened 2018-07-12
Type-check all Links example programs as part of test suite

frank-emrich:
It think it would be helpful to perform some automated regression testing on the Links example programs.
This PR adds a setting "typecheck_only", which makes Links exit after type-checking. Effectively, this turns the exit code into an indicator whether the given file parsed & type-checked successfully.

There are currently 72 example programs that are broken and therefore not tested.

Eventually, I would like to enable the IR type-checker for these tests too.


tests/typecheck_examples.tests
```diff
@@ -0,0 +1,1007 @@
...
+Typecheck example file examples/distribution/serverToClient2.links
+examples/distribution/serverToClient2.links
+filemode : args
+args : --config=tests/typecheck_examples.tests.config
+
+Typecheck example file examples/distribution/simpleServerAP.links
+examples/distribution/simpleServerAP.links
+filemode : args
+args : --config=tests/typecheck_examples.tests.config
+ignore : broken example
```
dhil: You may want to check with @SimonJF whether the "broken" distribution examples are indeed broken. Maybe they just need to be deleted...

tests/typecheck_examples.tests
```diff
@@ -0,0 +1,1007 @@
...
+examples/todo.links
+filemode : args
+args : --config=tests/typecheck_examples.tests.config
+
+Typecheck example file examples/todoDb.links
+examples/todoDb.links
+filemode : args
+args : --config=tests/typecheck_examples.tests.config
+
+Typecheck example file examples/tofix/dom-send.links
```
dhil: I see little point in adding examples under the directory `tofix` to the test suite. Those examples have been bit rotting for ages now.

tests/typecheck_examples.tests
```diff
@@ -0,0 +1,1007 @@
...
+filemode : args
+args : --config=tests/typecheck_examples.tests.config
+ignore : broken example
+
+Typecheck example file examples/games/twentyfortyeight.links
+examples/games/twentyfortyeight.links
+filemode : args
+args : --config=tests/typecheck_examples.tests.config
+
+Typecheck example file examples/handlers/alert.links
```
dhil: Do not add the `handlers/` directory to the test suite. Those examples that I want to maintain are already included in `handlers.test`.

jamescheney: If examples in `handlers/` and `tofix/` are not being maintained, maybe they are candidates for deletion?

dhil: Yes... I need to clean up `handlers/`.

tests/typecheck_examples.tests
frank-emrich: So far, my approach was to be very explicit and add all examples to the test file. This would ease adding new examples utomatically, because we only need to check for examples whose path is not in the .tests file, yet. However, as @dhil suggested, we may use a more clever script which already does type-checks all new examples and only adds new + working examples.

tests/typecheck_examples.tests
frank-emrich: Fun fact: The tofix directory hasn't been touched in 9 years. I suggest we just get rid of it.

tests/typecheck_examples.tests
dhil: What I suggested was a little different, actually, just a slight variation of your previous script. I think it is fine to conservatively assume an example that fails to type check is "broken". However, you should filter the generated `.tests` file such that it doesn't include directories like `tofix` or experimental stuff like `handlers/` (arguably experimental or incomplete examples should be placed elsewhere)

dhil:
Ping @frank-emrich.

frank-emrich:
@jamescheney marked this for 0.8, which I thought was a sign to delay this to the next mayor release?

jamescheney:
Actually it looks like @dhil added it to the 0.8 milestone.  However, I think 0.8 is the planned next release so we ideally want to close/merge anything with this tag as soon as we can (realistically anything that we don't have time for will get snoozed to the next minor release but if this is almost ready to go then it would be nice to get it done).

tests/typecheck_examples.tests
frank-emrich: @SimonJF could you have a look at this and determine which of the session types examples should be working?

tests/typecheck_examples.tests
frank-emrich: Okay. We removed tofix in the meantime. Are all of the examples in the handlers subdirectory experimental?

tests/typecheck_examples.tests
SimonJF: Good idea. Planning to spend a day hacking Links on Friday, so I'll take a look at the session types examples to tidy up / determine which should be retained.

bin/links.ml
```diff
@@ -51,6 +51,9 @@ let process_program ?(printer=print_value) (valenv, nenv, tyenv) (program, t) ex
   let tenv = (Var.varify_env (nenv, tyenv.Types.var_env)) in
 
   let (globals, _) = Backend.transform_program tenv program in
+
+  (if Settings.get_value BS.typecheck_only then exit 0);
```
dhil: It would be cleaner to define a separate entry point for "type check ir only". Alas, `links.ml` is a big mess that could do with some Spring cleaning. Some future PR....

tests/typecheck_examples.tests
dhil: I did some cleaning recently (as in a couple of months ago). They might just all work now... You can give it a try if you like.

frank-emrich: The following handler examples are currently broken:
!FAILURE: Typecheck example file examples/handlers/choose.links
!FAILURE: Typecheck example file examples/handlers/count_web.links
!FAILURE: Typecheck example file examples/handlers/deep_pipes.links
!FAILURE: Typecheck example file examples/handlers/delimitedcontinuations.links
!FAILURE: Typecheck example file examples/handlers/nim.links
!FAILURE: Typecheck example file examples/handlers/pipes.links
!FAILURE: Typecheck example file examples/handlers/shallow_pipes.links
!FAILURE: Typecheck example file examples/handlers/tests.links
!FAILURE: Typecheck example file examples/handlers/toggle.links
!FAILURE: Typecheck example file examples/handlers/toggle_web.links


using the following configuration and testing on the current master:
typecheck_only=true
modules=true
enable_handlers=true
shredding=true
session_exceptions=true

If that's fine, I'll just mark those as ignored. The actual output is as follows:
[output.txt](https://github.com/links-lang/links/files/2332711/output.txt)

dhil: Are those results from the HEAD of master?

bin/links.ml
frank-emrich: I agree that links.ml needs some cleanup

tests/typecheck_examples.tests
frank-emrich: yes

SimonJF:
This is an excellent idea.

There was one `distribution` example which had a typo -- I've fixed that, and added it to your PR branch. We have two copies of the `chatserver` example, one of which is out of date, so I will add a separate PR to remove that. That'll need to be merged first.

I expect you'll then have to rebase (sorry).

frank-emrich:
This should be ready now. The session examples have been fixed, and I've ignored those examples in the handlers subdirectory that are currently not working.

SimonJF:
There are a few broken session examples I missed, it seems. One sec.

SimonJF:
OK -- I'm happy for this to go in now.

PR #376, opened 2018-07-19
Fixes buggy implementation of getValue in jslib

dhil:


lib/js/jslib.js
```diff
@@ -1752,8 +1762,8 @@ function _getValue(nodeRef) {
   } else if (nodeRef.nodeType == document.ELEMENT_NODE ) {
     var children = Nil;
 
-    // @TODO
-    for (var i=nodeRef.childNodes.length -1; i <= 0; i++) {
+    const numChildren = nodeRef.childNodes.length;
+    for (let i = nodeRef.childNodes.length; i < numChildren; i++) {
```
fehrenbach: Do you mean `let i = 0;`? Also, they will be in reverse order, I think. Is that intended?

PR #378, opened 2018-07-27
Adding my demo projects to the example folder

Corneliusbusch:


jamescheney:
Is there anything holding up merging these?  Or do they belong in links-examples?

dhil:
I think we agreed during the previous meeting that they belong to links-examples.

slindley:
I think we did suggest putting them in links-examples initially, but then I suggested to Nicole that we just put them in the main repository instead, as they are all quite small, and some of them (certainly some of the handler examples) would be really good to have there. I'd be inclined to accept the pull request without many changes, but afterwards it might be worth doing some curation.

If others would prefer to use links-examples as a staging post, then that's also fine - as long as someone is happy to do that.

dhil:
OK. I have no problem with putting them into this repository. I'll merge.

PR #385, opened 2018-08-25
[WIP] Relational Lenses

rudihorn:
Pull request for adding support for relational lenses. Also adds an additional links library named `links-unit`, which is a wrapper for all the unit tests. To run the unit tests use `./run-unit-tests`. 

Most changes are protected by an extra flag `relational_lenses` which has to be set to true for the lenses to work. The flag doesn't affect the parser however.

The following required changes were made to the links code base:
- data structures and code for binary / unary operators were moved to `core/operators.ml` to avoid circular dependencies
- errors related to command line options are only thrown in `bin/links.ml`, but evaluated in `core/parseSettings.ml` to avoid side effect problems and to allow unit test command lines

While some further cleaning up would be nice, this is mainly constrained to the `core/lens*` files and doesn't affect the rest of the links source code much. I will be using this pull requests to make sure Travis works correctly, but this should be discussed at the next links meeting. 

Remaining known issues:
- JSON serialization of lens values, both in `core/json.ml` as well as `core/value.ml` (value compression).


core/sugartypes.ml
```diff
@@ -1,44 +1,7 @@
-(** The syntax tree created by the parser. *)
-
-type name = string [@@deriving show]
```
dhil: Why have you moved this to `Operators.ml`? It seems to me that you have just disorganised the sugared syntax module.

core/parser.mly
```diff
@@ -899,6 +901,31 @@ database_expression:
                                                                  pos() }
 | DATABASE atomic_expression perhaps_db_driver                 { `DatabaseLit ($2, $3), pos() }
 
+fn_dep_cols:
```
dhil: Have you checked that this has zero shift/reduce conflicts?

core/sugartypes.ml
rudihorn: I believe the problem was that `sugartypes.ml` depends on `types.ml` and `lens_phrase` in `types.ml` depends on the operators (which would be in `sugartypes.ml`). The lens types in `typ` depend on `lens_phrase`. The alternative would be copy pasting the operators, which I wanted to avoid.

core/parser.mly
rudihorn: Will try double checking this.

unit/unitTestsLensCommon.ml
```diff
@@ -10,7 +10,7 @@ open LensSetOperations
 
 let display_table_query_opt = Conf.make_bool "display_table_query" false "Show queries to take and manipulate tables."
 let leave_tables_opt = Conf.make_bool "leave_tables" false "Do not delete tables after run." 
-let db_host_opt = Conf.make_string "db_host" "localhost" "Database server hostname." 
+let database_args_opt = Conf.make_string "database_args_host" "links:localhost:5432::links" "Database connection args."
```
dhil: This looks suspicious to me. You should not hard code database connection information in the test, rather, you should read it from the provided configuration file.

rudihorn: So technically this is only the default config. This sets up an argument within the unit testing framework so that if you run `unitTests.exe -help` it will give you all possible command line options related to unit testing including:

`  -database-args-host str           Database connection args. (default: localhost:5432::links)`

Maybe it is worth discussing in the links meeting if we want to use the links config for unit testing, but I personally find it better for the unit testing things to rely on as little infrastructure as possible.

dhil: It is plain and simple bad software engineering practice. The database configuration differs from system to system. For example, I am running different setups on my server and my workstation. I should not have to edit the same information in two different configuration files to run the test suite.

rudihorn: There are already two different configs for testing (`tests/shredding/config.sample` and `tests/database/config.sample`), but I have now just made it fallback to a value in a settings file if there is one that can be loaded.

core/ir.ml
```diff
@@ -75,7 +75,13 @@ and binding =
   | `Module of (string * binding list option) ]
 and special =
   [ `Wrong of Types.datatype
-  | `Database of value
+  | `Database of value 
+  | `Lens of value * Types.lens_sort
+  | `LensDrop of value * string * string * value * Types.lens_sort
+  | `LensSelect of value * Sugartypes.phrase * Types.lens_sort
+  | `LensJoin of value * value * string list * Sugartypes.phrase * Sugartypes.phrase * Types.lens_sort
```
dhil: This doesn't look right. Why haven't `Sugartypes.phrase`s been translated? The IR should not contain instances of concrete syntax.

rudihorn: Hm yes one should be able to use `lens_phrase` at this point. I'll fix this tomorrow.

core/types.ml
```diff
@@ -124,6 +124,61 @@ type ('t, 'r) session_type_basis =
     | `End ]
       [@@deriving show]
 
+type lens_delete_mode =
+  [ `DeleteLeft
+  | `DeleteRight
+  | `DeleteBoth ]
+      [@@deriving show]
+
+type lens_phrase =
```
dhil: What is the relevance of `lens_phrase` to types? It seems to be some representation of term syntax.

core/types.ml
```diff
@@ -124,6 +124,61 @@ type ('t, 'r) session_type_basis =
...
+  (* the In operator checks if the tuple generated by the names list is 
+   * in the list of constant tuples *)
+  | `In        of Operators.name list * (Constant.constant list) list
+  | `Case      of lens_phrase option * (lens_phrase * lens_phrase) list * lens_phrase
+  | `TupleLit  of lens_phrase list
+  ]
+      [@@deriving show]
+
+(* Lenses *)
+module ColSet = struct
```
dhil: Relevance to types? Looks like something that is utility related.

core/types.ml
```diff
@@ -124,6 +124,61 @@ type ('t, 'r) session_type_basis =
...
+    include StringSet
+
+    let pp_pretty fmt cs = 
+        List.iter (fun c -> Format.pp_print_string fmt c; Format.pp_print_space fmt ()) (elements cs)
+end
+
+type colset = ColSet.t
+      [@@deriving show]
+
+module FunDep = struct
```
dhil: Relevance to types? Looks like something that is utility related.

core/types.ml
```diff
@@ -124,6 +124,61 @@ type ('t, 'r) session_type_basis =
...
+        if res = 0 then
+            ColSet.compare r1 r2 
+        else
+            res
+end
+
+type fundep = colset * colset 
+        [@@deriving show]
+
+module FunDepSet = struct
```
dhil: Relevance to types? Looks like something that is utility related.

unit/README.md
```diff
@@ -0,0 +1,14 @@
+To run the unit tests, simply call `jbuilder runtest`.
```
dhil: I suppose you mean `dune` here.

core/ir.ml
rudihorn: Fixed this.

core/desugarFors.ml
```diff
@@ -111,7 +111,7 @@ object (o : 'self_type)
   *)
   method qualifiers : Sugartypes.iterpatt list ->
     'self_type *
-      (Sugartypes.phrase list * Sugartypes.pattern list * Sugartypes.name list *
+      (Sugartypes.phrase list * Sugartypes.pattern list * Operators.name list *
```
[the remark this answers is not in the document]
rudihorn: This seems mostly specific to `SugarTypes.name`, so I have reintroduced that type alias and it can now be called `Sugartypes.name` again.

core/types.ml
rudihorn: The original work on relational lenses defines typing rules which involve the phrases. Not all of these checks are done yet, and most of this depends also on how possible it is to instantiate "dynamic" phrases. Having `lens_phrase` in the type allows some of those static type checks to be performed during compile time but limits functionality. These checks aren't quite implemented yet, but I have been trying to retain the concept of having a `lens_sort` describing known information about lenses.

core/types.ml
rudihorn: Moved into LensUtility.

core/types.ml
rudihorn: Moved into LensUtility.

unit/README.md
rudihorn: I have changed this to use the `./run-unit-tests` script now in any case, since it ensures that the `-runner sequential` tag is included and is necessary to prevent tests from failing.

core/lensHelpers.ml
```diff
@@ -0,0 +1,302 @@
+open Types
+open Utility
+open LensUtility
```
jamescheney: What is the difference between LensHelpers and LensUtility?  It sounds like LensHelpers contains the actual implementation of the relational lens operations (the core logic) so that LensHelpersClassic is the old version, while the other "helpers" are components used by the core functionality that work the same in either version.

core/lensHelpersClassic.ml
```diff
@@ -0,0 +1,139 @@
+open Value
```
jamescheney: If this is the old (i.e. pre-ICFP paper/correctness proof) version then I'd be inclined to archive it somewhere but remove from this PR.

If there is a reason to keep it, I'd suggest working out an interface that abstracts over which of LensHelpers and LensHelpersClassic is used, so that no other code winds up depending on anything specific to the implementation.

core/lensQueryHelpers.ml
```diff
@@ -0,0 +1,470 @@
...
+    failwith "Dummy database exec not supported."
+  method escape_string str =
+    "'" ^ Str.global_replace (Str.regexp "'") "''" str ^ "'"
+  method quote_field f =
+    "`" ^ Str.global_replace (Str.regexp "`") "``" f ^ "`"
+  method! make_insert_returning_query : (string * string list * string list list * string) -> string list =
+    failwith "Dummy database make_insert_returning_query not supported"
+end
+  
+let comma_seperated a b =
```
jamescheney: Where is this used?  I don't see it

core/lensQueryHelpers.ml
```diff
@@ -0,0 +1,470 @@
+open Types
+open Sugartypes
+open Operators
+open Utility
+open LensUtility
+open Value
+open LensRecordHelpers
+
+exception Runtime_error = Errors.Runtime_error
+
```
jamescheney: This overlaps to some extent with query.ml and queryshredding.ml, except that you are generating set-valued queries instead of multiset-valued ones.  I don't suggest any change for the time being but at some point we should extend the core query language with duplicate elimination and translate the RL-related queries to that instead of having two (or more) very similar query generators.

core/lensQueryHelpers.ml
```diff
@@ -0,0 +1,470 @@
...
+let construct_select_query (query : select_query) =
+  let db = query.db in
+  let colFn col = db#quote_field col.table ^ "." ^ db#quote_field col.name ^ " AS " ^ db#quote_field col.alias in 
+  let tableFn (table, alias) = db#quote_field table ^ " AS " ^ db#quote_field alias in
+  let cols = List.filter (fun c -> c.present) query.cols in
+  let cols = List.fold_left (fun a b -> a ^ ", " ^ colFn b) (colFn (List.hd cols)) (List.tl cols) in
+  let tables = query.tables in
+  let tables = List.fold_left (fun a b -> a ^ ", " ^ tableFn b) (tableFn (List.hd tables)) (List.tl tables) in
+  let sql = "SELECT " ^ cols ^ " FROM " ^ tables in
```
jamescheney: Should this be SELECT DISTINCT given that you want set-valued semantics?  Or does this never matter because the queries always return (at least) a key?

core/lensUtility.ml
```diff
@@ -0,0 +1,134 @@
...
+    include StringSet
+
+    let pp_pretty fmt cs = 
+        List.iter (fun c -> Format.pp_print_string fmt c; Format.pp_print_space fmt ()) (elements cs)
+end
+
+type colset = ColSet.t
+      [@@deriving show]
+
+module FunDep = struct
```
jamescheney: Could FunDep, FunDepSet and FunDepTree live in LensFDHelpers?

core/lensUtility.ml
```diff
@@ -0,0 +1,134 @@
+open Utility
+
+(* Lenses *)
+module ColSet = struct
```
jamescheney: ColSet seems like a very small delta over StringSet.  Would it make sense to just add this pretty-printing function to Utility.StringSet?

core/notfound.ml
```diff
@@ -53,6 +53,15 @@ end
 module List =
 struct
   include List
+
+  let find2 p l =
```
jamescheney: This doesn't seem to be used anywhere (and doesn't seem to do anything find doesn't).

core/notfound.ml
```diff
@@ -53,6 +53,15 @@ end
...
+    find p l
+
+  (* add for old OCaml < 4.0.5 *)
```
jamescheney: Is this still needed given that Links 0.7.3 requires OCaml > 4.0.5?

core/sugartypes.ml
```diff
@@ -218,6 +183,7 @@ and iterpatt = [
 ]
 and sec = [`Minus | `FloatMinus | `Project of name | `Name of name]
 and declared_linearity = [ `Lin | `Unl ]
+and fn_dep = string * string
```
jamescheney: This doesn't need to be part of the mutual recursion (though I see other things above that also don't need to be)

core/types.ml
```diff
@@ -133,11 +158,20 @@ type typ =
...
     | `Effect of row
     | `Table of typ * typ * typ
+    | `Lens of lens_sort 
     | `Alias of ((string * type_arg list) * typ)
     | `Application of (Abstype.t * type_arg list)
     | `MetaTypeVar of meta_type_var
     | `ForAll of (quantifier list ref * typ)
     | (typ, row) session_type_basis ]
+and lens_sort     = fundepset * lens_phrase option * (lens_col list)
+and lens_col      = {
```
jamescheney: Not sure I understand the reason for this type and using lens_col list instead of typ (where typ has to be a record / row type).  Could we use record/row types here, but disallow open rows in typechecking, like with tables?  Also not sure what "table" and "alias" are.

examples/relational_lenses/cds.links
```diff
@@ -0,0 +1,37 @@
+
+
+var db = database "links" "postgresql" "localhost:5432:links:links";
```
jamescheney: Config file?

unit/README.md
```diff
@@ -0,0 +1,14 @@
+To run the unit tests, call `./run-unit-tests`.
```
jamescheney: +1 to unit testing!  It would be good to also document (or point to external documentation) how to create new unit tests, since I'm sure there are other parts of Links that would benefit from this.

core/basicsettings.ml
```diff
@@ -129,6 +129,12 @@ module Shredding = struct
   let shredding = Settings.add_bool("shredding", false, `User)
 end
 
+module RelationalLenses = 
+struct
+    let relational_lenses = Settings.add_bool("relational_lenses", false, `User)
+    let classic_lenses = Settings.add_bool("relational_lenses_classic", false, `User)
```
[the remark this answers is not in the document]
rudihorn: This is more a poor choice of words. Classic refers to non incremental lenses, correct refers to the incremental version. I should probably rename this.

core/lensHelpers.ml
rudihorn: `LensHelpers` contains some of the core logic, while `LensUtility` is currently just datatypes that are standalone. `Value` and `Types` depend on `LensUtility`, while `LensHelpers` depends on `Value` and `Types`. `LensHelpersClassic` is the non-incremental implementation, while `LensHelpersCorrect` is the incremental one and both mainly just contain the put operations.

core/lensFDHelpers.ml
```diff
@@ -0,0 +1,104 @@
+open Utility
+open LensUtility
```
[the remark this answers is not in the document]
rudihorn: I like having them as submodules in a single Lens module, I'm less happy with putting them all in a single file. A single subdirectory seems nicer, but I'm not entirely sure how OCaml handles subfolders etc?. It is also difficult to have them as a seperate library (like core vs bin), since they depend on `Value.t` and `Types.typ`.

core/channelVarUtils.ml
```diff
@@ -86,6 +89,9 @@ let variables_in_computation comp =
         traverse_value v;
         traverse_stringmap (fun (_, c) ->
           traverse_computation c) clauses
+    | `LensDrop (v1, _, _, v2, _)
```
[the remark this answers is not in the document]
rudihorn: I was just reusing the other case classes, but otherwise not really.

core/evalir.ml
```diff
@@ -2,8 +2,10 @@ open Webserver_types
 open Ir
 open Lwt
 open Utility
+open LensUtility
 open Proc
 open Pervasives
+open LensHelpers
```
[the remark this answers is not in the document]
rudihorn: Will fix

core/lensQueryHelpers.ml
rudihorn: Can be deleted.

core/notfound.ml
rudihorn: Can be deleted now.

core/notfound.ml
rudihorn: Can also be deleted now.

core/sugartypes.ml
rudihorn: I'm happy to change this if it is preferred to be out of the mutual recursion.

core/types.ml
rudihorn: Columns have the name `alias` in links, and map to the table called `table` in the database which has a column `name`. This allows taking two tables with the same columns, and giving them differing aliases for instance.

examples/relational_lenses/cds.links
rudihorn: Will change this.

unit/README.md
rudihorn: Yes the OUnit documentation is admittedly not the best. I will try to maybe come around to this, but luckily doing simple unit testing is fairly straightforward (starts with `unit/unitTests.ml`). It would be nice if there was a better way to structure unit tests in submodules as currently I have all unit testing files start with UnitTests* to ensure they don't conflict with anything else.

core/parser.mly
yallop: Perhaps `ocamlyacc` could be invoked with `--strict` so that conflicts cause a build failure.

core/lensFDHelpers.ml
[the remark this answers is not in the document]
dhil: You can have your cake and eat it too. Here is how: Create a file `Lens.ml` that contains aliases of the modules Utility, FDHelpers, etc...
```ocaml
module Utility = LensUtility
module FDHelpers = LensFDHelpers
...
```
Create a signature file `Lens.mli` that includes the signatures of the modules Utility, FDHelpers, etc..
```ocaml
module Utility : sig
  include module type of type LensUtility
end

module FDHelpers : sig
  include module type of type LensFDHelpers
end
...
```

core/parser.mly
dhil: @yallop Good point! Thanks.

core/evalir.ml
```diff
@@ -590,6 +592,52 @@ struct
...
+    | `LensPut (lens, data, _rtype) ->
+        let _ = LensHelpers.ensure_lenses_enabled () in
+        let lens = value env lens in
+        let data = value env data in
+        let classic = Settings.get_value Basicsettings.RelationalLenses.classic_lenses in
+        if classic then
+            LensHelpersClassic.lens_put lens data 
+        else
+            LensHelpersCorrect.lens_put lens data;
+        apply_cont cont env (Value.box_unit ())
```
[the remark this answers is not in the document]
frank-emrich: yes, this looks like something where my approach would allow us define the evaluation of the Lens* constructors somewhere else.

core/sugartypes.ml
dhil: I seem to remember that type definition are recursive by default in OCaml. So
```ocaml
type t = int
and s = float
```
is only "marginally worse" than 
```ocaml
type t = int
type s = float
```
If you don't want a recursive type, you have to explicitly say so
```ocaml
type nonrec t = int
type nonrec s = float
```

core/basicsettings.ml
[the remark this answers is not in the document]
dhil: Do the "naive"/non-incremental lenses add any value? Is there a reason to ever use them rather than the incremental ones?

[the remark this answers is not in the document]
jamescheney: Rudi and I talked about this a bit the other day.  I think it is worth keeping both implementations for now:
- the "naive" version is what we compare against in the paper
- the "naive" version is less complicated and more directly
- I can imagine a future implementation strategy where we start by generating query expressions corresponding to the naive version and then incrementalize
- ultimately we would like to find an extensibility mechanism that allows us to define both naive and incremental versions more modularly (not in the Links core) 

That said, if there was a strong objection to including this we could probably live without it, as the reproducibility point is already But the naive version should certainly be documented as not something intended for end-use.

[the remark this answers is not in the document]
dhil: It is fine by me to include both. I suppose it was the name "naive" that alerted me, that we might be about to merge in a "dead-on-arrival" feature.

core/sugartypes.ml
jamescheney: @slindley, @SimonJF this is the part that Daniel was objecting to regarding the operators.  Please have a look.

core/parser.mly
jamescheney: @rudihorn is this checked/resolved?

core/lensFDHelpers.ml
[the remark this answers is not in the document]
jamescheney: @dhil we just discussed this in the meeting and decided to defer fine-tuning the organization for now; we can revisit later.

core/lensHelpersClassic.ml
jamescheney: This is the naive lenses implemtentaion which we want to keep (for now) for comparison purposes.

core/sugartypes.ml
SimonJF: This seems like a fairly standard DAG-ification refactoring that is necessary for this change.

core/parser.mly
rudihorn: `parser.mly` has been fixed and passes with the `--strict` tag.

jamescheney:
I am mostly happy except there still seem to be some random changes to permissions.

@dhil's review included some concerns about factoring out "operations.ml" from "sugarTypes.ml" which we discussed in the meeting and Sam and Simon said they'd have a look.  @dhil, do you still object to this and why?

Other than that and a couple of minor things this seems close to being ready to go.

SimonJF:
Reopened to activate Travis.

SimonJF:
The Travis unit tests seem to be failing: have you added `LensHelpersCorrect`?

core/sugartypes.ml
slindley: I don't see anything wrong with this. I'm guessing @dhil hadn't realised that the operators were pulled out because they are used in two places in order to avoid a cyclic dependency.

dhil:
I don't think I objected to the change regarding operators. I seem to remember adding it as a comment, and not as "request changes". Personally, I would probably have opted for another solution than splitting the syntax module.

I did request some changes, but I cannot see them any longer because @rudihorn force pushed. I think I just concerned about the hard wired database configuration in the unit test suite.

slindley:
This force push has made it quite hard to track the conversation around this pull request.

slindley:
On reflection, I think I now see @dhil's point. It seems that @rudihorn's changes requires breaking a cyclic dependency, but it may have been better to do so by pulling apart the experimental lens implementation rather than the core sugartypes.ml. This kind of issue is at the heart of the Skye project. Hopefully Skye will give rise to a more principled way of adding language extensions in future.

rudihorn:
I went through all issues and marked the ones where I have taken action as resolved. I'm sorry about the force pushing thing, that was probably not the best to do. The issue with moving settings out to config files have all been resolved. I've also added franks typechecking for the relational lens example and I think I have fixed the file permissions thing. I think the unit tests have been fixed (I had renamed a file and forgotten to test the unit testing), which should correct Travis failing.

unit/unitTestsLensPrimitives.ml
```diff
@@ -294,10 +318,17 @@ let test_join_lens_2 n test_ctx =
     let res = Query.filter (Query.lt 40 << Query.col "b") (lens_get l3 ()) in
     LensTestHelpers.print_verbose test_ctx (string_of_value (lens_get l3 ())); 
     LensTestHelpers.print_verbose test_ctx (string_of_value res);
+<<<<<<< HEAD
```
dhil: Looks like an unresolved merge conflict.

core/desugarSessionExceptions.ml
```diff
@@ -130,6 +129,56 @@ let contains_session_exceptions prog =
   (o#program prog)#satisfied
 
 
+(*
```
dhil: Looks like you are merging in @SimonJF's work.

unit/unitTestsLensPrimitives.ml
rudihorn: This should be fixed.

core/desugarSessionExceptions.ml
rudihorn: Damn. Wasn't paying attention while shuffling things around. I'll fix this tomorrow.

core/lensHelpers.ml
```diff
@@ -0,0 +1,302 @@
...
+        ) q2.tables in
+        let tables = List.append q1.tables q2.tables in
+        let cols = LensSort.cols sort in
+        if (q1.db <> q2.db) then
+            failwith "Only single database expressions supported."
+        else
+            {tables = tables; cols = cols; pred = get_lens_sort_pred sort; db = q1.db}
+  | _ -> failwith "Unsupported lens for query"
+
+(* BUG: Lists can be too big for List.map; need to be careful about recursion *)
```
dhil: It looks like you are mapping over a (potentially large) list of, say, length `N`. The `List.map` function uses `O(N)` stack space `O(N)`. So you indeed risk overflowing the stack for large lists. 

You can implement a space efficient `map` that uses `O(1)` stack space by using a "double fold". That is first transform the data using a `List.fold_left`, the elements will be reversed in the resulting list. You rectify this by using another `List.fold_left` on the result list.

What is the complexity of this? The time complexity is clearly `O(2N) = O(N)` (assuming that the transformation function is well-behaved and does not capture the list that we are mapping over). Since OCaml is a strict language, you have the property that `List.fold_left` is tail-recursive, allowing tail call optimisation to kick in, which means that `List.fold_left` uses `O(1)` space. 

I believe that the `List.rev_map` function is implemented in terms of `List.fold_left`, thus you ought to be able to implement the map described above as simple as

```ocaml
let space_efficient_map f xs = 
  List.(rev_map (fun x -> x) (rev_map f xs))
```

If you are willing to use side-effects and you know `N` in advance, then you can obviously do it in one linear pass using a for loop and mutation.

core/ir.ml
```diff
@@ -76,6 +76,12 @@ and binding =
 and special =
   [ `Wrong of Types.datatype
   | `Database of value
+  | `Lens of value * Types.lens_sort
+  | `LensDrop of value * string * string * value * Types.lens_sort
+  | `LensSelect of value * Types.lens_phrase * Types.lens_sort
+  | `LensJoin of value * value * string list * Types.lens_phrase * Types.lens_phrase * Types.lens_sort
```
dhil: I am somewhat skeptical about `lens_phrase` appearing here. It is the same representation that is used in the frontend. I would expect something that has been translated to appear here.

core/desugarSessionExceptions.ml
dhil: @frank-emrich's recent PR also appears in the delta.

core/lensHelpers.ml
jamescheney: If the lists here are being used/generated internally to represent sets of tuples (whose order doesn't matter) then just using rev_map would also be fine.

core/desugarSessionExceptions.ml
SimonJF: Git can be a joy sometimes :upside_down_face:

SimonJF:
Travis is failing because of `crop.links` no longer parsing correctly. Could you investigate?

jamescheney:
Approving since the remaining file permissions have been fixed.  So should be ready to merge once the keyword/parser fix is done.

jamescheney:
@rudihorn is this almost ready to merge?  Or if you don't have time do you object to someone else making the remaining minor changes?  I think we wanted to get this merged by now, so that the new version process can happen next week.

rudihorn:
I've just committed what may fix the problem. Will see what the result of travis is.

SimonJF:
Sounds good. If Travis passes then I'm happy to merge and release.

SimonJF:
CI has passed, and all reviewers approve. I'll now merge.

I imagine this is several years in the making: congratulations, @rudihorn!

PR #391, opened 2018-09-03
Execute custom scripts as part of test suite, used to validate example testing file

frank-emrich:
The run-tests file now executes all files in tests that end with .testscript, which allows adding little testscripts easily to the existing infrastructure.

Following a suggestion of @SimonJF, I've used this new mechanism to validate the typecheck_examples.tests file. We now impose that all .links files in the examples folder must occur in typecheck_examples.tests, even if they are just ignored. This has already detected a couple of files that were missing.


SimonJF:
Good idea! I like this approach. I'll properly review tomorrow.

tests/validate_typecheck_testfile.testscript
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
dhil: You could probably make this a commit hook instead. Or a sanity script that Travis runs.

frank-emrich: yes, but why? I think it is a desirable property that the test suite behaves exactly like what Travis does. This way I can run "make tests" before pushing and I'll know exactly what Travis will do and if it fails I can reproduce it 1:1 locally.

SimonJF: I prefer the proposed approach rather than a hook -- this approach tests the state of a system as it is being tested via the CI, as opposed to the state of the system when a commit is made.

A sanity script that's run by Travis is pretty much what this is, but without tying it into a particular CI. While there's a bit of "hand-rolling" going on, I think that's fine here.

tests/validate_typecheck_testfile.testscript
```diff
@@ -0,0 +1,13 @@
...
+
+typecheck_tests_file="tests/typecheck_examples.tests"
+example_folder="examples"
+
+while read file; do
+    if ! grep  -q -E "^$file$" "$typecheck_tests_file"; then
+        echo "FAILURE: Links example file $file is not mentioned in $typecheck_tests_file"
+        exit 1
+    fi
+done <<< "$(find "$example_folder" -iname "*.links")"
```
SimonJF: I would like some way of whitelisting directories. Could you make it so that, for example, there's either:

  1. Some convention, for example directories prefixed with `_`
  2. A whitelist file

where such directories are excluded from this check? A concrete use-case that already occurs is in a subdirectory of the `examples/AIPL14/LectureExamples` directory, where several files are purposely incomplete.

tests/validate_typecheck_testfile.testscript
SimonJF: Nit: as best practice, I tend to prefer `#!/usr/bin/env bash` to `#!/bin/bash`.

(I realise this convention isn't followed throughout the rest of the codebase, and it really is nit-picking)

tests/validate_typecheck_testfile.testscript
frank-emrich: A slight argument against this would be that then there are two places where we configure the type-checking tests: the .tests file and the validation script. So far the approach would have been that the examples not to be tested have to be included in the .tests file and set to ignore. But I do see the point that having to add many files just to ignore them is annoying and that it makes sense to have a menchanism to ignore whole folders in the validation script.

I would prefer a blacklist over a whitelist over a naming convention. Assume that someones adds a new folder of examples without adding them to the .tests file. In that case, my validation script would not complain, because the person wouldn't have added the new folder to the whitelist, either. So I guess it would be better to have a blacklist where people explicitly need to state that a certain folder need not be tested.

SimonJF: " So I guess it would be better to have a blacklist where people explicitly need to state that a certain folder need not be tested."

Yes, that's what I was thinking. So we'd add

`examples/AIPL/LectureExamples`

to a file, and then that folder would not be checked.

jamescheney: Can I vote against adding new, ad hoc pragmas hidden in comments?

rudihorn: I probably agree with @jamescheney on this. I like the idea of a `.no_typecheck_ir` file for the directory, or for single files to have an identically named empty file called `<filename>.no_typecheck_ir` or something.

tests/validate_typecheck_testfile.testscript
dhil: Good point. 

Might be worth mentioning why, namely, that writing `/usr/bin/env bash` in a shebang is portable whereas `/bin/bash` is not. However as we do not officially support any other platforms than Linux, the latter ought to work

tests/validate_typecheck_testfile.testscript
SimonJF: Yeah, I missed the bit about the pragma -- I'd be in favour of the `.no_typecheck_ir` though, which seems a nice solution.

tests/validate_typecheck_testfile.testscript
SimonJF: The latter will be fine. I think it's just habit that was beaten into me at St Andrews.

slindley: Given that it's a tiny change we should just insist on doing the portable thing everywhere.

tests/validate_typecheck_testfile.testscript
dhil: @jamescheney yes. I wasn't suggesting to implement the above, rather I was just outlining a possible alternative to the whitelist/blacklist discussed above.

That said, I would prefer a solution that doesn't involve a separate status file which needs to maintained. 

A possibly better solution is to bite the bullet and clean up the `examples/` directory once and for all, and subsequently enforce that every example has to pass the IR type checker. Then broken examples can either be deleted, moved somewhere else, kept in separate branches, etc. And tutorial materials can go in a separate directory, say `tutorials/`.

jamescheney: @dhil I agree, but my impression from @frank-emrich is that we are still some way off having a front-end that is capable of generating well-typed IR code in all cases that it should (i.e. existing Links programs that aren't obviously broken).  This may require work on both the frontend and IR.

frank-emrich: First of all, I would like to clarify that this whole test system is not (necessarily) for *IR* type-checking. The idea is to parse and *frontend*-type-check the examples and perform the backend transformations (which may later on involve IR type-checking as well). But still, the main idea for these tests is to make sure that any changes to Links do not break parsing or *frontend* type-checking of the examples.

With respect to the actual implementation of the tests: I agree with Daniel that ideally, we would remove all examples that currently fail (and are therefore ignored in the typecheck_examples.tests file). However, I think that the implementation of this testing feature should be as lightweight as possible. I do agree that it is cumbersome to have a big .tests file that lists all examples. However, I don't think that it would be advisable to invent yet another file format where you can specify exceptions from the testing and custom flags passed to individual examples, etc. For such a relatively simple testing feature, I would prefer having a big, explicit file using an existing format instead.

It may be a good compromise to adapt the script that validates the typecheck_examples.tests file such that whenver it detects that some example is not mentioned in the .tests file, it will output the necessary lines (assuming no special flags are required for running) that can be copy-pasted into the .tests file. This way, adding a new folder of examples does not involve manually creating a dozen new entries in the typecheck_examples.tests file.

frank-emrich:
I adressed the problems, the script now tells you what to do in case of an error. Let's see what Travis does (yay!) and then it's ready to go.

jamescheney:
So did travis integration magically just start working on its own or did someone turn it off and then on again?

SimonJF:
I fixed a small permissions error -- thankfully Travis is back now :)

SimonJF:
I've run this and am happy with the changes. I will merge once Travis passes.

PR #392, opened 2018-09-03
IR Type Checker

frank-emrich:
This PR contains the IR type checker. It lives in irCheck.ml and is invoked from backend.ml, which was merged in an earlier WIP PR. The type checker is disabled by default, the corresponding flag is typecheck_ir. There are currently still a lot of typing bugs around, so you better leave it switched off for now. I will continue fixing those bugs, but for now it seems desirable to have the current state merged in.

Design: 
The actual type checker is currently implemented as a sub-class of the existing type "reconstructor" found in IrTraversals.Transform. This is in order to make them interchangable.

Currently unsupported features:
- recursive types (pending the introduction of fold/unfold inserted by the frontend). This is currently reflected by the type equality test in irCheck returning true whenever it sees recursion.
- I'm still working on typing do/handle, but I'll commit that separately in the next week or so.
- There is currently no support for linearity or session types, meaning that linearity is not enforced and for session types, the type checker only does type reconstruction.

Other things in this PR:
I have reworked some of the pretty-printing of types. In particular, there is a new setting print_types_pretty. If it's enabled, printing of types is done by our own Types.Print module. Otherwise, printing is done by the (much more verbose) generated printing functions. I have finally worked out how to remove the cycles resulting from recursive types prior to invoking the generated printing functions. This means you can now use the string_of_* functions in Types and everything is taken care of for you.  In addition, printing a  Unionfind.point now shows you its phyiscal adress, in order to ease debugging the side effects during unification.


SimonJF:
Reopened to activate Travis.

core/irCheck.ml
```diff
@@ -75,7 +75,7 @@ end
 
 
 
-let eq_types : type_eq_context -> (Types.datatype * Types.datatype) -> bool =
+let eq_types occurence : type_eq_context -> (Types.datatype * Types.datatype) -> bool =
```
dhil: occurrence.

dhil:
Let's merge this in once you have rebased and updated the code correspondingly.

frank-emrich:
This is currently blocked by #399. I think it's okay to merge this without actual type-checking of lens constructs, but delegating to the type "reconstructor" instead. However, I think that my type-equality checker should have a proper implementation of type equality for lens types. And ideally, I would like to just re-use the implementation for #399 there.

dhil:
I think it is fine to merge as "feature incomplete". You can submit subsequent patches. Also, it shouldn't be only your responsibility to keep the IR typechecker up to date with what experimental features the rest of us have merged in.

SimonJF:
I agree with @dhil. I don't think it's necessarily blocked; lenses are currently an optional feature hidden behind a flag.

Why don't we merge this now, post-conflict-resolution, since the feature is useful as it is. We can open an issue connected to #399, and update the IR type checker once #399 is addressed.

frank-emrich:
Okay, I will rebase this today or tomorrow

frank-emrich:
It looks like everything is ready now

PR #395, opened 2018-09-11
Fixed closure conversion to generalize functions over their free type variables when hoisting

### LINKED ISSUE #384: Closure Conversion does not respect free type variables in hoisted functions
I had a look at the current closure conversion code and ran across a problem. Consider the following function

```
sig outer : ((a) {}~> a, [a]) {}~> [a]
fun outer(f, l) {
	fun twice (y){
		f(f(y))
	}
	map(twice,l)
}
```
which evokes that twice has type (a) {}~> a, where a refers to the type variable outer is quantified over (in particular, twice is monomorphic).
Our current implementation of closure conversion will hoist twice to the toplevel, as it should. However, it does not respect free type variables. What *should* happen is that when hoisting the function twice, it is made polymorphic, obtaining some type forall b. (b) {wild}-> b for twice. When calling twice from outer (after hoisting), we would then have to generate a type application that instantiates b as a.

For reference, the current result of closure conversion looks as follows. In particular, twice has type ((8264) {}~> 8264, where 8264 is an unbound type variable.

```
[`Fun (((2069, ((8264) {}~> 8264, "twice", `Global)),
         ([], [(2068, (8264, "", `Local))],
          ([`Let (((2066, (8264, "y", `Local)),
                   ([], `Return (`Variable (2068)))));
             `Let (((2067, (8264, "", `Local)),
                    ([],
                     `Apply ((`Project (("2065", `Variable (2073))),
                              [`Variable (2066)])))))
             ],
           `Apply ((`Project (("2065", `Variable (2073))), [`Variable (2067)])))),
         (Some (2073, ((2065:(8264) {}~> 8264), "env_2069", `Local))),
         `Unknown));
   `Fun (((2072,
           (forall 8280.((8280) {}~> 8280, [8280]) {}~> [8280], "outer",
            `Global)),
          ([8264],
           [(2070, ((8264) {}~> 8264, "", `Local));
             (2071, ([8264], "", `Local))],
           ([`Let (((2064, ([8264], "l", `Local)),
                    ([], `Return (`Variable (2071)))));
              `Let (((2065, ((8264) {}~> 8264, "f", `Local)),
                     ([], `Return (`Variable (2070)))))
              ],
            `Apply ((`TApp ((`TAbs (([8274; 8275; 8276],
                                     `TApp ((`Variable (285),
                                             [8274; { |8275 }; 8276])))),
                             [8264; { |wild }; 8264])),
                     [`Closure ((2069,
                                 `Extend (({2065 => `Variable (2065); }, None))));
                       `Variable (2064)])))),
          None, `Unknown))
   ]
```

frank-emrich:
I've adapted closures.ml such that not only free term variables are collected, but also free type variables occuring in functions. We then generalize the function over those free variables that are actually bound further up in the AST.
To make this type-check, I've changed the `Closure (f, v) constructor in IR values to  `Closure (f, targs, v), now carrying a list of type arguments. This means that `Closure now has a built-in type application. We cannot just enclose TApp (f, targs) in `Closure instead for the following reasons: 
- f is a variable, not a value. It is important to keep it that way, because the function f must be looked up in a map telling us the type of the closure environment it needs. In particular, the value `Variable f would not type-check to prevent using functions without providing the necessary closure environment.
- In order to type-check the closure, we need to know the type arguments that are applied.

I have implemented the necessary changes in the IR type checker. This PR is independent from #392, but I will commit an update to that PR to include the new code closure checking code, but commented out for now. 

This fixes #384

### DIFF
--- a/core/closures.ml
+++ b/core/closures.ml
@@ -43,11 +43,19 @@ struct
       method close_term x =
         if IntSet.mem x bound_term_vars then
           if IntMap.mem x fenv then
-            let zs = (IntMap.find x fenv).termvars in
-            List.fold_left
+            let freevars = IntMap.find x fenv in
+            let zs = freevars.termvars in
+            let typevars = freevars.typevars in
+            let o = List.fold_left
               (fun o (z, _) -> o#close_term z)
               o
-              zs
+              zs in
+            List.fold_left
+              (fun o (tv,_) ->
+                o#register_type_var tv
+              )
+              o
+              typevars
           else
             o
         else
--- a/core/closures.ml
+++ b/core/closures.ml
@@ -399,8 +399,8 @@ struct
             `Presence p
       ) tyvars
 
-  let close f zs tyvars =
-    `Closure (f, create_tyargs tyvars,`Extend (List.fold_right
+  let close f zs tyargs =
+    `Closure (f, tyargs, `Extend (List.fold_right
                             (fun (zname, zv) fields ->
                                StringMap.add zname zv fields)
                             zs
@@ -425,30 +425,34 @@ struct
         | `Variable x ->
           let x, t, o = o#var x in
 
-          let rec var_val x =
+          let rec var_val x : (Ir.value * Types.datatype ) =
             if IntSet.mem x cvars then
-              `Project (string_of_int x, `Variable parent_env)
+              `Project (string_of_int x, `Variable parent_env), t
             else if IntMap.mem x fenv then
               let zs = (IntMap.find x fenv).termvars in
               let tyvars = (IntMap.find x fenv).typevars in
-              match zs with
-              | [] -> `Variable x
+              match zs, tyvars with
+              | [], [] -> `Variable x, t
               | _ ->
+                let tyargs = create_tyargs tyvars in
+                let (remaining_type, instantiation_maps) = Instantiate.type_arguments_to_instantiation_maps false t tyargs in
+                let overall_type = Instantiate.datatype instantiation_maps remaining_type in
                 if List.mem_assoc x parents then
-                  `Closure (x, create_tyargs tyvars,`Variable parent_env)
+                  `Closure (x, tyargs,`Variable parent_env), overall_type
                 else
                   let zs =
                     List.map
                       (fun (z, _) ->
-                         let v = var_val z in
+                         let v = fst (var_val z) in
                          (string_of_int z, v))
                       zs
                   in
-                  close x zs tyvars
+                  close x zs tyargs, overall_type
             else
-              `Variable x
+              `Variable x, t
           in
-          var_val x, t, o
+          let overall_val, overall_type = var_val x in
+          overall_val, overall_type, o
         | v -> super#value v
 
       method set_context parents parent_env cvars =
--- a/core/closures.ml
+++ b/core/closures.ml
@@ -472,7 +472,6 @@ struct
         | [] -> [], o
         | b :: bs when Ir.binding_scope b = `Global ->
           let b, o = o#binding b in
-          (*o#print_bindings [b];*)
           let bs', o = o#pop_hoisted_bindings in
           let bs, o = o#bindings bs in
           bs' @ (b :: bs), o
@@ -656,7 +655,6 @@ struct
               ) xs [] in
             (*Debug.print ("function currently being hoisted, before instantiation:\n" ^ Ir.string_of_binding (`Fun (f, (tyvars, xs, body), z, location)));*)
             let body = IrTraversals.InstantiateTypes.computation (o#get_type_environment) inner_maps body in
-            (*o#print_bindings [(`Fun (f, (tyvars, xs, body), z, location))];*)
             (f, (tyvars, xs, body), z, location)
           end
 
--- a/core/closures.ml
+++ b/core/closures.ml
@@ -206,7 +206,7 @@ struct
           (* reset free and bound variables to be empty *)
           let o = o#reset in
 
-          let f, o = o#binder f in
+
           let o = List.fold_left (fun o q -> o#quantifier q) o tyvars in
           let (xs, o) =
             List.fold_right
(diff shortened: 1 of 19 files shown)


core/closures.ml
```diff
@@ -407,6 +564,87 @@ struct
           let bs, o = o#bindings bs in
           b :: bs, o
 
+
+      (** Given a list of free variables, return a tuple containing the following:
+        - a list of fresh quantifiers, each corresponding to one free variable
+        - Three maps mapping the old free variables to fresh ones (to be used with Instantiate)  **)
+      method create_substitutions_replacing_free_variables_with_fresh_ones (free_type_vars : Types.quantifier list)  =
```
jamescheney: This is a very long method name...

core/instantiate.ml
```diff
@@ -297,22 +297,29 @@ let datatype = instantiate_datatype
 
 module SEnv = Env.String
 
-let apply_type : Types.datatype -> Types.type_arg list -> Types.datatype =
-  fun pt tyargs ->
+let type_arguments_to_instantiation_maps : bool -> Types.datatype -> Types.type_arg list -> (datatype * (datatype IntMap.t * row IntMap.t * field_spec IntMap.t)) =
```
jamescheney: Consider putting the types on the next line(s)

core/instantiate.ml
```diff
@@ -297,22 +297,29 @@ let datatype = instantiate_datatype
...
+    if (not arities_okay) then
         (Debug.print (Printf.sprintf "# Type variables (total %d)" (List.length vars));
          let tyvars = String.concat "\n" @@ List.mapi (fun i t -> (string_of_int @@ i+1) ^ ". " ^ Types.show_quantifier t) vars in
          Debug.print tyvars;
          Debug.print (Printf.sprintf "\n# Type arguments (total %d)" (List.length tyargs));
          let tyargs' = String.concat "\n" @@ List.mapi (fun i arg -> (string_of_int @@ i+1) ^ ". " ^ Types.show_type_arg arg) tyargs in
          Debug.print tyargs';
          raise ArityMismatch);
+
+    let vars, remaining_quantifiers = if tyargs_length = vars_length
```
jamescheney: Suggest putting the "if" on the next line

core/irTraversals.ml
```diff
@@ -226,11 +226,21 @@ struct
             let (args, _, o) = o#list (fun o -> o#value) args in
               (* TODO: check arg types match *)
               `ApplyPure (f, args), deconstruct return_type ft, o
-        | `Closure (f, z) ->
+
+        | `Closure (f, tyargs, z) ->
+
             let (f, t, o) = o#var f in
+            let t = if tyargs = []
```
dhil: Consider using pattern matching rather than if-then-else. I'd probably format the code like this:
```ocaml
let t =
  match tyargs with
  | [] -> t
  | _ ->
     let (remaining_type, instantiation_maps) =
        Instantiate.type_arguments_to_instantiation_maps false t tyargs
     in
     Instantiate.datatype instantiation_maps remaining_type
in
...
```

frank-emrich:
@jamescheney This PR is relatively independent from #392. In particular, they can both be compiled on top of the current master. The only overlap is that this PR changes the type-checking of Closure. I have committed the necessary changes to #392, but commented them out so that everything still compiles independently. This means that we can merge either of those two PRs first and the other one can then easily be adapted.

dhil:
>  The only overlap is that this PR changes the type-checking of Closure. I have committed the necessary changes to #392, but commented them out

Clearly the right thing to do is to merge this first, and then afterwards merge in the type checker. It seems silly to commit a chunk of commented code.

jamescheney:
OK, sounds good to me.

core/instantiate.ml
```diff
@@ -329,7 +339,15 @@ let apply_type : Types.datatype -> Types.type_arg list -> Types.datatype =
                         mapstrcat ", " (fun t -> Types.string_of_type_arg t) tyargs))
         vars tyargs (IntMap.empty, IntMap.empty, IntMap.empty)
     in
-      instantiate_datatype (tenv, renv, penv) t
+    if remaining_quantifiers = [] then
```
dhil: I think it is cleaner to use pattern-matching instead of a bare if-then-else here.

core/instantiate.ml
```diff
@@ -298,22 +298,32 @@ let datatype = instantiate_datatype
...
-    let t, vars =
-      match concrete_type pt with
-        | `ForAll (vars, t) -> t, Types.unbox_quantifiers vars
-        | t -> t, [] in
-    let tenv, renv, penv =
-      if (List.length vars <> List.length tyargs) then
+    let vars, t = TypeUtils.split_quantified_type pt in
+    let tyargs_length = List.length tyargs in
+    let vars_length = List.length vars in
+    let arities_okay = if must_instantiate_all_quantifiers
```
dhil: Put the `if` on a new line. You can format it like
```ocaml
let arities_okay =
  if must_instantiate_all_quantifiers
  then tyargs_length = vars_length
  else tyargs_length <= vars_length
```

core/instantiate.ml
```diff
@@ -298,22 +298,32 @@ let datatype = instantiate_datatype
 
 module SEnv = Env.String
 
-let apply_type : Types.datatype -> Types.type_arg list -> Types.datatype =
-  fun pt tyargs ->
+let type_arguments_to_instantiation_maps :
+      bool -> Types.datatype -> Types.type_arg list -> (datatype * (datatype IntMap.t * row IntMap.t * field_spec IntMap.t)) =
```
dhil: Consider introducing a type alias for `datatype IntMap.t * row IntMap.t * field_spec IntMap.t` to give it an informative name.

core/instantiate.ml
```diff
@@ -298,22 +298,32 @@ let datatype = instantiate_datatype
 
 module SEnv = Env.String
 
-let apply_type : Types.datatype -> Types.type_arg list -> Types.datatype =
-  fun pt tyargs ->
+let type_arguments_to_instantiation_maps :
```
dhil: I suppose the "proper OCaml name" would be `instantiation_maps_of_type_arguments`.

jamescheney:
This looks fairly reasonable now, is it ready?  @dhil are your comments addressed?

dhil:
@jamescheney sure, I am happy. I only made some comments about style.

PR #401, opened 2018-09-17
Amend type check validation script

dhil:
It is mildly annoying that the `run-tests` script fails locally because some of my examples are not under revision control. Thus this PR modifies the type check validation script such that it only errors for tracked examples.


frank-emrich:
Looks good to me

PR #402, opened 2018-09-18
Move manual install/remove steps to dune

mseri:
`dune` does not yet support dynamic install targets, I have replaced those
with the preliminary generation of the necessary install stanzas. These
are written in install.inc that is then included in the dune file. An
empty install.inc is kept in the root to make sure that the project
compiles even outside the `opam` installation environment.

Signed-off-by: Marcello Seri <marcello.seri@gmail.com>


mseri:
This is more of a proof of concept, and maybe it's not worth doing it until dune is able to dynamically generate the install stanzas.

dhil:
Thank you Marcello! This looks like the sort of thing we need in order to comply with the OPAM linter.

PR #406, opened 2018-10-01
Less invasive lenses

dhil:
This PR removes the all the top-level `open Lens*` from the code base. It also gets rid of all the redundant references to `LensHelpers.ensure_lenses_enabled`.

The lenses implementation is still somewhat invasive as the front-end syntax and IR still depend on the lens implementation (it ought to be the other way around!). Fixing this issue is left for a future PR.

I will leave it to @rudihorn to clean up the Lens API.


core/lens/lensHelpers.ml
```diff
@@ -13,13 +13,7 @@ let query_count = ref 0
...
   (if false then print_endline message)
 
-let ensure_lenses_enabled () =
-  if Settings.get_value Basicsettings.RelationalLenses.relational_lenses then
-    ()
-  else
-    failwith "Code uses relational lenses, but relational lenses are not enabled. Please set the relational lenses flag."
-
-module Lens = struct
+module Lens' = struct
```
jamescheney: Is it a good idea to have primes in a module name?  (Or is there some reason it's bad to ahve "Lens.Helpers.Lens"?  

I tend to agree that "Lens" is not a very descriptive name for this submodule, anyway.

dhil: I agree that the module `Lens'` should have another name, but I will let @rudihorn decide on the API. I don't like the primed name either, but it needs some name that is different from `Lens` to avoid a clash with the encompassing module `Lens`.

PR #414, opened 2018-11-06
Do not use "wrapped false" in the dune files

### LINKED ISSUE #350: Avoid OCaml reserved module names
Some modeule names (Env, Parser, Parse) are also used in the OCaml compiler, which will cause problems when linking with the OCaml compiler library.

We should change or qualify these names to avoid conflicts.

frank-emrich:
This means that modules are not exported at the top-level of the module hierarchy anymore, but instead wrapped under the respective library name.

fixes #350 (somewhat)


jamescheney:
Hmm.  Presumably that means that whatever qualified name munging Dune is doing doesn't expect us to do something that stupid.  I'd tend to lean towards naming the source cod files corresponding to toplevel executable(s) with more functional names (main, toplevel, repl, compile, etc.)  

We are never importing the Links "module" anywhere else, right?

frank-emrich:
You mean the module corresponding to links.ml? No, we don't use that anywhere. Since links.ml  is declared to be part of an "executable" (whereas the files in core/ are part of a "library"), I am not sure if or how the corresponding module is exported. I just assume that seeing a reference to Links in links.ml confuses dune's internal magic.

SimonJF:
I think that going with `Links_core` instead of having to rename files in the source tree to allow the use of `compilerlibs` is a worthy tradeoff.

@dhil -- being someone who has actually used compilerlibs and Links together (albeit before Dune?), do you foresee any difficulties here?

frank-emrich:
@dhil says he sees no problem with this, so I guess we should go ahead with it

jamescheney:
Looks good to me, just needs to be updated wrt master.

jstolarek:
Sorry for joining in late, but I just realized that this breaks Menhir for reasons I don't understand. When trying to compile the code I get:

> Module Parser in directory /dane/projekty/links/_build/default/core depends on Links_core.
> This doesn't make sense to me.
>
> Links_core is the main module of the library and is the only module exposed
> outside of the library. Consequently, it should be the one depending
> on all the other modules in the library.

And the same for the other two parsers (JSON, XML). Looking at the generated parser file it is indeed the case that it now prefixes types with `Links_core`, e.g.:

```OCaml
let _v : (Links_core.Sugartypes.bindingnode Links_core.Sugartypes.with_pos) = let _endpos = _endpos__6_ in
```

Any ideas what is happening? I'm clueless. My Menhir migration code is on `js-T411-menhir-migration` branch.

frank-emrich:
Modules that are part of a library X should never access som Foo that is also in  X by X.Foo, but use Foo directly. If the generated files do that, we need to find a workaround

frank-emrich:
Here's an ugly workaround that fixes parser.mly: Put the following in line 30, right under the open statements:

```module Links_core =
struct
  module Sugartypes = Sugartypes
  module Types = Types
  module Operators = Operators
end
```

frank-emrich:
Good news, using the same workaround for Jsonparse and Xmlparser fixes your problem and makes your branch compile. You just need an additional `module Value = Value` in the struct for those two files

jstolarek:
Works indeed, thanks!

Issue #410, opened 2018-10-15
Move to OPAM 2.0


SimonJF:
With the release of OPAM 2.0, we will need to update the format of our OPAM files before we do the next release.

jamescheney:
just to further +1 this, at the moment Links depends transitively on a ssl package that doesn't install correctly on MacOS and the fix is an OPAM 2.0 package.

slindley:
I think MacOS is a red herring. If I understand correctly, that version of that package no longer installs anywhere because the package maintainer violated the basic assumptions of OPAM by updating a URL in an already released package to point to a non-existent file - @SimonJF can tell you all of the gory details.

jamescheney:
Ugh. Upgrading to opam 2 fixed the problem, (modulo blowing away opam 1.2 cruft) anyway.

SimonJF:
The package owner has re-uploaded `ssl` at the previous URL, so we shouldn't have any more issues with it thankfully.

Nonetheless, OPAM 2.0 compatibility is a priority.

jstolarek:
It makes me sad that due to a bug OPAM 2.0 does not work on a setup where `~/.opam` is a symlink to a directory on a partition with more disk space:

https://github.com/ocaml/opam/issues/3622

Am I the only one affected?

frank-emrich:
This should be fixed by #416, but @SimonJF would have to test at some point if we can actually build releases with opam2. I'm closing this for now.

Issue #54, opened 2016-08-01
Create Vagrant image to simplify Links development and deployment


jgbm:

jamescheney:
I have pretty much done this now (at least in a simple form) as part of making MSc students able to use Links.  Simon, maybe we should discuss.

jamescheney:
See also:
https://github.com/links-lang/links/wiki/Installing-Links-on-a-fresh-VM

jamescheney:
I will do this when I get a chance.

jamescheney:
Done: https://github.com/links-lang/links-tutorial

PR #416, opened 2018-11-07
Use opam 2.0 format in .opam files

frank-emrich:
This should move us towards #410. I am, however, unsure if we can now produce a release with opam 2 since I have insufficient knowledge of how that works.


frank-emrich:
The build server uses the opam version 1.x found in its distribution's repository. It looks like this gives us two options:
1) Keep the .opam files compatible with opam 1.x for now
2) adapt the travis script to manually download opam 2 as part of the initialisation

jamescheney:
It looks like we are getting the OPAM package from 

https://launchpad.net/~avsm/+archive/ubuntu/ppa

which hasn't been updated since 4.02.  I vaguely recall that we had to do this to get a stable version of camlp4. (I'm not sure if we depend anymore on camlp4 anyway.)

See also: 
https://askubuntu.com/questions/1080715/opam-2-0-on-ubuntu

frank-emrich:
Yes, it seems weird that the author claims his PPA had opam2 when in fact the PPA hasn't been updated in ages. However, I've noticed that opam2 only consists of a single executable, making its manual installation easy. We can just obtain the 2.0 release via wget/curl from github and execute it directly. Still hack-ish, but it just means adding a few lines to the Travis script or our Makefile until opam2 has landed in someone's PPA

jamescheney:
Hmm.  In links-tutorial I just do apt-get install opam (in xenial) which brings opam 1.2.2.  However it doesn't appear opam 2 has made it into any ubuntu distributions yet.  Welcome to the bleeding edge.

Maybe we should ask Anil if there's a reason the ppa is not being maintained (such as: there's a better way of doing the same thing)

frank-emrich:
In the meantime, I have reached out to avsm, the maintainer of the opam/ocaml PPA. He says that there won't be  opam2 packages for Ubuntu versions prior to bionic/18.04, because they are missing another package that is now required by opam2. The Ubuntu version currently offered by Travis CI is trusty/14.04 (quite old, actually). This leaves us with two alternatives:
1. As part of the travis setup, download the opam2 executable from github (no further setup required)
2. There are docker images of opam2 that contain all necessary dependencies, which we could use

I've tried the first alternative and it seems to work. If we decide to go for docker I would kindly ask @SimonJF  to do the setup, because he seems to have more experience with docker than me.

jamescheney:
Ah, OK.  So this is like when you have one foot on one boat and another foot on another boat and they are going different speeds.

jamescheney:
OK, but let's merge this when OPAM2/Travis are happy and make a note/issue to the effect that Dockerizing things at some point would be nice.  See also #54 (s/Vagrant/Docker)

frank-emrich:
Okay, I'll give people another day to object because merging this breaks compability with opam 1.x, effectively forcing everyone to upgrade

jamescheney:
It seems to me that this change will only require installing opam2 when creating a release (which is now the only option anyway), but people should upgrade to opam2 anyway.

Issue #411, opened 2018-10-19
Migrate the main parser to Menhir


dhil:
Our main parser / grammar `parser.mly` could do with some cleaning.

Some of the production rules such as `preamble` are essentially garbage. We should get rid of rules like those. In general we should probably simplify/refactor our grammar. The best solution might be to migrate the main parser from ocamlyacc to Menhir, since Menhir provides some convenient combinators that enable us to get rid of most (if not all) of the `_list` and `_perhaps` productions.

In doing so, we might want to consider whether my favourite Links program should remain syntactically valid
```links
links> {;};
() : ()
```

frank-emrich:
I just saw that Menhir has built-in support for grammar modularity (cf. [http://gallium.inria.fr/~fpottier/menhir/manual.html#sec%3Asplit]). This means that the definition of the overall  grammar can be split over multiple files and rules for shared non-terminals are merged. Other parser generators have more advanced modularity features, but Menhir's basic support for modularity is certainly nice to have!

jamescheney:
I don't have the feeling that this is high on anyone's to-do list at the moment.  It also seems like #411 ought to be mostly orthogonal to #187, so as long as people aren't working on them at the same time there shouldn't be a problem.  (Also, #187 seems higher priority and might give us useful information about how / if we want to move to Menhir / refactor the parser)

dhil:
I agree with @jamescheney. Migration to Menhir is unlikely to happen anytime soon. Though, I firmly believe that Menhir would simplify and generally make it easier to work with our parser(s).

jstolarek:
One nice feature to have would be the ability to record source positions of various elements of a production. Say I have a production:

```
foo:
| A B C       { ... }
```

Right now we only have access to `pos()`, which I understand records positon for the whole production. I would like to be able to say, e.g. `pos $2` to record position of `B`.

A separate questions are: 1) which elements of a parse tree should be annotated with position and; 2) how productions should attach position information. Re 1 I think it would make sense to be consistent and produce a fully annotated tree, i.e. every parsed element should carry a location information. Re 2, parser would probably be easier to understand and manage if we had a convention that every production returns an annotated element. So instead of having:

```
foo:
| A   { Foo }
| B   { Bar }

bar:
| C   { Quux }
| D   { Quuz }

baz:
| foo bar   { Baz ($1, pos $1) ($2, pos $2) }
```

we should have:

```
foo:
| A   { Foo, pos() }
| B   { Bar, pos() }

bar:
[...]

PR #417, opened 2018-11-13
Display source position for unbound type constructors

### LINKED ISSUE #187: No line number / position on error: Unbound type constructor
```
*** Fatal error : Unbound type constructor Foo
```
should give a position where the error occurred, preferably with filename, line and position

jstolarek:
This PR fixes #187.

Datatype declarations are now annotated with source code position. Following an already existing pattern in `sugartypes.ml`:

```
type phrasenode = [....]
and pharse = phrasenode * position

type patternnode = [....]
and pattern = patternnode * position
```
we now have:
```
type datatypenode = [....]
and datatype = datatypenode * position
```
with appropriate changes made to sugarTraversals. Positions are propagated to a place where an error message for unbound type constructor is raised. If I write a program:

```


sig f : (Foo) -> Foo
fun f(x) { x }
```

I now get an error message:

```
*** Fatal error : T187.links:3: Unbound type constructor Foo
```

It seems to me - and please correct me if I'm wrong - that we don't really have error messages that display source locations, so I took a liberty of designing my own format of the message.

I also removed some unused arguments from a couple of functions in the parser (separate commit).

I have some thoughts on how we could improve our parser. #411 looks like a good place to share them.

### DIFF
--- a/core/parser.mly
+++ b/core/parser.mly
@@ -7,11 +7,11 @@ open Sugartypes
 
 let type_variable_counter = ref 0
 
-let fresh_type_variable : subkind option -> datatype =
+let fresh_type_variable : subkind option -> datatypenode =
   function subkind ->
     incr type_variable_counter; `TypeVar ("_" ^ string_of_int (!type_variable_counter), subkind, `Flexible)
 
-let fresh_rigid_type_variable : subkind option -> datatype =
+let fresh_rigid_type_variable : subkind option -> datatypenode =
   function subkind ->
     incr type_variable_counter; `TypeVar ("_" ^ string_of_int (!type_variable_counter), subkind, `Rigid)
 
@@ -983,9 +983,9 @@ just_datatype:
 | datatype END                                                 { $1 }
 
 datatype:
-| mu_datatype                                                  { $1 }
-| straight_arrow                                               { $1 }
-| squiggly_arrow                                               { $1 }
+| mu_datatype                                                  { $1, pos() }
+| straight_arrow                                               { $1, pos() }
+| squiggly_arrow                                               { $1, pos() }
 
 arrow_prefix:
 | LBRACE RBRACE                                                { ([], `Closed) }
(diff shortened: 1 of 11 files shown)


core/refineBindings.ml
```diff
@@ -349,7 +349,7 @@ module RefineTypeBindings = struct
         if rts || List.length sccs > 1 then
           let muName = gensym ~prefix:"refined_mu" () in
             ((tyName, muName) :: env, `Mu (muName, sugaredDT))
-        else (env, sugaredDT) in
+        else (env, fst sugaredDT) in
```
jstolarek: This `fst` feels a bit "meh". It would probably be better to have a function like `un_position` or something to drop position annotation. I'll probably change that before a final merge.

core/parser.mly
```diff
@@ -1281,7 +1281,7 @@ regex_pattern_sequence:
  */
 pattern:
 | typed_pattern                                             { $1 }
-| typed_pattern COLON primary_datatype                      { (`HasType ($1, datatype $3), pos()) }
+| typed_pattern COLON primary_datatype                      { (`HasType ($1, datatype ($3, pos())), pos()) }
```
jstolarek: This new usage of `pos()` is incorrect - it annotates `$3` with position of the whole production, rather than just with position of `$3`. Fixing it would require pushing `pos()` into datatype production. Since this would require a lot of clutter I decided to avoid it for now - but see https://github.com/links-lang/links/issues/411#issuecomment-438246200

jstolarek:
> See format_exception in errors.ml.

Oh, I like that. So now it seems better to me to define a new type of exception, raise it during desugaring, and handle error formatting inside `errors.ml`. Which I will do, unless there are objections?

core/sugartypes.ml
```diff
@@ -92,6 +92,7 @@ type datatype =
   | `Choice          of row
   | `Dual            of datatype
   | `End ]
+and datatype = datatypenode * position
```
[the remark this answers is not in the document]
jstolarek: Yes, this sounds like a good idea, but not in scope of this PR. I'll create a new ticket for it.

frank-emrich:
Some of these errors are actually type errors in my opinion (e.g., "Unbound type constructor"). I guess they could re-use the existing exception for type errors. This should also make it easier to migrate to using the gripers there, too, later on.

jstolarek:
It is a type error, but the error message we would get from reusing `Type_error` would not be helpful in this case. It would say "Type error in expression Foo" without saying that the error comes from `Foo` being undefined.

frank-emrich:
Oh, okay, I agree that this message would be useless. Feel free to define a new exception for this case then

jstolarek:
I updated the PR - now there's a new exception type to raise when an unbounded tycon is encountered.

jstolarek:
Merged as d34644c2

Issue #418, opened 2018-11-13
Refactor attaching of position to source nodes


jstolarek:
In `sugartypes` we have a recurring pattern of annotating a datatype with source code location that corresponds to it:

```
type phrasenode = [....]
and pharse = phrasenode * position

type patternnode = [....]
and pattern = patternnode * position

type datatypenode = [....]
and datatype = datatypenode * position
```

We could refactor it like this:

```
type ('a) with_pos = {node:a, pos:pos}

type phrase = phrasenode with_pos
type pattern = patternnode with_pos
type datatype = datatypenode with_pos
```
This would make projections easier.

jamescheney:
I think it's a pretty standard workaround for the lack of something like "traits" for datatypes  in ML-like languages

PR #419, opened 2018-11-14
Refactor nodes with position

jstolarek:
This is a WIP on #418.

I did some prototype work on defining a separate datatype to attach source position to various syntax tree nodes (patterns, phrases, etc.). So far I have refactored definitions of `datatype` and `pattern`. The first one was easy, but the second one was quite a lot of work. There are several other datatypes that will most likely require a lot more work. I would be happy to hear your opinions at this point, before I commit to further work. Do you think this should be done differently? Do you think this shouldn't be done at all?

One thing that I'm planning to change is defining a separate function for attaching a dummy position to a node. This is a common pattern in various desugar* modules and it is worth having a function for it.

A separate thought is that if desugaring discards so many source positions and replaces them with dummies then we perhaps we could rethink our data representations. But I'd say this is not in scope of this ticket/PR.

This PR is based on #417 work.

### DIFF
--- a/core/chaser.ml
+++ b/core/chaser.ml
@@ -71,7 +71,7 @@ let rec add_module_bindings deps dep_map =
     | [module_name]::ys ->
       (try
         let (bindings, _) = StringMap.find module_name dep_map in
-        Sugartypes.mkWithDPos (`Module (module_name, bindings)) :: (add_module_bindings ys dep_map)
+        Sugartypes.with_dummy_pos (`Module (module_name, bindings)) :: (add_module_bindings ys dep_map)
       with Notfound.NotFound _ ->
         (failwith (Printf.sprintf "Trying to find %s in dep map containing keys: %s\n"
           module_name (print_list (List.map fst (StringMap.bindings dep_map))))));
--- a/core/chaser.ml
+++ b/core/chaser.ml
@@ -71,7 +71,7 @@ let rec add_module_bindings deps dep_map =
     | [module_name]::ys ->
       (try
         let (bindings, _) = StringMap.find module_name dep_map in
-        (`Module (module_name, bindings), Sugartypes.dummy_position) :: (add_module_bindings ys dep_map)
+        Sugartypes.mkWithDPos (`Module (module_name, bindings)) :: (add_module_bindings ys dep_map)
       with Notfound.NotFound _ ->
         (failwith (Printf.sprintf "Trying to find %s in dep map containing keys: %s\n"
           module_name (print_list (List.map fst (StringMap.bindings dep_map))))));
--- a/core/compilePatterns.ml
+++ b/core/compilePatterns.ml
@@ -75,7 +75,7 @@ let lookup_name name (nenv, _tenv, _eff, _penv) =
 let lookup_effects (_nenv, _tenv, eff, _penv) = eff
 
 let rec desugar_pattern : Ir.scope -> Sugartypes.pattern -> pattern * raw_env =
-  fun scope (p, pos) ->
+  fun scope {Sugartypes.node=p; Sugartypes.pos} ->
     let pp = desugar_pattern scope in
     let empty = (NEnv.empty, TEnv.empty, Types.make_empty_open_row (`Any, `Any)) in
     let (++) (nenv, tenv, _) (nenv', tenv', eff') = (NEnv.extend nenv nenv', TEnv.extend tenv tenv', eff') in
@@ -93,10 +93,10 @@ let rec desugar_pattern : Ir.scope -> Sugartypes.pattern -> pattern * raw_env =
             let p, env = pp p in
             let ps, env' = pp ps in
               `Cons (p, ps), env ++ env'
-        | `List [] -> pp (`Nil, pos)
+        | `List [] -> pp (Sugartypes.mkWithPos `Nil pos)
         | `List (p::ps) ->
             let p, env = pp p in
-            let ps, env' = pp (`List ps, pos) in
+            let ps, env' = pp (Sugartypes.mkWithPos (`List ps) pos) in
               `Cons (p, ps), env ++ env'
         | `Variant (name, None) -> `Variant (name, `Any), empty
         | `Variant (name, Some p) ->
@@ -130,8 +130,8 @@ let rec desugar_pattern : Ir.scope -> Sugartypes.pattern -> pattern * raw_env =
             in
               `Record (bs, p), env
         | `Tuple ps ->
-            let bs = mapIndex (fun (p, pos) i -> (string_of_int (i+1), (p, pos))) ps in
-              pp (`Record (bs, None), pos)
+            let bs = mapIndex (fun p i -> (string_of_int (i+1), p)) ps in
+              pp (Sugartypes.mkWithPos (`Record (bs, None)) pos)
         | `Constant constant ->
             `Constant constant, empty
         | `Variable b ->
--- a/core/compilePatterns.ml
+++ b/core/compilePatterns.ml
@@ -79,12 +79,12 @@ let rec desugar_pattern : Ir.scope -> Sugartypes.pattern -> pattern * raw_env =
     let pp = desugar_pattern scope in
     let empty = (NEnv.empty, TEnv.empty, Types.make_empty_open_row (`Any, `Any)) in
     let (++) (nenv, tenv, _) (nenv', tenv', eff') = (NEnv.extend nenv nenv', TEnv.extend tenv tenv', eff') in
-    let fresh_binder (nenv, tenv, eff) =
-      function
-        | (name, Some t, _) ->
-            let xb, x = Var.fresh_var (t, name, scope) in
-              xb, (NEnv.bind nenv (name, x), TEnv.bind tenv (x, t), eff)
-        | _ -> assert false
+    let fresh_binder (nenv, tenv, eff) bndr =
+      assert (Sugartypes.binder_has_type bndr);
+      let name = Sugartypes.name_of_binder bndr in
+      let t = Sugartypes.type_of_binder_exn bndr in
+      let xb, x = Var.fresh_var (t, name, scope) in
+      xb, (NEnv.bind nenv (name, x), TEnv.bind tenv (x, t), eff)
     in
       match p with
         | `Any -> `Any, empty
@@ -93,10 +93,10 @@ let rec desugar_pattern : Ir.scope -> Sugartypes.pattern -> pattern * raw_env =
             let p, env = pp p in
             let ps, env' = pp ps in
               `Cons (p, ps), env ++ env'
-        | `List [] -> pp (Sugartypes.mkWithPos `Nil pos)
+        | `List [] -> pp (Sugartypes.with_pos `Nil pos)
         | `List (p::ps) ->
             let p, env = pp p in
-            let ps, env' = pp (Sugartypes.mkWithPos (`List ps) pos) in
+            let ps, env' = pp (Sugartypes.with_pos (`List ps) pos) in
               `Cons (p, ps), env ++ env'
         | `Variant (name, None) -> `Variant (name, `Any), empty
         | `Variant (name, Some p) ->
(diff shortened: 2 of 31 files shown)


core/chaser.ml
```diff
@@ -71,7 +71,7 @@ let rec add_module_bindings deps dep_map =
     | [module_name]::ys ->
       (try
         let (bindings, _) = StringMap.find module_name dep_map in
-        (`Module (module_name, bindings), Sugartypes.dummy_position) :: (add_module_bindings ys dep_map)
+        Sugartypes.mkWithDPos (`Module (module_name, bindings)) :: (add_module_bindings ys dep_map)
```
dhil: Please rename to `Sugartypes.with_dummy_pos`

core/compilePatterns.ml
```diff
@@ -79,12 +79,12 @@ let rec desugar_pattern : Ir.scope -> Sugartypes.pattern -> pattern * raw_env =
...
     let (++) (nenv, tenv, _) (nenv', tenv', eff') = (NEnv.extend nenv nenv', TEnv.extend tenv tenv', eff') in
-    let fresh_binder (nenv, tenv, eff) =
-      function
-        | (name, Some t, _) ->
-            let xb, x = Var.fresh_var (t, name, scope) in
-              xb, (NEnv.bind nenv (name, x), TEnv.bind tenv (x, t), eff)
-        | _ -> assert false
+    let fresh_binder (nenv, tenv, eff) bndr =
+      assert (Sugartypes.bndrHasTy bndr);
+      let name = Sugartypes.bndrName bndr in
```
dhil: Please rename `Sugartypes.bndrName` to `Sugartypes.name_of_binder` or consider introducing a `Binder` module with a `to_name` function within `Sugartypes`.

core/compilePatterns.ml
```diff
@@ -79,12 +79,12 @@ let rec desugar_pattern : Ir.scope -> Sugartypes.pattern -> pattern * raw_env =
...
+      let t = Sugartypes.bndrTy bndr in
```
dhil: `Sugartypes.type_of_binder`

core/desugarCP.ml
```diff
@@ -36,50 +36,50 @@ object (o : 'self_type)
             let (o, e, t) = desugar_cp o p in
             let o = o#restore_envs envs in
             o, `Block
-                  ([add_pos (`Val ([], add_withPos (`Record ([("1", add_withPos (`Variable (x, Some u, pos)));
-                                                              ("2", add_withPos (`Variable (c, Some s, pos)))], None)),
-                                   add_pos (`FnAppl (add_pos (Sugartypes.tappl (`Var "receive", grab_tyargs)),
-                                                     [add_pos (`Var c)])),
-                                   `Unknown, None))],
-                  add_pos e), t
+                ([add_withPos (`Val ([], add_withPos (`Record ([("1", add_withPos (`Variable (mkBndr x u pos)));
```
dhil: Something like `add_with_position` and `make_binder`.

core/desugarHandlers.ml
```diff
@@ -162,8 +162,8 @@ let split_handler_cases : (pattern * phrase) list -> (pattern * phrase) list * (
     match ret with
     | [] ->
        let x = "x" in
-       let xb = (x, None, dp) in
-       let id = (mkWithPos (`Variable xb) dp, (`Var x, dp)) in
+       let xb = mkBndrNoTy x dp in
```
dhil: `make_untyped_binder`

jstolarek:
@dhil thanks, I will rename per your suggestions. I was wondering if instead of having:

```OCaml
let mkBndr       n ty pos                  = { node=(n   , Some ty); pos }
let mkBndrNoTy   n    pos                  = { node=(n   , None   ); pos }
```
it would make sense to have a single function with an optional `ty` argument.

jstolarek:
Also, are there any OCaml naming conventions for functions like `setBndrName` and `setBndrTy`? (Other than not using camel case.)

dhil:
>  I was wondering if instead of having:
> ```ocaml
> let mkBndr       n ty pos = { node=(n   , Some ty); pos }
> let mkBndrNoTy   n    pos = { node=(n   , None   ); pos }
> ```
> it would make sense to have a single function with an optional ty argument defaulting to None.

Indeed, for this particular case, using an optional parameter `?type` might be a good fit.

jstolarek:
I am also tempted to define position as optional:

```
type 'a with_pos = { node : 'a
                   ; pos  : position option }
```

and then instead of:

```
let with_pos       node pos = { node; pos }
let with_dummy_pos node     = { node; pos = dummy_position }
```
we could just have one function with default `pos=None` argument.

These are in fact two independent ideas. We can have a default `pos` argument to `with_pos` without changing the datatype itself.

frank-emrich:
For which nodes would such an optional position be None? I feel like the easier it is to not set the position, the likelier it will get in the future that people forget adding the positions to nodes.

jstolarek:
> For which nodes would such an optional position be None?

That is a good question. The goal of attaching a position is to identify *source code token* for the purpose of reporting error messages. The moment we start to transform a syntax tree and create nodes out of thin air these positions no longer make sense. Or at least they shouldn't - we probably want to report errors before or during AST transformations, not after we have created new nodes within the tree. So at some point during the compilation positions should go away - I am not familiar enough with Links source code to say where exactly this should happen.

All that being said, turning `position` into an `option` helps little in terms of enforcing an invariant that after some point in the pipeline annotations don't make sense. Moreover, in the absence of pattern synonyms it would cause an enormous amount of clutter during pattern matching. So for now I withdraw the idea of using an `option`. Perhaps a better approach would be to parametrize `with_pos` not only by the type of `node` but also by the type of `pos`? We could then have `pos : position` in the frontend and `pos : ()` in the backend.

> I feel like the easier it is to not set the position, the likelier it will get in the future that people forget adding the positions to nodes.

Note that this is completely independent from the definition of `pos` field in `with_pos`. Rather, it is a matter of using optional arguments with default values. Indeed, if I were to implement this idea it would be the case that `with_pos foo` would be valid an equivalent to current `with_dummy_pos foo`. That might potentially lead to errors or confusion. I wonder how others feel about it. Should we:

1. have separate functions for a) creating a binder with and without type annotation; b) creating node with a specified annotation and with a dummy annotation

2. handle above cases with just a one function that takes an optional argument with a default value: `dummy_position` when creating a node with position annotation; `None` when creating a binder with a type annotation.

frank-emrich:
> The moment we start to transform a syntax tree and create nodes out of thin air these positions no longer make sense.

Ahh, I see. I agree that as soon as we rewrite code, the newly created nodes should not have source locations associated with them, as they would not make sense. Therefore, it should be okay to have functions to create nodes that easily allow creating them without a source location set, for example by using optional arguments. I don't have a strong opinion on whether the position should therefore have an option type or if the default position should be a dummy value. 


It sounds like me as if you wanted to enforce the absence of source locations at some point using types. However, I think that this is already the case: The source locations become meaningless as soon as we translate to the IR. Until then, the various desugaring stages may have created many nodes without source locations, but the ones still present should still be valid (but maybe useless).

> So at some point during the compilation positions should go away

Exactly, which should currently be achieved by translating from the frontend ASTs defined in sugartypes.ml to the IR ASTs defined in ir.ml. The latter does not define source locations for its nodes.
The only thing shared between the forms of ASTs are the datatypes for Links types, as defined in types.ml. If we added source locations to those, these would indeed be nonsensical once we go from the frontend language to the IR. But to me it looks as if you haven't added source locations to types.ml, in which case we should be fine and we shouldn't need more sophisticated types to express when locations should be present or not.

> I wonder how others feel about it. Should we:

I don't quite understand the two options you present here. You suddenly mention adding type annotations to nodes. How is this related to the source locations?
Are type annotations currently provided as optional arguments of node creation functions? In that case, if we made source locations an optional argument for the creation functions for nodes, we would have source locations and type annotations as optional arguments at the same time. To me, that doesn't sound like a big deal.

jstolarek:
> It sounds like me as if you wanted to enforce the absence of source locations at some point using types. However, I think that this is already the case: The source locations become meaningless as soon as we translate to the IR. Until then, the various desugaring stages may have created many nodes without source locations, but the ones still present should still be valid (but maybe useless).

Indeed. I guess the big question is whether we completely discard source locations during desugaring (or could potentially do so), and if so how to reflect that in our data types without overcomplicating things.

> The only thing shared between the forms of ASTs are the datatypes for Links types, as defined in types.ml. If we added source locations to those, these would indeed be nonsensical once we go from the frontend language to the IR

Yes, absolutely. There should be a clear distinction between *source code tokens* that correspond to types in a source file, and actual representations of types within a compiler. This is the case at the moment and nothing changes with my refactoring.

> I don't quite understand the two options you present here. You suddenly mention adding type annotations to nodes. How is this related to the source locations?

I was referring to earlier discussion with Daniel about type annotations on **binders** - sorry if that was confusing. Right now I have two helper functions for creating binders:

```OCaml
let make_binder         n ty pos = { node=(n, Some ty); pos }
let make_untyped_binder n    pos = { node=(n, None   ); pos }
```

It looks like it would make sense to have just one function `make_binder` with `ty` argument being optional and having a default value of `None`.

> Are type annotations currently provided as optional arguments of node creation functions?

No, at the moment I have:

```OCaml
let with_pos       node pos = { node; pos }
let with_dummy_pos node     = { node; pos = dummy_position }
```

and the question is whether we want just one function `with_pos` with an optional argument `pos` having a default value of `dummy_position`.

jamescheney:
> Indeed. I guess the big question is whether we completely discard source locations during desugaring 
> (or could potentially do so), and if so how to reflect that in our data types without overcomplicating 
> things.

An alternative could be to make the "locations" richer, e.g.

```
type annot = RawPos of pos
  | DerivedFrom of string * annot list
```
where `DerivedFrom stage [annots]` says that this AST node was constructed as a result of processing by desugaring stage `stage` and based on other AST nodes annotated by `annots`.  As a special case the list could be empty indicating new nodes created from nothing by a given stage.

Then at the end they can all still be discarded but if we run into an error (due to a bug in desugaring/typecheckign) halfway through we might have more of a chance of understanding the problem.  (Also, the fact that we are generating type-incorrect IR code suggests that it might be helpful to carry the annotations even further along).

This is not a suggestion for this PR however :)  - maybe more of a research question.

jstolarek:
> An alternative could be to make the "locations" richer

But that still doesn't address the problem of enforcing invariants about annotations at the type level.

> the fact that we are generating type-incorrect IR code suggests that it might be helpful to carry the annotations even further along

I don't think I agree with that. If we are generating type-incorrect IR from correct programs we should fix our IR generation, which I understand Frank is working on. For incorrect programs we shouldn't attempt to generate IR but instead report errors earlier in the compilation pipeline.

frank-emrich:
> If we are generating type-incorrect IR from correct programs we should fix our IR generation, which I understand Frank is working on.

I agree. Making the IR type errors user-friendly hasn't been a priority for me. They indicate a bug in Links, not (necessarily) in the user program. Piping the source locations into the backend seems like too much effort with little gain to me (Debugging IR type errors means stepping through the translation anyways...)

jamescheney:
> >    If we are generating type-incorrect IR from correct programs we should fix our IR generation, which I understand Frank is working on.


>I agree. Making the IR type errors user-friendly hasn't been a priority for me. They indicate a bug in  Links, not (necessarily) in the user program. Piping the source locations into the backend seems like too much effort with little gain to me (Debugging IR type errors means stepping through the translation anyways...)

I think what I wrote wasn't clear - what I was suggesting was solely about debugging bugs in Links, as all user-level type errors should be caught before desugaring begins.  (at least in the fantasy version of Links I have in my head).  

Propagating richer annotations about what AST nodes were desugared from what source nodes (and which desugaring stage was involved) might be overkill because once we've eliminated all the bugs in desugaring they won't be needed anymore.  But if we were to provide the ability to add new desugaring stages as modules, then suddenly program code might introduce bugs in desugaring.  Anyway, this is turning into a science-fictional enhancement proposal way beyond the scope of this PR.

jstolarek:
I have rebased my branch on top of `master` and cleaned up commit history. I decided not to squash into one big commit - in case I made a mistake and introduced a bug somewhere it will be easier to bisect.

Also, for now I decided not to introduce optional arguments in `make_binder` and `with_pos`. I might do that at some point but first I'd like to come up with a better story for handling of positions in desugar modules.

If there are no further comments about this PR I'd like to merge it.

jamescheney:
Looks good to me also.

jstolarek:
Fixed by bb3318cc

PR #425, opened 2018-11-23
Migrate parsers to Menhir

### LINKED ISSUE #411: Migrate the main parser to Menhir

jstolarek:
Fixes #411.

This migrates our parsers to Menhir. Almost no grammar refactoring, other than to silence warnings. The only change to Links syntax is that opening a module now requires a semicolon at the end. See https://github.com/links-lang/links/issues/411#issuecomment-441234996.

I also flipped the order of arguments to `with_pos` - it is more convenient to have position first.

This PR doesn't build currently. Build broke after 30c0a0bc31369f29ea5a2ec7becc57328a76bc47, which turns of `wrapped` setting in dune. You can revert that commit locally and then everything builds and works as expected.


jstolarek:
Build fixed thanks to Frank.

core/jsonparse.mly
```diff
@@ -1,8 +1,15 @@
+(* See Note [Debugging grammar conflicts] in parser.mly *)
+
 %{
 open Utility
 open ProcessTypes
 open WebsocketMessages
 
+module Links_core =
```
jamescheney: It would be good to document why this declaration is here in case the workaround needed changes (or Ocaml gets sensible namespaces and we can stop).  Likewise with the other ones.

core/parser.mly
```diff
@@ -299,102 +333,98 @@ declaration:
...
+                                                                   with_pos $loc `Infix }
+| tlvarbinding SEMICOLON                                       { let (bndr,p,l) = $1
+                                                                 in with_pos $loc($1) (`Val ([],
+                                                                    (with_pos $loc($1) (`Variable (make_untyped_binder bndr))), p, l, None)) }
+| signature tlvarbinding SEMICOLON                             { annotate $loc($1) $1 $loc($2) (`Var $2) }
 | typedecl SEMICOLON                                           { $1 }
 
 | links_module                                                 { $1 }
-| links_open                                                   { $1 }
+| links_open SEMICOLON                                         { $1 }
```
jamescheney: I think this is fine, since other toplevel declaration forms are followed by semicolon so this is more uniform anyway.  

An alternative would be to require all "open"s to be at the beginning of the file with at least one normal declaration afterwards, but having to write `open Foo;` seems like a small price to pay for `open` being more first-class.

dhil:
Issue #411 also mentions refactoring and clean up of the grammar (independent of migrating to Menhir). A separate issue for Another issue for refactoring and cleaning up the grammar should be open. Alternatively, the "fixes" phrase should be remove and the title of #411 updated to reflect the more general nature of the issue.

jstolarek:
> A separate issue for Another issue for refactoring and cleaning up the grammar should be open.

I would propose to limit #411 to Menhir migration and have a separate ticket for refactoring Links grammar. BTW, this is a rather open-ended task and it might be hard to say when exactly can we consider it done. Anyway, this patch already does some minor refactoring of parser code and I am planning to have another pass through the grammar to see what can be improved. But I would prefer to have Menhir migration merged first. I find it easier to work with smaller patches - easier to debug when something goes wrong.

dhil:
> I would propose to limit #411 to Menhir migration and have a separate ticket for refactoring Links grammar. 

Sounds good to me!

jstolarek:
Merged as f8dbc830f31ef7f8fec891b0afe4d24030709080

Issue #426, opened 2018-11-26
Improve Links grammar


jstolarek:
This is a follow-up to #411:

> Some of the production rules such as preamble are essentially garbage. We should get rid of rules like those. In general we should probably simplify/refactor our grammar. (...) Menhir provides some convenient combinators that enable us to get rid of most (if not all) of the _list and _perhaps productions.

Another thing to consider is what positions to attach to nodes that don't correspond directly to source code tokens. For example, here `Unit` is annotated with a dummy position:

```
hear_arrow_prefix:
| LBRACE COLON datatype COMMA efields RBRACE
   { row_with ("wild", `Present (with_dummy_pos `Unit))
              (row_with ("hear", `Present $3) $5) }
```
but other productions assign a position corresponding to a whole production:
```
field:
| field_label   { $1, `Present (with_pos $loc `Unit) }
```
We should decide on a consistent approach and follow it.

jstolarek:
I wonder how others feel about refactoring these two productions:

```
| LPAREN labeled_patterns VBAR pattern RPAREN { with_pos $loc (`Record ($2, Some $4)) }
| LPAREN labeled_patterns RPAREN              { with_pos $loc (`Record ($2, None   )) }
```
into:
```
| LPAREN labeled_patterns preceded(VBAR, pattern)? RPAREN   { with_pos $loc (`Record ($2, $3)) }
```
The latter one shorter and fancier but I'm not sure whether it is easier to understand and, as such, whether it is a good refactoring.

Another thought is that to fully refactor the grammar we also need to change our internal representations. For example:

```
parenthesized_pattern:
| LPAREN RPAREN                        { with_pos $loc (`Tuple []) }
| LPAREN pattern RPAREN                { $2 }
| LPAREN pattern COMMA patterns RPAREN { with_pos $loc (`Tuple ($2 :: $4)) }
```
So parenthesized patterns are internally represented as a `Tuple`, unless it's just one pattern, in which case internal representation is something completely different. If we had a uniform internal representation, e.g. by allowing `Tuple`s with one component, we could have a single production instead of three. But I'll leave this as future work - for now I only plan to focus on refactoring that can be done without touching the AST and further stages of the compilation pipeline.

jstolarek:
The refactoring is going well but I am sometimes running into problems I don't understand. I'm not sure whether this is me not understanding how parsing works or some peculiarity of Menhir (I'm suspecting the former). Consider these two productions:

```
[...]

PR #434, opened 2018-11-30
Grammar cleanup

jstolarek:
Addresses #426.

Below is a parser refactoring that I have been working on for the past week or so. I mostly tried to compactify the rules by relying on Menhir combinators. So for example instead of productions like:

```
foos:
| foo      { $1       }
| foo foos { $1 :: $2 }
```

we now have:

```
foos:
| foo+      { $1 }
```

As a result `parser.mly` is now about 100 lines shorter and about 13% smaller in terms of file size in bytes. This is as far as I was able to go without altering the structure of Links AST and touching further stages of compilation pipeline. More improvements are possible if we did so, but I personally don't yet feel competent enough to do this. A good example of a possible improvement are the three production rules for `INSERT` already discussed in https://github.com/links-lang/links/issues/426#issuecomment-442583867:

```
| INSERT exp VALUES LBRACKET LPAREN labeled_exps RPAREN RBRACKET
  { with_pos $loc (`DBInsert ($2, labels $6,
     with_pos $loc($6) (`ListLit ([with_pos $loc($6) (`RecordLit ($6, None))], None)),
    None)) }

| INSERT exp VALUES LBRACKET LPAREN RPAREN RBRACKET RETURNING VARIABLE
  { with_pos $loc (`DBInsert ($2, [],
     with_pos $loc (`ListLit ([with_pos $loc (`RecordLit ([], None))], None)),
    Some (with_pos $loc($9) (`Constant (`String $9))))) }

| INSERT exp VALUES LBRACKET LPAREN labeled_exps RPAREN RBRACKET RETURNING VARIABLE
  { with_pos $loc (`DBInsert ($2, labels $6,
     with_pos $loc($6) (`ListLit ([with_pos $loc($6) (`RecordLit ($6, None))], None)),
    Some (with_pos $loc (`Constant (`String $10))))) }
```

These can be collapsed into a single rule:

```
| INSERT exp VALUES LBRACKET LPAREN loption(labeled_exps) RPAREN RBRACKET
         preceded(RETURNING, VARIABLE)?
  { with_pos $loc (`DBInsert ($2, labels $6,
     with_pos $loc($6) (`ListLit ([with_pos $loc($6) (`RecordLit ($6, None))], None)),
    OptionUtils.opt_map (fun v -> with_pos $loc (`Constant (`String v))) $9)) }
```
But this requries that later stages of compilation pipeline catch a case where `labeled_exps` is an empty list and there is no `RETURNING VARIABLE`.

There is one more thing I am tempted to do, but I first need to know what others think. Currently `parser.mly` is formatted in such a way that productions are placed on the left and semantic actions are placed in the same line on the right, starting in column 63. This leads to extremely long lines (as much as 140 columns). I am often working with two source files opened side by side, but with `parser.mly` formatted the way it is now this is impossible. Similalrly, I suspect many of you will experience problems viewing changes in this PR, because lines are too long to be displayed side by side and they need to be wrapped, which makes viewing difficult. I would like to reformat `parser.mly` so that semantic actions are placed below productions whenever necessary. An example of this is above - the three INSERT rules are reformatted using this style for the purpose of this comment. In the source code they currently look like this:

```
| INSERT exp VALUES
  LBRACKET LPAREN labeled_exps RPAREN RBRACKET                 { with_pos $loc (`DBInsert ($2,
                                                                          labels $6,
                                                                          with_pos $loc($6)
                                                                                   (`ListLit ([with_pos $loc($6)
                                                                                                        (`RecordLit ($6, None))], None)),
                                                                          None)) }
| INSERT exp VALUES
  LBRACKET LPAREN RPAREN RBRACKET
  RETURNING VARIABLE                                           { with_pos $loc (`DBInsert ($2,
                                                                          [],
                                                                          (with_pos $loc (`ListLit ([with_pos $loc (`RecordLit ([], None))],
                                                                                                    None))),
                                                                          Some (with_pos $loc($9) (`Constant (`String $9))))) }
| INSERT exp VALUES
  LBRACKET LPAREN labeled_exps RPAREN RBRACKET
  RETURNING VARIABLE                                           { with_pos $loc (`DBInsert ($2,
                                                                          labels $6,
                                                                          (with_pos $loc($6) (`ListLit ([with_pos $loc($6)
                                                                                                                  (`RecordLit ($6, None))],
                                                                                                        None))),
                                                                          Some (with_pos $loc (`Constant (`String $10))))) }

```
which, I think, is horrible to read.  Are there any strong opinions for or against such reformatting?

Finally, there is this note in `parser.mly`:

```
/* FIXME:
 *
 * recursive row vars shouldn't be restricted to vfields.
 */
row_var:
| nonrec_row_var                                               { $1 }
| LPAREN MU VARIABLE DOT vfields RPAREN                        { `Recursive ($3, $5) }
```

What should the other allowed productions be? We can now parameterize production rules so this should be easy to fix.

In general, it would be nice if these kind of TODO/FIXME notes were avoided and appropriate tickets with informative description be created instead. (I am guessing many such notes come from times when there was no bugtracker?) I believe this would greatly increase the chances that such TODOs ever get addressed.

### DIFF
--- a/core/parser.mly
+++ b/core/parser.mly
@@ -528,8 +528,7 @@ cp_expression:
 
 primary_expression:
 | atomic_expression                                            { $1 }
-| LBRACKET RBRACKET                                            { with_pos $loc (`ListLit ([], None)) }
-| LBRACKET exps RBRACKET                                       { with_pos $loc (`ListLit ($2, None)) }
+| LBRACKET perhaps_exps RBRACKET                               { with_pos $loc (`ListLit ($2, None)) }
 | LBRACKET exp DOTDOT exp RBRACKET                             { with_pos $loc (`RangeLit($2, $4))   }
 | xml                                                          { $1 }
 | FUN arg_lists block                                          { with_pos $loc     (`FunLit (None, `Unl, ($2,
@@ -602,10 +601,9 @@ spawn_expression:
                                                                           with_pos $loc($2) (`Block $2), None)) }
 
 postfix_expression:
-| primary_expression                                           { $1 }
+| primary_expression | spawn_expression                        { $1 }
 | primary_expression POSTFIXOP                                 { with_pos $loc (`UnaryAppl (([], `Name $2), $1)) }
 | block                                                        { with_pos $loc (`Block $1) }
-| spawn_expression                                             { $1 }
 | QUERY block                                                  { with_pos $loc (`Query (None, with_pos $loc($2) (`Block $2), None)) }
 | QUERY LBRACKET exp RBRACKET block                            { with_pos $loc (`Query (Some ($3, with_pos $loc (`Constant (`Int 0))),
                                                                                         with_pos $loc($5) (`Block $5), None)) }
@@ -615,19 +613,19 @@ postfix_expression:
 
 
 arg_spec:
-| LPAREN RPAREN                                                { [] }
-| LPAREN exps RPAREN                                           { $2 }
+| LPAREN perhaps_exps RPAREN                                   { $2 }
 
 exps:
-| exp COMMA exps                                               { $1 :: $3 }
-| exp                                                          { [$1] }
+| separated_nonempty_list(COMMA, exp)                          { $1 }
+
+perhaps_exps:
+| separated_list(COMMA, exp)                                   { $1 }
 
 unary_expression:
 | MINUS unary_expression                                       { with_pos $loc (`UnaryAppl (([], `Minus     ), $2)) }
 | MINUSDOT unary_expression                                    { with_pos $loc (`UnaryAppl (([], `FloatMinus), $2)) }
 | PREFIXOP unary_expression                                    { with_pos $loc (`UnaryAppl (([], `Name $1   ), $2)) }
-| postfix_expression                                           { $1 }
-| constructor_expression                                       { $1 }
+| postfix_expression | constructor_expression                  { $1 }
 | DOOP CONSTRUCTOR arg_spec                                    { with_pos $loc (`DoOperation ($2, $3, None)) }
 | DOOP CONSTRUCTOR                                             { with_pos $loc (`DoOperation ($2, [], None)) }
 
--- a/core/parser.mly
+++ b/core/parser.mly
@@ -315,16 +315,14 @@ interactive:
 | END                                                          { `Directive ("quit", []) (* rather hackish *) }
 
 file:
-| preamble declarations exp END                                { $1 @ $2, Some $3 }
-| preamble exp END                                             { $1, Some $2 }
-| preamble declarations END                                    { $1 @ $2, None }
+| declarations exp? END                                        { $1, $2 }
+| exp END                                                      { [], Some $1 }
 
 directive:
 | KEYWORD args SEMICOLON                                       { ($1, $2) }
 
 args:
-| /* empty */                                                  { [] }
-| arg args                                                     { $1 :: $2 }
+| arg*                                                         { $1 }
 
 arg:
 | STRING                                                       { $1 }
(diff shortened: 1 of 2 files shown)


jstolarek:
I will merge this on Monday if there are no comments/complaints.

core/parser.mly
```diff
@@ -1013,14 +968,11 @@ binding:
...
                                                                                   None)) }
-| typedecl SEMICOLON                                           { $1 }
 | typed_handler_binding                                        { with_pos $loc (`Handler $1) }
-| links_module                                                 { $1 }
-| alien_block                                                  { $1 }
-| links_open                                                   { $1 }
+| typedecl SEMICOLON | links_module | alien_block | links_open { $1 }
 
 bindings:
-| binding                                                      { [$1] }
```
frank-emrich: shouldn't this collapse to a single  rule  ` bindings -> binding+`  ?

core/parser.mly
```diff
@@ -1013,14 +968,11 @@ binding:
...
+| binding                                                      { [$1]      }
 | bindings binding                                             { $1 @ [$2] }
```
frank-emrich: shouldn't this collapse to a single  rule  ` bindings -> binding+`  ?

(sorry, the original comment was on the wrong line)

jstolarek: I will reply in main discussion - this is too long for an inline comment.

jstolarek:
Replying to Frank's inline comment:

> shouldn't this collapse to a single rule bindings -> binding+

I wish this was the case, but no - doing so leads to a shift/reduce conflict. I believe this has to do with the current definition being left-recursive, whereas `binding+` definition would be right-recursive. Here is what Menhir says about the conflict:

```
LBRACE block_contents RBRACE 
       (?)

** In state 922, looking ahead at VARIABLE, shifting is permitted
** because of the following sub-derivation:

bindings exp SEMICOLON 
nonempty_list(binding) 
binding nonempty_list(binding) 
        binding 
        exp SEMICOLON 
        lens_expression 
        database_expression 
        table_expression 
        formlet_expression 
        escape_expression 
        iteration_expression 
        case_expression 
        conditional_expression 
        session_expression 
        db_expression 
        typed_expression 
        logical_expression 
        infixl_0 
        infixr_0 
        infixl_1 
        infixr_1 
        infixl_2 
        infixr_2 
        infixl_3 
        infixr_3 
        infixl_4 
        infixr_4 
        infixr_5 EQUALSTILDE regex 
        infixl_6 
        infixr_6 
        infixl_7 
        infixr_7 
        infixl_8 
        infixr_8 
        infixl_9 
        infixr_9 
        unary_expression 
        postfix_expression 
        primary_expression 
        atomic_expression 
        . VARIABLE 

** In state 922, looking ahead at VARIABLE, reducing production
** nonempty_list(binding) -> binding 
** is permitted because of the following sub-derivation:

bindings exp SEMICOLON // lookahead token appears because exp can begin with VARIABLE
nonempty_list(binding) // lookahead token is inherited
binding .
```

So for a right-recursive definition, after reducing one `binding` production and encountering `VARIABLE` token it is not clear whether that token is another binding (then shift in `bindings` production) or an `exp` (then reduce `bindings`). I am not 100% sure why this problem does not arise for a left-recursive definition. I presume that for such a definition all such ambiguities are resolved in favour of reducing as many bindings to the right as possible.

There are several more such productions, but not many. Perhaps there is a way to restructure the grammar in a way that avoids this kind of ambiguities, but I admit I don't know Links well enough to do it yet.

jstolarek:
Merged as 951270883de0fb90988845c03ff91f13224a6a8a.

I will submit a separate PR with a proposed refactoring of `parser.mly` to 80-column format.

Issue #421, opened 2018-11-20
Cleaner source code with smart constructors


jstolarek:
Source code is full of abominations in the likes of:

```OCaml
let x = Utility.gensym ~prefix:"xml" () in
  with_pos (`FnAppl
  (with_pos (`TAppl (with_pos (`Var "plugP") pos, [`Row (o#lookup_effects)])) pos,
   [with_pos (`FunLit
               (Some ([Types.make_tuple_type [Types.xml_type], o#lookup_effects]),
                `Unl,
                ([[with_pos (`Variable (make_binder x Types.xml_type pos)) pos]],
                 with_pos (`Xml (name, attrs, dynattrs,
                                 [with_pos (`Block ([], with_pos (`Var x) pos)) pos])) pos), `Unknown)) pos;
    desugar_nodes pos children])) pos

```
These are especially common in `desugar*` modules and in the parser. GHC avoids this kind of clutter by using smart constructors, e.g.

```Haskell
mkFamDecl :: SrcSpan
          -> FamilyInfo RdrName
          -> LHsType RdrName                   -- LHS
          -> Located (FamilyResultSig RdrName) -- Optional result signature
          -> Maybe (LInjectivityAnn RdrName)   -- Injectivity annotation
          -> P (LTyClDecl RdrName)
mkFamDecl loc info lhs ksig injAnn
  = do { (tc, tparams, fixity, ann) <- checkTyClHdr False lhs
       ; mapM_ (\a -> a loc) ann -- Add any API Annotations to the top SrcSpan
       ; tyvars <- checkTyVarsP (ppr info) equals_or_where tc tparams
       ; return (L loc (FamDecl (FamilyDecl{ fdInfo      = info, fdLName = tc
                                           , fdTyVars    = tyvars
                                           , fdFixity    = fixity
                                           , fdResultSig = ksig
                                           , fdInjectivityAnn = injAnn }))) }
  where
    equals_or_where = case info of
```
and then the parser simply calls `(mkFamDecl (comb3 $1 $3 $4) DataFamily $3 (snd $ unLoc $4) Nothing)`. Perhaps we could make our code cleaner by following the same pattern?

jamescheney:
Maybe I'm missing something about how the Haskell code snippet works, but I'm not sure it's doing hte same thing.  In the Links/OCaml example we seem to be broadcasting the same position to several constructors (presumably, so that any later errors including newly generated constructors are easier to debug because we can relate them to the location in the source that somehow "caused" them.)  

[...]

Issue #423, opened 2018-11-22
Inconsistent handling of positions during desugaring


jstolarek:
In `desugarFuns` we have:

```OCaml
`Block
  ([with_dummy_pos (`Fun (unwrap_def ( make_binder f ft dp, lin, ([], lam)
                                     , location, None)))],
   (with_dummy_pos (`Var f)))
```
In `desugarCP` we have:

```OCaml
let add_pos x = with_pos x pos in
(...)
`Block
 ([add_pos (`Val ([], add_pos `Any,
                       add_pos (`FnAppl (add_pos (`Var "wait"),
                                        [add_pos (`Var c)])),
                       `Unknown, None))],
  add_pos e), t
```

Notice how the first module erases positions by replacing them with `dummy_position`, while the second one propagates positions to constructed nodes. This can be inconsistent even within a single module, e.g. `desugarHandlers`. I believe all `desugar*` modules should treat positions consistently. I believe that the erasing positions is the right thing to do here, since we create nodes that don't correspond to source tokens written by the user and so it does not seem to make sense to attach positions to them.

jstolarek:
Fixed in 73e5892.

PR #443, opened 2018-12-18
Use smart constructors in the parser

### LINKED ISSUE #442: Collapse three INSERT parsing rules into one
These three parser rules:

```
| INSERT exp VALUES LBRACKET LPAREN labeled_exps RPAREN RBRACKET
  { with_pos $loc (`DBInsert ($2, labels $6,
     with_pos $loc($6) (`ListLit ([with_pos $loc($6) (`RecordLit ($6, None))], None)),
    None)) }

| INSERT exp VALUES LBRACKET LPAREN RPAREN RBRACKET RETURNING VARIABLE
  { with_pos $loc (`DBInsert ($2, [],
     with_pos $loc (`ListLit ([with_pos $loc (`RecordLit ([], None))], None)),
    Some (with_pos $loc($9) (`Constant (`String $9))))) }

| INSERT exp VALUES LBRACKET LPAREN labeled_exps RPAREN RBRACKET RETURNING VARIABLE
  { with_pos $loc (`DBInsert ($2, labels $6,
     with_pos $loc($6) (`ListLit ([with_pos $loc($6) (`RecordLit ($6, None))], None)),
    Some (with_pos $loc (`Constant (`String $10))))) }
```

can be collapsed into a single rule:

```
| INSERT exp VALUES LBRACKET LPAREN loption(labeled_exps) RPAREN RBRACKET
         preceded(RETURNING, VARIABLE)?
  { with_pos $loc (`DBInsert ($2, labels $6,
     with_pos $loc($6) (`ListLit ([with_pos $loc($6) (`RecordLit ($6, None))], None)),
    OptionUtils.opt_map (fun v -> with_pos $loc (`Constant (`String v))) $9)) }
```
This requires a smart constructor to catch a case where `labeled_exps` is an empty list and there is no `RETURNING VARIABLE`. This can be done with a smart constructor - see #421.

jstolarek:
This is the first chunk of work on introducing smart constructors (#421). It fixes #442 and otherwise simplifies or refactors multiple productions in the parser. I introduced a new module to store smart constructors. Many of the already existing functions have been moved from parser to that module.

One particular thing I would particularly like to hear feedback on are the naming conventions for the constructors. Names of most of the functions I introduced begin with `make_`. Some of the already existing functions follow that convention - c.f. `make_untyped_handler` in sugartypes.ml or `make_tuple` - but others don't (e.g. `datatype`, `cp_unit`). I am thinking of dropping `make_` prefix altogether to make the code more concise. Do you think this would be a good idea?


jamescheney:
I wonder if it would make sense to remove the "make_" prefix, but import the smartConstructors module as a qualified module name such as `Make` or `Mk`, so that you would write `Make.untyped_handler" or "Mk.untyped_handler" - using the module's namespace to decrease the clutter in the identifiers?  Or is that an antipattern?

dhil:
The more modules the merrier. I think it is a good idea to collect those utility functions in a module named `Make` or something descriptive. Organising code into to separate modules is generally a good idea.

dhil:
You can embed the module `SugarConstructors` directly into `SugarTypes`, e.g.
```ocaml
module Make: sig
   val funlit : ppos -> declared_linearity -> pattern list list -> block_body with_pos -> phrase
end = struct
  let funlit ....
end
```
That way the constructors would be available whenever `SugarTypes` are in scope.

slindley:
What @jamescheney suggests is idiomatic OCaml - not an antipattern

jstolarek:
One more question. Functions in `Make` module currently use `ppos`, i.e. positions produced by the parser, which differ from positions used in Sugartypes. Since constructors module is currently used only by the parser it is not a problem. But it will become a problem once I want to apply smart constructors in desugar modules. What would be a canonical way of solving such a problem in OCaml? Parametrizing `Make` module with a type of positions?

dhil:
> Functions in Make module currently use ppos, i.e. positions produced by the parser, which differ from positions used in Sugartypes.

It may be that I don't remember the details, but each (term) node in AST is paired with a position produced by the parser. The positions produced by the parser have type `SourceCode.pos`. Each desugaring pass operate over the AST from `Sugartypes`, so they ought to use `SourceCode.pos` too. 

The type checker uses both `SourceCode.pos` and `SourceCode.position`, where the latter is the type of "resolved" positions. Resolved positions are used in error messages.

jstolarek:
> The positions produced by the parser have type `SourceCode.pos`

Before switch to Menhir that was indeed the case. We just had this one `pos` function in the parser that produced a position of type `SourceCode.pos` for a whole production:

```
let pos () : Sugartypes.position = Parsing.symbol_start_pos (), Parsing.symbol_end_pos (), None
```

With Menhir that function no longer exists and we have access built-in Menhir functions that produce locations both for a whole production and its sub-productions. The type of positions produced by Menhir is `SourceCode.lexpos * SourceCode.lexpos`. So after a switch to Menhir I just created convenient wrappers around `sugartypes.with_pos`:

```
(* Convert position produced by a parser to Sugartypes position *)
let pos (start_pos, end_pos)= (start_pos, end_pos, None)

(* Wrapper around Sugartypes.with_pos.  Accepts parser positions. *)
let with_pos p = Sugartypes.with_pos (pos p)
``` 

As long as these functions were limited in scope to the parser this design worked nicely.
But now when they are in `Make` module that will get used in other places this no longer looks like a good design. Even now you will notice there were several places in sugarConstructors.ml where I had to explicitly call `Sugartypes.with_pos` and not the `Make.with_pos`.

jstolarek:
I attempted to solve the problem outlined above by using a functor, but I've been defeated by OCaml (or rather my lack of knowledge of OCaml). My work is on `js-T421-smart-constructors-functor` branch. Here's what I attempted to do. Firstly, abstract positions using a module:

```
module type Pos = sig
  type t
  val pos      : t -> Sugartypes.position
  val with_pos : t -> 'a -> 'a Sugartypes.with_pos
end
```

In the parser this module will implement simple wrappers around definitions in sugartypes.ml:


```
module ParserPosition : Pos = struct
  type t = SourceCode.lexpos * SourceCode.lexpos
  let pos (start_pos, end_pos) = (start_pos, end_pos, None)
  let with_pos p = Sugartypes.with_pos (pos p)
end
```

In all other modules `pos` will be an identity and `with_pos` will be an alias for `Sugartypes.with_pos`.  All the smart constructors will be defined in a functor that takes module of type `Pos`:

```
module SmartConstructors (Position : Pos)
       : (SmartConstructorsSig with type t := Position.t) = struct

  let pos      = Position.pos
  let with_pos = Position.with_pos

  let variable_pat ppos name =
    with_pos ppos (`Variable (make_untyped_binder name))
  
  (...)
```

with the signature `SmartConstructorsSig` being defined as:

```
module type SmartConstructorsSig = sig

  (* Positions *)
  type t
  val pos      : t -> Sugartypes.position
  val with_pos : t -> 'a -> 'a with_pos

  val variable_pat : t -> name with_pos -> pattern
  (...)
end
```

However, if I say this in the parser:

```
module ParserConstructors  = SmartConstructors(ParserPosition)
open ParserConstructors
```
or even
```
module ParserConstructors
       : (SmartConstructorsSig with type t := ParserPosition.t) =
  SmartConstructors(ParserPosition)
open ParserConstructors
```
then I get a compilation error on a `$loc` argument in a `with_pos $loc ...` expression:

```
File "parser.mly", line 682, characters 74-78:
Error: This expression has type Lexing.position * Lexing.position
       but an expression was expected of type ParserPosition.t
```
Except that the two types are equal. I was hoping to expose that equality with a destructive substitution `with type t := ParserPosition.t` but clearly that's not how it's supposed to be done. So, my questions are:

1. Is this even a good approach to solving the problem?
2. If so, how do I make it work?

I would also like to get rid of `pos` and `with_pos` from `SmartConstructorsSig` - these are only aliases - and use definitions from `Pos` directly. I think I know how to do it but first I need to get the code to compile :-)

rudihorn:
Try replacing the code with:

```
module ParserPosition : Pos with type t = Lexing.Position * Lexing.Position = struct
...

end

```


By saying the module is of type `Pos`, you are abstracting away the information about what `t` is, and the rest of the code will treat `t` as completely different to its underlying type.

jstolarek:
That works, thanks! This is strange though. I could swear I tried this before and got an error message complaining about mismatched signatures due to `t` being undefined when I tried to instantiate `SmartConstructors(ParserPosition)`. I guess some other fragment of the code must've been different as well.

rudihorn:
You probably in that case used destructive substitution rather than type equality, which gets rid of the type t completely. This means that the module is not of module type `Pos` anymore.

jstolarek:
Yes, I believe this is what I did. Thanks for explanation.

jstolarek:
I updated the PR to include this functor refactoring. One more question is whether it is possible to avoid repeating the long `SmartConstructorsSig` signature in both the `ml` and `mli` files?

A cool thing about this design is that we can now define:

```
module DummyPosition : Pos with type t = unit = struct
  type t          = unit
  let pos ()      = Sugartypes.dummy_position
  let with_pos () = Sugartypes.with_pos Sugartypes.dummy_position
end

module DummyMake = SmartConstructors(DummyPosition)
```

And use this module when constructing values in `desugar*`modules. Or perhaps even make this the default implementation, since we mostly construct things with dummy values (see #423).

jstolarek:
I cleaned up the patch using Rudi's suggestions and added documentation. If there are no change requests then I would like to merge it before moving to refactor desugar* modules.

core/parser.mly
```diff
@@ -59,7 +60,9 @@ module ParserPosition
   let with_pos p = Sugartypes.with_pos (pos p)
 end
 
-module ParserConstructors = SmartConstructors(ParserPosition)
+(* Instatntiate SugarConstructors functor and open it to avoid explicit name
```
dhil: s/Instatntiate/Instantiate

dhil: s/Instatntiate/Instantiate

core/sugarConstructors.mli
```diff
@@ -1,16 +1,13 @@
...
-  val with_pos : t -> 'a -> 'a Sugartypes.with_pos
-end
+(* Import module signatures *)
+module type Pos                  = SugarConstructorsIntf.Pos
+module type SugarConstructorsSig = SugarConstructorsIntf.SugarConstructorsSig
 
-module type SmartConstructorsSig = SmartConstructorsIntf.SmartConstructorsSig
+(* Actual implementation of smart constructors as a functor on a Pos module *)
+module SugarConstructors (Position : Pos)
+       : (SugarConstructorsSig with type t := Position.t)
```
dhil: Comment: You should be able to drop the parentheses in `(SugarConstructorsSig with type t := Position.t)`. Personally, I don't care whether we have a few extraneous parentheses.

core/sugarConstructors.ml
```diff
@@ -2,18 +2,15 @@ open Sugartypes
...
 open Utility.OptionUtils
 
-module type Pos = sig
-  type t
-  val pos      : t -> Sugartypes.position
-  val with_pos : t -> 'a -> 'a Sugartypes.with_pos
-end
-
-module type SmartConstructorsSig = SmartConstructorsIntf.SmartConstructorsSig
+(* Import module signatures *)
```
dhil: Missing punctuation.

core/sugarConstructors.ml
```diff
@@ -2,18 +2,15 @@ open Sugartypes
...
+module type Pos                  = SugarConstructorsIntf.Pos
+module type SugarConstructorsSig = SugarConstructorsIntf.SugarConstructorsSig
 
-module SmartConstructors (Position : Pos)
-       : (SmartConstructorsSig with type t := Position.t) = struct
+(* Actual implementation of smart constructors as a functor on a Pos module *)
```
dhil: Missing punctuation.

core/sugarConstructors.ml
```diff
@@ -2,18 +2,15 @@ open Sugartypes
...
+module SugarConstructors (Position : Pos)
+       : (SugarConstructorsSig with type t := Position.t) = struct
 
-  (** Positions *)
+  (** Convenient aliases for functions operating on positions *)
```
dhil: Missing punctuation.

core/sugarConstructors.ml
```diff
@@ -227,6 +224,7 @@ module SmartConstructors (Position : Pos)
 
 end
 
+(* A default type of positions *)
```
dhil: Missing punctuation.

core/sugarConstructors.mli
```diff
@@ -1,16 +1,13 @@
-module type Pos = sig
-  type t
-  val pos      : t -> Sugartypes.position
-  val with_pos : t -> 'a -> 'a Sugartypes.with_pos
-end
+(* Import module signatures *)
```
dhil: Missing punctuation.

core/sugarConstructors.mli
```diff
@@ -1,16 +1,13 @@
...
-module type SmartConstructorsSig = SmartConstructorsIntf.SmartConstructorsSig
+(* Actual implementation of smart constructors as a functor on a Pos module *)
+module SugarConstructors (Position : Pos)
+       : (SugarConstructorsSig with type t := Position.t)
 
-module SmartConstructors (Position : Pos)
-       : (SmartConstructorsSig with type t := Position.t)
-
-module Make : (SmartConstructorsSig
+(* Modules for making nodes using various types of positions *)
```
dhil: Missing punctuation.

core/sugarConstructorsIntf.ml
```diff
@@ -1,9 +1,32 @@
...
-module type SmartConstructorsSig = sig
+(* An abstract type of positions and operations on them.  The core type of
+   positions used in the compilation pipeline is Sugartypes.position, which
+   again is an alias to SourceCode.pos.  However, in several places we operate
+   on positions of a different type.  Most importantly, the parser produces a
+   positions in a different format.  What is important is that there needs to be
+   a way of converting such positions to Sugartypes.position and this is what
+   modules implementing this signature provide. *)
+module type Pos = sig
+  (* Type of positions *)
```
dhil: Missing punctuation.

core/sugarConstructorsIntf.ml
```diff
@@ -1,9 +1,32 @@
...
+  type t
+  (* Convert a position to Sugartypes.position *)
```
dhil: Missing punctuation.

core/sugarConstructorsIntf.ml
```diff
@@ -1,9 +1,32 @@
...
+  val pos      : t -> Sugartypes.position
+  (* Produce a syntax tree node with a position attached *)
```
dhil: Missing punctuation.

core/sugarConstructorsIntf.ml
```diff
@@ -1,9 +1,32 @@
...
+   Putting them here allows to avoid repetition. *)
+
 open Sugartypes
 open Operators
 
-module type SmartConstructorsSig = sig
+(* An abstract type of positions and operations on them.  The core type of
+   positions used in the compilation pipeline is Sugartypes.position, which
+   again is an alias to SourceCode.pos.  However, in several places we operate
+   on positions of a different type.  Most importantly, the parser produces a
```
dhil: s/the parser produces a positions/the parser produces positions.

core/sugarConstructors.mli
jstolarek: Yes, these are redundant. I don't know which style is considered better. Personally I found having these parens more readable.

core/sugarConstructors.ml
jstolarek: I tend not to short (single-sentence) comments with punctuation - that's why it's consistently missing. Will fix.

jstolarek:
@dhil, thanks! All done, except for parens around type signatures in modules.

jstolarek:
Merged as f06ca1ad

PR #450, opened 2019-01-08
Ensure Mind example is IR-typechecked

SimonJF:
Fixes the CI


SimonJF:
I don't think this is controversial, so I'll merge it.

PR #451, opened 2019-01-08
Use smart constructors in desugar* modules

jstolarek:
This is a refactoring of desugar* modules in an attempt to cut down on clutter by using smart constructors. It addresses #421 and #423. 

For example, code that previously looked like this:

```OCaml
with_pos (`FnAppl
(with_pos (`TAppl (with_pos (`Var "plugP") pos, [`Row (o#lookup_effects)])) pos,
 [with_pos (`FunLit
             (Some ([Types.make_tuple_type [Types.xml_type], o#lookup_effects]),
              `Unl,
              ([[with_pos (`Variable (make_binder x Types.xml_type pos)) pos]],
               with_pos (`Xml (name, attrs, dynattrs,
                               [with_pos (`Block ([], with_pos (`Var x) pos)) pos])) pos), `Unknown)) pos;
  desugar_nodes pos children])) pos
```
now looks like this:
```OCaml
fn_appl "plugP" [`Row (o#lookup_effects)]
   [fun_lit ~args:[Types.make_tuple_type [Types.xml_type], o#lookup_effects]
            `Unl [[variable_pat ~ty:Types.xml_type x]]
            (xml name attrs dynattrs [block ([], var x)]);
    desugar_nodes children]
```

One important design decision is handling of locations in desugar modules. I assumed that since locations are meant to correspond to source code tokens they can only be meaningfully constructed by the parser. Therefore the `SugarConstructors.Make` module allows to attach only dummy positions to nodes:

```OCaml
module SugartypesPosition : Pos with type t = unit = struct
  type t = unit
  let pos ()           = Sugartypes.dummy_position
  let with_pos () node = Sugartypes.with_pos Sugartypes.dummy_position node
  let dp               = ()
end
```
This module is meant to be used everywhere except for the parser, which implements its own `Pos` module. To make things easier, all smart constructors have an optional position argument with a default dummy position.

A separate thing that caught my attention during refactoring is the usage of string constants. It seems to me that in some places we rely on string typing, which is something we should probably avoid. For now in some modules I have turned string constants into let-bound identifiers. Not sure if it was worth the effort though.

### DIFF
--- a/core/desugarAlienBlocks.ml
+++ b/core/desugarAlienBlocks.ml
@@ -12,6 +12,7 @@
 
 open Utility
 open Sugartypes
+open SugarConstructors.Make
 
 let rec flatten_simple = fun () ->
 object(self)
@@ -39,14 +40,13 @@ object(self)
     | {node=`AlienBlock (lang, lib, decls); _} ->
         self#list (fun o ((bnd, dt)) ->
           let name = name_of_binder bnd in
-          let pos = bnd.pos in
-          o#add_binding (with_pos pos (`Foreign (bnd, name, lang, lib, dt)))) decls
-    | {node=`Module (name, bindings); pos} ->
+          o#add_binding (with_dummy_pos (`Foreign (bnd, name, lang, lib, dt)))) decls
+    | {node=`Module (name, bindings); _} ->
         let flattened_bindings =
           List.concat (
             List.map (fun b -> ((flatten_bindings ())#binding b)#get_bindings) bindings
           ) in
-        self#add_binding (with_pos pos (`Module (name, flattened_bindings)))
+        self#add_binding (with_dummy_pos (`Module (name, flattened_bindings)))
     | b -> self#add_binding ((flatten_simple ())#binding b)
 
   method! program = function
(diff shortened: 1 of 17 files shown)


dhil:
> A separate thing that caught my attention during refactoring is the usage of string constants. It seems to me that in some places we rely on string typing, which is something we should probably avoid.

Yes, it is problematic, and as a result our frontend transformations are unhygienic. Shadowing some of the names (in a regular source file) makes Links go mad. We can probably address this issue in the module patch as it adds qualified names.

jstolarek:
I have rebased the patch on top of master and squashed all the commits. I'd like to merge this tomorrow unless there is someone who plans to do a thorough review and needs more time.

dhil:
It will be good to merge this soon. I have some patches I would like to rebase on this before continuing. In particular, I would like to change the representation of names in the frontend to something richer than raw strings. Hopefully this PR will make that work somewhat easier.

jstolarek:
Let me just remark that I have created constructors only for the use cases where I felt it is beneficial. A common criterion was that the constructor will be used more than once. That being said, do add new constructors if you find the need.

PR #459, opened 2019-01-23
[WIP]: hacky fix to issue225

### LINKED ISSUE #225: Function return type cannot be qualified by module name.
Consider the following example:
```
# one.links
typename One = [| Foo | Bar |];
```
```
# two.links
open One

sig foo : (One.One) ~> One.One
fun foo (a) {
  a
}
```
Results in 
```
*** Parse error: two.links:4
  sig foo : (One.One) ~> One.One
                              ^
```
whereas when omitting the module prefix it unsurprisingly gets confused by the identical module name:
```
*** Fatal error : Unbound type constructor One
```

jamescheney:
This is an attempt to fix #225 while allowing reuse of `.` for session types and qualified types.  It is hacky and wrong, but passes all of the existing tests.

Explanation:
1. Create `primary_datatype_loc` nonterminal for primary_datatypes annotated with their location (so that their semantic actions have the same type as `datatype`)
2. Remove `qualified_type_name` from `parenthesized_datatypes` and make a new nonterminal `qualified_or_primary_datatype` that allows either a non-parenthesized `primary_datatype` or a `qualified_or_primary_datatype`.
3. Move `TILDE datatype` from `session_type` to `primary_datatype` and restrict the second component to be `primary_datatype_loc`. 

The third change was the one I was least pleased about but seems necessary to parse existing type expressions using `TILDE` conventionally on variables, such as

```
# examples/sessions/forkbomb.links
mu a.!~a.a
``` 

Restricting what goes after the datatype is necessary so that a `session_type` cannot end with a tilde'd qualified type.  That is, you still have to write 
```
~(M.x)
```
but that is what we would have had to write before this change.

TODO:
- [x] Add tests from #225 to confirm this solves original problem
- [x] Discuss alternatives / clean up hacked grammar
- [x] Remove commented-out line and add comment documenting the reason(s) for this change

### DIFF
--- a/core/parser.mly
+++ b/core/parser.mly
@@ -931,18 +931,24 @@ forall_datatype:
      Parenthesised versions take priority over non-parenthesised versions.
 */
 session_datatype:
-| BANG datatype DOT datatype                                   { `Output ($2, $4) }
-| QUESTION datatype DOT datatype                               { `Input  ($2, $4) }
+| BANG primary_datatype_loc DOT datatype                                   { `Output ($2, $4) }
+| QUESTION primary_datatype_loc DOT datatype                               { `Input  ($2, $4) }
 | LBRACKETPLUSBAR row BARPLUSRBRACKET                          { `Select $2       }
 | LBRACKETAMPBAR row BARAMPRBRACKET                            { `Choice $2       }
 | TILDE datatype                                               { `Dual $2         }
 | END                                                          { `End             }
-| primary_datatype                                             { $1               }
+| qualified_or_primary_datatype                                { $1               }
 
 parenthesized_datatypes:
 | LPAREN RPAREN                                                { [] }
-| LPAREN qualified_type_name RPAREN                            { [with_pos $loc (`QualifiedTypeApplication ($2, []))] }
+/*| LPAREN qualified_type_name RPAREN                            { [with_pos $loc (`QualifiedTypeApplication ($2, []))] }*/
 | LPAREN datatypes RPAREN                                      { $2 }
+
+qualified_or_primary_datatype:
+| LPAREN qualified_type_name RPAREN                            { `QualifiedTypeApplication ($2, []) }
+
+primary_datatype_loc:
+primary_datatype                                           { with_pos $loc $1 }
 
 primary_datatype:
 | parenthesized_datatypes                                      { match $1 with
--- a/core/parser.mly
+++ b/core/parser.mly
@@ -930,7 +930,7 @@ forall_datatype:
 
      Parenthesised versions take priority over non-parenthesised versions.
 
-   jcheney: Modified grammar to allow unparenthesized qualified type names
+   24.1.19: Modified grammar to allow unparenthesized qualified type names
    in most places, but specifically forbid in ? . or ! .
    This also requires moving the TILDE production into primary_datatype
    since otherwise, ? ~M.x . y is ambiguous.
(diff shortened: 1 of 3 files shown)


jamescheney:
I've cleaned the change up a little and documented it, so I think it's now ready to merge.  Anyone want to give it a quick review?

jamescheney:
(just to clarify, the main change was just to inline `qualified_or_primary_datatype`)

core/parser.mly
```diff
@@ -929,22 +929,32 @@ forall_datatype:
...
 parenthesized_datatypes:
 | LPAREN RPAREN                                                { [] }
-| LPAREN qualified_type_name RPAREN                            { [with_pos $loc (`QualifiedTypeApplication ($2, []))] }
 | LPAREN datatypes RPAREN                                      { $2 }
 
+primary_datatype_loc:
+| primary_datatype                                             { with_pos $loc $1 }
+
 primary_datatype:
+| TILDE primary_datatype_loc                                   { `Dual $2 }
```
jstolarek: IIUC this will allow tilde in patterns, because:

```
pattern:
| typed_pattern                         { $1 }
| typed_pattern COLON primary_datatype  { with_pos $loc (`HasType ($1, datatype (with_pos $loc($3) $3))) }
```

Is that OK? If no then it needs to be fixed. If yes, then it would make sense for the `pattern` production to use `primary_datatype_loc` instead of currrent `with_pos $loc($3) $3`.

core/parser.mly
```diff
@@ -929,22 +929,32 @@ forall_datatype:
...
 | END                                                          { `End             }
 | primary_datatype                                             { $1               }
+| qualified_type_name                                          { `QualifiedTypeApplication ($1, []) }
 
 parenthesized_datatypes:
 | LPAREN RPAREN                                                { [] }
-| LPAREN qualified_type_name RPAREN                            { [with_pos $loc (`QualifiedTypeApplication ($2, []))] }
 | LPAREN datatypes RPAREN                                      { $2 }
 
+primary_datatype_loc:
```
jstolarek: Nitpick: across the compiler we use the term "position" to refer to a token's placement in a source file, but "location" seems to be a more commonly used name in the literature (e.g. Menhir uses "location" term). That being said, it would be nice to be consistent about naming and thus this production would probably be better named as `primary_datatype_pos`.

core/parser.mly
```diff
@@ -929,22 +929,32 @@ forall_datatype:
      S = !(ModuleA.ModuleB.Type).S
 
      Parenthesised versions take priority over non-parenthesised versions.
+
+   jcheney: Modified grammar to allow unparenthesized qualified type names
```
jstolarek: Hint: comment "ownership" can be always determined using `git blame`, so no need to sign comments. Especially that they can - and most likely will - be updated by others, which blurs the "ownership".

core/parser.mly
jamescheney: I don't know that part of the grammar well, but it looks like the effect of my change would be that parentheses are not required in a pattern of the form 
```
typed_pattern COLON TILDE primary_datatype_loc
```
whereas in the current grammar one would have to parenthesize:
```
typed_pattern COLON LPAREN TILDE primary_datatype_loc RPAREN
```

However, I don't have a clear idea whether either syntax leads to something meaningful, i.e. whether typechecking/kindchecking ever succeeds for programs containing either syntactic form.  I suspect not, but either way the effect on the grammar seems to be to make it more liberal in terms of parentheses without changing the set of possible abstract syntax trees that can be represented.

core/parser.mly
jamescheney: OK with me, I wasn't sure if there was a convention already for this.

jamescheney:
This seems to pass all tests, any objections to merging now?

core/parser.mly
jamescheney: OK.  I only introduced it because I didn't know you could do $loc($3) to get the location info of a specific RHS nonterminal though, so we could also inline it in both places.

I also wonder if it would help to establish a convention that `nt_pos` means nonterminal `nt` with position information (e.g. `datatype` should be `datatype_pos`) or the reverse, say `nt` means with position information and `nt_node` means without.  The fact that different "type" nonterminals sometimes have position information and sometimes don't led to confusing misunderstandings on my part (and the resulting type error messages were not that helpful).  

Alternatively, probably enough to document the types in the mly file.

jstolarek: > I also wonder if it would help to establish a convention that nt_pos means nonterminal nt with position information

I was wondering the same thing. I guess the only way to find out is trying it and seeing whether it improves readability of the code. BTW. most smart constructors in `SugarConstructors` produce values with positions, but on several occasions I have found the need to have constructors just for the nodes. These have names ending with `_node` suffic.

dhil:
Looks good to me! Let's merge!

Issue #460, opened 2019-01-24
Attaching of dummy positions inside the pipeline not enforced correctly


jstolarek:
73e5892b51823d9207759d680ec9332b53e132fd introduced `SugarConstructors.Make` module that was intended to prevent attaching of positions to nodes inside the compiler pipeline (only the parser is allowed to do this). However, that works only if `SugarConstructors.Make` module is open. If a module opens `Sugartypes` directly it is still possible to attach positions to nodes (and there are many places in the code that do so). This should be fixed by moving `with_pos` function from `Sugartypes` to `SugarConstructors` (it isn't really used in that module).

jstolarek:
`DesugarHandlers.desugar_handlers_early` does weird things with positions:

```OCaml
let desugar_handlers_early =
object
  inherit SugarTraversals.map as super
  method! phrasenode = function
    | `HandlerLit hnlit ->
       let (fnparams, body) = funlit_of_handlerlit hnlit in
       let funlit : Sugartypes.phrasenode = (fun_lit `Unl fnparams body).node in
       super#phrasenode funlit
    | e -> super#phrasenode e

  method! phrase {node; pos} =
    match node with
    | `Handle h ->
       let (val_cases, eff_cases) = split_handler_cases h.sh_effect_cases in
       with_dummy_pos (`Handle { h with sh_effect_cases = eff_cases;
                                        sh_value_cases  = val_cases })
    | _ -> super#phrase {node; pos}

  method! bindingnode = function
    | `Handler (binder, hnlit, annotation) ->
       let fnlit  = funlit_of_handlerlit hnlit in
       (fun_binding' ?annotation binder fnlit).node
    | b -> super#bindingnode b
end
```

  * `HandlerLit` is converted into a function literal and the position is kept. (Conversion itself introduces dummy positions into the body of a new literal.)
  * `Handle` has some of its arguments modified and the position is discarded. 
   * `Handler` binding is converted into a function binding, and again the position is kept.

Whatever we decide to do, it should be consistent.

dhil:
I wouldn't pay attention to the `DesugarHandlers` module, it is practically dead code. I want to plan to get rid of it.

jstolarek:
OK, in that case I'll leave it as is. I was planning on fixing positions handling in that module because it is generally too eager (my fault, I'm afraid).

dhil:
The `handler` and `shallowhandler` keywords are syntactic sugar, for example
```
[|handler h { 
    case Return(x) -> M
[...]

PR #461, opened 2019-01-24
Enforce dummy positions inside compiler pipeline

jstolarek:
A simple fix for #460, hopefully uncontroversial.

### DIFF
--- a/core/compilePatterns.ml
+++ b/core/compilePatterns.ml
@@ -11,6 +11,7 @@
 open CommonTypes
 open Utility
 open Ir
+open SugarConstructors.Make
 
 type pattern = [
 | `Any
@@ -76,7 +77,7 @@ let lookup_name name (nenv, _tenv, _eff, _penv) =
 let lookup_effects (_nenv, _tenv, eff, _penv) = eff
 
 let rec desugar_pattern : Ir.scope -> Sugartypes.Pattern.with_pos -> pattern * raw_env =
-  fun scope {Sugartypes.node=p; Sugartypes.pos} ->
+  fun scope {Sugartypes.node=p; _} ->
     let desugar_pat = desugar_pattern scope in
     let empty = (NEnv.empty, TEnv.empty, Types.make_empty_open_row (lin_any, res_any)) in
     let (++) (nenv, tenv, _) (nenv', tenv', eff') = (NEnv.extend nenv nenv', TEnv.extend tenv tenv', eff') in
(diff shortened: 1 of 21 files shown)


dhil:
I don't quite understand the rationale for #460. It seems to me that desugaring passes now suffer a loss of information.

jstolarek:
The rationale is the same as for #423. Positions should correspond only to source code tokens and so it makes no sense to attach positions to AST nodes created inside the compiler pipeline. Such nodes have no corresponding source code tokens, even though they are derived from one, and so I argue that it makes no sense to propagate positions to these derived nodes. These shouldn't even be a problem. Positions should only be used for error reporting and we should catch errors before or during desugaring, not after (we shouldn't desugar incorrect programs). I realize that this last condition might not actually hold, i.e. we first desugar and then report errors, but if that is the case then I argue this is a Bad Thing (TM) and it should be fixed.

Note also that sugarTraversals maintain positions without changing them.

dhil:
> Positions should only be used for error reporting and we should catch errors before or during desugaring, not after (we shouldn't desugar incorrect programs). I realize that this last condition might not actually hold, i.e. we first desugar and then report errors, but if that is the case then I argue this is a Bad Thing (TM) and it should be fixed.

Debugging Links programs is somewhat painful, I reckon taking away helpful information such as positions on desugared nodes will just make it even more painful to debug programs. Therefore I suggest we wait with merging this PR until the the invariant that you describe is true.

jstolarek:
> I suggest we wait with merging this PR until the the invariant that you describe is true.

I believe it will be easier to reach that invariant by merging this PR first. Why? Firstly, I don't know whether this invariant is true or not at the moment. I would have to inspect a significant chunk of Links source code to see whether that's the case - and I still might make a mistake. Now, if me merge this PR and then if we manage to get an error message with a dummy position attached to it then that's a clear bug that points us to something that should be fixed. If we keep things as they are, even if we get error messages generated after desugaring, we might not realize that this is something that shouldn't happen. 

BTW. my changes didn't break any of the tests, so perhaps there is no situation at the moment where we are using this position information.

jamescheney:
This is related to an earlier discussion somewhere about whether to generalize the type of "positions" to allow us to say "this ast node was created by desugaring pass XYZ using input ast nodes with the following "position" information" - where the position information of the previous ast nodes might be a plain source code location or another derivation step.  (I.e. a form of "how" provenance).  I think Janek's response to that was that we should not need that kind of information if everything works, i.e. in the end state mentioned above where everything that typechecks desugars properly.  I submit that since Links is a permanent work in progress, we will usually not be in that state and thus, at least optionally, maintaining more detailed information that might be helpful to get an idea of what went wrong might be a good idea.

slindley:
I agree with @jamescheney. This is a controversial change that I suggest we discuss at the next Links developer meeting.

One problem that this issue highlights is that our test suite does not cover (and possibly cannot cover) much of the diagnostic information that we may need when developing Links. Given that Links is primarily a research vehicle, this might be something we want to think about addressing.

jstolarek:
Ok, let's discuss this during the next dev meeting.

jstolarek:
I updated this PR following yesterday's discussion during dev meeting. It fixes #460 and #480, and is rebased on top of polymorphic variants branch.

The important bit is that now there are two concrete implementations of `SugarConstructors` functor available: one that allows attaching standard sugartypes positions (this one is used to traverse the AST and leave the positions untouched, but also to propagate positions to derived nodes) and one that only allows to attach dummy positions. See Note [Attaching positions to nodes] in sugarConstructors.ml

Issue #458, opened 2019-01-16
query.ml and queryshredding.ml misaligned?


jamescheney:
In commit a543381 which fixed #25, changes were made to query.ml.  At the time the query shredding code was in a parallel branch.  Later, it was merged in as a separate file queryshredding.ml, which does not have this change.  This may mean #25 was not fixed for shredding.  

- [ ] Add test that runs the test for bug  #25 with shredding enabled (and possibly all the other database tests?)
- [ ] Fix queryshredding if needed / resolve this by merging query.ml and queryshredding.ml (cf #431)

wricciot:
My current code (largely importing QueryShredding into Query) does indeed fail on the xpath.links and xpath-reduced.links tests with error DbEvaluationError("Error projecting from record").

More importantly, if I turn on shredding on those examples, I get the following errors:

cmnd=./links --config=tests/database/config tests/database/xpath.links
*** Error: Links_core.Queryshredding.Eval.DbEvaluationError("Error projecting from record")
FAILED TEST xpath
cmnd=./links --config=tests/database/config tests/database/xpath-reduced.links
FAILED TEST xpath-reduced:[(name = "f"), (name = "e"), (name = "c")]but got:[(name = "c"), (name = "e"), (name = "f")]

While the second message should be benign (tuples are reordered by shredding, as expected), the first one appears to show that #25, indeed, was not fixed for shredding.

Issue #431, opened 2018-11-28
refactor and merge Sql and ShreddedSQL modules 


jamescheney:
Currently `query.ml` and `queryshredding.ml` contain similar (but not identical) modules for representing SQL abstract syntax and generating concrete query strings.

These should be factored out and merged so that query.ml and queryshredding.ml both use one SQL module.

- [x] Move both modules (and any SQL-related code) out of `query*.ml` and into separate files
- [x] Compare and merge the differences between them so that we have just one Sql module that both `query*` modules can use

PR #462, opened 2019-01-24
Query refactoring/1

wricciot:
This is the first part of Query/Queryshredding refactoring. It merges common parts of Queryshredding into Query (up to the Eval module), and moves non-shredding parts of Query into Queryplain.
Fixes the bug in #458 and partially addresses #431.


jamescheney:
Thanks!  Looks like the CI is failing due to trailing whitespace somewhere.  We have an automated check that complains about that. It's probably worth having a look over:

https://github.com/links-lang/links/wiki/Working-conventions#trailing-whitespace

(and possibly also the rest of the wiki page, though I think most of it is now redundant given that we decided to use "squash and merge" in pull requests which gets rid of any merge commits that are present in the PR.)

core/types.mli
```diff
@@ -423,3 +424,4 @@ end
 
 module Transform : TYPE_VISITOR
 module ElimRecursiveTypeCyclesTransform : TYPE_VISITOR
+
```
jamescheney: Unnecessary trailing newline?

core/queryplain.ml
```diff
@@ -0,0 +1,706 @@
+open Utility
```
jamescheney: I'm not excited about the Query/Queryplain/Queryshredding nomenclature.  I gather that Query now has common code that both Queryplain and Queryshredding use, and that the latter differ in the "evaluation" (IR -> normalized QL) code?

I admit guilt for introducing the bad module name Queryshredding.  I suggest the following to expiate my sins:

- Query
- EvalQuery
- EvalNestedQuery

I also wonder if it might make sense to move query*.ml* into a subdirectory "query" or "database", as well.

wricciot:
I had indeed introduced trailing whitespace as a result of some cut and paste.
After the last commits, the various query* have been moved to a query/ subdirectory with new names.

core/types.mli
wricciot: Now fixed.

core/queryplain.ml
wricciot: Renamed and moved to query/ subdirectory.

core/query/evalQuery.mli
```diff
@@ -0,0 +1,9 @@
+open Utility;;
```
dhil: The double semicolons `;;` are unnecessary.

core/query/query.mli
```diff
@@ -0,0 +1,104 @@
...
+
+val used_database : t -> Value.database option
+    ;;
+
+val string_of_t : t -> string
+
+val type_of_expression : t -> Types.datatype
+    ;;
+
+val default_of_base_type : [> `Bool | `Char | `Float | `Int | `String ] ->
```
dhil: I would create an alias of variant row, and name it, say, `base_type`.

core/query/evalQuery.mli
wricciot: Yes, there are many instances of this in the code. I will remove them when I come across them (like everybody else should).

core/query/query.mli
wricciot: Makes sense. However I'm not entirely sure about why default_of_base_type has a catch-all case whose only purpose is to raise assert false.

wricciot: Changes made. I noticed there is a "Types.primitive" that's very similar to the new "Types.base_type", with extra cases for `XmlItem and `DB. I have chosen to keep the two types separate because, as the code changes, I would prefer to see a type error at compile time rather than a random assertion failure at runtime (raised e.g. by Query.default_of_base_type). If somebody feels these should be only one type, let me know.

jamescheney: I think they should probably be separate types, since XmlItem and DB values are not possible to return in the result of a (flat SQL) query  It might be good to document the difference in the name, e.g. rename `Types.base_type` to `Types.query_primitive` or (eventually) `Sql.primitive` or something like that.

dhil: I would probably implement the `base_type` (or whatever we call it) in the `Query` module to keep it local. If you want to keep it in the `Types` module, then you can avoid a bit of code duplication by
```
type base_type = [ `Bool | `Char | `Float | `Int | `String ]
type primitive_type = [ base_type | `XmlItem | `DB ]
```

wricciot: Yes, it makes sense to move this definition to Query. Changes made.

core/query/evalQuery.mli
jamescheney: If we really wanted to get rid of `;;` we could add an automated check that complains about them, same as trailing whitespace...  I am sure some of them have been there for over a decade.

jamescheney:
So this looks ready to be merged now, any objections?

PR #465, opened 2019-01-25
Whitelist for whitespace check (Fixes #464)

### LINKED ISSUE #464: Whitespace rule checker checks binary files
The whitespace rule checker currently runs the following command:

```
find "$dir"  -type f -not -name "*.pdf" -exec egrep -l " +$" {} \;
```
meaning that it is run on every file that isn't a PDF. Adding a few PNGs seemed to cause the check to fail.

I wonder whether it might be better to go for a "check these file types" approach, rather than a "don't check these file types" approach, allowing us to restrict attention to only files which are textual. For example, I would consider the following list:

- ml
- mli
- mly
- mll
- py
- pl
- html
- js
- txt
- md
- links
- css
- sh

SimonJF:
This PR only checks textual files defined in a whitelist, as opposed to all files except PDFs, for trailing whitespace.

Note that due to syntactically-significant trailing whitespace in Markdown files (as identified by @dhil),`md` files are excluded from this check.

Also, I toyed with the idea of storing the whitelist in an array, but decided against this as it wasn't pleasant to programmatically build up the `find` command, and the alternative was to run the `find` command `m * n` times, where `m` is the number of directories and `n` is the number of extensions.

### DIFF
--- a/tools/rules/trailing_whitespace
+++ b/tools/rules/trailing_whitespace
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/bash
 #
 # Checks whether source files contain trailing script.
 #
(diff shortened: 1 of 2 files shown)


dhil:
I think you can "build up" the find pattern, e.g. something along the lines of
```sh
WHITELIST=(".ml" ".links")
FIND_PATTERN=${WHITELIST[0]}
unset WHITELIST[0]
for ext in ${WHITELIST[@]}; do
  FIND_PATTERN="$FIND_PATTERN -o *.$ext"
done

for f in $(find "$dir" -type f $FIND_PATTERN); do ... done
```
I may not have gotten the syntax right.

SimonJF:
OK, I've added an updated version which uses an array. The string interpolation gets hairy because the paths need quotes, necessitating `eval`.

tools/rules/trailing_whitespace
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/bash
 #
 # Checks whether source files contain trailing script.
```
dhil: Oops. I suppose, I originally meant to type "Checks whether **any** source files contain trailing **white space**".

tools/rules/trailing_whitespace
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/bash
```
dhil: Why have you changed this?

SimonJF: Apologies and well-spotted, it crept in during debugging. Sorted now.

tools/rules/trailing_whitespace
SimonJF: Fixed :)

SimonJF:
@slindley quite possibly, but that's a little beyond the scope of #464, so let's do that separately?

PR #467, opened 2019-01-25
Make the `End` session type linear

### LINKED ISSUE #449: Session buffers are not garbage-collected
Session endpoints of type `End` are unrestricted and may be discarded. However, there is currently no logic on either the server or the client for removing the empty buffers once the endpoints have been discarded.

With a `close` primitive and a linear `End` type, this would be straightforward -- simply delete the buffer.

The way to solve this without such a primitive is to delete the entry in the buffers table whenever it is empty, and create it whenever a message is received and it doesn't exist. However, presence of a buffer in the local endpoints table is currently used to indicate that a communication is local or remote, for distributed session-typed communication. Thus, it'll be necessary to track presence of an endpoint separately.

SimonJF:
As identified in #449, while having an unrestricted `End` type is convenient for programming, it is difficult to see how buffers can be garbage-collected. In contrast, a linear `End` type coupled with a `close` operation makes this trivial. This PR should therefore fix #449.

This PR makes `End` linear, and introduces a `close` primitive which deletes the corresponding entry in the buffers table.

### DIFF
--- a/prelude.links
+++ b/prelude.links
@@ -1200,12 +1200,13 @@ typename EndQuery = ?().End;
 
 sig wait : (EndQuery) ~> ()
 fun wait(s) {
-  ignore(receive(s))
+  var (_, s) = receive(s);
+  close(s)
 }
 
-sig close : (EndBang) ~> ()
-fun close(s) {
-  ignore(send((), s))
+sig closeBang : (EndBang) ~> ()
+fun closeBang(s) {
+  close(send((), s))
 }
 
 sig makeEndBang : () ~> EndBang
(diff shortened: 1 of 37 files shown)


prelude.links
```diff
@@ -1226,7 +1227,7 @@ fun forkSync(f) {
   var _ = spawn {
     var c = accept(ap);
     var c = f(c);
-    close(c)
+    closeBang(c)
```
dhil: Maybe we should amend the lexical definition of variables to include symbols such as exclamation marks.

dhil: I did not mean to request a change for the above, rather it was more of a general comment.

PR #469, opened 2019-01-30
Adds support for XML comments. Resolves #270.

### LINKED ISSUE #270: Cannot write comments in XML mode
Suppose the I want to add comments to my code.

When doing that in between XML there are the following scenarios:
```
<div>
  # Links style comment
</div>
```
Prints the comment on the page

---
```
<div>
  <!-- XML style comment -->
</div>
```
```
Unexpected character : <
            <!-- -->
              ^
```
---
```
<div>{
  # Comment back in Links mode
}</div>
```
```
Type error: XML child nodes must have type `Xml'
```
---
The only thing that works is
```
<div>{
  # Comment with empty XML value
  <#/>
}</div>
```

dhil:
Implements something like https://www.w3.org/TR/REC-xml/#sec-comments. It is debatable whether we want to ignore XML comment or reify them as XML nodes. In this patch I have opted for the former. I believe with little extra effort we could implement the latter.

Fixes #270.

### DIFF
--- a/core/lexer.mll
+++ b/core/lexer.mll
@@ -295,6 +295,7 @@ rule lex ctxt nl = parse
   | "|>"                                { RIGHTTRIANGLE }
   | '<' (def_qname as id)               { (* come back here after scanning the start tag *)
                                           ctxt#push_lexer (starttag ctxt nl); LXML id }
+  | "<!--"                              { xmlcomment_lex ctxt nl lexbuf }
   | "[|"                                { LBRACKETBAR }
   | "|]"                                { BARRBRACKET }
   | '['                                 { LBRACKET }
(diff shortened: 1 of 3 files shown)


jamescheney:
At the moment it doesn't make a huge amount of sense to retain XML comments as nodes.  However I can think of realistic settings where you want to generate X(HT)ML that has comment nodes or other "unusual" constructs such as processing instructions or CDATA nodes.  This probably comes under #46.

dhil:
I've added a few more tests.

frank-emrich:
To me this looks as if we can now use XML comments outside of XML blocks. Is that what we want? It means we have two different syntactic ways of writing comments in non-XML Links code now. On the other hand, this may be a good thing as it introduced a way of writing multi-line comments.

dhil:
> To me this looks as if we can now use XML comments outside of XML blocks.

An XML comment is XML. I would argue that XML comments aren't "outside of XML blocks", but rather delimit an XML block.

> It means we have two different syntactic ways of writing comments in non-XML Links code now. On the other hand, this may be a good thing as it introduced a way of writing multi-line comments.

That is correct, as long as we ignore XML comment you can abuse this to have multi-line comments in Links.

SimonJF:
Thanks for addressing my comments, @dhil . I think this is ready to be merged, unless anyone has objections.

frank-emrich:
looks good to me

SimonJF:
Great, let's merge. Thanks everyone.

Issue #470, opened 2019-02-06
yojson deprecation 


jamescheney:
In recent CI builds e.g. https://travis-ci.org/links-lang/links/builds/489548988 it seems that the current version of yojson is giving us deprecation warnings:

```
Error (warning 3): deprecated: Yojson.Safe.json
 json types are being renamed and will be removed in the next Yojson major version. Use type t instead
```

which kill the build.  These seem to be due to whatever code for deriving yojson, not anything we can directly change, so we should probably just ignore these deprecation warnings for now, but keep this issue open to remind us (and possibly raise an issue somewhere wherever the yojson deriving code lives - but presumably that will change once the promised breaking change happens in yojson.

jamescheney:
Yeah, I saw that too, just a question of how long it will be until their next release.

jamescheney:
This seems to have been fixed by an unversioned patch to the `ppx_deriving_yojson` OPAM file, that restricts the version of `yojson` to < 1.6.0.  So we don't need the warning suppression, provided they do the right thing whenever `ppx_deriving_yojson` is updated.

PR #471, opened 2019-02-06
disable deprecation warnings due to yojson deriving for now

jamescheney:
This should temporarily work around #470, let's see what the CI says


jamescheney:
Merging

PR #472, opened 2019-02-07
Query SQL refactoring

### LINKED ISSUE #271: Table names are case-insensitive
When working with tables with case-sensitive table-names, Links crashes because the relation could not be found. This is unspurising, since the created query uses the unquoted table-name:
```SQL
SELECT (t3029."id") as "id" FROM myUsers as t3029;
```
Links-strings, used to name tables, are case-sensitive.
Attributes are translated quoted, so they case-sensitive as well.

This makes this unnecessarily confusing. 
All database-related strings should be treated the same with respect to case.

### LINKED ISSUE #431: refactor and merge Sql and ShreddedSQL modules

wricciot:
This is the second part of the Query/Queryshredding refactoring. It factorizes the SQL query generation into a separate Sql module, which is used by EvalQuery and EvalNestedQuery.
Fixes #271 for shredding (issue was closed but only addressed for regular queries). 
Closes #431.


core/query/sql.ml
```diff
@@ -0,0 +1,467 @@
...
+
+let string_of_query db range q =
+  let range =
+    match range with
+      | None -> ""
+      | Some (limit, offset) -> " limit " ^string_of_int limit^" offset "^string_of_int offset
+  in
+    string_of_query db false q ^ range
+
+let prepare_clauses : Q.t -> Q.t list =
```
jamescheney: THe code from here to about line 436 seems to be more about "converting Q.t to Sql.query" (with/without shredding-specific complications such as indexes) and not about SQL per se.  I wonder if this code should live in query.ml instead of sql.ml.  (I'd also like to get rid of any dependency of sql.ml on query.ml - if anything the dependency should be in the other direction, or the translation from query.t to sql.t should happen in a later module that imports both.)

wricciot: Ah yes, that makes sense. I'll look into it.

wricciot: It looks like `update` and `delete` will also need to be moved to `Query`, as they both depend on `Query.base`.
This is hardly surprising, as we already had `ordered_query` in `EvalQuery` and `unordered_query_package` in `EvalNestedQuery`

jamescheney: That seems like an improvement.  The (compile) insert/delete/update functions are really about Links's interface to SQL updates, we don't have any particular syntax for SQL updates themselves at the moment.  These might eventually deserve to be in a separate (Update) module at some point but I can live with this for now.

jamescheney:
I think this can be merged next assuming that updating w.r.t. master goes smoothly.

jamescheney:
Let's wait until #479 is merged in case there are any warnings that the CI is missing.  (There don't seem to be.)

Issue #468, opened 2019-01-29
Virtual modules: Reorganisation of built-in functions and values.


dhil:
Once the (proper) module system lands, the (front end) compilation pipeline will maintain knowledge of modules (rather than as today where they are desugared early on and entirely forgotten about). Thus, the compiler gains the power to treat modules as proper primitives. We can leverage this new power to organise the prelude and built-in functions and values into modules. 

Splitting the prelude into multiple (coherent) modules will be mostly straightforward (except for the cases where the compiler relies on the existence of particular functions in the prelude). 

Some of the built-in functions, such as the DOM functions, can be hoisted out of the compiler and defined using the foreign function interface in a user-definable module (like the prelude). However, presently, there are other functions that cannot be expressed using the foreign function interface (e.g. functions that lives both on the client and server). With a modest extension to the module system implementation, we could organise such functions and values into "virtual" modules.

As a result, we get coherent APIs for built-ins, and we localise their definitions (or in most cases declarations). Furthermore, we ought to be able to get rid of the monolithic `lib.ml`.

Issue #455, opened 2019-01-09
Typename shadowing


dhil:
Shadowing of type names is, to put it mildly, confusing.

Somewhat surprisingly, the following program type checks:
```links
typename Foo = Int;
typename Foo = Bool;

sig f : (Foo) {}~> Int
fun f(foo) { foo }

links> f(2)
2 : Int
```

Reading the program top-down, I would expect to see a type error. I expect the type of `f` to be structurally equivalent to `(Bool) {}~> Int`, it so happens that it isn't. It is equivalent to `(Int) {}~> Int`, which is telling that Links does not respect shadowing of type names. Either we should prohibit duplicate declarations or support shadowing in the same scope.

The reason why the first `Foo` is picked over the second `Foo`) is due to `refineBindings.ml`.

jamescheney:
Weirdly, if I do this one line at a time in the REPL, the right thing happens - i.e. you get an error 

```
links> typename Foo = Int;
Foo = Int
links> typename Foo = Bool;
Foo = Bool
links> sig f : (Foo) {}~> Int
...... fun f(foo) { foo };
<stdin>:2: Type error: The non-recursive function definition has return type
    `Foo'
but its annotation has return type
    `Int'
In expression: fun f(foo) { foo }.
```

dhil:
I have a patch now that makes duplicate type names in same scope an error.

For example, for the following program
```
typename Foo = Int;
typename Foo = Float;

sig f : (Foo) {}~> Int
fun f(foo) { foo }

typename Foo = Bool;
```
Links now produces the following error
```
Syntax error: The type name Foo is declared multiple times.
  First declaration at file test.links:8:
    - typename Foo = Int
  Subsequent declaration at file test.links:10:
    - typename Foo = Float
  Subsequent declaration at file test.links:17:
    - typename Foo = Bool
```

SimonJF:
Great, that sounds like a more sensible solution. Is this part of your larger nominal typename patch?

dhil:
> Great, that sounds like a more sensible solution. Is this part of your larger nominal typename patch?

I intend to separate them out. There is a little bit more work to be done on `refineBindings`. I intend to upstream that work first.

For value bindings I have adopted a similar rule, for example, the following program:
```
fun a() { println("a1") }
fun b() { a(); c() }
fun a() { println("a2") }
fun c() { a() }
```
fails to compile with the following error:
```
Syntax error: The function a is declared multiple times within the same binding group.
[...]

PR #476, opened 2019-02-12
Add a simple links-mode for Emacs

jstolarek:
Because I was bored. Supports:

  * highlighting of keywords, primitive types, builtins and comments
  * commenting/uncommenting using `M-;`

Nothing fancy. Install by putting `links-mode.el` where Emacs can find it and adding `(require 'links-mode)` to your `.emacs` file.


fehrenbach:
You might want to merge this in with https://github.com/links-lang/links/blob/master/links-mode.el

jstolarek:
Good point. Done now. So this essentially ads highlighting builtin primitive types and functions.

fehrenbach:
Is highlighting of built-in functions something people actually want? I always find it irritating when functions look differently just because they are built-in/have been defined in the prelude. Maybe some magical ones, like `sysexit`, but `take` and `drop`? (And *not* `map`, by the way.)

But then again, I'm not exactly an active Links user these days, so my opinion should probably not count for very much :)

dhil:
How does the highlighting of primitives interact with shadowing? 

As outlined in #468 I would like to hoist some of the primitives out of the compiler, and implement them using Links. At that stage, they may well cease to be considered as primitives.

jstolarek:
As you can see the list of builtins is currently hard-coded, so it changes we would need to update it.

> I always find it irritating when functions look differently just because they are built-in/have been defined in the prelude.

So I wonder what others think. I personally find highlighting of builtins helpful but that might be because I'm less experienced in Links and don't know all of them by heart.

dhil:
> As you can see the list of builtins is currently hard-coded, so it changes we would need to update it.

I understand that as it would interact badly with shadowing of primitives, e.g.
```links
fun error(x) { x + 1 }
fun main() {
  error(2)
}
```
then `error` would be wrongly highlighted as a primitive function.

jstolarek:
Yes, that's correct.

dhil:
> Yes, that's correct.

OK. Thanks. Then I will object to merging this into `links-mode.el`.

jstolarek:
I agree, this is a flaw I haven't thought about.

That leaves us with highlighting of primitive types (Int, Char, etc.). Is that an acceptable addition? If so I will update the PR.

fehrenbach:
Can Emacs "faces" inherit from each other? We could create a new built-in-function face that inherits from whatever the default face is. This way, people who want to highlight built-ins can just customize the appropriate face and others will never know the difference.

dhil:
Well, as discussed in #455 the intended semantics is to support shadowing of types, I don't know whether the intention is to cover primitive types too. The current implementation does not support this properly (that is evident from looking at how `parser.mly` handles data types declarations).
```links
links> typename Bool = Int;
Bool = Int
links> sig f : (Bool) -> Bool fun f(x) { x };
f = fun : (Bool) -> Bool
links> f(2);
<stdin>:1: Type error: The function
    `f'
has type
    `(Bool) -a-> Bool'
    `Int'
    `wild'
In expression: f(2).
```

fehrenbach:
Somewhat inconsistently I like highlighted primitive types. (Because I can never remember whether any particular given language spells it Bool or Boolean...)

jstolarek:
> Can Emacs "faces" inherit from each other?

An educated guess would be that this is possible, given how customizable Emacs is, but I'd have to spend some time to go through Emacs documentation to figure out how to do it.

jstolarek:
I'm sending this PR down the drain. The already existing links-mode does the most important thing that I need (block commenting/uncommenting) and the changes this PR introduces are controversial.

PR #481, opened 2019-02-14
Remove some polymorphic variants in Sugartypes

jstolarek:
A while back we had a discussion about possibly removing use of polymorphic variants in the frontend (Sugartypes). I've made another attempt at doing this with moderate success. There are some datatypes where we can remove the ticks and things Just Work. This patch redefines such easy-win datatypes as standard OCaml variants.

For the majority of the datatypes though the required changes are more involved. I'd like to discuss this during the next Links dev meeting.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,4 +19,4 @@ script:
   - make nc-release
   - make tests
   - ./run-tests db-only shredding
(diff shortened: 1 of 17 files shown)


dhil:
Brilliant! This looks like a step in the right direction.

Issue #439, opened 2018-12-05
Insert propagation backwards through join lens is broken


jamescheney:
In the lens example in `examples/relational_lenses`, if you `put` a new row that mentions a new album, only the new track row gets created, not the new album row.  For example:

```

var db = database "links";

var albumsTable =
	table "albums"
	with (album: String, quantity: Int)
	tablekeys [["album"]]
	from db;

var tracksTable =
	table "tracks"
	with (track: String, date: Int, rating: Int, album: String)
	tablekeys [["track", "album"]]
	from db;

var albumsLens = lens albumsTable with { album -> quantity };
var tracksLens = lens tracksTable with { track -> date rating };

var joinedLens = lensjoin albumsLens with tracksLens on album delete_left;

var tracks = lensget joinedLens;

var newTracks = tracks ++ [(track="It's the end of the world as we know it",
                            rating=5, album="Eponymous", date=1988, quantity=42)];

var r = lensput joinedLens with newTracks;
newTracks == (lensget joinedLens)
```

returns `false`.

rudihorn:
I've had a quick look at this issue and it seems to just be that the `get_changes` function is not quite correctly implemented for what should be the delta merge function. I'll make sure to fix this in #454.

rudihorn:
This should now be fixed in #454 (except for the fact that the order of newTracks in this example would not match that of tracks, but considered as a set they are equal).

jamescheney:
Thanks.  Yes, instead of `==` the code should have something that either tests set equality of the two lists (by sorting by key first, say) or just checks for the correct existence of the new track.

BTW, would tests like this be possible to add to the relational lens tests (using a similar strategy to the db and shredding tests?)  Then the CI would automatically check that this issue is resolved / not regressed.

rudihorn:
So adding these tests by rewriting them as unit tests would be possible, but adding them directly in links might be a little more difficult (because of a lack of code to handle set types etc.). Though I was wondering if it would be nice to expose the `lens_sorted_records` as a links module, as this would make it easier to write such tests (and also to play around with relational data) in links a bit easier.

jamescheney:
@rudihorn Did #454 fix this?

rudihorn:
@jamescheney I've created two tests in #482 which are now executed by travis to check this issue and the example in the RL paper. Unfortunately it is not using set semantics and is dependent on the order that postgres returns its results.

jamescheney:
[...]

PR #482, opened 2019-02-15
Relational Lens Tests

rudihorn:
Added explicit tests to run the example and ensure that it works as expected.

I will first only test this in Travis and may add some more unit tests. This is also to ensure #439 is fixed (which it should be).


rudihorn:
I'm happy for this to be merged.

PR #484, opened 2019-02-18
Remove redundant code???

jstolarek:
`desugarInners` has several places that say:

```OCaml
let extra_env = extra_env in
```
Looks redundant to me, but perhaps this is some smart OCaml hack that I don't understand? Thought I'd ask before merging.


dhil:
I don't think there is any subtle trick here. In other places, such as `transformSugar` you will see the pattern
```ocaml
let env = o#backup in (* a value copy of `env' *)
let (x, o) = o#do_something in
let o = o#restore env in
...
```
It could be that a similar pattern was used in this module in the past.

PR #485, opened 2019-02-18
The obvious thing works to allow qualified type constructor applications

### LINKED ISSUE #483: Cannot parse qualified parameterised typenames
On the master branch it seems that qualified parameterised type names cannot be parsed.
foo.links
```
typename T(a) = Bool;
typename S = Bool;
```
bar.links
```
open Foo;
sig f : (Foo.T(Int)) -> Foo.T(Int)
fun f(x) { not(x) }
```
```
$ links -m --path=. bar.links
*** Parse error: bar.links:2

  sig f : (Foo.T(Int)) -> Foo.T(Int)
                   ^
```

However, if the type name has no parameters, then it parses just fine.
baz.links
```
open Foo;
sig g : (Foo.S) -> Foo.S
fun g(x) { not(x) }
```
```
$ links -m --path=. baz.links
() : ()
```

I suppose #459 may not have accounted for qualified parameterised type names.

jamescheney:
Hopefully fixes #483

### DIFF
--- a/core/parser.mly
+++ b/core/parser.mly
@@ -944,6 +944,7 @@ session_datatype:
 | END                                                          { `End             }
 | primary_datatype                                             { $1               }
 | qualified_type_name                                          { `QualifiedTypeApplication ($1, []) }
+| qualified_type_name LPAREN type_arg_list RPAREN              { `QualifiedTypeApplication ($1, $3) }
 
 parenthesized_datatypes:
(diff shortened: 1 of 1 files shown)


dhil:
Thank you!

jamescheney:
I do sort of wonder if we should merge the "bare constructor" and "qualified constructor" cases.  I'll leave this to the real module system experts out there...

Issue #255, opened 2017-09-18
Investigate 'visitors' and 'alphalib' for building SugarTraversals?


yallop:
The [`SugarTraversals` module][sugar-traversals] module was once generated automatically from the datatype definition by a [Camlp4 extension][map-generator].  Nowadays it seems to be maintained by hand.

It may be worth investigating @fpottier's [visitors library][visitors] for generating visitor classes, and [alphaLib library][alphalib] for dealing with binding, which are both described in [_Visitors Unchained_](http://gallium.inria.fr/~fpottier/publis/fpottier-visitors-unchained.pdf) (ICFP 2017), and which support a principled approach to traversing ASTs with binding using modern OCaml tools such as [ppx][ppx], with the aim of returning to a situation where `SugarTraversals` no longer needs to be updated by hand.

[sugar-traversals]: https://github.com/links-lang/links/blob/c49f16135579f90d1932e534666d63a3c70c9210/sugarTraversals.ml
[map-generator]: https://github.com/ocaml/camlp4/wiki/Camlp4MapGenerator
[visitors]:  https://gitlab.inria.fr/fpottier/visitors
[alphalib]: https://gitlab.inria.fr/fpottier/alphaLib
[ppx]: https://whitequark.org/blog/2014/04/16/a-guide-to-extension-points-in-ocaml/

jamescheney:
We should do this for 0.8.

fpottier:
Hi! `visitors` is released and fully documented.
`alphaLib` is not really released or documented,
but I'd be happy to answer questions about it.

dhil:
Large portions of the code base depends on the visitors defined in SugarTraversals. To assess whether we want to use these libraries to derive visitors, it may be worthwhile, as a starting point, to use them to define visitors for the backend IR. Contrary to the front-end AST, the back-end IR has only a few (unused) optimisation passes/visitors, meaning that we can rapidly prototype the libraries for the IR. It may relevant for @frank-emrich's work on type checking the IR.

frank-emrich:
I have now had a more detailed look at the visitors library/preprocessor and identified a few problems:

1. The library does not support polymorphic variants. However, virtually all of Links (the surface syntax, IR syntax, types,...) is represented by polymorphic variants. I don't know if this is really necessary or if we could switch to "normal" variants, but it would require a lot of work to perform those changes.

[...]

PR #487, opened 2019-02-19
Remove polymorphic variants using module namespacing

jstolarek:
A while back I started [an email thread about warnings generated by the compiler when several datatypes share the same constructor](http://lists.inf.ed.ac.uk/pipermail/links-dev/2019-January/000133.html).  In such cases compiler uses type-directed disambiguation and for the most time it guesses wrong.  Ideas to solve the problem included:
 
 1) using type annotations
 2) putting datatype definitions into modules
 3) prefixing constructor names so that they are different

I wasn't happy with (1) because it meant awkward clutter. Sam pointed out that (3) is a poor-man's namespacing. So I decided to experiment with (2) to see how it works out. I placed `pattern` and `patternode` datatypes in a `Pattern` module like this:

```OCaml
module Pattern = struct
  type node =
  | Any
  | Nil
  | Cons     of t * t
  | List     of t list
  | Variant  of name * t option
  | Effect   of name * t list * t
  | Negative of name list
  | Record   of (name * t) list * t option
  | Tuple    of t list
  | Constant of constant
  | Variable of binder
  | As       of binder * t
  | HasType  of t * datatype'
   [@@deriving show]
  and t = node with_pos
   [@@deriving show]
end
```

and then changed all the code accordingly. Thoughts, questions and observations:

1. This approach adds some clutter to the code. Most pattern matches are now prefixed with `let open Pattern in`, unless we only match on one or two constructors, in which case I used name quantification.  Even with `let open` quantification is sometimes required because OCaml is not very good at type inference:

```OCaml
let open Pattern in
match p with
| Pattern.Any -> ()
| Nil -> ()
| _ -> ()
```

Here if `Any` isn't prefixed the compiler complains about type-directed disambiguation.  An alternative solution is to put `Nil` branch first, because `Nil` does not appear in other datatypes (yet):

```OCaml
let open Pattern in
match p with
| Nil -> ()
| Any -> ()
| _ -> ()
```
This generally works, but it might force us to order branches in an unintuitive order.  Another annoying part was the parser, where every use of constructor has to be prefixed.
 
Is all that clutter acceptable? Perhaps a shorter module name (`Pat`?) would be better?

2. `Pattern.t` is now a pattern with position, while `Pattern.node` is just a pattern. Is that intuitive? Now that I wrote this I am having thoughts that a more intuitive naming would be to have `Pattern.t` as the pattern datatype with all the constructors and `Pattern.with_pos` as pattern with position.

3. In Sugartypes there is a long series of definitions of mutually recursive datatypes that include `phrase`, `binding`, and all the accompanying datatype definitions. Within that group two constructors in `cp_phrasenode` (`Select`, `Offer`) are repeated in other datatypes. I am leaning strongly towards prefixing all constructors in `cp_phrasenode` with `CP`. This would mean that the whole recursive group can remain as is, without the need of putting all the datatype definitions into mutually recursive modules.

4. Unrelated to this PR, I think some datatypes would perhaps be better kept as polymorphic variants.  I am in particular thinking about `freedom`.  The reason is its interaction with `Types.Vars.flavour` and subtype casting from `freedom` to `flavour` (e.g. in `free_bound_type_vars`).  While a direct cast can be easily replaced with a function (a small runtime expense), we also seem to be using subtyping on data structures. From my understanding of code this would require rebuilding the whole data structure just to perform a cast on elemnents stored inside, which is potentially a large runtime expense.

Any thoughts and feedback welcome.

### DIFF
--- a/core/commonTypes.ml
+++ b/core/commonTypes.ml
@@ -1,15 +1,17 @@
 module Linearity = struct
   type t = Any | Unl
     [@@deriving eq,show]
-end
 
-let string_of_linearity = function
-  | Linearity.Any -> "Any"
-  | Linearity.Unl -> "Unl"
+  let isAny lin = lin == Any
+  let isUnl lin = lin == Unl
+
+  let string_of = function
+    | Any -> "Any"
+    | Unl -> "Unl"
+end
 
 let linUnl = Linearity.Unl
 let linAny = Linearity.Any
-let isUnl lin = lin == Linearity.Unl
 
 module Restriction = struct
   type t =
@@ -30,15 +32,15 @@ module Restriction = struct
   let isSession = function
     | Session -> true
     | _       -> false
+
+  let string_of = function
+    | Any     -> "Any"
+    | Base    -> "Base"
+    | Session -> "Session"
+    | Effect  -> "Eff"
 end
 
 let resAny     = Restriction.Any
 let resBase    = Restriction.Base
 let resSession = Restriction.Session
 let resEffect  = Restriction.Effect
-
-let string_of_restriction = function
- | Restriction.Any     -> "Any"
- | Restriction.Base    -> "Base"
- | Restriction.Session -> "Session"
- | Restriction.Effect  -> "Eff"
--- a/core/commonTypes.ml
+++ b/core/commonTypes.ml
@@ -10,6 +10,7 @@ module Linearity = struct
     | Unl -> "Unl"
 end
 
+(* Convenient aliases for constructing values *)
 let linUnl = Linearity.Unl
 let linAny = Linearity.Any
 
@@ -20,6 +21,10 @@ module DeclaredLinearity = struct
   let isLin lin = lin == Lin
   let isUnl lin = lin == Unl
 end
+
+(* Convenient aliases for constructing values *)
+let dlLin = DeclaredLinearity.Lin
+let dlUnl = DeclaredLinearity.Unl
 
 module Restriction = struct
   type t =
@@ -48,7 +53,26 @@ module Restriction = struct
     | Effect  -> "Eff"
 end
 
+(* Convenient aliases for constructing values *)
 let resAny     = Restriction.Any
 let resBase    = Restriction.Base
 let resSession = Restriction.Session
 let resEffect  = Restriction.Effect
+
+module PrimaryKind = struct
+  type t =
+    | Type
+    | Row
+    | Presence
+    [@@deriving show,eq]
+
+  let string_of = function
+    | Type -> "Type"
+    | Row -> "Row"
+    | Presence -> "Presence"
+end
+
+(* Convenient aliases for constructing values *)
+let pkType     = PrimaryKind.Type
+let pkRow      = PrimaryKind.Row
+let pkPresence = PrimaryKind.Presence
--- a/core/commonTypes.ml
+++ b/core/commonTypes.ml
@@ -1,0 +1,12 @@
+module Linearity = struct
+  type t = Any | Unl
+    [@@deriving eq,show]
+end
+
+let string_of_linearity = function
+  | Linearity.Any -> "Any"
+  | Linearity.Unl -> "Unl"
+
+let linUnl = Linearity.Unl
+let linAny = Linearity.Any
+let isUnl lin = lin == Linearity.Unl
--- a/core/commonTypes.ml
+++ b/core/commonTypes.ml
@@ -2,29 +2,39 @@ module Linearity = struct
   type t = Any | Unl
     [@@deriving eq,show]
 
-  let isAny lin = lin == Any
-  let isUnl lin = lin == Unl
+  let is_any = function
+    | Any -> true
+    | _   -> false
 
-  let string_of = function
+  let is_nonlinear = function
+    | Unl -> true
+    | _   -> false
+
+  let to_string = function
     | Any -> "Any"
     | Unl -> "Unl"
 end
 
 (* Convenient aliases for constructing values *)
-let linUnl = Linearity.Unl
-let linAny = Linearity.Any
+let lin_any = Linearity.Any
+let lin_unl = Linearity.Unl
 
 module DeclaredLinearity = struct
   type t = Lin | Unl
     [@@deriving show]
 
-  let isLin lin = lin == Lin
-  let isUnl lin = lin == Unl
+  let is_linear = function
+    | Lin -> true
+    | _   -> false
+
+  let is_nonlinear = function
+    | Unl -> true
+    | _   -> false
 end
 
 (* Convenient aliases for constructing values *)
-let dlLin = DeclaredLinearity.Lin
-let dlUnl = DeclaredLinearity.Unl
+let dl_lin = DeclaredLinearity.Lin
+let dl_unl = DeclaredLinearity.Unl
 
 module Restriction = struct
   type t =
@@ -34,19 +44,23 @@ module Restriction = struct
     | Effect
     [@@deriving eq,show]
 
-  let isAny = function
+  let is_any = function
     | Any -> true
     | _   -> false
 
-  let isBase = function
+  let is_base = function
     | Base -> true
     | _    -> false
 
-  let isSession = function
+  let is_session = function
     | Session -> true
     | _       -> false
 
-  let string_of = function
+  let is_effect = function
+    | Effect -> true
+    | _      -> false
+
+  let to_string = function
     | Any     -> "Any"
     | Base    -> "Base"
     | Session -> "Session"
@@ -54,10 +68,10 @@ module Restriction = struct
 end
 
 (* Convenient aliases for constructing values *)
-let resAny     = Restriction.Any
-let resBase    = Restriction.Base
-let resSession = Restriction.Session
-let resEffect  = Restriction.Effect
+let res_any     = Restriction.Any
+let res_base    = Restriction.Base
+let res_session = Restriction.Session
+let res_effect  = Restriction.Effect
 
 module PrimaryKind = struct
   type t =
@@ -66,38 +80,38 @@ module PrimaryKind = struct
     | Presence
     [@@deriving show,eq]
 
-  let string_of = function
-    | Type -> "Type"
-    | Row -> "Row"
+  let to_string = function
+    | Type     -> "Type"
+    | Row      -> "Row"
     | Presence -> "Presence"
 end
 
 (* Convenient aliases for constructing values *)
-let pkType     = PrimaryKind.Type
-let pkRow      = PrimaryKind.Row
-let pkPresence = PrimaryKind.Presence
+let pk_type     = PrimaryKind.Type
+let pk_row      = PrimaryKind.Row
+let pk_presence = PrimaryKind.Presence
 
 module Location = struct
   type t = Client | Server | Native | Unknown
     [@@deriving show]
 
-  let isClient = function
+  let is_client = function
     | Client -> true
     | _      -> false
 
-  let isServer = function
+  let is_server = function
     | Server -> true
     | _      -> false
 
-  let isNative = function
+  let is_native = function
     | Native -> true
     | _      -> false
 
-  let isUnknown = function
+  let is_unknown = function
     | Unknown -> true
     | _      -> false
 
-  let string_of = function
+  let to_string = function
     | Client -> "client"
     | Server -> "server"
     | Native -> "native"
@@ -105,7 +119,7 @@ module Location = struct
 end
 
 (* Convenient aliases for constructing values *)
-let locClient  = Location.Client
-let locServer  = Location.Server
-let locNative  = Location.Native
-let locUnknown = Location.Unknown
+let loc_client  = Location.Client
+let loc_server  = Location.Server
+let loc_native  = Location.Native
+let loc_unknown = Location.Unknown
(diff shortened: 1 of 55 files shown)


frank-emrich:
> Here if `Any` isn't prefixed the compiler complains about type-directed disambiguation. An alternative solution is to put `Nil` branch first, because `Nil` does not appear in other datatypes (yet)

I think this is a good argument to settle the debate on type-directed disambiguation by just accepting that we use it and disable the corresponding warning.


>     I am leaning strongly towards prefixing all constructors in `cp_phrasenode` with `CP`.

This sounds like a reasonable thing to do and much better than creating mutually recursive modules.



> Is all that clutter acceptable? Perhaps a shorter module name (`Pat`?) would be better?

I would rather have some clutter than trying to be clever about the ordering of patterns. Once we allow type-directed disambiguation, the local `open` and occasional prefexing of constructors with module names should be fine IMHO.

jstolarek:
> I think this is a good argument to settle the debate on type-directed disambiguation by just accepting that we use it and disable the corresponding warning.

We can't disable the warning because they are right. Like I said, with type-directed disambiguation the compiler picks one of the datatypes in scope, and it usually picks the wrong one.

I pushed the refactoring further and redefined almost all sugartypes as ordinary variants. Few minor definitions remain and I am working on them. I'd say that the amount of extra clutter is reasonable and acceptable.

jamescheney:
This looks like progress to me.  How close does it get us to being able to reconsider using `visitors` and `alphalib` (#255)?

jstolarek:
Note to self and @rudihorn: `lens_opeartors.ml` says

```
let from_links v =
   match v with
   | `Minus -> Minus
   | `FloatMinus -> Minus
   | `Not -> Not
   | `Name name -> Name name
```

But ``` `Not``` is not introduced anywhere in Links source code, and thus the corresponding `Not` operator seems dead.

jstolarek:
@rudihorn, I got stuck on refactoring unary operators in Sugartypes due to the way the code interacts with lenses stuff. The problem is with `Lens_operators` module. Previously it pulled operator tags out of thin air, but now it needs to rely on definitions in `Sugartypes`, which creates a cycle between `Sugartypes`, `Types` and `Lens_operators`. `Sugartypes` has to rely on `Types`. Since `Types` now relies on lenses stuff it seems that all lenses definitions should go from `Types` to `Sugartypes`. This looks messy. You are more familiar with the structure of lens code - what are your thoughts? Would you be able to take a look at the code or perhaps sit down with me tomorrow before the Links meeting? The code is currently on `js-polymorphic-variants-unary-ops-stuck-on-lenses` branch - changed up to a point where the compilation trips on the lenses stuff.

jstolarek:
@rudihorn, you can ignore the above question now. I've found a good way to break the module cycle.

jstolarek:
This PR is ready for review. Things of interest:

  * new module `CommonTypes` contains simple types used across the compiler + some helper functions. **See whether you like what you see there because you'll be encountering and using it a lot.**
  * `Sugartypes.freedom` remains as a polymorphic variant due to it's interaction with `Types.Vars.flavour`.
  * the interesting changes are in `Sugartypes.ml`. Operators have been moved to `Operators.ml` due to interaction with lenses.
  * all the remaining changes are fallout.

If there are no objections I'd like to merge on Monday.

core/sugarTraversals.ml
```diff
@@ -559,73 +560,74 @@ class map =
...
-      | `Unit -> `Unit
-      | `Tuple _x ->
-          let _x = o#list (fun o -> o#datatype) _x in `Tuple _x
-      | `Record _x -> let _x = o#row _x in `Record _x
-      | `Variant _x -> let _x = o#row _x in `Variant _x
-      | `Effect r -> let r = o#row r in `Effect r
-      | `Table (_x, _x_i1, _x_i2) ->
+          let _x_i1 = o#datatype _x_i1 in Mu (_x, _x_i1)
+      | Forall (_x, _x_i1) ->
+          let _x = _x in
```
frank-emrich: any reason for keeping this?

core/commonTypes.ml
```diff
@@ -0,0 +1,111 @@
...
+
+  let string_of = function
+    | Any     -> "Any"
+    | Base    -> "Base"
+    | Session -> "Session"
+    | Effect  -> "Eff"
+end
+
+(* Convenient aliases for constructing values *)
+let resAny     = Restriction.Any
```
frank-emrich: you're essentially introducing aliases for the more generically named variants like `Any`, by introducing `linAny`, `resAny`. In my opinion, this just shows that those constructors should be renamed. Effectively, only the aliases you define here are used in the rest of the code, so we could just rename the underlying constructors. What do the others think?

jstolarek: > only the aliases you define here are used in the rest of the code

To be more precise: only aliases are used for constructing values. For deconstructing them (pattern matching) we still use constructors, for these types usually with explicit qualification. As for renaming the constructors to `ResAny`, `LinAny`, etc. - because I understand that's what you're proposing - I wonder what others think. Sam criticized this as poor man's namespaces, so I went with modules. Perhaps a middle ground here is shortening the names of modules to `Lin`, `DecLin`, `Res`, `PrKind`? We could then remove the aliases because spelling out the module name wouldn't be that painful.

Rant: I wish OCaml had pattern synonyms.

dhil: I quite like the prefix, it is informative. But I wouldn't be opposed to the good ol' single letter prefix on constructor names either.

dhil:
> If there are no objections I'd like to merge on Monday.

I am slightly concerned about rebasing the proper-modules branch on top of this. Although, I suppose I'll just leave that task to @frank-emrich...

core/commonTypes.ml
```diff
@@ -0,0 +1,111 @@
+module Linearity = struct
+  type t = Any | Unl
+    [@@deriving eq,show]
+
+  let isAny lin = lin == Any
```
dhil: Should be `is_any`.

core/commonTypes.ml
```diff
@@ -0,0 +1,111 @@
...
+  let isUnl lin = lin == Unl
```
dhil: `is_unl`

core/commonTypes.ml
```diff
@@ -0,0 +1,111 @@
...
+
+  let string_of = function
+    | Any -> "Any"
+    | Unl -> "Unl"
+end
+
+(* Convenient aliases for constructing values *)
+let linUnl = Linearity.Unl
```
dhil: `lin_unl`

core/commonTypes.ml
```diff
@@ -0,0 +1,111 @@
...
+let linAny = Linearity.Any
```
dhil: `lin_any`

core/commonTypes.ml
```diff
@@ -0,0 +1,111 @@
...
+
+module DeclaredLinearity = struct
+  type t = Lin | Unl
+    [@@deriving show]
+
+  let isLin lin = lin == Lin
```
dhil: `is_lin` (maybe even `is_linear`)

core/commonTypes.ml
```diff
@@ -0,0 +1,111 @@
...
+  let isUnl lin = lin == Unl
```
dhil: `is_unl`

core/commonTypes.ml
dhil: You are using reference equality `==`, I think you should use structural equality `=`. Even better use pattern matching, that shouldn't use the polymorphic equality operator, e.g.
```ocaml
let is_lin =  function Lin -> true | _ -> false
```

core/commonTypes.ml
dhil:
```ocaml
let is_unl = function Unl -> true | _ -> false
```

core/commonTypes.ml
```diff
@@ -0,0 +1,111 @@
...
+end
+
+(* Convenient aliases for constructing values *)
+let dlLin = DeclaredLinearity.Lin
```
dhil: `let declared_lin = ...`

core/commonTypes.ml
```diff
@@ -0,0 +1,111 @@
...
+let dlUnl = DeclaredLinearity.Unl
```
dhil: `let delcared_unl = ...`

core/commonTypes.ml
```diff
@@ -0,0 +1,111 @@
...
+
+module Restriction = struct
+  type t =
+    | Any
+    | Base
+    | Session
+    | Effect
+    [@@deriving eq,show]
+
+  let isAny = function
```
dhil: `is_any`

core/commonTypes.ml
```diff
@@ -0,0 +1,111 @@
...
+    | Any -> true
+    | _   -> false
+
+  let isBase = function
```
dhil: `is_base`

core/commonTypes.ml
```diff
@@ -0,0 +1,111 @@
...
+    | Base -> true
+    | _    -> false
+
+  let isSession = function
```
dhil: `is_session`

core/commonTypes.ml
```diff
@@ -0,0 +1,111 @@
...
+    | Session -> true
+    | _       -> false
+
+  let string_of = function
```
dhil: `to_string`

core/commonTypes.ml
dhil: `res_any`

core/commonTypes.ml
```diff
@@ -0,0 +1,111 @@
...
+    | Any     -> "Any"
+    | Base    -> "Base"
+    | Session -> "Session"
+    | Effect  -> "Eff"
+end
+
+(* Convenient aliases for constructing values *)
+let resAny     = Restriction.Any
+let resBase    = Restriction.Base
```
dhil: `res_base`

core/commonTypes.ml
```diff
@@ -0,0 +1,111 @@
...
+let resSession = Restriction.Session
```
dhil: `res_session`

core/commonTypes.ml
```diff
@@ -0,0 +1,111 @@
...
+let resEffect  = Restriction.Effect
```
dhil: `res_effect`

core/commonTypes.ml
```diff
@@ -0,0 +1,111 @@
...
+
+module PrimaryKind = struct
+  type t =
+    | Type
+    | Row
+    | Presence
+    [@@deriving show,eq]
+
+  let string_of = function
```
dhil: `to_string`

core/commonTypes.ml
```diff
@@ -0,0 +1,111 @@
...
+    | Type -> "Type"
+    | Row -> "Row"
+    | Presence -> "Presence"
+end
+
+(* Convenient aliases for constructing values *)
+let pkType     = PrimaryKind.Type
```
dhil: `pk_type`

core/commonTypes.ml
```diff
@@ -0,0 +1,111 @@
...
+let pkRow      = PrimaryKind.Row
```
dhil: `pk_row`

core/commonTypes.ml
```diff
@@ -0,0 +1,111 @@
...
+let pkPresence = PrimaryKind.Presence
```
dhil: `pk_presence`

core/commonTypes.ml
```diff
@@ -0,0 +1,111 @@
...
+
+module Location = struct
+  type t = Client | Server | Native | Unknown
+    [@@deriving show]
+
+  let isClient = function
```
dhil: `is_client`

core/commonTypes.ml
```diff
@@ -0,0 +1,111 @@
...
+    | Client -> true
+    | _      -> false
+
+  let isServer = function
```
dhil: `is_server`

core/commonTypes.ml
```diff
@@ -0,0 +1,111 @@
...
+    | Server -> true
+    | _      -> false
+
+  let isNative = function
```
dhil: `is_native`

core/commonTypes.ml
```diff
@@ -0,0 +1,111 @@
...
+    | Native -> true
+    | _      -> false
+
+  let isUnknown = function
```
dhil: `is_unknown`

core/commonTypes.ml
```diff
@@ -0,0 +1,111 @@
...
+    | Unknown -> true
+    | _      -> false
+
+  let string_of = function
```
dhil: `to_string`

core/commonTypes.ml
```diff
@@ -0,0 +1,111 @@
...
+    | Client -> "client"
+    | Server -> "server"
+    | Native -> "native"
+    | Unknown -> "unknown"
+end
+
+(* Convenient aliases for constructing values *)
+let locClient  = Location.Client
```
dhil: `loc_client`

core/commonTypes.ml
```diff
@@ -0,0 +1,111 @@
...
+let locServer  = Location.Server
```
dhil: `loc_server`

core/commonTypes.ml
```diff
@@ -0,0 +1,111 @@
...
+let locNative  = Location.Native
```
dhil: `loc_native`

core/commonTypes.ml
```diff
@@ -0,0 +1,111 @@
...
+let locUnknown = Location.Unknown
```
dhil: `loc_unknown`

jstolarek:
@dhil: thanks for feedback, all fixed now.

Issue #486, opened 2019-02-19
Shredding for SQLite


fehrenbach:
[SQLite has window functions now.](https://www.sqlite.org/releaselog/3_25_0.html) I haven't checked the [details](https://www.sqlite.org/windowfunctions.html), but this might be enough to support shredding.

jamescheney:
[MySQL 8.0 also appears to have window functions](http://www.mysqltutorial.org/mysql-window-functions/).

If these are both SQL:1999 compatible then it "should" just be a matter of making sure a sufficiently recent MySQL or SQLite is installed.  (Famous last words)

dhil:
Resolved by #492.

Issue #322, opened 2018-01-26
Add MySql and Sqlite drivers/packages to CI testing


jamescheney:
So that we can say that they are "supported" (and decrease the likelihood that they bit rot).

dhil:
I believe the current status is as follows

- Both PostgreSQL and SQLite3 get installed during CI
- Only PostgreSQL is tested during CI
- MySQL remains unsupported

jamescheney:
I think that's correct.

Testing SQLite 3 could just be a matter of adapting the existing tests to create/initialize and connect to a SQLite 3 database - theoretically both flat and shredded queries should work now thanks to #495.

dhil:
Reeled in from #505 
> It seems Ubuntu 16.04 (as provided by Azure Pipelines) only offers SQLite3 version 3.11, whilst we require at least version 3.25 (c.f. [release notes]()). So I suppose we need to find a third-party repository which offers a more recent version of SQLite3 before we can enable the shredding tests.

jamescheney:
SQLite 3.25 (which added support for `ROW_NUMBER` etc.) is only needed if we want to test shredding for SQLite3 on the CI.  I suggest that we don't wait around indefinitely and add SQLite3 and MySQL CI support but just run the plain database tests, not the shredding ones.  Whenever the default SQLite3 and/or MySQL versions available have support for the needed features, enabling the shredding tests for them should be a one-or-two-liner.

jamescheney:
It also looks like some work will be needed for the OCaml MySQL library to become compatible with Mysql 8...

https://github.com/ygrek/ocaml-mysql/issues/13

 so for now the ambition will just be to get the old mysql driver working again with Mysql 5.*, and adapt the CI to run / test the flat query examples with mysql and sqlite3.

jamescheney:
WIP in branch issue322 has reinstated the links-mysql package, with changes needed to make it compile.  However, when attempting to dynamically load the links-mysql library I get this error:
```
links> @load "factorials.links";
compiling to IR
compiled IR
debug: found loadable database driver
- /Users/jcheney/src/links/_build/install/default/lib/links-mysql/links_mysql.cmxs
- /Users/jcheney/src/links/_build/install/default/share/links-mysql/links_mysql_dependencies.json

***: Error: Cannot load plugin dependency '/Users/jcheney/.opam/4.07.0/lib/mysql/mysql.cmxs' (link error: no implementation available for Stdlib__stdLabels)
 
```
[...]

PR #492, opened 2019-02-21
SQLite shredding

fehrenbach:
Seems to pretty much just work.
It's not extensively tested, though.

Related: #486, #322


core/evalir.ml
```diff
@@ -678,8 +678,8 @@ struct
            | None -> computation env cont e
            | Some (db, p) ->
              begin
-               if db#driver_name() <> "postgresql"
-                 then raise (Errors.Runtime_error "Only PostgreSQL database driver supports shredding");
+               if not (List.mem (db#driver_name()) ["postgresql"; "sqlite3"])
```
jamescheney: I wonder if we should instead add a test to `db` objects that tells us whether the db supports shredding, and report "this database driver XYZ does not support shredding".

links-sqlite3.opam
```diff
@@ -1,21 +1,20 @@
...
-  [ "jbuilder" "build" "-p" name "-j" jobs ]
+  [ "dune" "subst" ] {pinned}
+  [ "dune" "build" "-p" name "-j" jobs ]
 ]
 
 depends: [
-  "jbuilder" {build}
+  "ocaml" {>= "4.06.0"}
+  "dune" {build}
   "sqlite3"
```
jamescheney: Would be nice to be able to check that the actual installed SQLite version is > 3.25.0, but not sure if OPAM gives us a way to say this.  If not, then row_number queries will just mysteriously fail.

fehrenbach: Ideally this would tie in with the runtime database driver test above. We don't need 3.25.0 for building, just for shredding.

fehrenbach:
I agree with all of @jamescheney's comments above, but I'm unlikely to do anything about it. I was just curious whether it worked.

links-sqlite3.opam
dhil: There is a virtual OPAM package `conf-sqlite3`, but it doesn't seem to track the version number of system package.

dhil:
Superseded by #495.

Issue #493, opened 2019-02-22
DOS style EOL in irCheck.ml/mli


jstolarek:
I just realized `irCheck.ml(i)` files are in DOS format. Conversion is trivial but from git perspective it amount to rewriting the whole file, so doing any merge on that file will be a disaster. Does anyone have any changes they plan to commit to this file in foreseeable future? If no then I would make the conversion along with my polymorphic variants patch.

SimonJF:
Good spot. Perhaps we could make a CI test to check for this.

jstolarek:
Yes, this is definitely something CI should check.

jstolarek:
Thanks! But first I need to know no one is doing anything with that file - otherwise it'll be a mess. @dhil? @frank-emrich?

dhil:
I suppose this is due to @frank-emrich using Windows. You probably want to get all files created by him.
```
$ git log --name-status --diff-filter=A --format='> %aN' | awk '/^>/ {tagline=$0}
       /^A\t/ {print tagline "\t" $0}' | grep Frank
> Frank Emrich	A	tests/functions/closure-conv-type-abstr-only.links
> Frank Emrich	A	bin/driver.ml
> Frank Emrich	A	bin/repl.ml
> Frank Emrich	A	install_local_opam2.sh
> Frank Emrich	A	core/irCheck.ml
> Frank Emrich	A	core/irCheck.mli
> Frank Emrich	A	tests/functions/nested-functions-polymorphic.links
> Frank Emrich	A	tests/validate_typecheck_testfile.testscript
> Frank Emrich	A	tests/typecheck_examples.tests
> Frank Emrich	A	tests/typecheck_examples.tests.config
> Frank Emrich	A	core/backend.ml
> Frank Emrich	A	core/irTraversals.ml
> Frank Emrich	A	core/irTraversals.mli
> Frank Emrich	A	core/dbDriverDynload.ml
> Frank Emrich	A	core/dbDriverDynload.mli
> Frank Emrich	A	database/mysql-driver/links_mysql_dependencies.json
> Frank Emrich	A	database/pg-driver/links_postgresql_dependencies.json
> Frank Emrich	A	database/sqlite3-driver/links_sqlite3_dependencies.json
```

fehrenbach:
It doesn't have to be CI. We can configure git itself to Do An Approximation Of The Right Thing™.
https://help.github.com/en/articles/dealing-with-line-endings#per-repository-settings

dhil:
[...]

PR #494, opened 2019-02-22
Add check for CRLF line endings

SimonJF:
As identified by @jstolarek in #493. Naturally this should fail the CI at the moment. We should merge after #493 is fixed by @jstolarek's pass through.


dhil:
I think @fehrenbach's solution (#493) is smarter.

SimonJF:
Ooh, yes, that does look nicer. Anyone would have thought that developers had run into this issue before ;)

OK, let's go with @fehrenbach's solution.

PR #495, opened 2019-02-22
Sqlite shredding

dhil:
Update of #492.


fehrenbach:
I think what James had in mind was just to add something like `supports_shredding : unit -> bool` to the `database` virtual class in `value.ml`. The implementations could be `const true`, or `const false`, or actually opening a connection and asking the database for its version, or some other feature detection.

jamescheney:
Yeah.  I hadn't thought that far ahead, just was thinking that the check whether enough of SQL:1999 is supported could be done once, when we initialize the database driver/connection.

core/evalir.ml
```diff
@@ -673,29 +673,34 @@ struct
...
-                 EvalNestedQuery.Shred.pmap EvalNestedQuery.Stitch.build_stitch_map raw_results in
-                 apply_cont cont env
-                 (EvalNestedQuery.Stitch.stitch_mapped_query mapped_results)
-             end
-         end
+         match EvalNestedQuery.compile_shredded env (range, e) with
+         | None -> computation env cont e
+         | Some (db, p) ->
+            try
+              let driver = db#driver_name () in
```
jamescheney: I'd like if we could push the knowledge of whether a database supports shredding into the db object, so that when/if we add other database drivers (or support shredding for mysql) we don't have to keep changing this line of code to have more hard-wired constants.

core/evalir.ml
```diff
@@ -673,29 +673,34 @@ struct
...
+                   in
+                   let execute_shredded_raw (q, t) =
+                     Database.execute_select_result (get_fields t) q db, t in
+                   let raw_results =
+                     EvalNestedQuery.Shred.pmap execute_shredded_raw p in
+                   let mapped_results =
+                     EvalNestedQuery.Shred.pmap EvalNestedQuery.Stitch.build_stitch_map raw_results in
+                   apply_cont cont env
+                     (EvalNestedQuery.Stitch.stitch_mapped_query mapped_results)
+              else raise Not_found (* Static exception. *)
```
jamescheney: I do not understand the rationale for this control flow.  Why isn't
```
try (if M then N else raise exn) with Exn -> P
```
equivalent to 
```
if M then N else P
```

??

dhil: Oh yes, it is. I experimented with something more general; I forgot to remove the `try`.

jamescheney: OK.  I don't understand how this commit (699929b) improves matters though.  Also, GitHub user "jcheney" is not me.

jamescheney: Oh wait, that commit is from 4 hours ago, it looks like you haven't actually changed anything since my review.

dhil:
I think this implements everything but the version check. I am not quite sure how you want to reliably detect the version number of the database system.

fehrenbach:
> I am not quite sure how you want to reliably detect the version number of the database system.

`select sqlite_version();` should do for SQLite. MySQL probably has something similar.

jamescheney:
I'm offline until Monday and not terribly worried about the version check (as long as the needed versions are documented somewhere), as long as the abstraction is there I'm happy. Will try to check over as soon as I can.

dhil:
Postgresql
```
> SELECT version();
                                                            version                                                            
-------------------------------------------------------------------------------------------------------------------------------
 PostgreSQL 10.6 (Ubuntu 10.6-0ubuntu0.18.04.1) on x86_64-pc-linux-gnu, compiled by gcc (Ubuntu 7.3.0-16ubuntu3) 7.3.0, 64-bit
(1 row)
```
SQLite3
```
sqlite> select sqlite_version ();
3.22.0
```
MySQL (Mariadb)
```
MariaDB [(none)]> SELECT version();
+----------------------------------+
| version()                        |
+----------------------------------+
| 10.1.38-MariaDB-0ubuntu0.18.04.1 |
+----------------------------------+
1 row in set (0.00 sec)
```

So it would be easy to "parse" the version number for SQLite3, and we might be able to parse the MySQL version number using a regexp like "^[0-9]+\.[0-9]+\.[0-9]+". I suppose there is no need to try to parse the PostgreSQL version number as the postgres driver "always" had support for shredding.

database/sqlite3-driver/lite3_database.ml
```diff
@@ -81,14 +82,39 @@ end
...
+       let stmt = prepare connection "SELECT sqlite_version()" in
+       let result = new lite3_result stmt in
+       match result#status with
+       | `QueryOk ->
+          let version = result#getvalue 0 0 in
+          let is_number str =
+            Str.string_match (Str.regexp "^[0-9]+$") str 0
+          in
+          begin match Str.split (Str.regexp "[0-9]\\.") version with
+          | major :: minor :: _ when is_number major && is_number minor ->
```
jamescheney: This looks fine, but given that different databases have their own idiosyncratic way of reporting versions, it may be a pain to maintain.  If so I'm also happy with just having a hardwired true/false constant per driver and we just document the needed versions.

dhil: Indeed, that is I have opted for the conservative option. If the version number is not recognisable, then the driver will report that it does not support shredding. I suppose, we can improve this in the future.

PR #496, opened 2019-02-25
Polymorphic variants - further cleanup

jstolarek:
Who's ready for a second round of removing polymorphic variants? :-)

I redefined `primitive` and `constant` as ordinary variants because that would help in what I'm currently doing. Changes follow the previous pattern of putting a type into its own module in `CommonTypes`.

I also believe that we could easily redefine IR datatype definitions in ir.ml to be ordinary variants, but since I'm not working with the IR at the moment I am not really motivated to do that now.


frank-emrich:
> I also believe that we could easily redefine IR datatype definitions in ir.ml to be ordinary variants, but since I'm not working with the IR at the moment I am not really motivated to do that now.

Do you mean changing the types in types.ml or creating a separate set of type definitions, exclusively used by the IR?

The types in types.ml are not just used by the IR, but also by the frontend typechecker. Changing them to normal variant types will be tricky since `meta_type_var_basis` and `meta_row_var_basis` are one of the few types where I know that we actually rely on them being polymorphic variants.

On the other hand, creating a separate set of type definitions representing only the IR types has been floating around as an idea for some time, but would probably require some serious refactoring and re-creating many of the corresponding helpers (e.g., type instantiation, projection, row manipulation, etc).

jstolarek:
@frank-emrich, I was thinking about definitions in `ir.ml`, not `types.ml`.

```OCaml
type value =
  [ `Constant of constant
  | `Variable of var
  | `Extend of (value name_map * value option)
  | `Project of (name * value)
  | `Erase of (name_set * value)
  | `Inject of (name * value * Types.datatype)

  | `TAbs of tyvar list * value
  | `TApp of value * tyarg list

  | `XmlNode of (name * value name_map * value list)
  | `ApplyPure of (value * value list)

  | `Closure of var * tyarg list * value

  | `Coerce of (value * Types.datatype)
  ]
and tail_computation =
  [ `Return of (value)
  | `Apply of (value * value list)
  (* | `ApplyClosure of (value * value list) *)

  | `Special of special

  | `Case of (value * (binder * computation) name_map * (binder * computation) option)
  | `If of (value * computation * computation)
  ]
and fun_def = binder * (tyvar list * binder list * computation) * binder option * location
and binding =
  [ `Let of binder * (tyvar list * tail_computation)
  | `Fun of fun_def
  | `Rec of fun_def list
  | `Alien of (binder * name * language)
  | `Module of (string * binding list option) ]
and special =
  [ `Wrong of Types.datatype
  | `Database of value
  | `Lens of value * Types.lens_sort
  | `LensDrop of value * string * string * value * Types.lens_sort
  | `LensSelect of value * Types.lens_phrase * Types.lens_sort
  | `LensJoin of value * value * string list * Types.lens_phrase * Types.lens_phrase * Types.lens_sort
  | `LensGet of value * Types.datatype
  | `LensPut of value * value * Types.datatype
  | `Table of (value * value * value * (Types.datatype * Types.datatype * Types.datatype))
  | `Query of (value * value) option * computation * Types.datatype
  | `Update of (binder * value) * computation option * computation
  | `Delete of (binder * value) * computation option
  | `CallCC of (value)
  | `Select of (name * value)
  | `Choice of (value * (binder * computation) name_map)
  | `Handle of handler
  | `DoOperation of (name * value list * Types.datatype) ]
and computation = binding list * tail_computation
and effect_case = binder * binder * computation
and handler = {
    ih_comp: computation;
    ih_cases: effect_case name_map;
    ih_return: binder * computation;
    ih_depth: handler_depth;
}
and handler_depth = [`Deep of (binder * value) list | `Shallow]
  [@@deriving show]
```

This looks like an easy change.

frank-emrich:
Ahh, sorry, I thought you meant the OCaml types representing the IR types. But you're right, the type definitions in ir.ml look like they can be transformed easily.

frank-emrich:
I've looked at doing the transformation for the types in ir.ml. It's pretty annoying to do, since multiple other places (queries, lenses, the JS translator, values) define their own datatypes with similar constructors. I volunteer to do the translation for the IR, you have suffered enough in the frontend ;)

We should just make sure not to step on each other's toes there. I'll start once this PR is through and you don't have any other outstanding variant refactorings.

jstolarek:
> It's pretty annoying to do, since multiple other places (queries, lenses, the JS translator, values) define their own datatypes with similar constructors.

Yes, that is a common thing across the whole Links source code. For most of the time I left these datatypes unchanged and left them for a future refactoring. But use your best judgement of course.

This PR is now merged, so you can start with IR refactoring. Unless you are willing to wait a day or two so that we can fix #493 first. I checked and it looks that two largest pieces of work being done now (module system, MVU framework) won't be affected by this.

PR #497, opened 2019-02-25
Properly escape carriage returns.

SimonJF:
Before, carriage returns weren't escaped (and the escaping code wasn't extensible). Now, there's a slightly nicer way of doing substitutions, and carriage returns are handled properly.


dhil:
What about using [`Str.quote`](https://caml.inria.fr/pub/docs/manual-ocaml/libref/Str.html) instead?

SimonJF:
Thanks! Good idea, but it doesn't quite work:
```
# Str.global_replace (Str.regexp "'") (String.escaped "'") "'Hello', World";;
- : string = "'Hello', World"
```

SimonJF:
That said, maybe `Str.escaped` on the whole string, and a substitution for quotes might be the best idea.

SimonJF:
Much nicer, thanks for the suggestion!

Issue #420, opened 2018-11-20
Better errors for lenses?


jstolarek:
In `lens/lensTypes.ml` we have:

```
let var_name (var : phrase) =
    match var.node with
    | `Var name -> name
    | _ -> failwith "Expected a `Var type"
```
@rudihorn, would it make sense to report source position in the error message? It is stored as part of a `phrase`.

(On a master branch this code is slightly different, but position information is also available.)

jamescheney:
Special case of #525

jamescheney:
We discussed this on Wednesday and I think we agreed it can be closed. @rudihorn ?

rudihorn:
Yes this can be closed.

PR #498, opened 2019-02-26
[WIP] SourceCode and Binder module refactor

rudihorn:
I've been wanting to address #420 and stumbled into the issue of `with_pos` being in `sugartypes.ml`. I have refactored it and put it into `sourceCode.ml`, and given it it's own module along with an interface file `sourceCode.mli` and documentation. This makes the handling of withpos more similar to the recent changes by @jstolarek for `Location.t` and `DeclaredLinearity.t`.

At this stage this is more just a request for feedback / comments and possibly help, as for many places I'm using the `Legacy` module which is there for backwards compatibility. I decided to initially try to hide the internal with_pos type, but there are many places where we pattern match and having the type exposed proves very helpful, so it might be worth allowing it to be exposed by default.

I also did some refactoring of `binder` to a similar `Binder` module.

This currently still has some lens things snuck in as well, so it is not ready to be merged yet.


dhil:
> I decided to initially try to hide the internal with_pos type, but there are many places where we pattern match and having the type exposed proves very helpful, so it might be worth allowing it to be exposed by default.

If you want to allow pattern matching, but disallow construction by hand, then you can use a [private type](https://caml.inria.fr/pub/docs/manual-ocaml/extn.html#sec237).

core/sugarConstructors.ml
```diff
@@ -129,7 +133,7 @@ module SugarConstructors (Position : Pos)
 
   (** Fieldspec *)
 
-  let present        = Datatype.Present (Sugartypes.with_dummy_pos Datatype.Unit)
+  let present        = Datatype.Present (With_pos.make Datatype.Unit)
```
jstolarek: Somehow I find the name `make` a bit uninformative in this case. Previously the intention to attach a dummy position was clear.

core/sourceCode.ml
```diff
@@ -10,91 +10,205 @@ let trim_initial_newline s =
...
       (start,
        self#extract_line start.Lexing.pos_lnum,
        self#extract_substring start finish)
+  end
+
+module Lexpos = struct
+  type t = Lexing.position
+
+  module LexposType = struct
+    type a = t
```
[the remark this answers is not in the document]
rudihorn: I honestly have no idea. This was legacy code I took over. I don't understand why it had these modules.

core/sourceCode.ml
```diff
@@ -10,91 +10,205 @@ let trim_initial_newline s =
...
+  let resolve_expression t = Resolved.resolve t |> Resolved.source_expression
+
+  let resolve_start_expr t =
+    let r = Resolved.resolve t in
+    Resolved.start r, Resolved.source_expression r
+
+  let pp fmt _ = Utility.format_omission fmt
+end
+
+module With_pos = struct
```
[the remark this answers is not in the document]
rudihorn: So this is a slightly weird one. Normally I prefer camel case, but in the case of Ocaml and unix, having a file name of `my_file.ml` results in a module name of `My_file` and I think this is why others follow this convention for naming conventions. I'm fine either way.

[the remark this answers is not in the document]
dhil: Yes, we use camel case for module names in the Links code base (at least for file modules).

core/sugarConstructors.ml
```diff
@@ -214,9 +218,15 @@ module SugarConstructors (Position : Pos)
     list ~ppos [record ~ppos exps]
 
   (* Is the list of labeled database expressions empty? *)
-  let is_empty_db_exps : phrase -> bool = function
-    | {node=ListLit ([{node=RecordLit ([], _);_}], _);_} -> true
-    | _                                                    -> false
+  let is_empty_db_exps : phrase -> bool =
```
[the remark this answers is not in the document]
rudihorn: Yes, this was before I decided to introduce the legacy module. I agree that the pattern matching would probably be nicer so I think using private types as mentioned by @dhil is probably the best way forward.

core/sugarConstructors.ml
rudihorn: Hmm suggestions for alternatives? Maybe just `of`?

core/sugartypes.ml
```diff
@@ -1,42 +1,29 @@
...
+  type t = (name * Types.datatype option) With_pos.t
+  [@@deriving show]
 
-(* A type alias to be used inside modules that define a node t and a with_pos
-   type, which is a node with attached position.  Alias is required due to
-   with_pos name inside the module overlapping with top-level with_pos type. *)
-module WithPos = struct
-  type 'a t = 'a with_pos [@@deriving show]
-end
+  let name b = let (n, _) = With_pos.node b in n
```
dhil: Should be `to_name`.

core/sugartypes.ml
```diff
@@ -1,42 +1,29 @@
...
+  let typ b = let (_, ty) = With_pos.node b in ty
```
dhil: Should arguably be `to_type`.

rudihorn:
It would probably also be good to move some of the formatting and pretty printers into the `SourceCode` module as well.

core/sourceCode.ml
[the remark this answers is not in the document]
rudihorn: Okay should I change this then to `WithPos`?

core/types.ml
```diff
@@ -136,6 +136,7 @@ type lens_phrase =
   | In        of Lens_operators.name list * (Constant.constant list) list
   | Case      of lens_phrase option * (lens_phrase * lens_phrase) list * lens_phrase
   | TupleLit  of lens_phrase list
+and lens_phrase = lens_phrasenode SourceCode.With_pos.t
```
dhil: I am not a fan of this change. Now our type algebra depends on surface syntax matters.

core/sugarConstructors.ml
jstolarek: How about `WithPos.dummy`?

rudihorn: Currently the `make` function takes an optional argument `pos` for the position. So it isn't just a function for dummy positions. Or would you prefer this not to be the case?

core/sourceCode.ml
```diff
@@ -10,91 +10,205 @@ let trim_initial_newline s =
   else s
 
 class source_code =
-object (self)
-  val lines =
-    let tbl = Hashtbl.create default_lines in
+  object (self)
```
dhil: I think you should undo this change as it only introduces indentation. I don't think we have a coding convention for indentation at the moment, but it would be a silly if future commits went back and forth about the indentation level.

rudihorn: Yeah I can undo this, this was accidentally/automatically done by merlin. I think some day it would possibly be nice to find a coding convention for this and talk about it at the next links meeting?

core/sugarConstructors.ml
jstolarek: > So it isn't just a function for dummy positions. 

I know - that's why I said that the intention to attach a dummy position is not as clear as it was before. When I was introducing `with_pos` data type I had the same dilemma, i.e. whether to have a separate function for dummy positions or just use an optional argument. I went with the former because the burden of using named arguments seemed a bit to heavy syntactically. But that's just one person's opinion.

jstolarek: So yes, I would prefer to have a separate function to attach a dummy position.

core/sourceCode.ml
jstolarek: If anything, I think it is worth to separate large formatting changes into separate commits.

And personally I don't mind this extra indentation. I think this is how Merlin's indenter would format it, right?

dhil: I agree, it would be good to settle on a convention, but as we don't have one at the moment (at least I am not aware of one) I think we should avoid introducing such "trivial" changes.

core/sugartypes.ml
rudihorn: I personally feel like the `to_` convention should be reserved for type directed transformations, not necessarily simple projections like this. The `to_name` just somehow feels misleading here.

dhil: You are transforming a term of type `t` into a term of type `name`. I think it is irrelevant whether the transformation is trivial.

core/sourceCode.ml
rudihorn: @jstolarek yes this is merlin. I'm happy for this to be farmed out to a seperate commit, again this was more accidental than on purpose.

jstolarek:
Incidentally, I just realized that the thing I am working on requires me to perform the same kind of refactoring, i.e. taking `with_pos` and `binder` out of Sugartypes so that they can be used independently. So the priority of this PR went from "nice to have" to "must have".

rudihorn:
> Incidentally, I just realized that the thing I am working on requires me to perform the same kind of refactoring, i.e. taking with_pos and binder out of Sugartypes so that they can be used independently. So the priority of this PR went from "nice to have" to "must have".

Feel free to do any poking around you would like to help get it done quicker. I'll try to make requested changes soonish, but also have a little marking to do atm.

jstolarek:
@rudihorn, is there a way you could push your branch into the main Links repo and then change this PR to be based on that branch, rather than the one in your private repo? Perhaps then I could push some changes to it. But I guess that what I'm going to do for now is to simply rebase my work on top of your patch and continue my work from there.

jstolarek:
Ok, I pushed the `sourcecode_refactor` into the main repo but I don't see a way of changing the PR to a different branch.

rudihorn:
Huh, I think I clicked the checkbox allow edits from maintainers, which should give you access to my private repo?

jstolarek:
Ok, that works. Thanks.

dhil:
> @rudihorn, is there a way you could push your branch into the main Links repo and then change this PR to be based on that branch, rather than the one in your private repo?

You can check out a pull request locally, see https://help.github.com/en/articles/checking-out-pull-requests-locally. You can make changes locally, and then push them to the main repository, and open a new PR based on that branch.

jstolarek:
@dhil, but this would mean opening a new PR and I didn't want to lose any of the discussions we had here.

@rudihorn, I made a few simple changes (that'll mostly save me from conflicts during rebase) and merged the changes from master branch. I'll leave the rest to you.

dhil:
> @dhil, but this would mean opening a new PR and I didn't want to lose any of the discussions we had here.

The discussions wouldn't be lost; there would just be another layer of indirection :-)

jstolarek:
Yes, that's what I didn't want :-)

jstolarek:
I removed the `Legacy` module and created a separate function for attaching dummy positions. Now the question is whether we want to make `WithPos.t` private. Doing so will force us to say `WithPos.make ~pos node` instead of `{node;pos}` (or: `WithPos.make ~pos foo` instead of `{node=foo; pos}`. Pattern matching will work the way it used to, so we can now restore some of the code that was changed to explicitly use projections.

core/sugarConstructors.ml
```diff
@@ -129,7 +133,7 @@ module SugarConstructors (Position : Pos)
 
   (** Fieldspec *)
 
-  let present        = Datatype.Present (Sugartypes.with_dummy_pos Datatype.Unit)
+  let present        = Datatype.Present (WithPos.make Datatype.Unit)
```
jstolarek: Should change this to `WithPos.dummy`.

core/sugartoir.ml
```diff
@@ -909,16 +911,16 @@ struct
               in
                 I.switch env (ev e, cases, t)
           | DatabaseLit (name, (None, _)) ->
-              I.database (ev (with_pos pos (RecordLit ([("name", name)],
-                                          Some (with_pos pos (FnAppl (with_pos pos (Var "getDatabaseConfig"), [])))))))
+              I.database (ev (WithPos.make ~pos (RecordLit ([("name", name)],
```
jstolarek: I have experimentally replaced `with_pos` with `WithPos.make`, which is slightly more verbose. Does that look sensible? Or do we want to keep a local `with_pos` alias? Also, `WithPos.` prefix is not really mandatory here since the module is opened at the top level, but having only `make ~pos ...` looks a bit unintuitive.

rudihorn: I'm happy with this change. I think having the prefix is nice, as the only real reason we open the `WithPos` namespace is to pattern match the types.

rudihorn:
I don't recall the constructor being directly used that often, so I would say lets change it to private. I can try working on this if you would like @jstolarek.

jstolarek:
> I don't recall the constructor being directly used that often

Indeed. And most places that do probably should use a constructor instead.

>  I can try working on this if you would like @jstolarek.

Yes, please. I won't be touching that code today (and possibly tomorrow neither), but I'll try to provide feedback tomorrow if anything gets done.

rudihorn:
@jstolarek This should be sorted. Anything else that needs to be done on this PR before it can be merged? I think I may leave the small lens changes fragments for now, even though they should have been in a separate PR.

core/types.ml
frank-emrich: I agree that something in types.ml knowing about source code positions seems unclean in terms of separating the different stages of compilation

jstolarek:
I am doing a hopefully final pass through this patch. I'm not sure what to do about all the lens changes. Putting everything into one commit feels wrong, but I'm not sure whether I'll be able to separate binder/with_pos refactoring from lens changes. I will try though.

core/sugartypes.ml
```diff
@@ -1,42 +1,56 @@
...
+open SourceCode
 open Utility
 
 (** The syntax tree created by the parser. *)
 
 type name = string [@@deriving show]
 
-type position = SourceCode.pos
-let dummy_position = SourceCode.dummy_pos
+module Binder : sig
```
jstolarek: I really don't like having a module signature in `*.ml` file. Feels redundant. I understand this is for the sake of `traverse_map`, which has a non-trivial type. I changed this by removing the module signature and then just putting a type annotation on `traverse_map` only. Yell if you have a strong opinion on this.

core/errors.ml
```diff
@@ -1,59 +1,63 @@
...
-      let (pos,_,expr) = SourceCode.resolve_pos pos in
+      let pos, expr = Position.resolve_start_expr pos in
         Printf.sprintf "%s:%d: Redundant pattern match:\nIn expression: %s\n"
           pos.pos_fname pos.pos_lnum expr
   | Sugartypes.PatternDuplicateNameError(pos, name) ->
       (* BUG: this can't be right (the pattern and the expression are the same!) *)
-      let _,_,pattern = SourceCode.resolve_pos pos in
-      let (pos,_,expr) = SourceCode.resolve_pos pos in
+      let pos, expr = Position.resolve_start_expr pos in
+      let pattern = expr in
```
jstolarek: What the hell is going on here!?

PR #502, opened 2019-02-26
Remove polymorphic variants from IR, pattern matching compiler, and queries

frank-emrich:
exterminate, exterminate!!!
(to be read in Dalek voice)


core/ir.ml
```diff
@@ -39,103 +39,101 @@ type location = CommonTypes.Location.t
...
+  | Case of (value * (binder * computation) name_map * (binder * computation) option)
+  | If of (value * computation * computation)
+and fun_def = (binder * (tyvar list * binder list * computation) * binder option * location)
 and binding =
-  [ `Let of binder * (tyvar list * tail_computation)
-  | `Fun of fun_def
-  | `Rec of fun_def list
-  | `Alien of (binder * name * language)
-  | `Module of (string * binding list option) ]
+  | Let of (binder * (tyvar list * tail_computation))
```
dhil: I think we should get rid of the needless boxing before merging this, e.g. drop the outermost parentheses (with regular variants they have performance implications).

dhil: Or if you solemnly swear to remove the parentheses in a follow-up PR, then I will not object to merging this now.

frank-emrich:
I intended to do that in a separate PR, as the current state reflects how things were before. But I can also make the necessary changes part of this PR


________________________________

core/ir.ml
frank-emrich: done!

jstolarek: Why are the outermost parentheses a problem? I'm asking cause I introduced a fair amount of this in my refactoring. The reason was that with polymorphic variants I had something like:

```OCaml
type foo = [ `A of int * int ]

match f with 
  | `A a -> ()
```
After converting `foo` to an ordinary variant the pattern match `A a` was no longer valid, with compiler reporting an error that `A` expects two arguments but was only give one. My solution was to introduce parentheses around the arguments in the datatype definition:

```OCaml
type foo = A of (int * int)
```

dhil: > Why are the outermost parentheses a problem?

They introduce an extra box. For regular variants the code produced by following snippets differ
```
type t = Foo of int * int | Bar
let _ = Foo (2,3)
```
and
```ocaml
type t = Foo of (int * int) | Bar
let _ = Foo (2,3)
```
The former is represented `[0: 2 3]` at runtime, whilst the latter is represented as `[0: [0: 2 3]]`, that is there is an extra box. For polymorphic variants this doesn't matter because the constructor arguments are always boxed.

jstolarek: What if I said `type t = Foo of ((((((int * int)))))) | Bar`? Would it mean six additional layers of indirection?

dhil: > type t = Foo of ((((((int * int)))))) | Bar

No, only the first pair of parentheses have a semantic effect in this example.

dhil: However, 
```ocaml
type t = Foo of ((((((int * int)))) * bool)) | Bar
let _ = Foo ((2,3), true)
```
would be represented as `[0: [0: 2 3] 1a]` at runtime.

jstolarek: Good to know. I was about to ask about all of this on the last dev meeting but we ran out of time.

I'll remove these extra parentheses in Sugartypes in near future.

frank-emrich:
I have now extended this PR to also remove polymorphic variants from all query related files, yielding changes in:
query.ml, evalQuery.ml, evalNestedQuery.ml, and sql.ml

frank-emrich:
There was a pre-existing bug where the .mli files in the query subdirectory were ignored by dune and thus got out of sync with the existing implementations. I've updated the files (most notably query.mli) and re-enabled them in dune.

core/buildTables.ml
```diff
@@ -52,7 +53,7 @@ struct
         let rec bind o =
           function
             | [] -> o
-            | (`Let (x, _))::bs ->
+            | (Ir.Let (x, _))::bs ->
```
jstolarek: That constructor is prefixed with `Ir.` but other in this file are not. Accidental or on purpose?

core/closures.ml
```diff
@@ -373,11 +374,11 @@ struct
   let binder (x, (t, name, _)) = (x, (t, name, `Global))
   let fun_def (f, lam, z, location) = (binder f, lam, z, location)
   let binding = function
-    | `Let (x, body) -> `Let (binder x, body)
-    | `Fun def -> `Fun (fun_def def)
-    | `Rec defs -> `Rec (List.map fun_def defs)
-    | `Alien (x, n, language) -> `Alien (binder x, n, language)
-    | `Module _ -> failwith "unimplemented"
+    | Ir.Let (x, body) -> Ir.Let (binder x, body)
```
jstolarek: Are there name clashes only on some constructors? I tried to avoid prefixing only some of the constructors, i.e. I would either prefix all or none. Will adding a type signature on `binding` allow to avoid quantifiers (probably not)? Also, I only see `Ir` being open in this module, so what do the constructor names clash with? Are there local opens?

core/closures.ml
```diff
@@ -387,7 +388,7 @@ module ClosureConvert =
 struct
 
   let close f zs tyargs =
-    `Closure (f, tyargs, `Extend (List.fold_right
+    Ir.Closure (f, tyargs, Ir.Extend (List.fold_right
```
jstolarek: Same questions here.

core/compilePatterns.ml
```diff
@@ -13,21 +13,6 @@ open Utility
...
-| `Negative of StringSet.t
-| `Record   of (pattern StringMap.t) * pattern option
-| `Constant of Constant.t
-| `Variable of binder
-| `As       of binder * pattern
-| `HasType  of pattern * Types.datatype
-]
-    [@@deriving show]
-
 module Const = struct
```
jstolarek: Perhaps instead of having another module for representing constants we could extend existing module `CommonTypes.Constant` with `compare` function and use it instead?

core/compilePatterns.ml
```diff
@@ -218,40 +236,40 @@ open CompileEq
 let show_pattern_compilation = Basicsettings.CompilePatterns.show_pattern_compilation
 
 type annotation = [`Binder of binder | `Type of Types.datatype] list
```
jstolarek: Maybe untick this datatype?

core/evalir.ml
```diff
@@ -138,20 +138,20 @@ struct
 
   (** {0 Evaluation} *)
   let rec value env : Ir.value -> Value.t = function
-    | `Constant (Constant.Bool   b) -> `Bool b
-    | `Constant (Constant.Int    n) -> `Int n
-    | `Constant (Constant.Char   c) -> `Char c
-    | `Constant (Constant.String s) -> Value.box_string s
-    | `Constant (Constant.Float  f) -> `Float f
-    | `Variable var -> lookup_var var env
+    | Constant (Constant.Bool   b) -> `Bool b
```
jstolarek: `Value.t` is another candidate for conversion to a normal variant. But not in this patch, I think.

core/ir.ml
```diff
@@ -99,43 +97,43 @@ and handler = {
...
     | [] -> v
-    | _ -> `TApp (v, tyargs)
+    | _ -> TApp (v, tyargs)
 
-let letm (b, tc) = `Let (b, ([], tc))
-let letmv (b, v) = letm (b, `Return v)
-(*let letv (b, v) = `Let (b, `Return v)*)
+let letm (b, tc) = Let (b, ([], tc))
+let letmv (b, v) = letm (b, Return v)
+(*let letv (b, v) = Let (b, Return v)*)
```
jstolarek: Maybe just delete if unused?

core/buildTables.ml
frank-emrich: good catch!

core/closures.ml
frank-emrich: Agreed, this should be more consistent. I usually tried not to open entire modules globally inside of modules, but either qualify all constructors or open the necessary modules locally, similar to how you did it in the frontend. This spurious qualification is a result of me experimenting around, removed it.

core/closures.ml
frank-emrich: removed this

core/compilePatterns.ml
frank-emrich: yepp, overlooked this one, done now

core/ir.ml
frank-emrich: Done

core/evalir.ml
frank-emrich: Indeed! However, it is actually used as a polymorphic variant in value.ml and will require some additional work in order to become a regular datatype.

core/compilePatterns.ml
frank-emrich: Okay, I've given CommonTypes.Constant a derived compare function. This should actually be faster than the polymorphic Pervasives.compare function.

jstolarek:
I have this in my review queue, but I'm afraid it'll have to wait until tomorrow - still working on #498.

frank-emrich:
No worries, this PR isn't urgent

core/evalir.ml
```diff
@@ -549,26 +549,26 @@ struct
...
+          | Alien _ ->
             computation env cont (bs, tailcomp)
-          | `Module _ -> failwith "Not implemented interpretation of modules yet"
+          | Module _ -> failwith "Not implemented interpretation of modules yet"
   and tail_computation env (cont : continuation) : Ir.tail_computation -> result = function
-    | `Return v      -> apply_cont cont env (value env v)
-    | `Apply (f, ps) -> apply cont env (value env f, List.map (value env) ps)
-    | `Special s     -> special env cont s
-    | `Case (v, cases, default) ->
+    | Ir.Return v      -> apply_cont cont env (value env v)
```
jstolarek: Redundant or required? Also, alignment of `->` is off.

core/ir.ml
```diff
@@ -39,58 +39,56 @@ type location = CommonTypes.Location.t
   [@@deriving show]
 
 type value =
-  [ `Constant of Constant.t
-  | `Variable of var
-  | `Extend of (value name_map * value option)
-  | `Project of (name * value)
-  | `Erase of (name_set * value)
-  | `Inject of (name * value * Types.datatype)
+  | Constant of Constant.t
```
jstolarek: I'd align `of` in one column so that the declaration is visually more pleasant to read. Hint: in Emacs you can use `M-x align-regexp` on a block of text.

core/ir.mli
```diff
@@ -40,57 +40,56 @@ type location = CommonTypes.Location.t
 (* INVARIANT: all IR binders have unique names *)
 
 type value =
-  [ `Constant of Constant.t                   (* constant: c *)
-  | `Variable of var                          (* variable use: x *)
-  | `Extend of value name_map * value option  (* record extension: (l1=v1, ..., lk=vk|r) or (l1=v1, ..., lk=vk) *)
-  | `Project of name * value                  (* record projection: r.l *)
-  | `Erase of name_set * value                (* erase fields from a record: r\{ls} *)
-  | `Inject of name * value * Types.datatype  (* variant injection: L(v) *)
+  | Constant of Constant.t                      (* constant: c *)
```
jstolarek: Same here

core/irtojs.ml
```diff
@@ -108,7 +109,7 @@ let __kappa = "__kappa"
   _plus, _minus, etc.
     curried function versions of the standard arithmetic operators
   LINKS.XML(tag, attrs, children)
-    create a DOM node with name `tag'
+    create a DOM node with name tag'
```
jstolarek: I think this was meant to be a quote :-)

core/query/evalQuery.ml
```diff
@@ -50,9 +51,14 @@ struct
      duplicated and ensuring that we order by all occurences of them -
      e.g. by converting tail generators to non-tail generators. *)
 
-  type order_index = [ `Val of Q.t | `Gen of gen | `TailGen of gen
-                     | `DefVal of Primitive.t | `DefGen of gen | `DefTailGen of gen
-                     | `Branch of int ]
+  type order_index =
```
jstolarek: Align `of`

core/query/query.ml
```diff
@@ -1,56 +1,71 @@
...
-let _show_env = show_env (* Generated by ppx_deriving show *)
+module Lang =
+struct
+
+  type base_type = | Bool | Char | Float | Int | String
+
+  type tag = int
+      [@@deriving show]
+
+  type t =
```
jstolarek: Align `of`.

core/query/query.ml
```diff
@@ -85,122 +100,132 @@ let used_database v : Value.database option =
...
     match v with
-      | `Concat vs -> comprehensions vs
+      | Q.Concat vs -> comprehensions vs
       | v -> used v
 
 module S =
 struct
+
   (** [pt]: A printable version of [t] *)
   type pt =
```
jstolarek: Align `of`.

core/query/query.mli
```diff
@@ -1,74 +1,70 @@
...
-    | `Primitive of string
-    | `Var of (Var.var * Types.datatype StringMap.t) | `Constant of Constant.constant ]
-and env = Value.env * t Env.Int.t
-  [@@deriving show]
+  type base_type = | Bool | Char | Float | Int | String
+
+  type tag = int
+      [@@deriving show]
+
+  type t =
```
jstolarek: Align `of`.

core/query/sql.ml
```diff
@@ -2,17 +2,17 @@ open Utility
 open CommonTypes
 
 type query =
```
jstolarek: Align `of`.

core/compilePatterns.ml
```diff
@@ -13,42 +13,59 @@ open Utility
...
-| `Constant of Constant.t
-| `Variable of binder
-| `As       of binder * pattern
-| `HasType  of pattern * Types.datatype
-]
+module type CONSTSET = Set with type elt = Constant.t
+module ConstSet = Set.Make(Constant)
+module ConstMap = Map.Make(Constant)
+
+module Pattern =
```
jstolarek: I tend to write:

```
module Patter = struct
  type t = ...
```
and I think this is formatting used in many places in our code.

PR #503, opened 2019-02-27
changed gitattributes file for crlf

rudihorn:
For some reason my git was suddenly showing me that the `tests/dos-newlinks.links` file was modified because of CRLF settings. Seemingly also setting the `text=auto` flag rather than just `text` helps.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -3,4 +3,4 @@
 
 # An important exception is a file for testing Links' support for DOS line
 # endings
-tests/dos-newlines.links text eol=crlf
+tests/dos-newlines.links text=auto eol=CRLF
--- a/.gitattributes
+++ b/.gitattributes
@@ -3,4 +3,4 @@
 
 # An important exception is a file for testing Links' support for DOS line
 # endings
-tests/dos-newlines.links text=auto eol=CRLF
+tests/dos-newlines.links text=auto eol=crlf


jstolarek:
Strange. What system and which version of git? I'm on Debian with git 2.11. I tested this locally before committing and now I'm working on another machine and everything works as expected.

rudihorn:
It mostly worked as expected, but when I did a git reset and other fiddling it broke and just wouldn't fix itself (git stash, git checkout <file>) did not work. Changing `.gitattributes` then suddenly fixed it, even in branches where I did not have the `.gitattributes` file. This seems to be just more stable all-round.

jstolarek:
Please don't merge yet. There's one thing I want to test but don't the time to do it at the moment.

rudihorn:
Okay, I'll leave you to merge it when you are ready and if others are okay with this.

jstolarek:
I merged the fix but there is something I don't understand. Firstly, I can confirm that I was able to reproduce the problem by cloning a remote repo. I don't know why my tests with cloning a local repo didn't catch it yesterday - sorry about that. Secondly, a similar problem was revealed for *.pdf and *zip files in the repo. These are now marked as binary in `.gitattributes`.

Now, the part I don't understand is why saying `text=auto eol=crlf` works. Manual only mentions `text=auto` or `text eol=crlf` and does not say anything about possibility of combining the two. My understanding was that `eol=...` overrides the default behaviour of `auto`.

Furthermore, it seems possible to convert `tests/dos-newlines.links` to Unix line endings and commit it to the repo, but the file will be converted to DOS format on clone. This is a bit weird, since the working tree will differ from what's actually in the repo, and git emits a warning when committing a converted file:

> warning: LF will be replaced by CRLF in tests/dos-newlines.links.
> The file will have its original line endings in your working directory.

Anyway, the fix seems to work on a Linux machine. Can someone test on Windows? @frank-emrich, could you try cloning a repo and seeing whether the working tree is clean?

Issue #501, opened 2019-02-26
Check code style using OCaml format


rudihorn:
Following the whitespace / indentation comments in #498, I also just realised that we could extend this to formatting as well.  We could extend @dhil scripts to check the source code by using ocamlformat (https://github.com/janestreet/ocamlformat) to run on a source code and ensuring it makes no changes to that file during CI tests.

Can discuss at the next links meeting, I'm happy to help sort this out.

dhil:
I think this has been on @frank-emrich's road map for quite some time. I recall discussing it last year. I am generally in favour of using some **automatic** code formatting tool.

jamescheney:
We can close this now!

PR #504, opened 2019-02-27
[Proposal] Ocamlformat

rudihorn:
In reference to #501 I have made a modification to the whitespace script to include a script for executing `ocamlformat` on all the links files. I don't believe the tool works for `*.mly` and `*.mll` files yet, but it is still useful. To see how it formats the tools, install ocamlformat using `opam install ocamlformat` and then run `./tools/rules/ocamlformat_check fix`.

Many files need their comments fixed, because they inadvertently include documentation comments, i.e. `(** some comment *)` not properly attached to a node. These would need to be manually fixed.


rudihorn:
For reference:

@dhil Asked if there is an emacs plugin. There is an emacs plugin which can be installed by following the instructions: https://github.com/janestreet/ocamlformat#installation

jstolarek:
While I fully support the idea of having some sort of code formatting conventions I don't like the idea of handing this to an automated tool. I think human can do a better job of code formatting than a tool, whereas tools can do things that don't always improve readability.

That being said, I think the biggest problems with code readability and formatting come from poor quality of the code itself. Here's a typical excerpt from `typeSugar.ml`:

```OCaml
                        begin
                          match Instantiate.apply_type t tyargs with
                            | `Function (fps, fe, rettyp) ->
                                let rettyp = Types.for_all (rqs, rettyp) in
                                let ft = `Function (fps, fe, rettyp) in
                                let f' = erase f in
                                let e = tabstr (rqs, FnAppl (with_dummy_pos (tappl (f'.node, tyargs)), List.map erase ps)) in
                                  unify ~handle:Gripers.fun_apply
                                    ((exp_pos f, ft), no_pos (`Function (Types.make_tuple_type (List.map typ ps),
                                                                         context.effect_row,
                                                                         rettyp)));
                                  e, rettyp, merge_usages (usages f :: List.map usages ps)
                            | `Lolli (fps, fe, rettyp) ->
                                let rettyp = Types.for_all (rqs, rettyp) in
                                let ft = `Function (fps, fe, rettyp) in
                                let f' = erase f in
                                let e = tabstr (rqs, FnAppl (with_dummy_pos (tappl (f'.node, tyargs)), List.map erase ps)) in
                                  unify ~handle:Gripers.fun_apply
                                    ((exp_pos f, ft), no_pos (`Lolli (Types.make_tuple_type (List.map typ ps),
                                                                      context.effect_row,
                                                                      rettyp)));
                                  e, rettyp, merge_usages (usages f :: List.map usages ps)
                            | _ ->
                                assert false
                        end
```

The main body of the code is indented to start at column 30 and ends at column 125. Both branches are identical and differ only in `Function`/`Lolli` constructor. These two branches share a common piece of code that should be parametrized and extracted to top-level function. It will then have a small indentation and will easily fit 80 columns.

All of `typeSugar.ml` (and a few other files) is an example of this kind of repetitious code. Even if functions are declared, they are often declared locally in a branch and same declarations can be repeated in different branches of a `match`, again resulting in code duplication. These kind of problems need to be fixed by refactoring the code and I feel this would be something that would have a major impact on readability of Links source code.

slindley:
I quite agree with @jstolarek. It might be useful to use a tool to check whether code complies with a particular formatting convention on occasion, but blindly applying it to existing source code is doomed to failure.

The most important coding convention we have always aimed to follow (and not always succeeded in) is to try to use internally consistent formatting within each file.

jstolarek:
BTW does OCaml have an equivalent of Java's [checkstyle](http://checkstyle.sourceforge.net/)? (Fantastic tool!)

rudihorn:
While automatic code formatting isn't always perfect, my experience with `ocamlformat` is that it offers a minimum of readability that probably is on average better than the worst of the files in links. I doubt that ocamlformat will produce anything "doomed for failure" and if that did somehow happen to be the case, it is possible to configure and even disable ocamlformat in the file itself (https://github.com/ocaml-ppx/ocamlformat#options).

I agree with @jstolarek that part of the main problem is simply poorly written code, but even that particular snippet becomes slightly more readable with ocamlformat, without having to manually refactor the entire codebase by hand. And when things do get refactored, it will automatically join things in single lines that do fit in the 80 column limit.

```
            match Instantiate.apply_type t tyargs with
            | `Function (fps, fe, rettyp) ->
                let rettyp = Types.for_all (rqs, rettyp) in
                let ft = `Function (fps, fe, rettyp) in
                let f' = erase f in
                let e =
                  tabstr
                    ( rqs
                    , FnAppl
                        ( with_dummy_pos (tappl (f'.node, tyargs))
                        , List.map erase ps ) )
                in
                unify ~handle:Gripers.fun_apply
                  ( (exp_pos f, ft)
                  , no_pos
                      (`Function
                        ( Types.make_tuple_type (List.map typ ps)
                        , context.effect_row
                        , rettyp )) ) ;
                (e, rettyp, merge_usages (usages f :: List.map usages ps))
            | `Lolli (fps, fe, rettyp) ->
                let rettyp = Types.for_all (rqs, rettyp) in
                let ft = `Function (fps, fe, rettyp) in
                let f' = erase f in
                let e =
                  tabstr
                    ( rqs
                    , FnAppl
                        ( with_dummy_pos (tappl (f'.node, tyargs))
                        , List.map erase ps ) )
                in
                unify ~handle:Gripers.fun_apply
                  ( (exp_pos f, ft)
                  , no_pos
                      (`Lolli
                        ( Types.make_tuple_type (List.map typ ps)
                        , context.effect_row
                        , rettyp )) ) ;
                (e, rettyp, merge_usages (usages f :: List.map usages ps))
            | _ -> assert false )
```

jamescheney:
Instead of automatically reformatting, is it possible to set things up so that if something is not done the way `ocamlformat` would, you get a warning/report?  (I guess this is what an OCaml `checkstyle` would do, if it existed.)  Something like

```
for each file f1
  ocamlformat f > f2
  diff f1 f2
```
?

that one can run on individual files or on everything?

rudihorn:
This is how it is currently done. Travis checks if `ocamlformat` outputs the same file and otherwise does not pass CI.

jamescheney:
OK.  My suggestion was to provide the check but not require everything to pass it in order to commit.  I don't think there's a consensus in favor of forcing everything to match `ocamlformat`.  Perhaps we can maintain a whitelist of files that are checked in this way, and people can opt-in for files they maintain without requiring everyone to do it all at once?

jstolarek:
I think this PR is becoming something I would like to discuss at a meeting. I have several questions and concerns that will probably be much faster to formulate and answer without having to write.

frank-emrich:
> Perhaps we can maintain a whitelist of files that are checked in this way, and people can opt-in for files they maintain without requiring everyone to do it all at once?

Yes, `ocamlformat` allows enabling or disabling it on a per-file basis using a dedicated config file, which would allow a gradual migration or using it only on some files in the long term.
In addition, like many formatters, it can be disabled locally on parts of a file, where we feel that custom formatting does a better job (e.g., a large expression where we would like to align things cleverly).

jamescheney:
In the dev meeting we decided that if this can be put in an opt-in form (i.e. allowing people to try it out on specific files without renaming everything wholesale) then this can be merged, people can try it out and we can have a more informed discussion whether to do automate or check formatting.

jstolarek:
@rudihorn, is this PR in a state where we can try it out? What is the exact plan anyway? Do we merge the PR into master or are we expected to use a separate branch for experimenting?

rudihorn:
As it stands I have changed the rules check to not run files called `*.skip` and have used this to disable the ocamlformat check. I have added the `.ocamlformat` config file to this PR, and have started working on a wiki page but have not finished this yet. I am otherwise happy for this to be merged to the master though, as it won't touch or break anything else of links.

rudihorn:
I think the most important things to know about ocamlformat are now listed in https://github.com/links-lang/links/wiki/Ocamlformat. Feel free to edit or let me know what else should be added.

frank-emrich:
Thanks, that looks good. I have extended the description for disabling `ocamlformat` by explaining that we could also have a `.ocamlformat-ignore` file (for disabling per-file) and that one can disable formatting for individual bindings/expressions/etc.

frank-emrich:
@rudihorn 
> As it stands I have changed the rules check to not run files called *.skip and have used this to disable the ocamlformat check.

I suggest that instead your script just greps the `.ocamlformat-ignore` file to see if a file should be ignored. Of course, this means that we can't use glob patterns like `bin/*` in there, but that's fine.

rudihorn:
@frank-emrich You are right this is a better way to do this. I have now added all files that violate ocamlformat into the `.ocamlformat-ignore` file. No changes are necessary to the script as ocamlformat already takes care of this. It does not support glob patterns anyway, but this is fine.

jstolarek:
I just checked out this PR branch to try it out. However, if I run, say `ocamlformat -i core/typeSugar.ml` nothing happens, i.e. the file remains unchanged. What am I doing wrong?

rudihorn:
@jstolarek you need to remove `core/typeSugar.ml` from `.ocamlformat-ignore`, then it will work.

jstolarek:
Ok, so now I get a bunch of warnings:

```
ocamlformat: ignoring "core/typeSugar.ml" (misplaced documentation comments - warning 50)
File "core/typeSugar.ml", line 3191, characters 11-67:
Warning 50: unattached documentation comment (ignored)
File "core/typeSugar.ml", line 3197, characters 11-167:
Warning 50: unattached documentation comment (ignored)
File "core/typeSugar.ml", line 3421, characters 11-122:
Warning 50: unattached documentation comment (ignored)
File "core/typeSugar.ml", line 3439, characters 11-59:
Warning 50: unattached documentation comment (ignored)
File "core/typeSugar.ml", line 3441, characters 13-191:
Warning 50: unattached documentation comment (ignored)
File "core/typeSugar.ml", line 3450, characters 11-46:
Warning 50: unattached documentation comment (ignored)
File "core/typeSugar.ml", line 3454, characters 11-82:
Warning 50: unattached documentation comment (ignored)
File "core/typeSugar.ml", line 3652, characters 10-76:
Warning 50: unattached documentation comment (ignored)
```
but the file still remains unchanged.

rudihorn:
This is because someone thought that having comments with two starts, i.e. `(** comment*)` looks nicer, even though this syntax is reserved for documentation comments. If documentation comments aren't correctly placed at an element the formatter doesn't know what to do and just leaves the file. The trick is to just change them into normal comments by removing one of the stars (i.e. `(* comment *)`).

jstolarek:
Yes, so I understand that. The question is whether this is all ocamlformat pays attention to at the moment? How do I make it reformat a file like you showed in https://github.com/links-lang/links/pull/504#issuecomment-468219082?

rudihorn:
Should work with:

```
sed -i '/typeSugar.ml$/d' .ocamlformat-ignore
sed -i 's/(\*\*/(\*/;s/(\*\*)//' core/typeSugar.ml
./tools/rules/ocamlformat_check fix
```

rudihorn:
You can also format all files in core by doing:

```
sed -i 's/(\*\*/(\*/;s/(\*\*)//' core/*.ml
sed -i '/core\/*.ml/d' .ocamlformat-ignore
./tools/rules/ocamlformat_check fix
```

jstolarek:
Works. Thanks.

jstolarek:
I played a bit with ocamlformat.  Here are some examples of code before and after ocamlformat formatting (all examples come from typeSugar.ml).

Before:
```OCaml
and is_pure p = match p.node with
  | QualifiedVar _
  | Constant _
  | Var _
  | FunLit _
  | DatabaseLit _
  | TableLit _
  | TextNode _
  | HandlerLit _
  | Section _ -> true
```

After:
```OCaml
  and is_pure p =
    match p.node with
    | QualifiedVar _ | Constant _ | Var _ | FunLit _ | DatabaseLit _
     |TableLit _ | TextNode _ | HandlerLit _ | Section _ ->
        true
```

Before:
```OCaml
val extend_record : griper
val record_with : griper

val list_lit : griper

val table_name : griper
val table_db : griper
val table_keys : griper
```

After:
```OCaml
val extend_record : griper

val record_with : griper

val list_lit : griper

val table_name : griper

val table_db : griper

val table_keys : griper
```

Before:
```OCaml
let but2things (lthing, (lexpr, lt)) (rthing, (rexpr, rt)) =
  build_tyvar_names [lt;rt];
  let ppr_lt = show_type lt in
  let ppr_rt = show_type rt in
  ", but the " ^ lthing ^ nli () ^
   code lexpr           ^ nl  () ^
  "has type"            ^ nli () ^
   code ppr_lt          ^ nl  () ^
  "while the " ^ rthing ^ nli () ^
   code rexpr           ^ nl  () ^
  "has type"            ^ nli () ^
   code ppr_rt
```

After:
```OCaml
let but2things (lthing, (lexpr, lt)) (rthing, (rexpr, rt)) =
  build_tyvar_names [lt; rt] ;
  let ppr_lt = show_type lt in
  let ppr_rt = show_type rt in
  ", but the " ^ lthing ^ nli () ^ code lexpr ^ nl () ^ "has type" ^ nli ()
  ^ code ppr_lt ^ nl () ^ "while the " ^ rthing ^ nli () ^ code rexpr ^ nl ()
  ^ "has type" ^ nli () ^ code ppr_rt
```

Before:
```OCaml
(* FIXME:

   we need to explicitly instantiate quantifiers
   like this *whenever* we do any kind of
   elimination
*)
```

After:
```OCaml
(* FIXME:

           we need to explicitly instantiate quantifiers
           like this *whenever* we do any kind of
           elimination
        *)
```

I think all of the above are regessions.  I admit though that there are fragments of code that, I think, look better:

Before:
```OCaml
let () = unify ~handle:Gripers.extend_record
  (pos_and_typ r, no_pos (`Record (absent_field_env, Types.fresh_row_variable (lin_any, res_any), false))) in
```

After:
```OCaml
let () =
  unify ~handle:Gripers.extend_record
    ( pos_and_typ r
    , no_pos
        (`Record
          ( absent_field_env
          , Types.fresh_row_variable (lin_any, res_any)
          , false )) )
```

Before:
```OCaml
let field_env' =
  StringMap.fold (fun label f field_env' ->
                    match f with
                      | `Absent ->
                          if StringMap.mem label field_env then
                            field_env'
                          else
                            StringMap.add label `Absent field_env'
                      | `Present t ->
                          if StringMap.mem label field_env then
                            failwith ("Could not extend record "^ expr_string (erase r)^" (of type "^
                                        Types.string_of_datatype rtype^") with the label "^
                                        label^
                                        " (of type"^Types.string_of_datatype (`Record (field_env, Unionfind.fresh `Closed, false))^
                                        ") because the labels overlap")
                          else
                            StringMap.add label (`Present t) field_env'
                      | `Var _ -> assert false) rfield_env field_env in
```

After:
```OCaml
let field_env' =
  StringMap.fold
    (fun label f field_env' ->
      match f with
      | `Absent ->
          if StringMap.mem label field_env then field_env'
          else StringMap.add label `Absent field_env'
      | `Present t ->
          if StringMap.mem label field_env then
            failwith
              ( "Could not extend record "
              ^ expr_string (erase r)
              ^ " (of type "
              ^ Types.string_of_datatype rtype
              ^ ") with the label " ^ label ^ " (of type"
              ^ Types.string_of_datatype
                  (`Record
                    (field_env, Unionfind.fresh `Closed, false))
              ^ ") because the labels overlap" )
          else StringMap.add label (`Present t) field_env'
      | `Var _ -> assert false )
    rfield_env field_env
```

Given there is a fair number of places where ocamlformat makes things worse (in my humble opinion) I am still reluctant to just blindly reformat all code with it.  I also tried ocamlformat with Emacs, hoping there is a way to format a region, rather than the whole buffer, but that does not seem to be possible.  So at the moment I would say "no" to ocamlformat, but of course I am open to further discussion.

frank-emrich:
@jstolarek Thanks for taking the time to play around with this!

Fortunately, some of your concerns can be solved by changing the `ocamlformat` configuration.

Setting `break-cases` to `all` turns the code in the first example to

```
match p.node with
    | QualifiedVar _
     |Constant _
     |Var _
     |FunLit _
     |DatabaseLit _
     |TableLit _
     |TextNode _
     |HandlerLit _
     |Section _ ->
        true
```
So the or-patterns are still visibly marked, but also still have their own line.

In your second example, setting `module-item-spacing=preserve` will preserve the original number of lines between items, so the fragment remains as it was before:

```
  val extend_record : griper
  val record_with : griper

  val list_lit : griper

  val table_name : griper
  val table_db : griper
  val table_keys : griper
```


Regarding your third example: Yes, that looks worse. However, the function is part of the `Griper` module within `TypeSugar`. I think this module should be in its own file (independent from any formatting discussions) and we should disable ocamlformat for this file as a whole, because it has many custom-formatted functions. Note that in other cases where we have hand-crafted formatting (like the table in `parseSettings.ml`), we can disable formatting locally using `[@@ocamlformat "disable"]` on a particular binding.


Regarding the formatting of comments: I think there is no silver bullet about what to do with them. I think enabling `ocamlformat` for comments would destroy many hand-crafted comments. So we should set `wrap-comments=false`. Interestigly, this doesn't fix the funny-looking FIXME comment you mentioned. This seems to be a genuine bug in `ocamlformat`, the lines after the FIXME are indented by one level when they shouldn't. There is a different comment starting with `FIXME:` in the same file that formats correctly.

In general, `ocamlformat` has a safety mechanism built-in where it will compare the AST before and after formatting. The formatter will fail fatally if the two ASTs are not identical (modulo comments, whose string contents may genuinely be different). This just means that we don't have to be too worried about `ocamlformat` bugs breaking our code.

jstolarek:
> In general, ocamlformat has a safety mechanism built-in where it will compare the AST before and after formatting.

Interestingly, I noticed one place where ocamlformat removed a `begin ... end`. That was actually nice since it was redundant.

> some of your concerns can be solved by changing the ocamlformat configuration.

In that case I would ask to provide a better configuration to try out!

frank-emrich:
> Interestingly, I noticed one place where ocamlformat removed a `begin ... end`. That was actually nice since it was redundant.

Note that `begin ... end` is part of the concrete syntax, not the abstract one. But still, the check would catch if something went wrong there. If `ocamlformat` would have messed up the  `begin ... end` or other kinds of parantheses in the file, the checker would detect this after re-parsing the file and comparing the ASTs.


> In that case I would ask to provide a better configuration to try out!

I guess we didn't do that initially because much of this depends on personal taste. I'll update the configuration with the settings I mentioned regarding your concerns.

rudihorn:
So the first example @jstolarek mentioned I see as a bit of a personal preference. It is set to make match cases in the top level to be grouped which in small functions results in fewer case lines which I get. Setting it to all is a bad idea as it tends to add lots of extra lines I don't like, but one could maybe do this for individual functions where it makes it easier to find specific cases. 

The second example is doing this because at the same time it is screaming "document me with ocamldoc comments". With comments I think it would be doing the right thing.

The third example is in my eyes a case of bad practice tabstopesque formatting, which in most cases can be ridden of by writing better code and maybe having it poorly formatted would motivate us to do so. An example of code which is formatted by ocamlformat:

```

  let but2things (lthing, (lexpr, lt)) (rthing, (rexpr, rt)) =
    build_tyvar_names [lt; rt] ;
    let ppr_lt = show_type lt in
    let ppr_rt = show_type rt in
    Format.sprintf
      ", but the %s\n\
      \  %s\n\
       has type\n\
      \  %s\n\
       while the %s\n\
      \  %s\n\
       has type\n\
      \  %s"
      lthing (code lexpr) (code ppr_lt) rthing (code rexpr) (code ppr_rt)
```

jstolarek:
> An example of code which is formatted by ocamlformat: (...)
Yes, that also looks nice. What ocamlformat settings does this require?

> I guess we didn't do that initially because much of this depends on personal taste.

I thought that the goal of this PR is to work out a set of coding conventions that we can all agree on (even if they don't fully align with our own personal preferences), so having concrete proposals on what these conventions should be is certainly a desirable thing!

jamescheney:
We decided yesterday to ~~kick the can down the road~~ see if @rudihorn or others can configure this in a whitelist, opt-in way such that users of `ocamlformat` will not affect non-users.  That is;
- [ ] formatting is opt-in: only happens when invoked explicitly by the user
- [ ] there is a whitelist/opt-in mechanism (for example covering only `core/lens/*`) and if someone not using ocamlformat creates a new file it is not formatted unless whitelisted

Open question: should there be a (non-enforced) check to warn if a whitelisted file is not formatted according to ocamlformat?

rudihorn:
So this is what I committed yesterday. It modifies the CI check script to only check the lens folder and everything else can just be run by the user. It won't allow anyone to use write on save, but it will stop someone uploading a whitelisted file.

rudihorn:
I had another look and it turns out it is possible to set `disable=true` in the global `.ocamlformat` file, and then set `disable=false` in `lens/.ocamlformat` as well as `tests/unit/.ocamlformat`. This allows write on save as well. It is then possible to enable individual files using `[@@@ocamlformat "disable=false"]`.  (I think just using `enable` instead of `disable=false` should also work). If this passes CI I'm happy to merge.

rudihorn:
Seems to pass the CI tests fine. @slindley @jstolarek are you fine with this being merged?

jstolarek:
So the current state is that if anyone makes changes to lens or unit test code and it doesn't conform to ocamlformat rules then the CI will break? Is that correct?

rudihorn:
Yes, for the lens code I would like this. This can be fixed by running `./tools/rules/ocamlformat_check fix`.

jstolarek:
Ok, this is fine with me. I see no objections to merging.

dhil:
> [...] are you fine with this being merged?

No. During the previous meeting, I understood that we would defer the decision until the next meeting on the condition it is possible to get build a whitelist rather than a blacklist.

rudihorn:
Do you have any specific concerns that would need to be discussed at the next links meeting @dhil?

dhil:
> Do you have any specific concerns [...]

No. As I said, I understood that we agreed to make no further attempts to merge this into master until we meet next.

rudihorn:
At the last meeting there seemed to be agreement to introduce a whitelist only system, which we now have figured out and is in this PR. If there are still legitimate reservations that need to be considered at the meeting then this is fine, but it would be good to mention these so that is some chance to address them before next meeting. If there are no more reservations / concens then it seems like pointless bureaucracy to wait for the next meeting, which is not to say we can't wait till then, but it should be in a manner that we can wave it off then.

jamescheney:
I believe we did agree not to merge anything before the next meeting so that everyone who has an opinion has a chance to try it out.  I agree with Rudi that there should be an expectation that people who have reservations about merging this make some effort before the meeting to try it out.  I'll send a reminder a week or so before and if there aren't specific/constructive objections or discussion then this should be merged.

SquidDev:
One thing possibly worth exploring is [Dune's built-in support for ocamlformat](https://dune.readthedocs.io/en/latest/formatting.html).

This allows you to run `dune build @fmt` to format all ml/mli files in the project and display a diff, and then `dune promote` to apply that diff. It might be possible to simplify the `ocamlformat_check ` script to use that instead.

jamescheney:
I think we agreed this can go in, pending making consistent with master and configuring so that only the lens directory is whitelisted.  Later, when ocamlformat is released with finer-grained whitelisting, the whitelisting configuration should be updated as well.

azure-pipelines.yml
```diff
@@ -53,9 +49,14 @@ steps:
                   echo "opam \$@" > opam2_local &&
                   chmod a+x opam2_local ;;
         *) exit 1 ;;
-      esac
+      esac;;
+      opam install ocamlformat
```
dhil: Make this a separate step; and you need to use the wrapper script `opam2_local` rather than `opam`.

azure-pipelines.yml
```diff
@@ -52,16 +52,16 @@ steps:
...
-      make rule-check
-    displayName: 'Rule check'
-
   - script: |
       ./opam2_local switch create ocaml-base-compiler.4.06.1
       eval `./opam2_local config env`
     displayName: 'OCaml compiler'
 
+  - script: |
+      ./opam2_local install ocamlformat
```
dhil: By a separate step, I meant it should be in its own `script` section, i.e.
```yaml
- script: |
     ./opam2_local install ocamlformat
  displayName: OCamlformat
```

rudihorn: Okay will do this.

azure-pipelines.yml
```diff
@@ -58,11 +58,11 @@ steps:
     displayName: 'OCaml compiler'
 
   - script: |
-      eval `./opam2_local env`
```
dhil: Technically, I think you should keep this line. Whenever you use `opam2` you likely want to interact with the custom switch.

rudihorn: As long as you aren't using any installed opam products it shouldn't really matter?

dhil: Correct, but you are going to install `ocamlformat` under whatever switch is active. We should probably also run `eval $(./opam2_local env)` in the rule check script.

rudihorn: The rule check script already does this (as it needs it). I have added it back to this switch.

Issue #233, opened 2017-07-18
Missing DOCTYPE on HTML pages


elordin:
As the spec says:

> A DOCTYPE is a required preamble.

Links-generated documents don't have one.

SimonJF:
This should be easy; add it to `irtojs.ml`?

elordin:
Yes. 
In theory at least.
But ...

Adding the `<!DOCTYPE html>` (which I would argue is the only reasonable one) kind of implies that the generated documents are compliant with that doctype.

Since `page` takes abitrary `Xml`, there is no guarantee that it only includes elements that are part of that doctype. Kind of meh.

And then there is additional fun with self-closing tags: There is a list of void elements that must be self-closing, e.g. `img`, and then there are all other elements, including custom elements registered via `customElements.define`, that _must not_ be self-closing, even if empty, to be valid.

jamescheney:
There was some discussion early on about finer-grained typechecking for the XML fragments a la XDuce/CDuce, which I think went by the wayside a while ago (back when it seemed more likely that future HTMLs would align with XML).  

Not sure what to do / if anything should be done about support for HTML5 tag soup style, other than perhaps wait for the Web to collapse.  Does anyone volunteer to write an embedded HTML5 parser/validity checker?  Maybe a lighter-weight alternative might be to keep track of whether a tag is self-closing in our XML representation, and preserve this, making it the programmer's problem...

elordin:
> Maybe a lighter-weight alternative might be to keep track of whether a tag is self-closing in our XML representation, and preserve this, making it the programmer's problem...

This would work for me as a programmer.

slindley:
This didn't "go by the wayside". We consciously decided not to do anything as messy/heavyweight as XDuce/CDuce.

rudihorn:
My workaround for the moment was to just add a comment to force it to not self close, e.g. 
> <script src="/bootstrap/js/bootstrap.min.js">/* test */</script>

elordin:
> My workaround for the moment was to just add a comment to force it to not self close, e.g.

That sure works but having to add a comment for every empty element, except for the 14 void elemnts, is _very_ cumversome.

SimonJF:
Closed in #506

PR #506, opened 2019-02-27
Add <!doctype html> to each page

SimonJF:
Did you know that not having a `doctype` on a page can cause subtle CSS bugs? Neither did I, until I spent an hour and a half trying to work out why on earth my CSS was displaying differently despite the class structure being exactly the same!

### DIFF
--- a/core/page.ml
+++ b/core/page.ml
@@ -77,6 +77,7 @@ module Make_RealPage (C : JS_PAGE_COMPILER) (G : JS_CODEGEN) = struct
       script_tag("  var cgiEnv = {" ^
                     mapstrcat "," (fun (name, value) -> "'" ^ name ^ "':'" ^ value ^ "'") cgi_env ^
                     "};\n  _makeCgiEnvironment();\n") in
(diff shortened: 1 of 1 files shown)


jamescheney:
See also: #233

SimonJF:
Hm, well-spotted / remembered. 

Given that this is causing obscure issues, I'd be in favour of having this merged and then scoping out the self-closing tags issue later, but I'm probably biased.

It's probably also worth keeping in mind that in the future, I'm not entirely sure how the frontend web stuff will look. I'm currently exclusively using the Links-MVU (Elm-style) library (modulo a little legacy preamble), and ideally I'd like to see that as the primary method of constructing Links pages in future.

jamescheney:
I think we should be adding the doctype tags and shouldn't treat solving the side-issues from #233 as a prerequisit.  I think whether / how to handle HTML capabilities that are not possible using our XML datatype is a separate issue - if we care about these we should create new issues summarizing the discussion in #233.

PR #509, opened 2019-02-28
Remove extra parentheses in Sugartypes datatypes

jstolarek:
As promised, I am removing redundant parentheses in Sugartypes data declarations. Two question.

1. Previously I would say:

```
generator:
| list_generator    { List  $1 }
| table_generator   { Table $2 }
```
Now I have:
```
generator:
| list_generator    { List  (fst $1, snd $1) }
| table_generator   { Table (fst $1, snd $1) }
```
Feels awkward. Is there a better way?

2. Similarly:
```OCaml
Fun (unwrap_def (binder ~ty:ft f, lin, ([], lam), location, None))
```
became:
```OCaml
let (bndr, lin, tvs, loc, ty) =
   unwrap_def (binder ~ty:ft f, lin, ([], lam), location, None) in
Fun (bndr, lin, tvs, loc, ty)
```
Is that OK?

### DIFF
--- a/core/desugarModules.ml
+++ b/core/desugarModules.ml
@@ -137,8 +137,8 @@ let rec rename_binders_get_shadow_tbl module_table
       | Fun (bnd, lin, (tvs, fnlit), loc, dt_opt) ->
           let (o, bnd') = self#binder bnd in
           (o, Fun (bnd', lin, (tvs, fnlit), loc, dt_opt))
-      | Type (name, tvs, ty) -> (self, Type (name, tvs, ty))
-      | Val (pat, tvs, loc, ty) -> (self, Val (pat, tvs, loc, ty))
+      | (Type _) as ty -> (self, ty)
+      | (Val  _) as v  -> (self, v )
       | Exp b -> (self, Exp b)
       | Foreign (bnd, raw_name, lang, ext_file, dt) ->
           let (o, bnd') = self#binder bnd in
@@ -210,12 +210,9 @@ and perform_renaming module_table path term_ht type_ht =
           (self, (xs', rv'))
 
     method! bindingnode = function
-      | Module (n, bs) ->
-          (self, Module (n, bs))
-      | AlienBlock (lang, lib, decls) ->
-          (self, AlienBlock (lang, lib, decls))
-      | Foreign (bndr, fname, lang, file, ty) ->
-         (self, Foreign (bndr, fname, lang, file, ty))
+      | (Module     _) as m  -> (self, m )
+      | (AlienBlock _) as ab -> (self, ab)
+      | (Foreign    _) as f  -> (self, f )
       | Type (n, tvs, dt) ->
           (* Add type binding *)
           let fqn = make_path_string path n in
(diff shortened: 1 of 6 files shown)


core/desugarModules.ml
```diff
@@ -212,9 +212,10 @@ and perform_renaming module_table path term_ht type_ht =
     method! bindingnode = function
       | Module (n, bs) ->
           (self, Module (n, bs))
-      | AlienBlock ab ->
-          (self, AlienBlock ab)
-      | Foreign f -> (self, Foreign f)
+      | AlienBlock (lang, lib, decls) ->
```
dhil: An opportunity for a micro-optimisation:
```ocaml
| (AlienBlock _) as ab -> (self, ab)
```
This generates slightly better code, as `AlienBlock` is reused rather than recreated.

core/desugarModules.ml
```diff
@@ -212,9 +212,10 @@ and perform_renaming module_table path term_ht type_ht =
...
+          (self, AlienBlock (lang, lib, decls))
+      | Foreign (bndr, fname, lang, file, ty) ->
```
dhil: Same micro-optimisation applies here.

core/desugarModules.ml
```diff
@@ -137,8 +137,8 @@ let rec rename_binders_get_shadow_tbl module_table
       | Fun (bnd, lin, (tvs, fnlit), loc, dt_opt) ->
           let (o, bnd') = self#binder bnd in
           (o, Fun (bnd', lin, (tvs, fnlit), loc, dt_opt))
-      | Type t -> (self, Type t)
-      | Val v -> (self, Val v)
+      | Type (name, tvs, ty) -> (self, Type (name, tvs, ty))
```
dhil: Opportunity for micro-optimisation.

core/desugarModules.ml
```diff
@@ -137,8 +137,8 @@ let rec rename_binders_get_shadow_tbl module_table
...
+      | Val (pat, tvs, loc, ty) -> (self, Val (pat, tvs, loc, ty))
```
dhil: Opportunity for micro-optimisation.

jamescheney:
The issue is that constructor syntax is slightly non-compositional here (unlike in Standard ML): a constructor application is of the form `Foo(a1,a2,...,an)` where `n` is the arity of `Foo` (i.e. width of the top-level unparenthesized product) and if you just write `Foo a` then it is an arity mismatch.

You could get around this by introducing "smart constructors" for `List`, `Table`, `Fun` if they don't exist already, that take tuples as arguments.  If these get inlined (which they should, given no recursion) then there should be no overhead, but that might require more aggressive optimizations than currently happen.

jstolarek:
I see. There is a smart constructor for `Fun` already but its interface is designed for usage in other places of the compiler (optional arguments, etc.), so I guess it is not worth the effort. Code would be longer overall.

That being said, are there any objections to merging this?

jamescheney:
Looks OK to me.

PR #513, opened 2019-03-01
Validate XML nodes during desugaring

### LINKED ISSUE #511: Validate XML attributes during desugaring
XML attributes are validated in `sugartoir.ml`:

```OCaml
(* check for duplicates *)
let () =
  let rec dup_check names =
    function
      | [] -> ()
      | (name, _) :: attrs ->
          if StringSet.mem name names then
            raise (Errors.SugarError (Position.dummy, (* JSTOLAREK: dummy position substituted here *)
                                   "XML attribute '"^name^"' is defined more than once"))
          else
            dup_check (StringSet.add name names) attrs
  in
    dup_check StringSet.empty attrs
in
  if tag = "#" then
    if List.length attrs != 0 || attrexp <> None then
      raise (Errors.SugarError (Position.dummy, (* JSTOLAREK: dummy position substituted here *)
                             "XML forest literals cannot have attributes"))
    else ...
```
This should really be done during desugaring.

jstolarek:
This patch moves XML node validation from Sugar->Ir stage into the desugaring stage. It also ensures that smart constructor performs the same validation.

Fixes #511

### DIFF
--- a/core/sugarConstructors.ml
+++ b/core/sugarConstructors.ml
@@ -70,9 +70,9 @@ module SugarConstructors (Position : Pos)
     | Sig {node=({node=signame; _}, datatype); pos} ->
        (* Ensure that name in a signature matches name in a declaration. *)
        if signame <> name then
-         raise (ConcreteSyntaxError
-               (Printf.sprintf "Signature for `%s' should precede definition of `%s', not `%s'."
-                  signame signame name, pos));
+         raise (ConcreteSyntaxError (pos,
+               Printf.sprintf "Signature for `%s' should precede definition of `%s', not `%s'."
+                  signame signame name));
        Some datatype
     | NoSig -> None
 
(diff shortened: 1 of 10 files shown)


core/sugarConstructors.ml
```diff
@@ -250,17 +250,36 @@ module SugarConstructors (Position : Pos)
...
+         | Some (opening, closing) ->
+            raise (ConcreteSyntaxError
+                     ("Closing tag '" ^ closing ^ "' does not match start tag '"
+                      ^ opening ^ "'.", pos))
+         | _ -> () in
+       (* Check uniqueness of attributes *)
+       let () =
+         let attr_names = fst (List.split attr_list) in
+         if ListUtils.has_duplicates attr_names then
+           raise (Errors.SugarError (pos, "XML tag '" ^ name ^
```
dhil: You can use `Printf.sprintf` to interpolate a value into a string, e.g.
```ocaml
Printf.sprintf "XML tag '%s' has duplicate attributes" name
```

jstolarek:
One small change that I'd like to push with this PR (as a separate commit though) is flipping order of arguments in `Sugartypes.ConcreteSyntaxError`. All other errors take `(Position.t * string)`, this one takes `(string * Position.t)`.

Issue #515, opened 2019-03-01
Code duplication between parse.ml and parseXml.ml


jstolarek:
The most important `read` function is identical between the two files, except that in one file it calls Links lexer and in another an XML lexer. This can be easily merged into one file. Perhaps there is a common interface that can be isolated with a module.

PR #517, opened 2019-03-01
Eliminate code duplication in parse and xmlParse

jstolarek:
I've taken a stab #515. Repetitions eliminated but I'm still not very good with functors, so critical feedback is welcome.

### DIFF
--- a/core/desugarDatatypes.ml
+++ b/core/desugarDatatypes.ml
@@ -6,6 +6,7 @@ open Sugartypes
 open Utility
 open List
 open Errors
+open Parse
 
 module SEnv = Env.String
 
@@ -545,7 +546,7 @@ let sentence typing_env = function
   | Directive   d  -> typing_env, Directive d
 
 let read ~aliases s =
-  let dt, _ = Parse.parse_string ~in_context:(Parse.fresh_context ()) Parse.datatype s in
+  let dt, _ = parse_string ~in_context:(LinksLexer.fresh_context ()) datatype s in
   let vars, var_env = Desugar.generate_var_mapping (typevars#datatype dt)#tyvar_list in
   let () = List.iter Generalise.rigidify_quantifier vars in
     (Types.for_all (vars, Desugar.datatype var_env aliases dt))
--- a/core/parse.ml
+++ b/core/parse.ml
@@ -1,6 +1,5 @@
 open Utility
-open SourceCode
-open Lexing
+open Scanner
 
 (* NB : For now, positions are resolved eagerly.  It might be better
    to resolve them lazily, i.e. keep the source around and only find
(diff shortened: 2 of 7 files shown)


core/parseXml.ml
```diff
@@ -1,66 +1,17 @@
...
-   try
-      let p = parse (XmlLexer.lexer context ~newline_hook:(Utility.from_option Utility.identity nlhook)) lexbuf in
-        (p, code)
+module XmlLexer : (LexerSig with type token         = XmlParser.token and
+                                 type lexer_context = XmlLexer.lexer_context) =
+  struct
+    type token         = XmlParser.token
+    type lexer_context = XmlLexer.lexer_context
+    type 'a grammar    = (Lexing.lexbuf -> XmlParser.token) -> Lexing.lexbuf ->
+                         'a
```
dhil: looks a bit weird that `'a` is on its own line.

core/desugarDatatypes.ml
```diff
@@ -545,7 +545,7 @@ let sentence typing_env = function
   | Directive   d  -> typing_env, Directive d
 
 let read ~aliases s =
-  let dt, _ = Parse.parse_string ~in_context:(Parse.fresh_context ()) Parse.datatype s in
+  let dt, _ = Parse.parse_string ~in_context:(Parse.LinksLexer.fresh_context ()) Parse.datatype s in
```
dhil: I'd probably define `let fresh_context () = LinksLexer.fresh_context ()` in `Parse`.

core/parse.ml
```diff
@@ -115,9 +140,12 @@ let reader_of_readline ps1 =
   (accessor_fun, populate_fun)
 
 
-let interactive : Sugartypes.sentence grammar = Parser.interactive
-let program : (Sugartypes.binding list * Sugartypes.phrase option) grammar = Parser.file
-let datatype : Sugartypes.Datatype.with_pos grammar = Parser.just_datatype
+let interactive : Sugartypes.sentence LinksLexer.grammar =
+  Parser.interactive
+let program : (Sugartypes.binding list * Sugartypes.phrase option)
+                LinksLexer.grammar = Parser.file
```
dhil: I first thought this was a comparison. But `grammar` takes `binding list` and `phrase option` as arguments. I would move `LinksLexer.grammar` up on the same line as its arguments.

core/parseXml.ml
jstolarek: Done

core/parse.ml
jstolarek: Done

core/desugarDatatypes.ml
jstolarek: `open Parse` seems to do a better job here.

core/parse.ml
```diff
@@ -8,28 +8,35 @@ open Lexing
...
-let read : context:Lexer.lexer_context
-        -> ?nlhook:(unit -> unit)
-        -> parse:('intermediate grammar)
-        -> infun:(bytes -> int -> int)
-        -> name:string
-        -> 'result * source_code =
-fun ~context ?nlhook ~parse ~infun ~name ->
-  let code = new source_code in
-  let lexbuf = {(from_function (code#parse_into infun))
+module type LexerSig = sig
```
jamescheney: From here to line 79, I'm not sure why this code should be in parse.ml.  Since it is reused in parseXml.ml which does not otherwise need anything from parse.ml, I suggest it may be moved to a standalone file, perhaps something like `lex.ml` or `lexerIntf.ml`.

core/parse.ml
```diff
@@ -8,28 +8,35 @@ open Lexing
...
+  type lexer_context
+  type 'a grammar = (Lexing.lexbuf -> token) -> Lexing.lexbuf -> 'a
+
+  val lexer : lexer_context
+           -> newline_hook:(unit -> unit)
+           -> (Lexing.lexbuf -> token)
+  val fresh_context : unit -> lexer_context
+end
+
+module Reader (Lex : LexerSig) = struct
```
jamescheney: The functor name `Reader` is a little generic, though natural given that the main function was already called `read`.  Maybe something like `Tokenizer`/`tokenize`?  Or maybe I'm mistaking what is going on here,

core/parse.ml
jstolarek: Frankly speaking, I had similar thoughts. I'll put it in a separate file.

core/parse.ml
jstolarek: All the names where a bit ad-hoc. I've really ran out of ideas for names, especially in the presence of very similar names like `Parse` and `Parser`. Better ideas are certainly welcome.

dhil: I suppose `Scanner` fits the bill.

jstolarek:
I renamed `Reader` to `Scanner` and moved it to a separate source file.

PR #508, opened 2019-02-27
Lens changes

jstolarek:
This is a continuation of #498.

@rudihorn, I split the #498 patch into two commits. First one contains refactorings of with_pos and binders and is already merged into master. Second commit is contained in this PR and contains all the lens changes. At the very least it needs a meaningful commit message, since I don't know what the lens changes are about. Please take over from here.

PR #519, opened 2019-03-05
Remove Relational Lenses Links dependencies

rudihorn:
This is a work in progress pull request for removing almost all references to the links core library from the relational lenses code. This is in preparation for moving the relational lenses code to a separate library that is referenced by `Links_core`.

Instead of directly using links types, the relational lens code now mostly defines its own datatypes for sugar syntax, types and values for any lens phrases, and a separate data type for lens values. Additional modules perform value conversion between these two types.

This simplifies the links code in some cases, because I have wrapped all lens values into the type `Lens of Lens_value.t`, reducing the number of cases required in Links. I have not done this for types and sugartypes, but it would be nice to do this as well. I believe there is still some legacy code which does depend on `Utility` and `Sugartypes` a little, but I will slowly get rid of this.

Todo:
- [X] Add separate lens phrase sugartypes which are type checked and reduced to lens phrase without source position information
- [X] Remove remaining dependencies on links core / reduce legacy code
- [x] Fix unit tests
- [x] Move to separate library

For a future PR:
- Reduce lens `sugartype` and `ir` into single type


jstolarek:
> I believe there is still some legacy code which does depend on Utility and Sugartypes a little, but I will slowly get rid of this.

That would be very nice. During my recent refactorings I have found lens code problematic in a sense that many Sugartypes changes propagated into the lens code. It would be nice to have a layer that isolates this.

Also, can I close #508?

jamescheney:
This looks helpful, but I think it would be good to have an architectural discussion eventually regarding how to "decouple" the lens code (so that there aren't explicit dependencies on the core language implementation) without necessitating code duplication, e.g. for database access.  Eventually, I would like the relational lens code to be generating `Query.t` queries and running them the same way we run ordinary queries, for example.  This suggests identifying parts of the lens code that duplicate (subsets of) functionality of the core library, and defining interfaces that the lens code can be abstracted over (using functors).  Perhaps I'm not correctly understanding what is going on yet, though (still have to do a proper read-through of the lens code).

rudihorn:
My latest commit actually manages to move lenses to its own lens library. I will probably try and clean things up and remove the `lens_` prefix from most files.

rudihorn:
I think I am generally happy for this to be merged now, will remove the [WIP] tag. The further changes to sugartypes and ir I will leave as future work for a separate PR. The examples compile again and all unit tests psas.

rudihorn:
Any objections to merging?

dhil:
I still don't quite understand the rationale behind this change. Links now depends on your Lens package. I firmly believe it should be the other way around.

jamescheney:
@dhil I don't understand what alternative architecture you have in mind by "the other way around".  Do you mean something like this:

- Links_core as is but with no `lens`-related code whatsoever
- lens library refers to Links_core for things like database access, or possibly is functorized over some common interfaces
- Links executable builds on both libraries / instantiates the lens functors with implementations from Links_core

That's what my earlier comment was asking about.  I think something like this is still an aspiration but not something we know how to do yet, e.g. how to factor out the `Lens` cases of `evalir`, how to modularize the parsing rules for lens constructs (or support them in some more EDSL-style way).

rudihorn:
It prevents a dependency of `links <- lenses <- links` which stops the lens code from being separated out and simplifies the dependency graph to `lenses <- links`. The major benefit of this is that the lens code becomes standalone which, as @jstolarek mentioned, means that refactoring to the links code means that the lens code doesn't need to be touched. It also allows for a slightly cleaner abstraction within the lens code. It can be thought of links providing access to the lens library the same way links exposes functionality of the `cohttp` library, and the `cohttp` library should not be dependent on links. 

It would possibly be nice to have the lens code be a dynamic dependency similar to how we do it with database drivers, but unlike with databases we don't have libraries with weird system dependencies that can't just be installed with opam. The net benefit of doing this would be very little for a lot of work, and even if something like this was implemented, it should ideally be three libraries, one for `links` one for `lens` and one for `links_lens` which depends on links and lens and exposes the functionality of lenses to links, but as @jamescheney mentioned this is something we aren't really sure of how we can achieve it.

jamescheney:
Another thing that would be cool would be to factor out the DB query stuff (`query/sql` and maybe others) into a separate library that Links can just import, but that is otherwise Links-agnostic.  (And `lens` could potentially use that for the database stuff as well.)

core/dune
```diff
@@ -13,13 +13,9 @@
   (modes native)
   (libraries str yojson ppx_deriving_yojson.runtime unix safepass base64
              ANSITerminal linenoise cohttp lwt websocket websocket-lwt.cohttp
-             findlib menhirLib)
+             findlib menhirLib links_lens)
```
dhil: I suppose this should be `links.lens` rather than `links_lens`, as the former denotes a subpackage, unless I am mistaken.

rudihorn: Changed this.

rudihorn:
If theres no objections I'll be merging this later today. @dhil ?

PR #520, opened 2019-03-06
Update and document the Makefile.

dhil:
It's all in the title.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,10 +8,8 @@
 
 # LATEST STABLE RELEASE
 # =====================
-# To install the latest stable release of Links please use opam, e.g.
-# $ opam install links
-# .. and to enable database support
-# $ opam install links-postgresql
+# To install the latest stable release of Links please see the INSTALL
+# file.
 
 # To build the bleeding-edge version of Links please refer to the
 # instructions below.
@@ -27,8 +25,8 @@
 # NOTE: The rule `all' may skip the building the database drivers if
 # the necessary system prerequisites are absent.
 
-# The following command builds Links in development mode with database
-# support
+# The following command builds Links in development mode without
+# database support
 # $ make no-db
 
 # RELEASE
--- a/Makefile
+++ b/Makefile
@@ -1,63 +1,139 @@
+#
+# Makefile for the Links Programming Language.
+#
+
+# This file contains instructions for building a bleeding-edge version
+# of Links in release or development mode with and without database
+# support.
+
+# LATEST STABLE RELEASE
+# =====================
+# To install the latest stable release of Links please use opam, e.g.
+# $ opam install links
+# .. and to enable database support
+# $ opam install links-postgresql
+
+# To build the bleeding-edge version of Links please refer to the
+# instructions below.
+
+# DEVELOPMENT
+# ===========
+# By default the build rule `all' attempts to build the latest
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -1,63 +1,139 @@
...
+
+# To build the bleeding-edge version of Links please refer to the
+# instructions below.
+
+# DEVELOPMENT
+# ===========
+# By default the build rule `all' attempts to build the latest
+# development version of Links with database support, thus the
+# following command suffices to build everything
+# $ make all
```
jstolarek: If `all` is the default target why not just instruct users to run `make` instead of `make all`?

Makefile
```diff
@@ -1,63 +1,139 @@
+#
+# Makefile for the Links Programming Language.
+#
+
+# This file contains instructions for building a bleeding-edge version
```
jstolarek: Maybe make lines 80 columns wide? In Emacs `C-x f`to set the fill width to 80 (in case it is not set) and then `M-q` on a paragraph. This will require inserting newlines between text and code snippets. Otherwise snippets will be treated as part of a paragraph and wrapped as well.

Makefile
```diff
@@ -1,63 +1,139 @@
...
+# By default the build rule `all' attempts to build the latest
+# development version of Links with database support, thus the
+# following command suffices to build everything
+# $ make all
+# For legacy reasons the rule `nc' (native compile) is an alias of
+# `all'.
+# NOTE: The rule `all' may skip the building the database drivers if
+# the necessary system prerequisites are absent.
+
+# The following command builds Links in development mode with database
```
jstolarek: "with no datatabase support" ?

Makefile
```diff
@@ -1,63 +1,139 @@
...
+# To build the bleeding-edge version of Links please refer to the
+# instructions below.
+
+# DEVELOPMENT
+# ===========
+# By default the build rule `all' attempts to build the latest
+# development version of Links with database support, thus the
+# following command suffices to build everything
+# $ make all
+# For legacy reasons the rule `nc' (native compile) is an alias of
```
frank-emrich: is there anything preventing us from just getting rid of the `nc` target?

Makefile
```diff
@@ -1,63 +1,139 @@
+#
+# Makefile for the Links Programming Language.
+#
+
+# This file contains instructions for building a bleeding-edge version
+# of Links in release or development mode with and without database
+# support.
+
+# LATEST STABLE RELEASE
```
frank-emrich: This is now slightly duplicated from the INSTALL file. I feel like we should just refer readers there for how to install the stable version

Makefile
dhil: I prefer to be explicit here. The intended audience is you, me, and the rest of the Links developers. However, I did leave a note about how to install the latest stable version in case some bewildered user happens to look inside this file.

Makefile
dhil: Good catch. Thanks.

Makefile
dhil: Done.

Makefile
dhil: Well, I often type `make nc`. It seems to be imprinted in my muscle memory.

frank-emrich: time to break with this habit then? ;)

PR #523, opened 2019-03-12
Relational Lenses emacs keywords

rudihorn:
This extends the emacs file with a few more keywords for relational lenses. I'm happy for this to be merged if noone opposes.


fehrenbach:
You might also want to add them to `links.vim`.

Issue #278, opened 2017-09-24
Broken resolution of mutual recursive type names in modules


dhil:
Opening an module which declares mutually recursive type names results in an error. 
Minimal example `mutual.links`
```
module T {
  typename A = B;
  typename B = A;
}

open T
()
```
Links cannot resolve `B`:
```
$ ./links -m mutual.links
*** Fatal error : Unbound type constructor B
```

SimonJF:
Fixed in #528.

Issue #184, opened 2017-06-05
No line number / position on error: Type alias applied with incorrect arity


elordin:
```
*** Fatal error : Type alias Alias applied with incorrect arity (a instead of b)
```

should give a position where the error occurred, preferably with filename, line and position

jstolarek:
@dhil there was a suggestion during Links meeting that this might be related to your work on module system. Is that the case? I am unassigning myself for the moment because I'm not actively working on fixing this.

dhil:
@jstolarek we can probably fix this in the new module system patch.

jamescheney:
Fixed by #528

jstolarek:
Hmm... I was pretty sure I commented on this one somewhere in #528 but can't seem to find my comment now. Anyway, does #528 really fix #184? Meaning: is the error being raised in OP an actual internal error in the compiler (an equivalent of `assert false`) or is this an error message that should be displayed to the user to inform that her program is incorrect? If the former, then #184 was never really a bug, if the latter then I don't think #184 is fixed by #528.

jamescheney:
I thought it was fixed by the change in `instantiate.ml` line 432:

``` 
-       failwith (Printf.sprintf
-                    "Type alias %s applied with incorrect arity (%d instead of %d)"
-                    name (List.length tyargs) (List.length vars))
+        internal_error (Printf.sprintf
+               "Type alias %s applied with incorrect arity (%d instead of %d). This should have been checked prior to instantiation."
+          name (List.length tyargs) (List.length vars))
```

Indeed, in the current version this now seems to do the right thing:

```
links> typename Foo(a,b) = (a,b);
Foo = a,b.(a, b)
links> fun bar(x:Foo(Int)) {x};
<stdin>:1: Arity mismatch: Type Foo expects 2 type arguments, but 1 arguments were provided. In: Foo(Int)
```

I guess that the error is caught earlier than the `instantiate.ml` one, though.  So hopefully that one is dead code now.

SimonJF:
Good question, @jstolarek. @jamescheney is correct in saying that the error is caught earlier than in `instantiate.ml`:

https://github.com/links-lang/links/pull/528/files#diff-22ddaf26a647d1c52db61fb8076d2e6dR219

The reason that this is treated as an internal error is that the check should have been done prior to `instantiate.ml` -- if it hasn't been, then this is because the `instantiate` API has been misused somewhere (therefore this is treated as an internal error).

jstolarek:
[...]

Issue #457, opened 2019-01-11
Should mutually recursive functions be indicated in syntax?


jamescheney:
Summarizing side-discussion in #455:  

Currently Links allows multiple adjacent functions to be mutually recursive ("binding groups"), as long as they aren't separated by something else (like a variable binding or typename), and it should (but currently isn't) an error to define the same function name more than once in such a group.

For example:
```
fun a() { println("a1") }
fun b() { a(); c() }
fun c() { a() }
var x = 42;
fun a() { println("a2") }
```
This has the (potential) drawback that commenting-out or rearranging code might lead to surprising errors, because.  In principle it also allows the REPL syntax to coincide with source program syntax, by allowing later bindings to shadow earlier ones; however, the above example is not writable as one "entry" in the REPL, I think this is because each REPL entry consists of either a single "var", a "fun" function group (with / without signatures), a typename, or a directive (etc.)  

[Incidentally, it's also already not the case that the REPL syntax = source file syntax even just for variable declarations
```
links> var x = 1; var y = 2;
x = 1 : Int
links> y;
<stdin>:1: Type error: Unknown variable y.
In expression: y.
```
as it appears everything after the first `;`-terminated declaration is ignored.]

The current design is intermediate between the ML-style approach (mutually-recursive bindings using "and"; later bindings shadow earlier ones; no function signatures) and the Haskell-style approach (all functions in a given module/toplevel scope are potentially mutually recursive; order doesn't matter; names have to be unique and shadowing isn't allowed; function signatures are optional).

One additional observation made in #455 is that if we were to add `and`, then it isn't clear what to do with the function signatures.  I'll give a proposal below.

This issue is to collect (further) discussion as to whether to change the current approach, and if so how.  This doesn't seem urgent but might be good to resolve before / as nontrivial changes are happening to the module system.

SimonJF:
This is related to #359 and #412, which (as someone who is writing larger-scale Links programs and libraries), are both becoming problematic.

@dhil and I met on Monday and decided that explicit recursive functions and types as in this proposal, coupled with nominal typing for recursive types, would be very helpful in solving the two above issues. A (large) advantage of this is that we can dispense with `refineBindings` altogether.

Since this is impacting me and I'd rather see it fixed sooner rather than later, I'm planning on working on this as a priority.

SimonJF:
[...]

Issue #412, opened 2018-10-22
Report errors about un-unrolled types


SimonJF:
I just received the following error message:

```
simon@ultralife:~/git-repos/distributed-links-examples/extended-chatserver$ links chatServer.links
chatServer.links:103: Type error: The cases of an offer should match the type of its patterns, but the pattern
    `s'
has type
    `RoomReceive'
while the subsequent patterns have type
[...]

PR #528, opened 2019-03-19
Explicit mutually-recursive functions and types

### LINKED ISSUE #455: Typename shadowing

SimonJF:
# Overview

Currently, `refineBindings.ml` does two things:

  1. Implicitly determines groups of mutually-recursive functions by building a call graph and topologically sorting it
  2. Hacks mutually-recursive types by inlining type aliases and inserting `mu` binders

While 1 might be reasonably sensible, 2 was a huge mistake, leading to massive slowdowns and incomprehensible error messages.

This patch implements the design we decided upon in #457, where mutually-recursive types and functions are explicitly defined in a `mutual` block. For example, consider the canonical example of odd / even numbers. This can be expressed as follows:

```
mutual {
  typename Even = [| Z | SuccE:Odd |];
  typename Odd = [| SuccO:Even |];

  sig three : () -> Odd
  fun three() {
    SuccO(SuccE(SuccO(Z)))
  }
}

three()
```

The key idea behind the patch is to adopt a degree of nominal typing when dealing with mutually-recursive types. Instead of the inlining pass, which would construct something like:

```
typename Even = mu X . [| Z | SuccE:[| SuccO:X |] |];
typename Odd = mu X . [| SuccO:[| Z | SuccE:[| SuccO:X |] |] |];
```

we now retain the type information, and compare recursive type applications directly. This has the advantage of making unification faster, as we are no longer comparing huge types, and also means that error messages no longer fill the screen with junk. Having explicit mutually-recursive functions and types is also useful when doing other passes (such as the new module system), and fixes some odd behaviour.

## Relation to issues
This patch addresses #457, and also fixes #455 and #412. It should also go some way to helping with #359, although there are other issues at play there which could probably be fixed by considering types as minimised automata.

# Technical design

## Frontend

New AST node, `Mutual`, which contains a list of mutually-defined bindings.
Type alias constructor with `Type` is replaced with `Typenames`, which contains a list of mutually-recursive types. Types not defined in a `mutual` block are represented by a singleton list.

## `refineBindings.ml`
Redundant: removed completely. Rejoice!

## `types.ml`
Each type group is associated with some metadata:
```
type tygroup = {
  id: int;
  type_map: ((quantifier list * typ) Utility.StringMap.t);
  linearity_map: bool Utility.StringMap.t
}
```
Here, `id` is a unique ID for the type groups; `type_map` stores data about each type in the group (in particular, the quantifiers and the desugared datatype); and `linearity_map` stores linearity information.

A new `tycon_spec` constructor, `Mutual of (quantifier list * tygroup ref)` is used when desugaring types in the same group. As we cannot count on having desugared datatypes for mutually-defined types since they can be cyclic, type applications to a mutually-defined type are compiled to a new semantic type `RecursiveApplication of rec_appl`, where `rec_appl` is defined as:

```
and rec_appl = {
  r_name: string;
  r_dual: bool;
  r_unique_name: string;
  r_args: type_arg list;
  r_unwind: type_arg list -> bool -> typ;
  r_linear: unit -> bool option
}
```

Here, `r_name` is the user-visible name, and `r_unique_name` is used internally for unification. `r_dual` is used to track whether the type application is dualised (in the context of session types) -- my treatment of dualisation follows that of Lindley & Morris (2016). `r_args` contains the type arguments. `r_unwind` is a function which takes a list of type arguments and a boolean stating whether the unwinding should be dualised, and returns the one-step unfolding of the type: this is used
during unification. Finally, `r_linear` is a function which returns `None` if linearity has not yet been calculated; `Some true` if the recursive application is linear; and `Some false` if not.

## `desugarDatatypes.ml`

Constructs the semantic types for mutually-recursive syntactic types by creating a new `tygroup`; populating the alias environment with `Mutual` `tycon_spec`s for types in the same mutual blocks; and then desugaring each type. After this, it becomes possible to calculate the linearity of each type by calculating recursive applications; topologically-sorting the SCCs; and propagating linearity through the application graph. Finally, desugaring of the remaining program is calculated with an
updated `alias_env` with each of the types promoted to a "proper" `Alias`.

## `desugarMutual.ml`
A simple pass which desugars `mutual` blocks into `Typenames` and `Funs` blocks.

## `unify.ml`
Modified to track recursive applications as well as mu-bound function names. When unifying two `RecursiveApplication`s, checks to see whether their (unique) internal name is equal.
If so, unification succeeds.
If a recursive application is already witnessed, unification succeeds. Unification with a primitive fails. Otherwise, unwinds the recursive type, and records that this has been done in order to avoid nontermination.

## `instantiate.ml`
Some duplicate code refactored out; also some functionality exposed which allows us to instantiate recursive aliases.

## Testing
Test cases:
  * Lists (basic recursive types and functions on them)
  * Odd/even numbers (mutually-recursive types)
  * Only functions and typenames may be in a recursive block
  * Unguarded recursive types cannot unify with primitives
  * Type variables are not shared in a mutual block
  * Linearity: Linearity is propagated correctly, but also unrestricted types can still be treated as unrestricted
  * Duplicate bindings disallowed


core/desugarDatatypes.ml
```diff
@@ -77,7 +84,7 @@ object (self)
   method! bindingnode = function
     (* type declarations bind variables; exclude those from the
        analysis. *)
-    | Type _    -> self
+    | Typenames _    -> self
```
jstolarek: Maintain `->` alignment.

core/desugarDatatypes.ml
```diff
@@ -111,6 +118,31 @@ exception UnexpectedOperationEffects of string
 
 module Desugar =
 struct
+  (* Desugars quantifiers into Types.quantifiers,
+   * returning updated variable environment.
+   * Lifted / deduplicated from `typename` and `Forall` desugaring. *)
+  let desugar_quantifiers (var_env: var_env) (qs: Sugartypes.quantifier list) :
+      (Types.quantifier list * var_env) =
+      ListLabels.fold_right ~init:([], var_env) qs
+      ~f:(fun q (args, {tenv=tenv; renv=renv; penv=penv}) ->
```
jstolarek: Shorten this code by using field punning and avoiding repetition between the branches:

```OCaml
      ~f:(fun (name, (primarykind, subkind), _freedom)
              (args, {tenv; renv; penv}) ->
            let var = Types.fresh_raw_variable () in
            let subkind = concrete_subkind subkind in
            match primarykind with
              | PrimaryKind.Type ->
                  let point = Unionfind.fresh (`Var (var, subkind, `Rigid)) in
                    ((var, subkind, `Type point)::args,
                     {tenv=StringMap.add name point tenv; renv; penv})
              | PrimaryKind.Row ->
                  let point = Unionfind.fresh (`Var (var, subkind, `Rigid)) in
                    ((var, subkind, `Row point)::args,
                     {tenv; renv=StringMap.add name point renv; penv})
              | PrimaryKind.Presence ->
                  let point = Unionfind.fresh (`Var (var, subkind, `Rigid)) in
                    ((var, subkind, `Presence point))::args,
                     {tenv; renv; penv=StringMap.add name point penv})
```
Is there a trick to float `point` binding out of the branches? The expression is the same in all the three branches but it has a different type in each of them.

core/desugarDatatypes.ml
```diff
@@ -180,45 +185,69 @@ struct
...
+                  let ts = ListUtils.zip' qs ts in
+                  List.map
+                    (fun (q,t) ->
+                      let (q, t) = match_kinds (q, t) in
+                      match subkind_of_quantifier q with
+                      | (_, Restriction.Effect) -> type_arg' var_env alias_env t
+                      | _ -> type_arg var_env alias_env t) ts
+                with
+                | ListUtils.Lists_length_mismatch ->
+                   failwith (Printf.sprintf
```
jstolarek: In line with what was said in #525, would it be possible to replace `failWith` with `Errors`?

core/desugarDatatypes.ml
```diff
@@ -180,45 +185,69 @@ struct
...
+                     "Arity mismatch: the type constructor %s expects %d arguments, but %d arguments were provided"
+                     tycon (List.length qs) (List.length ts))
+                | Kind_mismatch ->
+                   failwith "Kind mismatch"
```
jstolarek: Same here.

core/desugarMutual.ml
```diff
@@ -0,0 +1,153 @@
...
+      |  Fun (bndr, lin, (tvs, fnlit), location, dt) ->
+          let fnlit = self#funlit fnlit in
+          if is_recursive bndr fnlit then
+            WithPos.make ~pos:position (Funs [(bndr, lin, ((tvs, None), fnlit), location, dt, position)])
+          else
+            WithPos.make ~pos:position (Fun (bndr, lin, (tvs, fnlit), location, dt))
+      | _ -> super#binding b
+end
+
+let check_dups funs tys =
```
jstolarek: There's a `has_duplicates` function in `utility.ml`. Would it be possible to rely on it instead of duplicating/reinventing the logic of duplicate checking?

core/desugarMutual.ml
```diff
@@ -0,0 +1,153 @@
...
+    else () in
+  check funs_map; check tys_map
+
+
+let rec flatten_simple = fun () ->
+object(self)
+  inherit SugarTraversals.map as super
+
+  method flatten_bindings bs =
+    List.concat (
```
jstolarek: There is `concat_map` in `utility.ml`.

core/instantiate.ml
```diff
@@ -410,47 +419,14 @@ let alias name tyargs env =
         failwith (Printf.sprintf "Unrecognised type constructor: %s" name)
     | Some (`Abstract _) ->
         failwith (Printf.sprintf "The type constructor: %s is abstract, not an alias" name)
+    | Some (`Mutual _) ->
+        failwith (Printf.sprintf "The type constructor: %s is a mutually-defined datatype, not an alias" name)
```
jstolarek: Avoid `failWith`?

core/moduleUtils.ml
```diff
@@ -174,13 +174,25 @@ let create_module_info_map program =
...
+          get_binding_names ms @ get_binding_names bs
       | _ :: bs -> get_binding_names bs in (* Other binding types are uninteresting for this pass *)
 
     (* Getting type names -- we're interested in typename decls *)
     let rec get_type_names = function
       | [] -> []
-      | { node = Type (n, _, _); _} :: bs -> n :: (get_type_names bs)
-      | _ :: bs -> get_type_names bs in
+      | b :: bs ->
+          begin
```
jstolarek: Out of curiosity, why do you use `begin ... end` block here? Is it a matter of style or is there a different reason?

core/parser.mly
```diff
@@ -871,6 +876,12 @@ block_contents:
 | exp                                                          { ([], $1) }
 | SEMICOLON | /* empty */                                      { ([], with_pos $loc (TupleLit [])) }
 
+bindings_block:
```
jstolarek: I wonder whether it is worth to introduce a production that is used only once. Also, I find the name slightly misleading, since it doesn't produce a `Block`.

core/transformSugar.ml
```diff
@@ -808,15 +811,23 @@ class transform (env : Types.typing_environment) =
...
-      | Type _ -> failwith "Unannotated type alias"
+      | Typenames ts ->
+          let (o, _) = listu o (fun o (name, vars, (x, dt')) ->
+            begin
+              match dt' with
+                | Some dt ->
+                   let o = o#bind_tycon name
+                     (`Alias (List.map (snd ->- val_of) vars, dt)) in
+                   (o, (name, vars, (x, dt')))
+                | None -> failwith "transformSugar.ml: Unannotated type alias"
```
jstolarek: Avoid `failWith`?

core/typeSugar.ml
```diff
@@ -3851,9 +3854,16 @@ and type_binding : context -> binding -> binding * context * usagemap =
...
-      | Type (name, vars, (_, Some dt)) as t ->
-          t, bind_tycon empty_context (name, `Alias (List.map (snd ->- val_of) vars, dt)), StringMap.empty
-      | Type _ -> assert false
+      | Typenames ts ->
+          let env = List.fold_left (fun env (name, vars, (_, dt')) ->
+            begin
+              match dt' with
+                | Some dt ->
+                    bind_tycon env (name, `Alias (List.map (snd ->- val_of) vars, dt))
+                | None -> failwith "typeSugar.ml: unannotated type"
```
jstolarek: Avoid `failWith`?

core/types.ml
```diff
@@ -1774,15 +1887,19 @@ struct
...
-          let bound_vars, vars =
-            List.fold_left
-              (fun (bound_vars, vars) tyvar ->
-                 let var, spec = varspec_of_tyvar tyvar in
-                   (TypeVarSet.add var bound_vars, (var, spec)::vars)) (bound_vars, []) tyvars
-          in
-            (List.rev vars) @ (free_bound_type_vars ~include_aliases bound_vars body)
+          let (bound_vars, vars) = split_vars tyvars in
+          vars @ (free_bound_type_vars ~include_aliases bound_vars body)
+      | `Mutual (tyvars, _) -> snd @@ split_vars tyvars
```
jstolarek: Am I the only one who dislikes `@@`?

core/unify.ml
```diff
@@ -785,6 +830,10 @@ let rec unify' : unify_env -> (datatype * datatype) -> unit =
        end
     | `End, `End -> ()
     | _, _ ->
+        (*
+        let () = Printf.printf "Match fail in unify: \n%s \n and \n%s"
```
jstolarek: Is this for debugging purposes? In other places I've seen `Debug.print` being used instead of `Printf.printf`.

SimonJF:
Thanks for your review, Janek -- I really appreciate you taking the time. Those comments will indeed make the PR better. The `failwith`s were mainly from previous code, but I agree that if I'm touching the code anyway, I may as well make a start on doing them properly.

core/types.ml
SimonJF: `@@` is rubbish, yes. `<|` is a synonym. Would that be OK? Alternatively I could just be explicit and bracket it.

jstolarek: My personal preference is to use brackets, but I don't really know what is more common in OCaml.

core/parser.mly
```diff
@@ -850,7 +854,8 @@ binding:
 | signature linearity VARIABLE arg_lists block                 { fun_binding ~ppos:$loc (Sig $1) ($2, $3, $4, loc_unknown, $5) }
 | linearity VARIABLE arg_lists block                           { fun_binding ~ppos:$loc  NoSig   ($1, $2, $3, loc_unknown, $4) }
 | typed_handler_binding                                        { handler_binding ~ppos:$loc NoSig $1 }
-| typedecl SEMICOLON | links_module | alien_block | links_open { $1 }
+| typedecl SEMICOLON | links_module | alien_block
+| links_open SEMICOLON | mutual_block                          { $1 }
```
dhil: I'd inline `mutual_block`.

core/parser.mly
```diff
@@ -871,6 +876,12 @@ block_contents:
 | exp                                                          { ([], $1) }
 | SEMICOLON | /* empty */                                      { ([], with_pos $loc (TupleLit [])) }
 
+bindings_block:
+| LBRACE bindings RBRACE                                       { $2 }
+
+mutual_block:
+| MUTUAL bindings_block                                        { with_pos $loc (Mutual $2) }
```
dhil: I'd inline `bindings_block`.

core/sugartypes.ml
```diff
@@ -274,10 +274,11 @@ and bindingnode =
                    (Types.datatype * Types.quantifier option list) option)
                    * funlit) * Location.t * datatype' option * Position.t) list
   | Handler of Binder.with_pos * handlerlit * datatype' option
+  | Mutual of binding list
```
dhil: An alternative design would be to use a richer structure than a list, say
```ocaml
type mutual_defs = {
  mut_funs : binding;
  mut_types : binding;
}

let add_mutual_binding ({ mut_types = Typenames ts; mut_funs = Funs fs } as block) binding =
  match WithPos.node binding with
  | Fun _ -> { block with mut_funs = Funs (binding :: fs) }
  | Type _ -> { block with mut_types = Typenames (binding :: ts) }
  | _ -> *raise a syntax error*
```
With the help of `add_mutual_binding` the parser could partition the bindings; if I am not mistaken that would render `flatten_bindings` above obsolete, and safe at least a post-traversal.

Note this design abuses the parser, but I think it is justified as the partitioning logic wouldn't be contained directly within the parser.

core/sugartypes.ml
```diff
@@ -483,6 +484,22 @@ struct
...
        name, (diff (funlit fn) name)
+    | Mutual bnds ->
+        (* Traverse, get names and RHSes *)
+        let names, fnlits, hlits =
+          List.fold_right
+            (fun b (names, fnlits, handlerlits) ->
+              match WithPos.node b with
+                | (Fun (bndr, _, (_, rhs), _, _)) ->
+                   (add (Binder.to_name bndr) names, rhs::fnlits, handlerlits)
+                | (Handler (bndr, hl, _)) ->
```
dhil: I want to get rid of the `Handler` node. It was a poor attempt at simulating first-class handlers using functions. I wouldn't be opposed to you just doing `assert false` here. Anyways, I will get rid of the `Handler` node in a subsequent patch.

core/types.ml
```diff
@@ -143,6 +176,7 @@ type typ =
     | `Lens of lens_sort
     | `Alias of ((string * type_arg list) * typ)
     | `Application of (Abstype.t * type_arg list)
+    | `RecursiveApplication of rec_appl
```
dhil: I was hoping to use `Application` rather than introducing `RecursiveApplication`. Perhaps it matters not here, but for the module system, I think it shouldn't be necessary as we'll resolve names. I think we will try to unify the two in a subsequent module patch.

core/types.ml
dhil: Yes, just use parentheses.

core/types.ml
```diff
@@ -2170,19 +2292,27 @@ struct
...
-                | [] -> datatype bound_vars p body
-                | _ -> mapstrcat "," (quantifier p) tyvars ^"."^ datatype bound_vars p body
-            end
+          let bvs = bound_vars tyvars in
+          begin
+            match tyvars with
+              | [] -> datatype bvs p body
+              | _ -> mapstrcat "," (quantifier p) tyvars ^"."^ datatype bvs p body
+          end
+      | `Mutual (tyvars, _) ->
```
dhil: Can you dump an example of what this produces?

core/instantiate.ml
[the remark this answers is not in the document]
SimonJF: Thanks -- that is the intention. I'm currently addressing @jstolarek's suggestions, and I'm incorporating an "InternalError" exception to handle cases like this.

I think you're right in that they should be wildcarded.

SimonJF:
Thanks for your comments, @dhil. I am working through both sets of comments now.

Regarding your question: I tested the web examples as I went through them, and nothing changed.

dhil:
> Thanks for your comments, @dhil. I am working through both sets of comments now.

My pleasure. I don't request any changes per see, they are just comments.

core/instantiate.ml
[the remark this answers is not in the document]
jstolarek: Another observation is that cases that are expected to be impossible are often mark with `assert false`, though admittedly an error message is more informative.

core/sugartypes.ml
SimonJF: I'm not entirely sure about this -- saving a traversal would be nice, as would removing the binding flattening, but did we not recently discuss at a meeting that the parser is already doing too much work?

core/types.ml
SimonJF: As it stands, `RecursiveApplication` has a lot more structure (for example, dualisation, unwinding, linearity tracking amongst a group, etc.) than `Application` at the moment. Maybe a better design would unify `Application` and `RecursiveApplication` to something like:

```
type application = | Abstract of Abstype.t | RecursiveApplication of rec_appl
...
| `Application of application * type_arg list
```

core/sugartypes.ml
dhil: Yes we did, but my point is that while this design would shift the work to the parser, the parser is somewhat oblivious to that fact. It is sort of like using a smart constructor.

core/types.ml
dhil: I did not mean to imply that `Application` would suffice as is. In time we'll get rid of the special handling of abstract types too. The next iteration of the module system is likely to introduce abstract types.

core/desugarMutual.ml
SimonJF: I considered this, but `has_duplicates` does not retain the information we need in order to provide an informative error message.

jstolarek: Personally, I would be in favour of recovering this information when an error message is to be generated. This might require adding some new general-purpose list functions in `utility.ml`, but I would rather see new functionality being added rather than existing functionality being duplicated.

examples/AIPL14/SMTP/Problem3.links
```diff
@@ -20,52 +20,53 @@ typename SMTPServer =
...
-}
-
 # Returns true, if the given String is a valid e-mail address.
 # The bit before @ has to be a single word!
 sig valid : (String) ~> Bool
 fun valid(address) {
   address =~ /.[a-z0-9]+.@.[a-z]+.[a-z]+/ # Regex to match 1+characters/numbers @ 1+characters . 1+characters
 }
 
+mutual {
```
jamescheney: The sheer number of changes due solely to indentation (inside `mutual` blocks) makes me very sad.  There are also a lot of inevitable changes due to the need to reorder things that are not mutually recursive.

core/types.mli
```diff
@@ -168,14 +201,19 @@ val dual_row : row -> row
...
+  | `Abstract of Abstype.t
+  | `Mutual of (quantifier list * tygroup ref) (* Type in same recursive group *)
+]
 
 type environment        = datatype Env.String.t
  and tycon_environment  = tycon_spec Env.String.t
  and typing_environment = { var_env   : environment ;
                             tycon_env : tycon_environment ;
-                            effect_row : row }
+                            effect_row : row}
```
jamescheney: This change seems accidental and asymmetric

core/types.ml
```diff
@@ -131,7 +131,40 @@ type lens_phrase =
 
 (* End of Lenses *)
 
-type typ =
+type rec_id =
```
jamescheney: Maybe add a line or two of documentation explaining the need for these two cases?

core/desugarDatatypes.ml
```diff
@@ -206,19 +208,16 @@ struct
...
                       | (_, Restriction.Effect) -> type_arg' var_env alias_env t
                       | _ -> type_arg var_env alias_env t) ts
                 with
                 | ListUtils.Lists_length_mismatch ->
-                   failwith (Printf.sprintf
-                     "Arity mismatch: the type constructor %s expects %d arguments, but %d arguments were provided"
-                     tycon (List.length qs) (List.length ts))
-                | Kind_mismatch ->
-                   failwith "Kind mismatch"
+                    raise (TyAppArityMismatch {pos; name=tycon;
```
dhil: I'd prefer if you were to spell out the name `TypeApplicationArityMismatch`.

core/desugarDatatypes.ml
```diff
@@ -206,19 +208,16 @@ struct
...
+                      expected=(List.length qs); provided=(List.length ts)})
```
dhil: I think you can drop the parentheses here.

core/desugarMutual.ml
```diff
@@ -0,0 +1,152 @@
...
+    let dups =
+      StringMap.filter (fun _ poss -> List.length poss > 1) map in
+    if StringMap.cardinal dups > 0 then
+      raise (MultiplyDefinedMutualNames dups)
+    else () in
+  check funs_map; check tys_map
+
+
+let rec flatten_simple = fun () ->
+object(self)
```
jamescheney: Why not 
```
let rec flatten_simple () = 
...
and flatten_bindings () = 
```

??

Also, I'd prefer if the body of both "let"s were indented, so that the `open` and `object ... end` constructs are clearly not toplevel ones.

core/desugarMutual.ml
```diff
@@ -0,0 +1,152 @@
...
+ * Typenames [Foo; Bar]
+ * Funs [foo; bar]
+ *
+ * Any other binding occurring in a mutual block is forbidden.
+ *)
+
+(* Simple check to see whether a function is recursive: checks whether the
+ * binder occurs within the free variables of the function body. *)
+let is_recursive bnd fnlit =
+  StringSet.mem (Binder.to_name bnd) (Freevars.funlit fnlit)
```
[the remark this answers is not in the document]
SimonJF: Yes - but this check is only used for single `Fun` bindings, to check whether they need to be lifted to a singleton `Funs` block (I.e., a recursive function which is not mutually recursive and only refers to itself). Mutually recursive functions would be in a `Funs` block already by this point. 

It's worth spelling this out in a comment though, thanks.

SimonJF:
@jamescheney: thanks for the review; all of those sound sensible.

I was also sad about the huge ripple effect and subsequent large changes; the alternative proposal with `and` may have kept this down a little, but the reordering needed would have still kept the modification count reasonably high.

SimonJF:
I think the CI failure was transient, so I'm starting it again. I've (mostly) now managed to remove the `Mutual` AST node, following @dhil 's suggestion. Still bashing my head against the wall getting it to work with the current module system but hopefully will have something soon.

core/sugartypes.ml
SimonJF: I went with this design in the end. It does introduce complexity in the parser (there are some intricacies not appearing in the above sketch) -- but I came round to the opinion that the additional parser complexity was worth removing the `Mutual` node and therefore most of `desugarMutual` completely. Good suggestion -- thanks.

core/errors.ml
```diff
@@ -62,12 +63,17 @@ let format_exception =
...
+  | MultiplyDefinedMutualNames duplicates ->
+      "Duplicate mutually-defined bindings\n" ^
         StringMap.fold (fun name positions message ->
                           message^" "^name^":\n  "^
 			    (mapstrcat "\n  " show_pos (List.rev positions)))
           duplicates ""
+  | InvalidMutualBinding pos ->
+      let pos, expr = Position.resolve_start_expr pos in
+      Printf.sprintf
+        "%s:%d: Mutual blocks can only contain `fun` and `typename` bindings, but the block contained: %s.\n"
```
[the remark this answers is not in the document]
SimonJF: I went with your phrasing suggestion in the end -- this check is now moved to `parser.mly`.

core/moduleUtils.ml
SimonJF: It turned out to be unnecessary -- good spot.

core/desugarDatatypes.ml
SimonJF: I couldn't find a way to float `point` out, unfortunately. Thanks for your comments in shortening the code, though, which I've added.

core/moduleUtils.ml
jstolarek: For the record, there were several such `begin .. end` blocks in the patch that I originally reviewed but I only flagged this one.

SimonJF: Ah, I see. I tend to use them whenever I do nested matches (although this one was a `function` and a `match`) to avoid errors due to ambiguity. I'm generally rather conservative with adding them in as the resulting errors can be confusing to work out.

SimonJF:
Thanks, everyone, for your comments so far: I have incorporated them and certainly think that they have improved the PR.

The largest change has been eliminating the `Mutual` node from `sugartypes` altogether by following @dhil's suggestion to move the work to the parser. Amusingly, I accidentally fixed #278 in the process.

core/typevarcheck.ml
```diff
@@ -206,6 +210,8 @@ and is_positive : TypeVarSet.t -> int -> datatype -> bool =
         | `Alias (_, t) -> isp t
         | `Application (_, ts) ->
             List.exists (is_positive_type_arg bound_vars var) ts
+        | `RecursiveApplication { r_args ; _ } ->
```
frank-emrich: I'm not sure if this is correct. When determining whether some type `Foo(a)` is positive w.r.t. `a`, we should check whether `a` is positive w.r.t. the type that results from unrolling Foo and substituting `a` in. At least that's what the old code would have done if `Foo` is  an alias for a recursive variant type. But that's something our resident type experts have a better grasp on than me...

core/unify.ml
```diff
@@ -586,6 +609,28 @@ let rec unify' : unify_env -> (datatype * datatype) -> unit =
...
        List.iter2 (fun lt rt -> unify_type_args' rec_env (lt, rt)) ls rs
+    | `Primitive t, `RecursiveApplication a
+    | `RecursiveApplication a, `Primitive t ->
+       raise (Failure
+                (`Msg ("Cannot unify primitive type '"^string_of_datatype (`Primitive t) ^
+                         "' with recursive type '"^ a.r_name ^"'")))
+    | `RecursiveApplication a1, `RecursiveApplication a2 ->
+        let (n1, args1) = (a1.r_unique_name, a1.r_args) in
+        let (n2, args2) = (a2.r_unique_name, a2.r_args) in
+        if n1 <> n2 then
```
frank-emrich: This means that two named types are only unifiable if they are the same. I would have expected this to fall back to a structural unification if the names don't match. 

The following was valid before, but fails to type-check now:
```
mutual {
  typename T1(a) = [| Nil | Cons: (a,T1(a)) |];
}
mutual {
  typename T2(a) = [| Nil | Cons: (a,T2(a)) |];
}

sig nil : T1(Int)
var nil = Nil;

sig test :(T2(Int)) -> Int
fun test(x) {
  123
}

test(nil)
```
Here, T1 and T2 are the same type with a different name, but we cannot use them interchangeably.

Of course, this is a design decision, so the others should probably chime in on this.

core/liftRecursive.ml
```diff
@@ -0,0 +1,30 @@
...
+open SourceCode
+
+(* A simple pass to lift (self)-recursive functions from a `Fun` node to a `Funs` node. *)
+
+(* Simple check to see whether a function is recursive: checks whether the
+ * binder occurs within the free variables of the function body. *)
+(* Note that we do not need to handle mutually-recursive functions
+ * here since they will be constructed explicitly during parsing. *)
+let is_recursive bnd fnlit =
+  StringSet.mem (Binder.to_name bnd) (Freevars.funlit fnlit)
```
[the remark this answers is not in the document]
jamescheney: In OCaml (and Scheme) there is a distinction between `let` and `letrec`, but in Standard ML (and probably most languages) there is only one function definition construct and it is implicitly recursive.  In the absence of such a distinction I think it's OK for the default to be recursive.  One can fake `let` by doing `var f = fun (...) {...}` if one wants to name a non-recursive function that refers to a previously defined function of the same name, if one really wants this.

I think here, all that changes is that previously two consecutive `fun` definitions would not be allowed at all, without intervening non-function bindings, and if separated by something else each one would have the same meaning as now.  So I'm OK with this implication because it does not change the behavior of existing working code.

core/irCheck.ml
```diff
@@ -210,10 +209,16 @@ let eq_types occurrence : type_eq_context -> (Types.datatype * Types.datatype) -
...
-                context, prev_equal && eq)
-              (context, Types.Abstype.equal s s') ts ts'
+             let (context, args_ok) = check_type_args context ts ts' in
+             (context, Types.Abstype.equal s s' && args_ok)
+         | _ -> (context, false)
+         end
+      | `RecursiveApplication a1 ->
+         let open Types in
+         begin match t2 with
+         | `RecursiveApplication a2->
```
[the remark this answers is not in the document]
jamescheney: Indeed, I was wondering how we are handling this now.  If all of the examples work with the current sound but incomplete check that's fine but we should create an issue recording this limitation (and fix it).  Relates to the later comment about typename's semantics I think.

core/typevarcheck.ml
```diff
@@ -71,6 +71,8 @@ let rec is_guarded : TypeVarSet.t -> int -> datatype -> bool =
         | `Application (_, ts) ->
             (* don't treat abstract type constructors as guards *)
             List.for_all (is_guarded_type_arg bound_vars var) ts
+        | `RecursiveApplication { r_args ; _ } ->
```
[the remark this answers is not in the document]
jamescheney: I am now somewhat confused about `typename` (see later comment).  We seem to be overriding it to sometimes mean "type synonym" (alias), and other times more like Haskell's "newtype" or "data".  I think that if we are now introducing generative nominal typing (i.e. ML `datatype` or Haskell `data` that introduces a new type name distinct from all others) --- which is perfectly reasonable as a way of simplifying recursive types! --- then the keywords should reflect this difference.

core/unify.ml
jamescheney: Relatedly to previous comments: When did we start treating recursive types/typename in this way?  Is it a change to behavior as of this PR only, or has this been different for a while (e.g. since introducing mutually recursive types) and this PR is just doing the same thing in a cleaner way?  It's not necessarily a problem, but seems like a nontrivial change - in particular it contradicts what the (old old old) documentation says about `typename`, so if we adopt this PR I request we update the documentation at least enough to reflect this.

jamescheney: (Incidentally, Frank's example above works in master if we leave out the `mutual` blocks; both types are translated to alpha-equivalent mu types.)

frank-emrich: Before this PR, the following was already valid:
```
typename T1(a) = [| Nil | Cons: (a,T1(a)) |];
```

However, this just introduced an alias for a \mu type before. Those aliases made writing complicated (recursive) types easier and were carried around to obtain nice error messages, but were completely irrelevant for type-checking; everything was typed structurally.
(My understanding of) the idea of this PR was to be a bit more clever about type-checking: If we carry around the type names for debugging purposes anyways, we can use them as a shortcut for type-checking: If we compare two types and both turn out to be named types with the same name and same type arguments, they must be the equal, and we don't have to do a structural type comparison. One prerequisite for doing so is to give named types proper identifiers (as done in this PR), whereas previously we just used strings with no consideration for scoping, shadowing, etc.


However, my understanding was that we would still keep the structural semantics: If we compare named types and the names don't match or we compare a named type to something different, we would fall back to a structural equality check. This way, the type system would stay structural, but the fact that some types have names would be used in a more disciplined way to increase performance of the type-checker.

My comment here shows that what I thought the new behavior should be (fall back to structural typing) is not what actually happens in this PR. I don't know if this is a bug or if my understanding of how to combine structural and (pseudo-)nominal typing is different from what Simon had in mind.

When adopting my understanding, `newtype` never actually creates new types, just aliases (which we may then use cleverly).

dhil: > However, my understanding was that we would still keep the structural semantics: If we compare named types and the names don't match or we compare a named type to something different [...]

Yes, this is one of the things we decided to try a year ago (or something like that) when we discussed how speed up the type checking in wake of #359. However, I thought that would be a separate PR on top of this, whose only goal I thought was to make recursion explicit in the syntax.

dhil:
> Thanks, everyone, for your comments so far: I have incorporated them and certainly think that they have improved the PR.
> 
> The largest change has been eliminating the `Mutual` node from `sugartypes` altogether by following @dhil's suggestion to move the work to the parser. Amusingly, I accidentally fixed #278 in the process.

OK. I didn't actually suggest anything, I only meant to outline an alternative design. As I said my comments were just that, comments. But getting rid of `Mutual` is interesting, and going somewhat further than I had in mind. I don't mind either design, to be honest. My priority would to get this PR into an acceptable state as soon as possible so it can be merged.

core/unify.ml
SimonJF: Thanks, @frank-emirch: your understanding is correct, and this is a bug. Thanks very much for catching it - I'll patch it and add a regression test.

core/sugarTraversals.ml
```diff
@@ -714,6 +698,28 @@ class map =
       fun p ->
         WithPos.map2 ~f_pos:o#position ~f_node:o#bindingnode p
 
+    method recursive_function : recursive_function -> recursive_function =
```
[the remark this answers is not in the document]
SimonJF: I thought I needed them yesterday when I was fixing a bug with the old module system. This turned out not to be necessary, so I'll revert this change.

core/unify.ml
SimonJF: Just to clarify: in my mind, the biggest issue fixed by this PR is to remove recursive type inlining (i.e., hacking recursive types using mu-types), which led to substantial problems in the programs I was writing in practice. The explicit mutual syntax is handy, and I think a good design decision.

core/typevarcheck.ml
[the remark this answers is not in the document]
jamescheney: Based on discussion below I think that we still intend that `typename` remains a transparent type synonym (like `type` in Haskell) for now, in which case what Frank wrote above should work (or at least have an informative error message explaining why it is a silly thing to do).

That is, if we are imposing a side-condition that the occurrences of recursively defined types in the RHSs of `typename`s inside of `mutual` blocks all need to be guarded by variant constructors, we should check for this somewhere and have an error message.  I'd be fine with this restriction if it makes life easier for now, though I think in Links things like
```
mu a. (a) -> Int

mu a. (a) -> a
```
are perfectly well defined using `mu` so if we can handle mutual recursion with arbitrary RHSs that would be fine too.

core/instantiate.ml
```diff
@@ -400,57 +412,28 @@ let replace_quantifiers t qs' =
...
-        failwith (Printf.sprintf "The type constructor: %s is abstract, not an alias" name)
+        internal_error (Printf.sprintf "Unrecognised type constructor: %s" name)
+    | Some (`Abstract _)
+    | Some (`Mutual _) ->
+        internal_error (Printf.sprintf "The type constructor: %s is not an alias" name)
     | Some (`Alias (vars, _)) when List.length vars <> List.length tyargs ->
-        failwith (Printf.sprintf
-                    "Type alias %s applied with incorrect arity (%d instead of %d)"
-                    name (List.length tyargs) (List.length vars))
+        internal_error (Printf.sprintf
```
jamescheney: This fixes #184 by the way.  Actually a lot of lingering error-message issues may be subissues of #525.

SimonJF: Hehe, splendid! I love accidentally-fixed issues.

core/errors.ml
```diff
@@ -10,11 +10,18 @@ type synerrspec = {filename : string; linespec : string;
...
+exception MultiplyDefinedMutualNames of ((Position.t list) stringmap)
+exception InvalidMutualBinding of Position.t
 exception RichSyntaxError of synerrspec
 exception SugarError of (Position.t * string)
 exception Runtime_error of string
 exception UnboundTyCon of (Position.t * string)
+exception InternalError of { filename: string; message: string }
+exception TypeApplicationArityMismatch of
+  { pos: Position.t; name: string; expected: int; provided: int}
+exception TypeApplicationKindMismatch of
```
jstolarek: Oh, a record. This is nice, though inconsistent with other errors that store just tuples. Would it be a good idea to refactor other errors to also store records `{ pos: Position.t; message: string }`?

SimonJF: I think so, yes. This points to a bigger issue, actually -- inline records are (relatively) recent in OCaml, and I think a lot of the nested tuples (for example, throughout the AST) may benefit from being redefined as inline records to make the source more descriptive.

core/sugartypes.ml
```diff
@@ -296,7 +292,17 @@ and cp_phrasenode =
   | CPLink        of Binder.with_pos * Binder.with_pos
   | CPComp        of Binder.with_pos * cp_phrase * cp_phrase
 and cp_phrase = cp_phrasenode WithPos.t
-                  [@@deriving show]
+and typename = (name * (quantifier * tyvar option) list * datatype' * Position.t)
+(* SJF: It would be nice to make these records at some point. *)
```
jstolarek: I suggest making a ticket also. I very much like the idea. In fact I really missed the record syntax and the ability to nicely lay out comments next to the fields. Here's how things look in GHC source code:

```Haskell
  | RecordCon
      { rcon_con_name :: Located id         -- The constructor name;
                                            --  not used after type checking
      , rcon_con_like :: PostTc id ConLike  -- The data constructor or pattern synonym
      , rcon_con_expr :: PostTcExpr         -- Instantiated constructor function
      , rcon_flds     :: HsRecordBinds id } -- The fields
```
I think it would be easier to understand Links source code if data types had these kind of documentation.

core/transformSugar.ml
```diff
@@ -7,6 +7,9 @@ open Utility
 
 module TyEnv = Env.String
 
+let internal_error message =
```
jstolarek: Yes, this is a recurring pattern. Please define it just once.

core/transformSugar.ml
```diff
@@ -691,7 +697,7 @@ class transform (env : Types.typing_environment) =
         (o, (pss, e), t)
 
     method handlerlit : Types.datatype -> handlerlit -> ('self_type * handlerlit * Types.datatype) =
-      fun _ _ -> failwith "transformSugar.ml: method handlerlit not yet implemented!" (*
+      fun _ _ -> internal_error "transformSugar.ml: method handlerlit not yet implemented!" (*
```
jstolarek: File name if it is already stored in `internal_error`, no need to repeat.

core/types.ml
```diff
@@ -131,7 +131,43 @@ type lens_phrase =
...
+(* Types *)
+and rec_appl = {
+  r_name: string;
+  r_dual: bool;
+  r_unique_name: string;
+  r_args: type_arg list;
+  r_unwind: type_arg list -> bool -> typ;
+  r_linear: unit -> bool option
+}
+and rec_unifier =
```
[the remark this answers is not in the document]
SimonJF: Thanks; I fixed this in bfd99594fa3235973e0ef7741db07fea83fef991.

core/unify.ml
SimonJF: Resolved in 4271aba98fb5c54cda9fd5f9daa2240e367b84ad.

core/typeSugar.ml
```diff
@@ -3851,9 +3855,16 @@ and type_binding : context -> binding -> binding * context * usagemap =
            (bind_var empty_context (Binder.to_name bndr, datatype)),
            StringMap.empty)
       | Foreign _ -> assert false
-      | Type (name, vars, (_, Some dt)) as t ->
-          t, bind_tycon empty_context (name, `Alias (List.map (snd ->- val_of) vars, dt)), StringMap.empty
-      | Type _ -> assert false
+      | Typenames ts ->
+          let env = List.fold_left (fun env (name, vars, (_, dt'), _) ->
+            begin
```
jstolarek: Redundant `begin ... end`?

core/transformSugar.ml
```diff
@@ -808,10 +814,17 @@ class transform (env : Types.typing_environment) =
       | Foreign (f, raw_name, language, file, t) ->
          let (o, f) = o#binder f in
          (o, Foreign (f, raw_name, language, file, t))
-      | Type (name, vars, (_, Some dt)) as e ->
-         let tycon_env = TyEnv.bind tycon_env (name, `Alias (List.map (snd ->- val_of) vars, dt)) in
-         {< tycon_env=tycon_env >}, e
-      | Type _ -> failwith "Unannotated type alias"
+      | Typenames ts ->
+          let (o, _) = listu o (fun o (name, vars, (x, dt'), pos) ->
+            begin
```
jstolarek: Redundant `begin ... end`?

core/types.ml
```diff
@@ -2170,19 +2288,21 @@ struct
...
-                 TypeVarSet.add (var_of_quantifier tyvar) bound_vars)
-              bound_vars tyvars
-          in
-            begin
-              match tyvars with
-                | [] -> datatype bound_vars p body
-                | _ -> mapstrcat "," (quantifier p) tyvars ^"."^ datatype bound_vars p body
-            end
+          let bvs = bound_vars tyvars in
+          begin
```
jstolarek: Redundant `begin ... end`?

core/liftRecursive.ml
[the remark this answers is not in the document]
SimonJF: Good observation. I don't think it's problematic enough to go with non-recursive functions by default (as I say, the ripple would be more like a tidal wave), but it's worth knowing about (and would be a good warning candidate). Thanks.

core/types.ml
```diff
@@ -2379,10 +2501,9 @@ let string_of_quantifier ?(policy=Print.default_policy) ?(refresh_tyvar_names=tr
 
 type environment       = datatype Env.t
 and tycon_environment  = tycon_spec Env.t
-and typing_environment = { var_env    : environment
-                         ; tycon_env  : tycon_environment
-                         ; effect_row : row }
-     [@@deriving show]
+and typing_environment = { var_env   : environment ;
```
jstolarek: I would align `:`

core/types.ml
```diff
@@ -2396,7 +2517,9 @@ let normalise_typing_environment env =
 let extend_typing_environment
     {var_env = l ; tycon_env = al ; effect_row = _  }
     {var_env = r ; tycon_env = ar ; effect_row = er } : typing_environment =
-  {var_env = Env.extend l r ; tycon_env = Env.extend al ar ; effect_row = er }
+  { var_env = Env.extend l r;
```
jstolarek: I would align `=`

core/unify.ml
```diff
@@ -95,7 +101,15 @@ let rec eq_types : (datatype * datatype) -> bool =
          end
       | `Application (s, ts) ->
           begin match unalias t2 with
-              `Application (s', ts') -> s = s' && List.for_all2 (Utility.curry eq_type_args) ts ts'
+              `Application (s', ts') ->
+                s = s' && List.for_all2 (Utility.curry eq_type_args) ts ts'
+            | _ -> false
+          end
+      | `RecursiveApplication a1 ->
+          begin match unalias t2 with
```
jstolarek: Redundant `begin ... end`?

core/unify.ml
```diff
@@ -95,7 +101,15 @@ let rec eq_types : (datatype * datatype) -> bool =
          end
       | `Application (s, ts) ->
           begin match unalias t2 with
```
jstolarek: Redundant `begin ... end`?

core/unify.ml
```diff
@@ -164,24 +178,33 @@ and eq_type_args =
...
 type quantifier_stack = int * int IntMap.t * int IntMap.t
 
 let compatible_quantifiers (lvar, rvar) (_, lenv, renv) =
   match IntMap.lookup lvar lenv, IntMap.lookup rvar renv with
     | Some ldepth, Some rdepth when ldepth=rdepth -> true
     | _ -> false
 
-type unify_env = {tenv: unify_type_env; renv: unify_row_env; qstack: quantifier_stack}
+type unify_env =
+  {tenv: unify_type_env;
```
jstolarek: I would:

```OCaml
  { tenv   : unify_type_env
  ; renv   : unify_row_env
  ; qstack : quantifier_stack
  }
```
(Related to #504)

core/unify.ml
```diff
@@ -1257,10 +1304,17 @@ and unify_type_args' : unify_env -> (type_arg * type_arg) -> unit =
      raise (Failure (`Msg ("Couldn't match "^ string_of_type_arg l ^" against "^ string_of_type_arg r)))
 
 let unify (t1, t2) =
-  unify' {tenv=IntMap.empty; renv=IntMap.empty; qstack=(0, IntMap.empty, IntMap.empty)} (t1, t2)
+  unify'
+    {tenv=RecIdMap.empty;
```
jstolarek: Formatting?

core/unify.ml
```diff
@@ -1257,10 +1304,17 @@ and unify_type_args' : unify_env -> (type_arg * type_arg) -> unit =
...
+     renv=IntMap.empty;
+     qstack=(0, IntMap.empty, IntMap.empty) } (t1, t2)
+
 (* Debug.if_set (show_unification) (fun () -> "Unified types: " ^ string_of_datatype t1) *)
 and unify_rows (row1, row2) =
-  unify_rows' {tenv=IntMap.empty; renv=IntMap.empty; qstack=(0, IntMap.empty, IntMap.empty)} (row1, row2)
+  unify_rows'
+    {tenv=RecIdMap.empty;
```
jstolarek: Formatting?

core/irCheck.ml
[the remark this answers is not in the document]
SimonJF: Good point. I've taken your suggestion in 10a58261a23dc491689ada3e39a3b0f6f81d5ceb -- is this what you were meaning?

core/typevarcheck.ml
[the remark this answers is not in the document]
SimonJF: Well-spotted. This appears to be a bug in `master` too, since I get the same error with:

```
typename Even(a) = (a) -> Int;

4
```

core/typevarcheck.ml
SimonJF: Would it suffice to copy the behaviour of `Recursive` `MetaTypeVar`s here? (Additionally, I think this also means that the definition of `is_guarded_type` and `is_positive_type` should also change accordingly).

Additionally, does anyone have any good literature references on positive / negative types? I feel I could benefit from reading up on this a bit more.

core/liftRecursive.ml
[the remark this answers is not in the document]
frank-emrich: okay, I'm perfectly fine with keeping things as they are

core/irCheck.ml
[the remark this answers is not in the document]
frank-emrich: yep, that looks good

core/typevarcheck.ml
frank-emrich: Yes mimicking the current behavior for a `Recursive` `MetaTypeVar` should do the trick

core/typevarcheck.ml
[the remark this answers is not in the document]
dhil: There is a free variable in your program, namely the implicit flexible effect variable.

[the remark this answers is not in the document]
SimonJF: Thanks, @dhil. I have been caught many a time by this sneaky implicit variable.

@frank-emrich -- thinking about your example, this will never yield a `RecursiveApplication` since it's not self-referential.

Perhaps a better example would be:
```
  typename Even(a) = Even(a) -> Int;
```

where `Even` occurs unguarded. I think extending the current guardedness checker, following how it's done for recursive type variables, and adding a guardedness check for each entry in a `Typenames` block would be the best idea.

core/liftRecursive.ml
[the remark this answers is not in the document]
jstolarek: In `typeSugar.ml` when typechecking `Fun` there is a hack:

```
(* HACK: Place a dummy name in the environment in
   order to ensure that the generalisation check
   does the right thing (it would be unsound to use
   the original name as the function is not
   recursive) *)
let v = Utils.dummy_source_name () in
bind_var context (v, fti), ft
```
It is still necessary given that the functions are recursive? I tried using `bind_var context (Binder.to_name bndr, fti), ft` instead and tests don't break.

core/errors.ml
```diff
@@ -62,12 +69,29 @@ let format_exception =
           "%s:%d: Syntax Error: Duplicate name `%s' in pattern\n  %s\nIn expression: %s"
           pos.pos_fname pos.pos_lnum name (xml_escape pattern) (xml_escape expr)
   | Failure msg -> "*** Fatal error : " ^ msg
-  | MultiplyDefinedToplevelNames duplicates ->
-      "Duplicate top-level bindings\n" ^
+  | MultiplyDefinedMutualNames duplicates ->
+      "*** Error: Duplicate mutually-defined bindings\n" ^
```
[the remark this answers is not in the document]
SimonJF: I don't think there is, but I think it would make sense for all to have the prefix.

core/instantiate.ml
```diff
@@ -2,6 +2,11 @@ open CommonTypes
 open Utility
 open Types
 
+exception ArityMismatch of (int * int)
+
+let internal_error message =
```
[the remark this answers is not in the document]
SimonJF: Excellent idea.

core/errors.ml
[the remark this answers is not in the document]
SimonJF: Ah, I see what it is now. If the error appears on a particular line number, then the error is prefixed with the position. If not (or, in this case, there are multiple positions to report), we have the prefix.

That said, `RichSyntaxError` and `SugarError` do not follow this pattern: I think I will streamline this, but in another patch.

core/types.ml
SimonJF: This one isn't redundant, as without the `begin ... end`, the `Mutual` and `Abstract` cases are parsed as being part of the inner `match`.

core/unify.ml
SimonJF: This one isn't redundant, since the next expression is another case branch.

core/unify.ml
SimonJF: Same.

core/sugartypes.ml
SimonJF: Thanks -- I've made a ticket: #533

core/instantiate.ml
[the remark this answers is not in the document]
SimonJF: I ended up refactoring this so that the exception wasn't constructed directly in each file, but one recurring thing I noticed was that I had to specify the same filename multiple times in a single file, which felt wrong. 

So each file needing `internal_error` still has it defined, which calls `internal_error` in `errors.ml`.

core/errors.ml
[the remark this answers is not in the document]
jstolarek: Yes, this is something to think about separately. Perhaps as part of #525?

core/instantiate.ml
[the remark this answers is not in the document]
jstolarek: Yes, that sounds good.

core/liftRecursive.ml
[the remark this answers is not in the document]
jstolarek: I'd like to point out to the above question to make sure it doesn't go unnoticed.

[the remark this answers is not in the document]
SimonJF: Sorry, I missed this. After the `liftFunctions` pass, all recursive functions are promoted to singleton `Funs` blocks, so we can assume that all `Fun` functions are explicitly non-recursive.

I wouldn't be able to give you a specific answer to your question (maybe @dhil, @frank-emrich, @jamescheney, or @slindley would be more familiar with this part of the typechecker?)

core/typevarcheck.ml
SimonJF: This should now be resolved in 15db7100b0015db7138f7afca86ea5a52ffa0443.

core/typevarcheck.ml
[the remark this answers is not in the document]
SimonJF: I've looked into this, and I don't believe we need to handle this in any special way. As @dhil clarified, the error that @frank-emrich came across was because of the sneaky implicit effect variable. I can confirm that

```
typename Even(a) = (Even(a)) {}-> Int;
```
is valid.

The possible soundness bugs due to unguardedness are checked in `tests/mutual/unguarded.links` and `tests/mutual/unguarded2.links`.

In 15db7100b0015db7138f7afca86ea5a52ffa0443, I fixed the guardedness and positivity checkers. Since the only place this is invoked is when introducing a (mu-bound) recursive type during type inference (`unify.ml`, `rec_intro_point`), I don't think we need to do anything special here.

I may be missing something though, so I'd appreciate any comments.

SimonJF:
Thanks again, everyone, for your careful reviews. I hope that everything is addressed now.

@frank-emrich: could you please confirm that my fixes raise the typing concerns you raised?

core/liftRecursive.ml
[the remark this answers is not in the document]
frank-emrich: In the HM generalization step, you need the current variable environment because you don't want to generalize type variables that already occur in the variable environment (i.e., in the type of any of the variables bound in the current context).
I'm quite sure that this is why we add a dummy variable to the context here, whose type is the type of the current function. However, I don't understand the exact reason why this is needed here...
In any case, I think that this PR doesn't change anything this piece of code relies on, in particular since `Fun` and `Funs` should be used in the same way as they were before.

frank-emrich:
Your changes look good to me :)

SimonJF:
Wonderful, thanks, @frank-emrich!

@jamescheney, have the revisions to the PR addressed the requests in your review? If so, can we go ahead and merge?

jamescheney:
I think so - aside from the transparency/opaqueness of recursive type names they were mostly cosmetic

SimonJF:
Amazing. I'm delighted this is being merged.

Thanks everyone -- this is a real triumph of code review.

PR #532, opened 2019-03-22
Require recent ppx_deriving_yojson version?

frank-emrich:
Janek realized that we don't need to suppress warning 39 (about unused rec flags) anymore (see 381fdee26314cc4ae). The warning originated from code generated by earlier versions of ppx_deriving_yojson. Since version 3.3, the generated code doesn't seem to yield those warnings any more. I've therefore un-suppressed those warnings in two additional files.

I'm creating this PR because I'm unsure how we deal with versioning of our dependencies. After my and Janek's change, building a release version of links (or installing a potential future release via opam) would error if one has a pre-3.3 version of ppx_deriving_yojson installed. I therefore added a version constraint to our opam file. Or do we generally give up on tracking the exact version requirements of our dependencies and just hope that users have recent versions installed?

### DIFF
--- a/links.opam
+++ b/links.opam
@@ -18,7 +18,7 @@ depends: [
   "ocaml" {>= "4.06.0"}
   "dune" {build}
   "ppx_deriving"
-  "ppx_deriving_yojson"
+  "ppx_deriving_yojson" {> "3.1"}
   "base64"
   "linenoise"
   "ANSITerminal"
--- a/links.opam
+++ b/links.opam
@@ -18,7 +18,7 @@ depends: [
   "ocaml" {>= "4.06.0"}
   "dune" {build}
   "ppx_deriving"
-  "ppx_deriving_yojson" {> "3.1"}
+  "ppx_deriving_yojson" {>= "3.3"}
   "base64"
   "linenoise"
   "ANSITerminal"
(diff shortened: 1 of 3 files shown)


links.opam
```diff
@@ -18,7 +18,7 @@ depends: [
   "ocaml" {>= "4.06.0"}
   "dune" {build}
   "ppx_deriving"
-  "ppx_deriving_yojson"
+  "ppx_deriving_yojson" {> "3.1"}
```
dhil: I'd pick `{>= "3.3"}` which is the current release. It should work with `ocaml >= 4.04`, which is fine since we require at least `4.06`.

frank-emrich: there is no version 3.2, so those two constraints are equivalent. I have no particular preference about expressing it with > or >=. Do you?

jamescheney:
This makes sense.  I was assuming that the default versions of yojson and ppx_deriving_yojson that were causing the problem was the most recent ones (i.e. 1.6.0 / 3.3) but it appears that ppx_deriving_yojson 3.3. avoids this problem by requiring yojson <= 1.5.0.  So requiring ppx_deriving_yojson >= 3.3 seems fine to me.

links.opam
jamescheney: I would go with the more specific constraint >= 3.3 just in case our assumption that version numbers are monotonic over time turns out to be optimistic.  (and it better documents that 3.3 is known to work without relying on context.)

Issue #446, opened 2018-12-30
IR typing: Miscellanea


jamescheney:
This issue collects some simple examples that illustrate ill-typed IR being generated:

The (anonymous identity function or a constant function:
```
fun (x) {x};
fun (x) {1};
```
both result in IR with unbound type variables.

```
for (x <- []) []
```
yields an "inorrect type annotation on a (unused) variable" error.  

There is also
```
fun () {query {[]}};
```
has  an "incompatible effects" error.

The anonymous composition function:
```
fun (f,g) {fun (x){f (g(x))}};
```
produces several unbound type variables, plus a "Providing closure to untracked function" error.

```
fun (x) {<a>{x}</a>}
```
produces a type mismatch between `[XmlItem]` and `Xml`, which may indicate a limitation in the IR typing rather than wrong behavior, since `Xml` is supposed to be a type alias for `[XmlItem]`.

frank-emrich:
The unbound type variable problems for the first two functions sound like instances of the problem originally reported here:
https://github.com/links-lang/links/issues/377

In short, when generating a polymorphic IR function, the type variable used for the type abstraction (e.g., /\ X. ....) is different from the type variable that is used in the type annotation (\forall Y. ....) for that function. This seems to be by design and shouldn't be a problem in general. However, the frontend mixes those two variables up occasionally. This means that the variable from the type annotation (Y) is suddenly used inside of terms (e.g., by type applications), where X should be used instead. This results in the IR typechecker (rightfully) reporting unbound type variables.

This problem occurs even in the simplest examples. However, I'm hestitant to touch the corresponding frontend code since I don't feel like I'm understanding all the brittle invariants sufficiently.

jamescheney:
In the simplest cases (identity/constant functions) I'm seeing compiled IR like this:
```
([`Fun (((2113, ((_) -> Int, "_fun__g142", `Local)),
         ([], [(2112, (_, "", `Local))],
          ([`Let (((2111, (_, "x", `Local)), ([], `Return (`Variable (2112)))))
             ],
           `Return (`Constant (`Int (1))))),
         None, `Unknown))
   ],
 `Return (`TAbs (([a; a::Row(Any,Any)], `Variable (2113)))))
```

One problem is that the type printing is printing out user-friendly type syntax and hiding the actual type variable IDs.  However, it also looks like the expression generated starts with a function binding, and the end result is a type abstraction expression that binds the type variables over the function name. 

[...]

Issue #256, opened 2017-09-18
Xml and XmlItem are printed differently


elordin:
When running links in web-mode:
```
links> hd(<div id="myId" class="a-class" />);
<div id="myId"class="a-class"/> : XmlItem
```
```
links> <div id="myId" class="a-class" />;
<div id="myId" class="a-class" /> : Xml
```
Clearly the `XmlItem` version should have a space between attributes as well.

jamescheney:
Also:

```
links> hd(<foo a="1" b="2"><bar a="3" b="4"><baz a="5" b="6"></baz></bar><bar a="42" b="17"></bar></foo>);
<foo b="2" a="1"><bar b="4" a="3"><baz b="6" a="5"/></bar><bar b="17"a="42"/></foo> : XmlItem
```

That is, things work right for `XmlItem`s that have content, hence their close tags must be printed, but wrong for `XmlItem`s that have attributes but no element content.

So perhaps the bug is conditionally guarded somewhere by a `close_tags` test.

jamescheney:
I suggest we defer this to 0.7.3 and release 0.7.2 ASAP.

jamescheney:
I think fixing this should be a part of the general list/xml cleanup that will hopefully fall out naturally if we make lists and XML into user-defined types #46, so happy to kick to >0.7.3.

jamescheney:
I've looked into this and it seems to be due to some weird stateful behavior regarding how `pp_print_space` hints are treated by the `Format` library.  For some reason if an `Xml = [XmlItem]` is printed by printing a list using the general value printer the right thing happens, but if the `p_xmlitem` printer is called directly on a singleton `XmlItem` then certain space hints are not printed as spaces.  

The reason spaces seem to be being dropped between attributes is that zero-width space hints are being generated, and the code in `Value.string_of_pretty` is setting the buffer formatter's `out_spaces` function to print zero spaces if there is a zero-width hint, or one space (instead of many) otherwise.  This code is also only being called because the `print_pretty` setting tested in  `Repl.print_value` is false.  If we set `print_pretty` to true, the problem seems to go away, but probably we still want XML to be printed correctly even if pretty-printing is disabled.  (And I'm not sure why pretty-printing is disabled by default in the first place.)

jamescheney:
The problem is that in the custom formatter defined in `string_of_pretty`, newlines were being ignored, so when for whatever reason the formatter was deciding to handle space hints by inserting newlines, no space was being inserted.

PR #534, opened 2019-03-26
Generate type applications for desugared anonymous functions

jstolarek:
This is a speculative fix for some of the IR typing problems reported in #446. Namely, this patch fixes unbound type variable errors for anonymous functions. These two examples now compile without IR errors:

```
fun (x) {x};
fun (x) {1};
```
Moreover, compilation of anonymous composition function `fun (f,g) {fun (x){f (g(x))}};` no longer generates these errors but still has some other errors.

The idea behind a fix is that if we see an anonymous function wrapped in a type abstraction we generate a corresponding type application. So previously we would generate:

```
TAbs Y. { 
  fun freshfun(x:Z) {x}
  freshfun
} 
```
whereas now we generate:
```
TAbs Y. { 
  fun freshfun(x:Z) {x}
  TApp freshfun [Y]
}
```

There is some voodoo coding here, so please review carefully. If this fix is correct then I plan on also updating the comments in `desugarFuns.ml`, because at least some of them are inaccurate.

Now, how would I write a test for this patch? Should I create a custom config file to enable IR typechecking or should I do it some other way?

### DIFF
--- a/core/desugarFuns.ml
+++ b/core/desugarFuns.ml
@@ -5,7 +5,12 @@ open SourceCode.WithPos
 open Sugartypes
 open SugarConstructors.DummyPositions
 
-(*
+(* This module desugars multi-argument functions, anonymous functions, and
+   operator sections.
+
+  1. Multi-argument functions are turned into a series of nested one-argument
+     functions:
+
     fun f[qs](xs1)...(xsk) {e}
   -->
     fun f[qs](xs1) {
@@ -20,16 +25,29 @@ open SugarConstructors.DummyPositions
       f2
     }
 
+  2. Anonymous functions are named and then turned into a series of nested
+     one-argument functions:
+
 
     fun [qs](xs1)...(xsk) {e}
   -->
-    {fun f[qs](xs1)...(xsk) {e};
-     f}
+    {fun f[qs](xs1) {
+       fun f1[](xs2) {
+       ...
+         fun fk[](xsk) {
+           e
+         }; fk
+       ...
+       }; f1
+    }; f}
 
+
+  3. Operator sections are turned into named functions:
 
     (.l)
   -->
-    fun (r) {r.l}
+    { fun f(x:r) {x.l}; f }
+
 *)
 
 
@@ -67,19 +85,38 @@ class desugar_funs env =
 object (o : 'self_type)
   inherit (TransformSugar.transform env) as super
 
+  method private desugarFunLit argss lin lam location tvs =
+    let inner_mb     = snd (last argss) in
+    let (o, lam, rt) = o#funlit inner_mb lam in
+    let ft = List.fold_right (fun (args, mb) rt -> `Function (args, mb, rt))
+                             argss rt in
+    let f = gensym ~prefix:"_fun_" () in
+    let body, tvs, ft = match tvs with
+      | [] -> Var f, [], ft
+      | _  ->
+         (* If a FunLit is surrounded by a type abstraction we need to generate
+            a corresponding type application.  Type abstractions are inserted
+            during type checking.  This really shouldn't happen since it couples
+            typechecking with this desugaring stage.  If we didn't insert type
+            application here the resulting AST would be ill-typed. *)
+         let (tvs, tyargs), ft = Generalise.generalise env.Types.var_env ft in
+         let ft = Instantiate.freshen_quantifiers ft in
+         tappl (Var f, tyargs), tvs, ft in
+    let (bndr, lin, tvs, loc, ty) =
+      unwrap_def (binder ~ty:ft f, lin, (tvs, lam), location, None) in
+    let e = block_node ([with_dummy_pos (Fun (bndr, lin, tvs, loc, ty))],
+                         with_dummy_pos body)
+    in (o, e, ft)
+
   method! phrasenode : Sugartypes.phrasenode -> ('self_type * Sugartypes.phrasenode * Types.datatype) = function
+    | TAbstr (tvs', {node = TAppl ({node = TAbstr (tvs, {node =
+         FunLit (Some argss, lin, lam, location); _ } ); _}, tyargs); _})
+      when Settings.get_value Instantiate.quantified_instantiation ->
+       let (o, e, ft) = o#desugarFunLit argss lin lam location !tvs in
+       (o, TAbstr (tvs', with_dummy_pos (TAppl (with_dummy_pos (
+            TAbstr (tvs, with_dummy_pos e)), tyargs))), ft)
     | FunLit (Some argss, lin, lam, location) ->
-        let inner_mb = snd (try last argss with Invalid_argument s -> raise (Invalid_argument ("!"^s))) in
-        let (o, lam, rt) = o#funlit inner_mb lam in
-        let ft =
-          List.fold_right (fun (args, mb) rt -> `Function (args, mb, rt))
-                          argss rt in
-        let f = gensym ~prefix:"_fun_" () in
-        let (bndr, lin, tvs, loc, ty) =
-          unwrap_def (binder ~ty:ft f, lin, ([], lam), location, None) in
-        let e = block_node ([with_dummy_pos (Fun (bndr, lin, tvs, loc, ty))],
-                            var f)
-        in (o, e, ft)
+       o#desugarFunLit argss lin lam location []
     | Section (Section.Project name) ->
         let ab, a = Types.fresh_type_quantifier (lin_any, res_any) in
         let rhob, (fields, rho, _) = Types.fresh_row_quantifier (lin_any, res_any) in
--- a/core/transformSugar.ml
+++ b/core/transformSugar.ml
@@ -243,7 +243,7 @@ class transform (env : Types.typing_environment) =
       | Constant c -> let (o, c, t) = o#constant c in (o, Constant c, t)
       | Var var -> (o, Var var, o#lookup_type var)
       | FunLit (Some argss, lin, lam, location) ->
-          let inner_e = snd (try last argss with Invalid_argument s -> raise (Invalid_argument ("@" ^ s))) in
+          let inner_e = snd (last argss) in
           let (o, lam, rt) = o#funlit inner_e lam in
           let (o, t) =
             List.fold_right
--- a/tests/empty_prelude.links
+++ b/tests/empty_prelude.links
@@ -1,0 +1,5 @@
+# This is an empty prelude file for the purpose of running IR typing tests.
+# Since standard Links prelude does not currently produce a type-correct IR we
+# to use an empty one so that we only focus on IR type errors arising from tests
+# programs and not the prelude functions.
+()
--- a/tests/typed_ir.tests
+++ b/tests/typed_ir.tests
@@ -1,0 +1,15 @@
+Anonymous constant function
+fun (x) {1}
+stdout : fun : (_) -> Int
+
+Anonymous identity function
+fun (x) {x}
+stdout : fun : (a::Any) -> a::Any
+
+Anonymous composition function
+fun (f,g) {fun (x){f (g(x))}}
+stdout : fun : ((a::Any) -b-> c::Any, (d::Any) -b-> a::Any) -> (d::Any) -b-> c::Any
+
+XML typing
+fun (x) {<a>{x}</a>}
+stdout : fun : (Xml) -> Xml
--- a/tests/typed_ir.tests.config
+++ b/tests/typed_ir.tests.config
@@ -1,0 +1,3 @@
+typecheck_ir=true
+fail_on_ir_type_error=true
+prelude=tests/empty_prelude.links


core/desugarFuns.ml
```diff
@@ -67,19 +67,30 @@ class desugar_funs env =
...
+    let inner_mb     = snd (last argss) in
+    let (o, lam, rt) = o#funlit inner_mb lam in
+    let ft = List.fold_right (fun (args, mb) rt -> `Function (args, mb, rt))
+                             argss rt in
+    let f = gensym ~prefix:"_fun_" () in
+    let body, tvs, ft = match tvs with
+      | [] -> Var f, [], ft
+      | _  ->
+         let (tvs, tyargs), ft = Generalise.generalise env.Types.var_env ft in
+         let ft = Instantiate.freshen_quantifiers ft in
```
jstolarek: Voodoo coding here. Do I need to call `freshen_quantifiers`? It seems to be called right after `generalise` in other places in the code so I inserted it here as well. But is it always required? (And if yes, why isn't it folded into `generalise`?)

jamescheney: My guess would be: necessary if you (or whatever code eventually receives `ft`) are planning on substituting something with free variables into `ft` inside the scope of a binder, but shouldn't hurt either way.  (Alternatively perhaps whatever does the substitution should do this.)

core/desugarFuns.ml
```diff
@@ -67,19 +67,30 @@ class desugar_funs env =
...
+         tappl (Var f, tyargs), tvs, ft in
+    let (bndr, lin, tvs, loc, ty) =
+      unwrap_def (binder ~ty:ft f, lin, (tvs, lam), location, None) in
+    let e = block_node ([with_dummy_pos (Fun (bndr, lin, tvs, loc, ty))],
+                         with_dummy_pos body)
+    in (o, e, ft)
+
   method! phrasenode : Sugartypes.phrasenode -> ('self_type * Sugartypes.phrasenode * Types.datatype) = function
+    | TAbstr (tvs , {node = FunLit (Some argss, lin, lam, location); _ }) ->
```
jamescheney: Can `TAbstr`s be nested?  (Or are we completely certain that they won't ever be?)

I can imagine that in an arbitrary `Sugartypes.phrasenode` that there might be an outer `TAbstr`, then something else, then another `TAbstr`, then a `FunLit` (maybe with more other stuff in between).  So it seems like in the general case we should keep track of all of the (distinct?) `tvs` encountered so far and pass them all into `o#desgarFunLit` when we encounter a `FunLit`.  Otherwise we will get mysterious match failures and/or missing type variables.

jstolarek: The intention here is to catch the `TAbstr` inserted by typeSugar, because that `TAbstr` contains tyvars that are abstract in the `FunLit`. Here's the relevant code from `typeSugar.ml`:

```OCaml
let e = FunLit (Some argss, lin, (List.map (List.map erase_pat) pats, erase body), location) in
  if Settings.get_value Instantiate.quantified_instantiation then
    let (qs, _tyargs), ftype = Utils.generalise context.var_env ftype in
    let _, ftype = Instantiate.typ ftype in
      tabstr (qs, e), ftype, StringMap.filter (fun v _ -> not (List.mem v vs)) (usages body)
  else
    e, ftype, StringMap.filter (fun v _ -> not (List.mem v vs)) (usages body)
```

jamescheney: Hmm.  So I guess this relates to what we just discussed regarding whether the closing should happen in `typeSugar` or a separate desugaring pass.  Now that I've looked at this I lean towards the former - would that be possible or is there some obstacle?  It would be a good invariant that after `typeSugar` does its thing, the resulting AST is still well-typed.

jstolarek: To be able to do this in typeSugar we would have to desugar anonymous functions there, because type application needs to go inside the block currently generated in desugarFuns. I agree that "resulting AST is still well-typed" is a perfectly reasonable invariant for a typechecker :-) But I think that in this case it is typeSugar that does more than it actually should, i.e. it should just type check the existing AST, rather than also transform it.

core/desugarFuns.ml
frank-emrich: I think this is necessary here. `Generalise.generalise` binds type variables without changing their names. 
In your example
```
TAbs Y. { 
  fun freshfun(x:Z) {x}
  TApp freshfun [Y]
}
```
Here, `freshfun` must not have type `forall Y. Y -> Y`, because then we would have two binding occurences for type variable `Y`, which violates an invariant of the IR (unique names). Instead, it must freshen the quantifiers (e.g., to `Z`, as you did in the example).

jstolarek:
@jamescheney, does my reply address your concerns? Do you have objections to merging (modulo tests/documentation)?

Also, @frank-emrich has kindly fixed the IR typing error in the anonymous composition function (more generally: in any nested anonymous functions).

jamescheney:
OK.  Could we merge after
- adding a comment documenting the reason for the (i.e. that `typeSugar` does too much / too little)?
- add checkboxes to the bottom of the issue description of #446, and check off the ones that are fixed here?  (so we can keep track of which problems are still to be fixed)?

jstolarek:
> adding a comment documenting the reason for the

OK. I've added some comments already but didn't mention that typeSugar should be doing less. Will amend.

> add checkboxes to the bottom of the issue description of #446

I have a plan on replacing #446, which I think is way too general to be useful for tracking what is done and what is not, and replacing it with separate bug reports for different types of IR typing errors that we see.

jstolarek:
@frank-emrich, I've run into problems with writing tests. I've added the tests in a way we discussed yesterday. The problem is that test harness tests for standard output and exit code. However, IR typing errors are now currently printed on the stderr and do not cause compilation failure. Thus the tests always pass, even in the presence of IR typing errors. That being said, how do you feel about turning `fail_on_ir_type_error` into a setting now? I can't think of other way of fixing the tests.

jstolarek:
@frank-emrich, see also #539

frank-emrich:
> how do you feel about turning `fail_on_ir_type_error` into a setting now?

Yes, that sounds like exactly the right thing to do now.


In general, I'm a bit unsure about this PR. It's doing the right thing, but it heavily relies on knowing exactly what `typesugar.ml` does. In particular, `TAbs` are not the only term form that binds type variables, as our binding forms (e.g., `Fun` and `Val`) have builtin type abstractions.

However, in order to do a more general solution, we would need to use the `Instantiate` and `Generalise` modules. I'm not sure if it's a good idea to do that outside of `typesugar`.

So in total, I think that this PR gets the job done, but couples `desugarFuns` tightly with `typesugar`. That may already be the case for many other desugarings. If we want to prevent this, however, we should consider moving this part of `desugarFuns` into `typesugar`, which seems paradox if it is the result of trying to make things more independent from each other.

I'll spend some more time experimenting with this today.

jstolarek:
Ok, so I turned `fail_on_ir_type_error` into a setting and I still can't get the tests to work. I added one test that should be failing (XML typing - not yet fixed) and it passes. I am doing something wrong with the testsuite here but can't see what.

>  we should consider moving this part of desugarFuns into typesugar

Why not the other way around? Why not insert TAbs in desugarFuns? This is an element of the desugaring, and it is meant to be a typed translation, so I think it is OK that it operates on types.

jamescheney:
@jstolarek Sounds good for now

@frank-emrich I think this is in general an argument in favor of defining and (re?)rationalizing what `typeSugar` and `desugarFuns` (and all of the desugaring stages!) do or should do.  It sounds like `typeSugar` is not simply a traversal that adds type information but also does some desugaring/transformation which leaves technical debt that `desugarFuns` is now paying off.  However, I'm not sure yet that we know enough about what should be happening to make IR typechecking happy besides the simple issues enumerated in #446, especially since `desugarFuns` is the very last desugaring stage and presumably can't be done earlier than others that introduce functions (I think?).

I would like some subset of us to have a better handle on that before we commit to reorganizing.

jstolarek:
@jamescheney, I expanded the comments as you requested. I'd say let's fix the tests (@frank-emrich - would you be able to take a look?) and merge for now. I'll then prepare a more detailed list of IR typing errors.

frank-emrich:
@jamescheney 
I am not aware of major rewrite steps that `typeSugar` does, except inserting type application and abstraction terms. But this may just the result of me not being overly familiar with `typeSugar`. If there are currently no other desugarings happening in `typeSugar`, we should probably keep it that way. I'm mostly worried about the potentially unforeseen effects of generalizing and instantiating things outside of `typesugar`. 

@jstolarek 
> Why not the other way around? Why not insert TAbs in desugarFuns? 

It looks to me as if the type `ftype` that results from generalizing the type of a `FunLit` is returned by `typeSugar` when it handles a `FunLit`. I'm not sure if removing this part from `typeSugar` and moving it to `desugarFuns` will break things in `typeSugar`.

If this blocks you, @jstolarek, I'm happy with merging this now. In any case, I'll have another look at it and see if I can convince myself that this is definitely fine or if I find problems.

frank-emrich:
Regarding the tests: Yes, I'll do that now.

frank-emrich:
The problem is that `test-harness`
1. Ignores the second argument that is passed to it, which is supposed to be the config file
2. Ignores the `args : ...` line in a test definition when running an inlined test program, which makes it ignore the config file you wanted it to use.

I'll try to fix this now.

jamescheney:
@jstolarek Sounds good.  

I had a look at `typeSugar` and `desugarFuns` and my understanding is:
- `typeSugar typechecks and abstracts functions `FunLit`, `Fun`, `Funs` based on the free/unconstrained type variables of the function body.
- This is the right place for this work to be done, since if we wanted to do it later we'd have to re-do type inference/generalization
- For `Fun` and `Funs`, the abstracted variables are reflected both in the function type(s) and in the updated declaration, because `Fun` and `Funs` have a `tyvar list` (initially empty)
- For `FunLit`, there is no place to put a `tyvar list` in the expression to indicate what type variables are abstracted, so instead a single `TAbstr` node is inserted which we now handle in `desugarFuns`
- The real job of `desugarFuns` is to translate functions that have several argument lists into nested (anonymous) functions.  These will necessarily have type abstractions only at the outer level where generlaization took place in `typeSugar`.

This is overall somewhat unsatisfying.  Two better strategies might be:

1. add a `tyvar list` to `FunLit`, which is initially empty and filled in during `typeSugar`, so that we can handle the abstracted type variables the same way as `Fun` and `Funs`, i.e. without a `TAbstr`.
2. or: at some early stage, translate `FunLit`s to locally declared, named functions, so that they get handled automatically and correctly by the existing `Fun` code.  (This happens eventually anyway due to closure conversion.)
3. or: disaggregate the notion of type abstraction from function literals/declarations entirely.  This feels like a better/more orthogonal solution but would entail a larger refactor of many things whose ramifications are hard to predict.

Would it be possible to try out approach 1 or 2 above before merging?  If it isn't straightforward then happy with status quo for now.

frank-emrich:
@jstolarek 
I pushed a hotfix that makes the test suite do the right thing. The CI will now break (as intended) because of your XML test case. I'll create an issue for fixing `test-harness` properly, but my hotfix is merge-able.

frank-emrich:
@jamescheney 
Regarding your second proposed strategy:
It is my understanding that the setting `quantified_instantiation` determines (among other things) whether we type lambdas/`FunLit`s polymorphically (unlike standard HM) or not. 

If we re-wrote `FunLit`s  to normal, named functions prior to type-checking, we would always attempt to generalize them, which would not be the expected behavior if `quantified_instantiation` is `false`. I'm not sure if we could generate a clever type annotation in this case that would force the named function to be typed monomorphically. Otherwise, I'm afraid this strategy is not viable.

jamescheney:
@jstolarek Ah, good point.  I had the impression that `quantified_instantiation` is basically the default now so the cases that handle when it is disabled are basically dead code / not being tested.  If we wanted to preserve this behavior we could make it conditional on `quantified_instantiation` whether `FunLit` is desugared, I suppose.  But I sort of think we should pick one design and there are other sensible reasons to have `FunLit` separate from `Fun`.  

So what about option 1?

frank-emrich:
> So what about option 1?

Yes, that seems worth a try. Janek actually suggested doing this earlier, but we didn't follow through with it because it's not what HM does. Now that it turns out that we're *deliberately* not doing standard HM tying for lambdas (with default settings), it looks like a good idea.

jamescheney:
Indeed, though I am still uncertain as to whether that behavior is really intentional - i.e. does anything break if we don't generalize `FunLit`s' abstractable type variables?

frank-emrich:
> Does anything break if we don't generalize FunLits' abstractable type variables?

In the case where we don't want to generalize them we would have to make sure that they are all `Flexible. I'm not sure if that's already guaranteed to be the case.

frank-emrich:
I've fixed the issues with `test-harness` properly in #541, just discard the changes made to that file in this branch when rebasing this.

jstolarek:
@jamescheney, just like @frank-emrich  says option (1) was my original idea of fixing this bug but there was this "this is not what HM does" argument against it.

So I just tried to implement approach (1) and failed. If I do an IR typecheck of an anonymous constant function I get an exception:

```
# Type variables (total 0)


# Type arguments (total 2)
1. _
2. { |_::Any }
Raised at file "stdlib.ml", line 33, characters 17-33
Called from file "core/irCheck.ml", line 998, characters 24-84
Called from file "core/irCheck.ml", line 892, characters 28-37
Called from file "list.ml", line 117, characters 24-34
Called from file "core/irCheck.ml", line 890, characters 10-161
Called from file "core/irCheck.ml", line 901, characters 20-30
Called from file "core/irCheck.ml", line 1097, characters 18-45
Called from file "list.ml", line 117, characters 24-34
Called from file "list.ml", line 117, characters 24-34
Called from file "core/backend.ml" (inlined), line 91, characters 2-61
Called from file "bin/driver.ml", line 34, characters 4-64
Called from file "camlinternalLazy.ml", line 27, characters 17-27
Re-raised at file "camlinternalLazy.ml", line 34, characters 4-11
Called from file "bin/driver.ml", line 69, characters 20-75
Called from file "camlinternalLazy.ml", line 27, characters 17-27
Re-raised at file "camlinternalLazy.ml", line 34, characters 4-11
Called from file "camlinternalLazy.ml", line 27, characters 17-27
Re-raised at file "camlinternalLazy.ml", line 34, characters 4-11
*** Fatal error : Unknown exception during IR type-checking: Links_core.Instantiate.ArityMismatch
```
And since it doesn't even give me a syntax tree I can't really figure out what goes wrong. Perhaps there is something obviously wrong that I am doing here?

frank-emrich:
What program exactly are you testing? The program `fun (x) {1}` IR type-checks for me on this branch

jamescheney:
I don't have particular insight into the type error.  

From looking at the last commit, I would expect to see changes somewhere to handle freshening / free variables correctly for `FunLit`s with type variables - basically, grep for `FunLit` / `Fun` / `Funs` and see what utility functions there already are that deal with binding/scope and make sure `FunLit` does the analogous right thing.  

For example, maybe somewhere we are calling the `funlit` smart constructor with its default arguments for `tyvars`, resulting in an empty list.

frank-emrich:
@jstolarek 
I'm improving the IR error handling a bit at the moment. Just making sure we're not doing the same thing right now...

jstolarek:
@frank-emrich I am testing `js-typed-ir-hacking`. Compiling `fun (x) {1}` with a following config:

```
typecheck_ir=true
prelude=empty.links
print_types_pretty=false
debug=true
```
gives me:
```
[killy@wau : /dane/projekty/links] make && ./links --config=../sandbox/links/T446.config ../sandbox/links/T446.links
dune build --only-packages links,links-postgresql,links-sqlite3 --build-dir=./_build --profile=development @install
ln -fs links linx
registering driver for null
compiling to IR
compiled IR
Finishing process MAIN
compiling to IR
compiled IR
# Type variables (total 0)


# Type arguments (total 2)
1. _
2. { |_::Any }
*** Fatal error : Unknown exception during IR type-checking: Links_core.Instantiate.ArityMismatch
```

> I would expect to see changes somewhere to handle freshening / free variables correctly for FunLits with type variables

There is code *intended* to do this in `typeSugar.ml` at line 2337 (all line numbers given for `js-typed-ir-hacking` branch):

```OCaml
            let tyvars, ftype =
              if Settings.get_value Instantiate.quantified_instantiation then
                let (tyvars, _tyargs), ftype =
                  Utils.generalise context.var_env ftype in
                let _, ftype = Instantiate.typ ftype in
                (tyvars, ftype)
              else
                [], ftype in
```

> make sure FunLit does the analogous right thing.

Yes, that's what I tried to do. More precisely I mirrored treatment of `Fun` on line 3698 in `typeSugar.ml`

jamescheney:
OK, thanks.  I had a look over the code to see where FunLit and fun_lit are used and didn't see any obvious problems, it may be necessary to inspect the ASTs to see where the type abstraction / tyvar lists are getting dropped.  If this is too much of a hassle now I'm happy with reverting to the version that worked and leaving it as an issue to figure out how to do it more cleanly - since there might be other things we need to keep in mind to fix other problems that lead to ill-typed IR.

jstolarek:
Warning: I just did an upstream rebase.

jstolarek:
I've been looking at this from the morning and I am out of ideas. I plan to merge after #541 gets merged so that I can just drop the workaround @frank-emrich implemented for my tests. In the meantime if anyone wants to take a look feel free to do so.

frank-emrich:
I discovered what the problem is: You're adding two variables to the`tyvars' list of the function, but you're not quantifiying over those variables in the function's type:

This is the IR code for the example, with the following additional settings (just affecting printing, nothing else):
```
print_types_pretty=true
show_compiled_ir=true
hide_fresh_type_vars=false
show_raw_type_vars=true
show_quantifiers=true
```

```
([(Ir.Fun
     ((233, ((378) -379-> Int, "_fun__g1", Var.Scope.Local)),
      ([378; 379::Row(Any,Any)], [(232, (378, "", Var.Scope.Local))],
       ([(Ir.Let ((231, (378, "x", Var.Scope.Local)),
            ([], (Ir.Return (Ir.Variable 232)))))
          ],
        (Ir.Return (Ir.Constant (CommonTypes.Constant.Int 1))))),
      None, CommonTypes.Location.Unknown))
   ],
 (Ir.Return (Ir.Variable 233))
```

We can see variable 378 and 379 in the tyvar list, but the type of the function (namely `(378) -379-> Int`) is missing a `forall 378, 379.`.

jstolarek:
Yes, but wouldn't doing that require calling `generalize` in `desugarFuns`?

frank-emrich:
I would have hoped that the answer is "no, because we already call `generalise` in `typeSugar`." But you're right: The generalized type created in `typeSugar` is returned as the type of the `FunLit`, but not otherwise preserved. It is therefore not available in `desugarFuns`. This means that we would have to repeat the generalization in `desugarFuns'.

So I guess there are multiple options:
1. Like named function bindings, carry the overall (generalized) type in `FunLits`. The typing of `FunLit` and `Fun` bindings would then be identical, and the only thing `desugarFuns` has to do is a minimal change without touching the types at all.
2. Repeat the necessary generalization in `desugarFuns`. I've just verified that right now, `typeSugar` is the only place in Links that performs generalization. This would also raise the question of whether or not we need to carry around the `tyvar` list in `FunLit` before that.
3. Surrender and do the `FunLit` desugaring in `typeSugar.ml`.


I don't see a clear favorite between those three. Thoughts?

jstolarek:
(2) is more or less what we did before and I think the whole point of @jamescheney's suggestion was to avoid generalization in `desugarFuns`.

I will think about (1) on Monday. I was considering changing the definition of `fun_lit` so that it carries tvs as well, because now, i.e. after fixing generalization of anonymous functions, every constructor that carries a `fun_lit` also carries `tvars list`. So perhaps it would make sense to carry a `Types.datatype option` in `fun_lit` as well.

jstolarek:
Oh, and this:

> `typeSugar` is the only place in Links that performs generalization

is a very strong argument against (3).

jamescheney:
I remain uncertain what (if anything) breaks if we do *not* generalize `FunLit`s (i.e. adopt the non-`quantified_instantiation` behavior for `FunLit` only).  A function literal that appears in the middle of an expression presumably only needs to typecheck where it is; if it is going to be used at multiple types, then it can be put on the RHS of a `var` and be generalized there, right?

If we do need/want this behavior, then doing what we do for `Fun` bindings (i.e., decorating the AST with the inferred polymorphic function type) seems like the most sensible course of action.

jamescheney:
So having looked at the code again more closely, it seems like (1) should work - basically record the result of generalization for later.  Right now, when we desugar to a function declaration we just reuse the (unquantified) function type.  I'm not totally convinced that this is much better than the `TAbstr` approach was but at least it seems more uniform.

I had been thinking that it ought to be straightforward to just quantify the function type with the `tyvars` but since the `tyvar list` doesn't include e.g. the kinds, this is not as straightforward as I expected.

I am now curious what breaks if we disable `quantified_instantiation` just for function literals, will try.

jamescheney:
Here is what breaks:

```
!FAILURE: Operation parameter pattern-matching (7)
  error: Operation parameter pattern-matching (7)
      (Unexpected stdout: expected `fun : (() {Move:([|Alice|Bob|_|]) {}-> _|c}~> Char) {Move{_}|c}~> Char'; got `fun : (() {|Move:([|Alice|Bob|_|]) {}-> _,wild|c}~> Char) {Move{_}|c}~> Char')
```

```
!FAILURE: Type ascription, parameterised handlers (1)
  error: Type ascription, parameterised handlers (1)
      (Unexpected stdout: expected `fun : (Int) -> (Float) -> (String) -> (() {Op:(d) {}-> d|e}~> f) -> () {Op{_}|e}~> f'; got `fun : (Int) -> (Float) -> (String) -> (() {|Op:(d) {}-> d,wild|e}~> f) -> () {Op{_}|e}~> f')
```

```
!FAILURE: Type ascription, parameterised handlers (2)
  error: Type ascription, parameterised handlers (2)
      (Unexpected stdout: expected `fun : (Float, String, Int) -> (() {Op:(b) {}-> b|c}~> d) -> () {Op{_}|c}~> d'; got `fun : (Float, String, Int) -> (() {|Op:(b) {}-> b,wild|c}~> d) -> () {Op{_}|c}~> d')
```

```
!FAILURE: Omission of resumption for nullary operations (3)
  error: Omission of resumption for nullary operations (3)
      (Unexpected stdout: expected `fun : (() {Foo:() {}-> _|b}~> Int) {Foo{_}|b}~> Int'; got `fun : (() {|Foo:() {}-> _,wild|b}~> Int) {Foo{_}|b}~> Int')
```
These are all in the `handlers` test cases. In each case, the example typechecks but has a slightly wilder type than expected.   But I have no clue whether the expected or inferred type is what is desired (or why the difference).

```
!FAILURE: Unwrap polymorphism
  error: Unwrap polymorphism
      (Unexpected return code: expected `0'; got `1')
  error: Unwrap polymorphism
      (Unexpected stdout: expected `3 : Int'; got `')
  error: Unwrap polymorphism
      (Unexpected stderr: expected `'; got `<string>:1: Type error: The value has type
    `[(a::Any) -b-> a::Any]'
but it is annotated with type
    `[forall c,d::Row.(c) -d-> c]'
In expression: var ids = [fun (x) {x}].')
```

```
!FAILURE: Bad polymorphism behaviour (1)
  error: Bad polymorphism behaviour (1)
      (Unexpected return code: expected `0'; got `1')
  error: Bad polymorphism behaviour (1)
      (Unexpected stdout: expected `[fun] : [forall a,b::Row.(a) -b-> a]'; got `')
  error: Bad polymorphism behaviour (1)
      (Unexpected stderr: expected `'; got `<string>:1: Type error: The value has type
    `[(a::Any) -b-> a::Any]'
but it is annotated with type
    `[forall c,d::Row.(c) -d-> c]'
In expression: var ids = [fun (x) {x}].')
```

```
!FAILURE: Bad polymorphism behaviour (2)
  error: Bad polymorphism behaviour (2)
      (Unexpected return code: expected `0'; got `1')
  error: Bad polymorphism behaviour (2)
      (Unexpected stdout: expected `[fun] : [(a::Any) -> a::Any]'; got `')
  error: Bad polymorphism behaviour (2)
      (Unexpected stderr: expected `'; got `<string>:1: Type error: The value has type
    `[(a::Any) -b-> a::Any]'
but it is annotated with type
    `[forall c,d::Row.(c) -d-> c]'
```

These three tests are in `polymorphism.tests`.  They appear to test that generalization of a list of anomymous functions yields a type with the `forall` inside the list constructor.  This makes sense if we generalize `FunLit`s locally but I think generalizing the whole value is reasonable; if we really want the 

Finally, in `records.tests` this fails:

```
!FAILURE: Possibly absent label in a closed row
  error: Possibly absent label in a closed row
      (Unexpected stdout: expected `fun : (()) -> ()'; got `fun : ((a-)) -> ()')
```
and I have no idea why, or what the inferred function type means.

However, no actual (non-test) programs/examples fail to typecheck, so I conjecture that the above tests fail because of differences in behavior if we don't locally generalize `FunLit`s, not wrong behavior.

jamescheney:
FWIW, the `polymorphism.tests` failures seem correctable by making the result type `forall a, b. [a -b-> a]` which is what you would get in Haskell (with RankNTypes).  Haskell rejects an attempt to typecheck `[fun x -> x]` at type `[forall a. a -> a]` even with RankNTypes enabled.

For the handlers and effects tests, I think what is happening is that with `quantified_instantiation` enabled, the generalize/quantify step has a side effect of normalizing the type expression, so that the redundant `wild` in the handlers types, and the vacuous absence constraint `(a-)` in the record example, go away.  So perhaps fixable by doing an appropriate generalize or type normalization/simplification step at the top level.

dhil:
Instead of regeneralizing, why not extend the notion of closures (IR) to close over type variables?

jamescheney:
@dhil I believe they already do (cf #384); this issue is about fixing the frontend typechecking/desugaring passes so the right thing happens to pass this information through to the IR.

slindley:
I've missed most of this discussion because I've been travelling for the last two weeks, but I'm pretty much entirely to blame for the undocumented hacky implementation of first-class polymorphism in Links, which seems to be what underlies much of the above discussion.

Please tread carefully:

 * I expect our test coverage may be poor, so I don't think it is a good idea to just hack on the code without understanding what's going on, and then conclude that everything's OK just because the existing tests pass.

 * The results of the tests in `polymorphism.tests` are as intended. The model is *not* analogous to GHC's `RankNTypes` but more like the now unsupported `ImpredicativeTypes`, Daan Leijen's HMF, the implementation of first-class polymorphism in Koka, and the bidirectional type system for polymorphism of  Dunfeld and Krishnaswami.

 * I suspect the reason why handler tests are failing is because they actually make use of first-class polymorphism in interesting ways.

 * Important functionality will break if `quantified_instantiation` is disabled. Setting it to false is not guaranteed to work at all. I added it for debugging purposes when I was implementing support for first-class polymorphism.

 * TAbstr is necessary to support polymorphic values that are not functions - notably the empty list.

 * Rather than people randomly hacking, I propose that I sit down with someone to try to work out how to fix the typing issues without breaking support for first-class polymorphism.

 * It would also be good to try to generate some documentation of how polymorphism works in Links, and ideally formalise the implementation, so that we can get a handle on how we might improve / simplify it.

jamescheney:
Thanks for this context @slindley.

As far as I am aware, no one is suggesting changes to the type inference/checking, removing `TAbstr/`TApp` or to the `quantified_instantiation` setting.  This PR is about causing the right terms and types for anonymous functions (`FunLit`) to be generated by `typeSugar`/`desugarFuns` so that we generate well-typed IR for at least the simplest examples reported in #446.

Janek experimetned with two approaches:
1. Catching the `TAbstr` constructs around `FunLit` added by `typeSugar` and handling them appropriately (up to d527064).  I complained about this but it seemed to work.
2. At my suggestion, in c86ddcd Janek tried avoiding creating the `TAbstr` constructs in `typeSugar` and instead decorating `FunLit` with the abstracted type variables.  This seems to require additionally recording the polymorphic function type for later use, as is already done for `Fun`/`Funs`.

I was *not* suggesting above disabling `quantified_instantiation` globally, or any change in this PR, I was just reporting on the result of an experiment disabling it locally just for `FunLit`, since I am still unclear on the reason why Links generalizes anonymous functions, to see whether we rely on this feature.  

I will try to have a quick look at the three type systems you mentioned.  

The fact that while a few tests fail, all actual Links example programs typecheck without generalizing anonymous functions suggests to me that this is not a feature we are relying on, but if there's a use case for generalizing anomymous functions that we can make to work with the typed IR then I'm agnostic.  I think that if we want this feature we should treat anonymous `FunLit`s exactly the same as we treat named, nonrecursive `Fun`s (i.e. approach 2 above).

Overall I tend to agree that it's better to drag the existing design out into the daylight and rationalize it.  So we shoudl probably discuss on Thursday before Janek/Frank putting more effort into it.

jstolarek:
>  I propose that I sit down with someone to try to work out how to fix the typing issues without breaking support for first-class polymorphism.

I guess that someone would be me and @frank-emrich. I suspect that we might not have enough time to discuss all the details during the dev meeting on Thursday, so perhaps we could have a follow-up meeting immediately after? @slindley, @frank-emrich, what do you think?

frank-emrich:
> so perhaps we could have a follow-up meeting immediately after?

I'd be up for that! I'm happy to come up with a list of questions as a starting point.
One important step would be to establish which invariants should hold during frontend processing and in particular within `typeSugar`. Ideally, we would codify the invariants and have the debugging option of checking them, e.g. between each frontend pass.

core/desugarFuns.ml
```diff
@@ -20,66 +25,78 @@ open SugarConstructors.DummyPositions
...
+let unwrap_def ({node=f, ft; _}, linearity, lam, location, t) =
   let ft = val_of ft in
   let rt = TypeUtils.return_type ft in
   let lam =
     let rec make_lam t : funlit -> funlit =
       function
-        | ([_ps], _body) as lam -> lam
-        | (ps::pss, body) ->
+        | (_, [_ps], _body) as lam -> lam
+        | ([], ps::pss, body) ->
```
jamescheney: It looks like there are many assertion failures on line 71 due to this pattern match being non-exhaustive.  Is there a particular reason we can assume that the tyvar list is empty if the first pattern doesn't match?  What would be wrong with binding it to a variable and using that variable again on line 67 below?

jamescheney: OK, so I tried this (binding `tyvars` instead of [], and passing `tyvars` through on line 67) and while more tests pass, there are still a bunch of ArityMismatch errors.  So not clear this is the right thing to do either, even by randomly-hacking-and-seeing-if-tests-pass standards.

jstolarek: Yes, good point. This is basically a work in progress towards a proper solution and I haven't really bothered to test before pushing :-)

jstolarek: > there are still a bunch of ArityMismatch errors

Yes, these are expected at this point.

jstolarek:
Following today's discussion with @slindley and @frank-emrich I have reverted this PR to the originally implemented solution and rebased it on top of master. I had to slightly adapt my solution to align with changes made by @slindley in 1477d3e32399490a06e023756b7f236e7e6dcdd8.  Note also that this patch is only relevant in the presence of `quantified_instantiation`. I added an extra `when Settings.get_value Instantiate.quantified_instantiation` clause on the relevant guard in `desugarFuns`. I believe in the current state of code this clause is a no-op, but it is there so that we remember to remove this whole thing if we remove `quantified_instantiation`.

One thing I couldn't figure out is how to make XML typing test be expected to fail (XML typing does not work correctly now, so it is supposed to throw an exception). I did:

```
XML typing
fun (x) {<a>{x}</a>}
stdout : fun : (Xml) -> Xml
exit : 1
```

but that still shows up as a failed test rather than a success. I commented out that test for now. If anyone knows how to fix it (i.e. mark it as expectd to fail) please let me know.

If there are no complaints I'd be happy to finally have this merged.

jamescheney:
It seems that the XML test typechecks in the frontend, but the IR typing fails because the IR typing does not know that `Xml` equals `[XmlItem]`, see related #256 where this causes other problems.

When IR typing is enabled I don't see the `fun : (Xml) -> Xml` in the output when run from the REPL (probably because of `fail_on_ir_type_error=true`).  So I suspect the test is failing because this expected string is not appearing.

jstolarek:
@jamescheney, thanks! BTW the test is supposed to be broken exactly because of the reasons you describe.

frank-emrich:
> It seems that the XML test typechecks in the frontend, but the IR typing fails because the IR typing does not know that `Xml` equals `[XmlItem]`

I remember us talking about this issue some time ago, but I'm not sure why it is an issue. In `DefaultAliases`,  `Xml` is defined as an alias for `[XmlItem]`. Thus, the IR type-checker should handle this correctly. I will have a look at it.

frank-emrich:
Good news: This was not a problem with XML per se, but a problem with aliases in the IR type-checker that Sam had fixed already in his branch. I have cherry-picked the fix on master.

jstolarek:
@frank-emrich, thanks! PR rebased onto current master.

PR #535, opened 2019-03-26
Saner printing of sugar ASTs; settings for dumping ASTs before and after frontend pipeline

frank-emrich:
This PR addresses two printing related things that frequently annoy me while debugging:

**Dumping ASTs**
 I often find myself in a position where I would like to see the sugar AST for a program and hacking something into `frontend.ml` to achieve this.
I've introduced the two settings `show_pre_frontend_sugar_ast` and `show_post_frontend_sugar_ast`. I've switched on, the program AST will be dumped right before, resp. after, executing the frontend pipeline in `frontend.ml`.

**Annoying `pos` fields**
Dumping sugar ASTs (e.g., using Sugartypes.show_program) currently yields a lot of ugly output of the following form:
```
 { SourceCode.WithPos.node =
                           (Sugartypes.Block
                              ([],
                               { SourceCode.WithPos.node =
                                 (Sugartypes.FnAppl (
                                    { SourceCode.WithPos.node =
                                      (Sugartypes.Var "f"); pos = ... },
                                    [{ SourceCode.WithPos.node =
                                       (Sugartypes.Var "x"); pos = ... }
                                      ]
                                    ));
                                 pos = ... }));
                           pos = ... }),
                        CommonTypes.Location.Unknown));
                     pos = ... }))
```
Here, all of the shown `pos` fields of the `WithPos` are unnecessary clutter (in particular because their content is only ever printed as `...`.

In order to address this, the new setting `show_sugar_positions` determines whether or not to show the `pos` blocks at all. I've also implemented some logic to sanely print positions.

With  `show_sugar_positions` enabled we get the following:
```
{ SourceCode.WithPos.node =
                           (Sugartypes.Block
                              ([],
                               { SourceCode.WithPos.node =
                                 (Sugartypes.FnAppl (
                                    { SourceCode.WithPos.node =
                                      (Sugartypes.Var "f");
                                      pos =
                                      File ../my_examples/application.links, line 3, chars 26 to 27
                                      },
                                    [{ SourceCode.WithPos.node =
                                       (Sugartypes.Var "x");
                                       pos =
                                       File ../my_examples/application.links, line 3, chars 28 to 29
                                       }
                                      ]
                                    ));
                                 pos =
                                 File ../my_examples/application.links, line 3, chars 26 to 30
                                 }));
                           pos =
                           File ../my_examples/application.links, line 2, char 4, to line 20, char 34
                           }),
```

without it, we get the following:
```
([],
 (Some (Sugartypes.FunLit (None, CommonTypes.DeclaredLinearity.Unl,
          ([[(Sugartypes.Pattern.Variable ("f", None))]],
           (Sugartypes.Block
              ([],
               (Sugartypes.FunLit (None, CommonTypes.DeclaredLinearity.Unl,
                  ([[(Sugartypes.Pattern.Variable ("x", None))]],
                   (Sugartypes.Block
                      ([],
                       (Sugartypes.FnAppl ((Sugartypes.Var "f"),
                          [(Sugartypes.Var "x")]))))),
                  CommonTypes.Location.Unknown))))),
          CommonTypes.Location.Unknown)))
```

Note that this is supposed to only affect debugging/dumping ASTs, not user-facing stuff. The setting is disabled by default.

However, I've noticed that many of the messages in errors.ml try to do their own formatting of source code positions and ranges. We may want to consolidate that and only use `Position.show` in the future.

### DIFF
--- a/core/frontend.ml
+++ b/core/frontend.ml
@@ -76,12 +76,12 @@ struct
 
 
 let program tyenv pos_context program =
-  if Settings.get_value Basicsettings.show_pre_frontend_sugar_ast then
+  if Settings.get_value Basicsettings.show_pre_frontend_ast then
     Debug.print ("Pre-Frontend AST:\n" ^ Sugartypes.show_program program);
 
   let ((post_program, _, _), _) as result = program_pipeline tyenv pos_context program in
 
-  if Settings.get_value Basicsettings.show_post_frontend_sugar_ast then
+  if Settings.get_value Basicsettings.show_post_frontend_ast then
     Debug.print ("Post-Frontend AST:\n" ^ Sugartypes.show_program post_program);
 
   result
@@ -110,12 +110,12 @@ let program tyenv pos_context program =
 
 
 let interactive tyenv pos_context sentence =
-  if Settings.get_value Basicsettings.show_pre_frontend_sugar_ast then
+  if Settings.get_value Basicsettings.show_pre_frontend_ast then
     Debug.print ("Pre-Frontend AST:\n" ^ Sugartypes.show_sentence sentence);
 
   let (post_sentence, _, _) as result = interactive_pipeline tyenv pos_context sentence in
 
-  if Settings.get_value Basicsettings.show_post_frontend_sugar_ast then
+  if Settings.get_value Basicsettings.show_post_frontend_ast then
     Debug.print ("Post-Frontend AST:\n" ^ Sugartypes.show_sentence post_sentence);
 
   result
--- a/core/frontend.ml
+++ b/core/frontend.ml
@@ -36,7 +36,7 @@ struct
   let after_typing_ext enabled = extension enabled after_typing
   let before_typing_ext enabled f x = if enabled then f x else x
 
-  let program =
+  let program_pipeline =
     fun tyenv pos_context program ->
       let program = (ResolvePositions.resolve_positions pos_context)#program program in
       let program = DesugarAlienBlocks.transform_alien_blocks program in
@@ -73,6 +73,18 @@ struct
        ->- after_typing ((DesugarPages.desugar_pages tyenv)#program ->- snd3)
        ->- after_typing ((DesugarFuns.desugar_funs tyenv)#program ->- snd3))
         program), ffi_files)
+
+
+let program tyenv pos_context program =
+  if Settings.get_value Basicsettings.show_pre_frontend_sugar_ast then
+    Debug.print ("Pre-Frontend AST:\n" ^ Sugartypes.show_program program);
+
+  let ((post_program, _, _), _) as result = program_pipeline tyenv pos_context program in
+
+  if Settings.get_value Basicsettings.show_post_frontend_sugar_ast then
+    Debug.print ("Post-Frontend AST:\n" ^ Sugartypes.show_program post_program);
+
+  result
 
 
   let interactive =
--- a/core/frontend.ml
+++ b/core/frontend.ml
@@ -87,7 +87,7 @@ let program tyenv pos_context program =
   result
 
 
-  let interactive =
+  let interactive_pipeline =
     fun tyenv pos_context sentence ->
     let sentence = (ResolvePositions.resolve_positions pos_context)#sentence sentence in
     let _sentence = CheckXmlQuasiquotes.checker#sentence sentence in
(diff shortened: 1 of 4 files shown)


core/sourceCode.ml
```diff
@@ -71,9 +71,9 @@ end
 module Lexpos = struct
   type t = Lexing.position
 
-  let pp fmt _ =
-    (** Supress lexpos output**)
-    Utility.format_omission fmt
+  let pp fmt lpos =
+    Format.fprintf fmt
+      "File %s, line %d, char %d" lpos.Lexing.pos_fname lpos.Lexing.pos_lnum lpos.Lexing.pos_cnum
```
jstolarek: I would say `character` or `column` instead of `char`.

core/sourceCode.ml
```diff
@@ -86,7 +86,35 @@ module Position = struct
...
+
+      let start_line  = pos.start.Lexing.pos_lnum  in
+      let start_char  = pos.start.Lexing.pos_cnum  in
+      let finish_line = pos.finish.Lexing.pos_lnum in
+      let finish_char = pos.finish.Lexing.pos_cnum in
+
+      if start_line = finish_line then
+        if start_char = finish_char then
+          Format.fprintf fmt
+            "line %d, char %d"  start_line start_char
```
jstolarek: Same here and below.

core/sourceCode.ml
```diff
@@ -86,7 +86,35 @@ module Position = struct
...
+      let file = pos.start.Lexing.pos_fname in
+
+      Format.fprintf fmt "File %s, "  file;
+
+      let start_line  = pos.start.Lexing.pos_lnum  in
+      let start_char  = pos.start.Lexing.pos_cnum  in
+      let finish_line = pos.finish.Lexing.pos_lnum in
+      let finish_char = pos.finish.Lexing.pos_cnum in
+
+      if start_line = finish_line then
```
jstolarek: I don't like these kind of nested `if`s. Perhaps it would be better to:
```
match (start_line = finish_line, start_char = finish_char) with
| (true, true ) -> ...
| (true, false) -> ...
| (false, _   ) -> ...
```

Not sure though. This isn't a critical fragment of the code so I guess either way will work.

core/frontend.ml
```diff
@@ -75,6 +75,18 @@ struct
         program), ffi_files)
 
 
+let program tyenv pos_context program =
```
SimonJF: Is there any way to have this for REPL `sentence`s as well?

jstolarek:
Yes, I like that, especially the option to avoid position printing.

> I've noticed that many of the messages in errors.ml try to do their own formatting of source code positions and ranges. We may want to consolidate that and only use Position.show in the future.

I suggest creating a ticket once this patch is merged.

core/sourceCode.ml
dhil: I personally don't mind the nested `if`, but if you really want to rewrite them, then you can avoid an allocation by doing something along the lines of.
```ocaml
match start_line = finish_line with
| true when start_char = finish_char -> ...
| true -> ...
| false -> ...
```

core/frontend.ml
frank-emrich: yep, done :)

frank-emrich:
@SimonJF 
> isn't `sugar` redundant? Could you get away with `show_pre_frontend_ast`

Yes, originally I had `frontend` in there because I'm hooking into the pipelines in `frontend.ml` and `sugar` because it's the `sugartypes.ml` AST. But those are kind of redundant, I'll rename the settings.

core/sourceCode.ml
frank-emrich: done

core/sourceCode.ml
SimonJF: I have an aversion to pattern matching on booleans and prefer the nested ifs, but that's just personal preference.

SimonJF:
Thanks, @frank-emrich -- more than happy for this to go in now.

core/sourceCode.ml
frank-emrich: I see the point in trying to avoid nested `if`s, but I'd say in this case where the branches are quite short it should be okay.

jstolarek: Ok, dismiss my remark then.

jstolarek:
Ok for me as well.

Issue #514, opened 2019-03-01
ConcreteSyntaxError vs SugarError


jstolarek:
In `Sugartypes` we have:

```OCaml
exception ConcreteSyntaxError       of (Position.t * string)
```
In `Errors.ml` we have:

```OCaml
exception SugarError of (Position.t * string)
```
What is the intended difference between the two?

SimonJF:
From what I can tell, `ConcreteSyntaxError` is a syntax error (for example, an invalid primary kind annotation), whereas `SugarError` is a program which parses correctly, but a semantic problem has been found (for example, a duplicate attribute definition).

Generally, `ConcreteSyntaxError` appears in `parser.mly`, and `SugarError` appears during the desugaring stages.

There's also `RichSyntaxError` which is raised as a result of catching a `ConcreteSyntaxError`, containing more positional information.

I think there is a case for keeping them separate: the errors that are printed are distinct (in particular, the syntax errors provide a parse error message, whereas `SugarError` returns the parsed but semantically-incorrect expression).

I propose renaming `SugarError` to `DesugaringError`, and changing the error display of "Syntactic sugar error: " to "Error <description of stage>", for example "Error compiling regular expression: ...".

SimonJF:
This terminology is clarified in #536, so closing.

PR #536, opened 2019-03-27
Some error message easy wins

### LINKED ISSUE #531: Remove HTML formatting in `errors.ml`
A pain point in adding new errors to `errors.ml` is having to add formatting code for both HTML and console errors. Some errors (for example, type errors) will never be printed as HTML since the web server won't run until the program is typechecked. In addition, the formatting of errors does not necessarily match up.

I propose simply having the console error function, and having the HTML function wrap around it (merely adding a header, rather than doing any specific formatting of its own).

SimonJF:
The result of two train-rides' work refactoring how we do error messages in the frontend.

This firstly fixes #531 by removing the logic duplication in formatting HTML -- in turn, this should make it less painful to add new errors (and encourage people to do that, instead of using `failwith`).

This secondly addresses #514 by renaming the confusingly-named `SugarError` to `DesugaringError` and attaching a variant describing a stage to it. In doing this, two things came up:
  1. In some places, `SugarError` was used where an `InternalError` was more appropriate
  2. In `desugarLAttributes`, dummy positions were used, so I've refactored it to use proper position information now

There are also some other minor refactorings to make things nicer.

### DIFF
--- a/core/checkXmlQuasiquotes.ml
+++ b/core/checkXmlQuasiquotes.ml
@@ -52,10 +52,10 @@ let check mode pos e =
         Printf.sprintf "%s %s in %s quasiquote" node_type expr kind in
       raise (desugaring_error ~pos ~stage:CheckQuasiquotes ~message) in
     match o#get_error with
-    | None -> ()
-    | Some (`FormletBinding, pos') -> raise_error "Formlet binding" pos'
+    | None                           -> ()
+    | Some (`FormletBinding, pos')   -> raise_error "Formlet binding" pos'
     | Some (`FormletPlacement, pos') -> raise_error "Formlet placement" pos'
-    | Some (`PagePlacement, pos') -> raise_error "Page placement" pos'
+    | Some (`PagePlacement, pos')    -> raise_error "Page placement" pos'
 
 (* traverse a whole tree searching for and then checking quasiquotes *)
 let checker =
--- a/core/checkXmlQuasiquotes.ml
+++ b/core/checkXmlQuasiquotes.ml
@@ -45,17 +45,17 @@ let check mode pos e =
     | `Formlet -> "formlet"
     | `Page -> "page"
   in
+    let raise_error node_type pos =
+      let open Errors in
+      let expr = Position.resolve_expression pos in
+      let message =
+        Printf.sprintf "%s %s in %s quasiquote" node_type expr kind in
+      raise (desugaring_error ~pos ~stage:CheckQuasiquotes ~message) in
     match o#get_error with
     | None -> ()
-    | Some (`FormletBinding, pos') ->
-      let expr = Position.resolve_expression pos' in
-        raise (Errors.SugarError (pos, "Formlet binding " ^ expr ^ " in " ^ kind ^ " quasiquote"))
-    | Some (`FormletPlacement, pos') ->
-      let expr = Position.resolve_expression pos' in
-        raise (Errors.SugarError (pos, "Formlet placement " ^ expr ^ " in " ^ kind ^ " quasiquote"))
-    | Some (`PagePlacement, pos') ->
-      let expr = Position.resolve_expression pos' in
-        raise (Errors.SugarError (pos, "Page placement " ^ expr ^ " in " ^ kind ^ " quasiquote"))
+    | Some (`FormletBinding, pos') -> raise_error "Formlet binding" pos'
+    | Some (`FormletPlacement, pos') -> raise_error "Formlet placement" pos'
+    | Some (`PagePlacement, pos') -> raise_error "Page placement" pos'
 
 (* traverse a whole tree searching for and then checking quasiquotes *)
 let checker =
@@ -87,7 +87,9 @@ object (o)
       o#phrase_with `Quasi body
     | (Formlet _ | Page _) when mode = `Quasi ->
       (* The parser should prevent this from ever happening *)
-      raise (Errors.SugarError (pos, "Malformed quasiquote (internal error)"))
+      let message =
+        Printf.sprintf "Malformed quasiquote (%s)" (Position.show pos) in
+      raise (Errors.internal_error ~filename:"checkXmlQuasiquotes.ml" ~message)
     | _ when mode = `Quasi ->
       o#phrase_with `Exp phrase
     | _ when mode = `Exp ->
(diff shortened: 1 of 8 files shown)


core/checkXmlQuasiquotes.ml
```diff
@@ -45,17 +45,17 @@ let check mode pos e =
...
-    | Some (`FormletBinding, pos') ->
-      let expr = Position.resolve_expression pos' in
-        raise (Errors.SugarError (pos, "Formlet binding " ^ expr ^ " in " ^ kind ^ " quasiquote"))
-    | Some (`FormletPlacement, pos') ->
-      let expr = Position.resolve_expression pos' in
-        raise (Errors.SugarError (pos, "Formlet placement " ^ expr ^ " in " ^ kind ^ " quasiquote"))
-    | Some (`PagePlacement, pos') ->
-      let expr = Position.resolve_expression pos' in
-        raise (Errors.SugarError (pos, "Page placement " ^ expr ^ " in " ^ kind ^ " quasiquote"))
+    | Some (`FormletBinding, pos') -> raise_error "Formlet binding" pos'
```
jstolarek: Alignment:

```OCaml
    | Some (`FormletBinding  , pos') -> raise_error "Formlet binding"   pos'
    | Some (`FormletPlacement, pos') -> raise_error "Formlet placement" pos'
    | Some (`PagePlacement   , pos') -> raise_error "Page placement"    pos'
```

core/desugarLAttributes.ml
```diff
@@ -41,14 +46,15 @@ let desugar_lhref : phrasenode -> phrasenode = function
                 apply "pickleCont" [fun_lit ~location:loc_server dl_unl [[]]
                                             target]])
               :: rest
-          | _ -> assert false (* multiple l:hrefs, or an invalid rhs;
-                                 NOTE: this is a user error and should
-                                 be reported as such --ez.*)
-      in Xml (a, attrs, attrexp, children)
+          | _ ->
+              desugaring_error pos
+                ("Invalid l:href: check that there are not " ^
```
jstolarek: "not" -> "no" ?

core/desugarLAttributes.ml
```diff
@@ -60,15 +66,17 @@ let desugar_laction : phrasenode -> phrasenode = function
...
-            in Xml (form, action::rest, attrexp, hidden::children)
-        | _ -> assert false (* multiple l:actions, or an invalid rhs;
-                               NOTE: this is a user error and should
-                               be reported as such --ez. *)
+            and action = ("action", [constant_str "#"]) in
+            WithPos.make ~pos
+              (Xml (form, action::rest, attrexp, hidden::children))
+        | _ ->
+            desugaring_error pos
+              ("Invalid l:action: check that there are not " ^
```
jstolarek: "not" -> "no" ?

core/desugarLAttributes.ml
```diff
@@ -78,33 +86,36 @@ let desugar_lonevent : phrasenode -> phrasenode =
...
+  let attr : Position.t -> string * phrase list -> (string * phrase list) list =
+    fun pos attribute ->
+    match attribute with
+      | "l:name", [{node=Constant (Constant.String v); _}] ->
+          let id, name = fresh_names () in
+            add v (id,name);
+            [("name", [constant_str name]);
+             ("id"  , [constant_str id  ])]
+      | "l:name", _ ->
+          desugaring_error pos "The value l:name binding must be a string constant"
```
jstolarek: "value of"?

core/desugarLAttributes.ml
```diff
@@ -78,33 +86,36 @@ let desugar_lonevent : phrasenode -> phrasenode =
...
+      | a -> [a] in
+  let rec aux : phrase -> phrase  = function
+    | { node=Xml (tag, attrs, attrexp, children); pos } ->
+        let attrs = concat_map (attr pos) attrs in
         let children =
-          List.map (fun {node;pos} -> with_pos pos (aux node)) children in
-          Xml (tag, attrs, attrexp, children)
+          List.map (fun child -> aux child) children in
```
jstolarek: Would OCaml allow `List.map aux children`?

core/desugarRegexes.ml
```diff
@@ -90,7 +90,9 @@ object(self)
           self#phrase (fn_appl libfn tyargs
                             [e1; desugar_regex self#phrase regex_type r])
     | InfixAppl ((_tyargs, BinaryOp.RegexMatch _), _, _) ->
-        raise (Errors.SugarError (pos, "Internal error: unexpected rhs of regex operator"))
+        let (_, expr) = SourceCode.Position.resolve_start_expr pos in
```
jstolarek: This looks strange. Can't we recover the RHS expression from the pattern match:

```
InfixAppl ((_tyargs, BinaryOp.RegexMatch _), _, rhs_expr)
```
?

core/errors.ml
```diff
@@ -7,14 +7,30 @@ type synerrspec = {filename : string; linespec : string;
...
-exception UndefinedVariable of string
+type sugar_error_stage =
+  | DesugarFormlets
+  | DesugarRegexes
+  | CheckQuasiquotes
+  | DesugarLAttributes
+  | DesugarPages
+  | CheckXML
+
+let string_of_stage = function
```
jstolarek: Alignment. In Emacs: `M-x align-regexp RET -> RET`

SimonJF: Vim user, but I think using something like https://www.vim.org/scripts/script.php?script_id=294 is likely overdue :)

core/desugarRegexes.ml
SimonJF: Sure, but `expr` here is the string representation from the source code rather than an AST itself, so is arguably more useful to report in the error.

SimonJF:
Thanks for the review, @jstolarek! Will address on my next train ride.

core/errors.ml
jstolarek: > Vim user

Shame on you!

core/desugarRegexes.ml
jstolarek: Ah, I see.

core/desugarLAttributes.ml
SimonJF: Yes, why I didn't eta-contract that remains a mystery to me.

PR #541, opened 2019-03-27
Fixing test-harness config/argument passing

frank-emrich:
I came across some bugs/inconsistencies in `test-harness`:

1. `run-tests` checks if a `.tests` file comes with a corresponding `.config` file and if so passes it to `test-harness` as its second argument. `test-harness` never used/checked its second argument.

2. There were two ways of specifying what (extra) arguments should be passed to Links when executing a test case: Either a `flags : ...` entry in a test specification, or, if `filemode : args` is part of a test specification, then a line `args : ...` would indicate command line arguments to pass.

After my changes, there is no `filemode : args` anymore, only `filemode : true` is left. The one and only way to pass arguments to Links is by specifying a `args : ...` line. There is no need to append the `-e` manually, this is done by the script based on whether `filemode` is active or not.

The script also checks whether a global config file was passed to the script and there is a `--config` parameter at the same time, which is probably not what you wanted.


jstolarek:
This PR breaks Travis:

```
cmnd=./test-harness tests/typecheck_examples.tests tests/typecheck_examples.tests.config
Test "Typecheck example file examples/AIPL14/LectureExamples/calc.links"
comes with an args entry to specify the config file, but a config file was also
passed as a command-line argument to the test harness
Traceback (most recent call last):
  File "./test-harness", line 126, in <module>
    main()
  File "./test-harness", line 118, in main
    evaluate(name, code, config_file, **opts)
  File "./test-harness", line 68, in evaluate
    exit(1)
TypeError: 'str' object is not callable
```

frank-emrich:
The error message was buried under the noise:
Test "Typecheck example file examples/AIPL14/LectureExamples/calc.links"
comes with an args entry to specify the config file, but a config file was also
passed as a command-line argument to the test harness

All the tests in `typecheck_examples.tests` came with an `args` entry  `--config=tests/typecheck_examples.tests.config` even though `tests/typecheck_examples.tests.config` is passed as the second argument to `test-harness`, which is now not only superfluous but forbidden (to have both)

tests/handlers.tests
```diff
@@ -1,490 +1,491 @@
 Identity handler (1)
 { handle(42) {case Return(x) -> x} }
 stdout : 42 : Int
-flags : --enable-handlers -e
```
jstolarek: Why is the `-e` parameter removed from all these tests? I understand that it is now added automatically if `filemode : true`, but filemode does not seem to be enabled here.

test-harness
```diff
@@ -57,15 +57,24 @@ def check_expected(name, item, got, expected, errors):
...
+        arg_array += ["--config=" + config_file]
+
+
     if filemode.startswith('true') :
-        proc = Popen([links, code], stdout=PIPE, stderr=PIPE, env=env)
-    elif filemode.startswith('args'):
-        arg_array = str.split(args, ' ')
-        popen_list = [links] + arg_array + [code]
-        proc = Popen(popen_list, stdout=PIPE, stderr=PIPE, env=env)
+        proc = Popen([links] + arg_array + [code], stdout=PIPE, stderr=PIPE, env=env)
```
jstolarek: I'd avoid the repetition with something like (I don't know Python, so syntax may be wrong):

```Python
if filemode.startswith('true') :
    arg_array = arg_array + "-e"

proc = Popen([links] + arg_array + [code], stdout=PIPE, stderr=PIPE, env=env)
```

tests/handlers.tests
frank-emrich: It's the other way around: The script adds `-e` as the last argument iff  `filemode : true` is *not* given.

test-harness
frank-emrich: Good point. Done!

jamescheney:
This looks ready to go now.  Is it?

PR #543, opened 2019-03-28
Nicer IR type-checking error reporting

### LINKED ISSUE #539: Broken error message when fail_on_ir_type_error is enabled
Something is horribly broken with errors generated when `fail_on_ir_type_error` is enabled:

```
*** Fatal error : Unknown exception during IR type-checking: (Failure
   "IR Type Error: Type mismatch:\
  \n Expected:\
  \n`Application (({ Types.Abstype.id = \"List\"; name = \"List\";\
  \n                 arity =\
  \n                 [(CommonTypes.PrimaryKind.Type,\
  \n                   (CommonTypes.Linearity.Unl, CommonTypes.Restriction.Any))]\
  \n                 },\
  \n               [`Type (`MetaTypeVar (Point @ 0x7f1861672d80 `Var ((397,\
  \n                                                                   (CommonTypes.Linearity.Unl,\
  \n                                                                    CommonTypes.Restriction.Any),\
  \n                                                                   `Rigid))))\
  \n                 ]))\
  \n Actual:\
  \n `ForAll ((ref ([]),\
  \n          `Application (({ Types.Abstype.id = \"List\"; name = \"List\";\
  \n                           arity =\
  \n                           [(CommonTypes.PrimaryKind.Type,\
  \n                             (CommonTypes.Linearity.Unl,\
  \n                              CommonTypes.Restriction.Any))\
  \n                             ]\
  \n                           },\
  \n                         [`Type (`MetaTypeVar (Point @ 0x7f1861673788 `Var (\
  \n                                               (378,\
  \n                                                (CommonTypes.Linearity.Unl,\
  \n                                                 CommonTypes.Restriction.Any),\
  \n                                                `Rigid))))\
  \n                           ]))))\
```
`fail_on_ir_type_error` is at the moment a hard-coded setting in `irCheck.ml`.

frank-emrich:
This is just a cleanup of the error handling in the IR type-checker. In particular, there should be a nicer separation between error handling done inside of the type-checker and the handling done outside of it.

Most of the changes are just because I got rid of some polymorphic variants.

Fixes #539 (together with  3062b583627b3ba750bed5e426593ab33b43ecff).

### DIFF
--- a/core/irCheck.ml
+++ b/core/irCheck.ml
@@ -67,9 +67,6 @@ let handle_ir_type_error lazy_val alternative occurrence =
         Debug.print (msg ^ occurrence_string);
         alternative
 
-
-
-
 (* Some of the helper functions we use raise their own exceptions.
    We translate them to IR type exceptions here *)
 let translate_helper_exns (f : 'a -> 'b) (x : 'a) occurrence : 'b =
@@ -83,9 +80,6 @@ let translate_helper_exns (f : 'a -> 'b) (x : 'a) occurrence : 'b =
         Printf.sprintf "Arity mismatch during type application/instantiation. \
                         Providing %d type arguments to function that takes %d." given takes in
       raise_ir_type_error msg occurrence
-
-
-
 
 let ensure condition msg occurrence =
   if condition then () else raise_ir_type_error msg occurrence
@@ -366,9 +360,6 @@ let eq_types occurrence : type_eq_context -> (Types.datatype * Types.datatype) -
       eqt  (context, t1, t2)
 
 
-
-
-
 let check_eq_types (ctx : type_eq_context) et at occurrence =
   if not (eq_types occurrence ctx (et, at)) then
     raise_ir_type_error
@@ -382,8 +373,6 @@ let check_eq_type_lists = fun (ctx : type_eq_context) exptl actl occurrence ->
       List.iter2 (fun  et at ->
           check_eq_types ctx et at occurrence
        )  exptl actl
-
-
 
 
 let ensure_effect_present_in_row ctx allowed_effects required_effect_name required_effect_type occurrence =
@@ -598,10 +587,8 @@ struct
                                          (string_of_datatype vt) (string_of_datatype t)) (SVal orig);
               Coerce (v, t), t, o
               end
-
       in
       fun value -> translate_helper_exns value_inner value (SVal value)
-
 
 
     method! tail_computation :
@@ -938,10 +925,8 @@ struct
         | LensGet _
         | LensJoin _
         | LensPut _ -> (* just do type reconstruction *) super#special special
-
       in
       fun special -> translate_helper_exns special_inner special (SSpec special)
-
 
 
    method! bindings : binding list -> (binding list * 'self_type) =
@@ -1119,9 +1104,6 @@ struct
               Module (name, defs), o
       in
       fun b -> translate_helper_exns binding_inner b (SBind b)
-
-
-
 
 
     method! binder : binder -> (binder * 'self_type) =
--- a/core/irCheck.ml
+++ b/core/irCheck.ml
@@ -18,21 +18,39 @@ type ir_snippet =
   | SProg  of program
   | SNone
 
-let string_of_occurrence : ir_snippet -> string = function
-  | STC tc -> "\noccurring in IR tail computation:\n" ^ Ir.string_of_tail_computation tc
-  | SVal v -> "\noccurring in IR value:\n" ^ Ir.string_of_value v
-  | SSpec s -> "\noccurring in IR special tail computation:\n" ^ Ir.string_of_special s
-  | SBind b -> "\noccurring in IR binding:\n" ^ Ir.string_of_binding b
+let string_of_occurrence : ir_snippet -> string =
+  let nl = "\n" in function
+  | STC tc ->
+    "occurring in IR tail computation:" ^
+    nl ^
+    Ir.string_of_tail_computation tc
+  | SVal v ->
+    "occurring in IR value:" ^
+    nl ^
+    Ir.string_of_value v
+  | SSpec s ->
+    "noccurring in IR special tail computation:" ^
+    nl ^
+    Ir.string_of_special s
+  | SBind b ->
+    "occurring in IR binding:" ^
+    nl ^
+    Ir.string_of_binding b
   | SBinds bs ->
-    "\noccurring in IR binding list:\n" ^
-    String.concat "\n" (List.map Ir.string_of_binding bs)
-  | SProg p ->  "\noccurring in IR program:\n" ^ Ir.string_of_program p
+    "occurring in IR binding list:" ^
+    nl ^
+    String.concat nl (List.map Ir.string_of_binding bs)
+  | SProg p ->
+    "occurring in IR program:" ^
+    nl ^
+    Ir.string_of_program p
   | SNone -> ""
 
 
 let raise_ir_type_error msg occurrence =
+  let nl = "\n" in
   let occurrence_string = string_of_occurrence occurrence in
-  raise (Errors.IRTypeError (msg ^ occurrence_string))
+  raise (Errors.IRTypeError (msg ^ nl ^ occurrence_string))
 
 
 (* Evalutes a lazy value and handles exceptions raised while doing so.
--- a/core/irCheck.ml
+++ b/core/irCheck.ml
@@ -6,30 +6,85 @@ open Ir
 
 let fail_on_ir_type_error =
   Settings.get_value Basicsettings.Ir.fail_on_ir_type_error
-exception IRTypeError of string
+
+(* Artificial "supertype" of all IR types,
+   so we can provide arbitrary IR fragments to the error handling functions *)
+type ir_snippet =
+  | STC    of tail_computation
+  | SVal   of value
+  | SSpec  of special
+  | SBind  of binding
+  | SBinds of binding list
+  | SProg  of program
+  | SNone
+
+let string_of_occurrence : ir_snippet -> string = function
+  | STC tc -> "\noccurring in IR tail computation:\n" ^ Ir.string_of_tail_computation tc
+  | SVal v -> "\noccurring in IR value:\n" ^ Ir.string_of_value v
+  | SSpec s -> "\noccurring in IR special tail computation:\n" ^ Ir.string_of_special s
+  | SBind b -> "\noccurring in IR binding:\n" ^ Ir.string_of_binding b
+  | SBinds bs ->
+    "\noccurring in IR binding list:\n" ^
+    String.concat "\n" (List.map Ir.string_of_binding bs)
+  | SProg p ->  "\noccurring in IR program:\n" ^ Ir.string_of_program p
+  | SNone -> ""
+
 
 let raise_ir_type_error msg occurrence =
-  let occurrence_string = match occurrence with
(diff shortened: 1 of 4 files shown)


jamescheney:
#301 is a pull request, are you sure it's not some other issue this fixes?

frank-emrich:
Oops, you're right. Edited the description.

dhil:
I think, in the future we should define IR errors in a separate module from the front end errors.

core/irCheck.ml
```diff
@@ -5,30 +5,85 @@ open Ir
...
+  | STC    of tail_computation
+  | SVal   of value
+  | SSpec  of special
+  | SBind  of binding
+  | SBinds of binding list
+  | SProg  of program
+  | SNone
+
+let string_of_occurrence : ir_snippet -> string = function
+  | STC tc -> "\noccurring in IR tail computation:\n" ^ Ir.string_of_tail_computation tc
```
jstolarek: I don't like inserting `\n` into strings. Is there no function or combinator suited to do just that?

Also, maybe align `->`?

core/irCheck.ml
```diff
@@ -5,30 +5,85 @@ open Ir
...
+  | SVal v -> "\noccurring in IR value:\n" ^ Ir.string_of_value v
+  | SSpec s -> "\noccurring in IR special tail computation:\n" ^ Ir.string_of_special s
```
jstolarek: Doesn't fit 80 columns.

core/irCheck.ml
```diff
@@ -1050,6 +1116,12 @@ struct
                       Some defs, o
             in
               Module (name, defs), o
+      in
+      fun b -> translate_helper_exns binding_inner b (SBind b)
+
```
jstolarek: Why so many empty lines?

core/irCheck.ml
```diff
@@ -990,7 +1053,9 @@ struct
                 let o = OptionUtils.opt_app o#remove_binder o z in
                 let o = List.fold_right (fun b o -> o#remove_binder b) xs o in
                 f, tyvars, xs, body, z, location, o
-              with e -> handle_ir_type_error e (f, tyvars, xs, body, z, location, o) in
+              ) in
+              let f, tyvars, xs, body, z, location, o =
+               handle_ir_type_error lazy_check (f, tyvars, xs, body, z, location, o) (SBind binding) in
```
jstolarek: Does not fit 80 columns.

core/irCheck.ml
```diff
@@ -881,6 +938,11 @@ struct
         | LensJoin _
         | LensPut _ -> (* just do type reconstruction *) super#special special
 
+      in
+      fun special -> translate_helper_exns special_inner special (SSpec special)
+
```
jstolarek: Why so many empty lines?

core/irCheck.ml
```diff
@@ -618,36 +669,42 @@ struct
...
             let right, rt, o = o#computation right in
-            o#check_eq_types vt (`Primitive Primitive.Bool) (`TC orig);
-            o#check_eq_types lt rt (`TC orig);
+            o#check_eq_types vt (`Primitive Primitive.Bool) (STC orig);
+            o#check_eq_types lt rt (STC orig);
             If (v, left, right), lt, o
 
+      in
+      fun tc -> translate_helper_exns tail_computation_inner tc (STC tc)
+
```
jstolarek: Why so many empty lines?

core/irCheck.ml
```diff
@@ -547,14 +593,19 @@ struct
...
-              ensure (eq_types (`Value orig) (o#extract_type_equality_context ()) (vt, t) || is_sub_type (vt, t)) (Printf.sprintf "coercion error: %s is not a subtype of %s"
-                                         (string_of_datatype vt) (string_of_datatype t)) (`Value orig);
+              ensure (eq_types (SVal orig) (o#extract_type_equality_context ()) (vt, t) || is_sub_type (vt, t)) (Printf.sprintf "coercion error: %s is not a subtype of %s"
+                                         (string_of_datatype vt) (string_of_datatype t)) (SVal orig);
               Coerce (v, t), t, o
               end
 
+      in
+      fun value -> translate_helper_exns value_inner value (SVal value)
+
```
jstolarek: Why so many empty lines?

core/irCheck.ml
```diff
@@ -547,14 +593,19 @@ struct
               end
             else
               begin
-              ensure (eq_types (`Value orig) (o#extract_type_equality_context ()) (vt, t) || is_sub_type (vt, t)) (Printf.sprintf "coercion error: %s is not a subtype of %s"
-                                         (string_of_datatype vt) (string_of_datatype t)) (`Value orig);
+              ensure (eq_types (SVal orig) (o#extract_type_equality_context ()) (vt, t) || is_sub_type (vt, t)) (Printf.sprintf "coercion error: %s is not a subtype of %s"
+                                         (string_of_datatype vt) (string_of_datatype t)) (SVal orig);
               Coerce (v, t), t, o
               end
```
jstolarek: Why an empty line here?

core/irCheck.ml
```diff
@@ -5,30 +5,85 @@ open Ir
...
+    f x
+  with
+    | TypeUtils.TypeDestructionError msg ->
+      raise_ir_type_error msg occurrence
+    | Instantiate.ArityMismatch (takes, given) ->
+      let msg =
+        Printf.sprintf "Arity mismatch during type application/instantiation. \
+                        Providing %d type arguments to function that takes %d." given takes in
+      raise_ir_type_error msg occurrence
+
```
jstolarek: Why so many empty lines?

core/irCheck.ml
```diff
@@ -5,30 +5,85 @@ open Ir
...
+           "\n-------------------------------------------------------------------\n" );
+        alternative
+      | exn ->
+        let msg =
+          "Encountered exception during IR type-checking, saying:\n" ^
+          Printexc.to_string exn in
+        let occurrence_string = string_of_occurrence occurrence in
+        Debug.print (msg ^ occurrence_string);
+        alternative
+
```
jstolarek: Why so many empty lines?

core/irCheck.ml
```diff
@@ -5,30 +5,85 @@ open Ir
...
-let handle_ir_type_error error alternative =
-  match error with
-    | IRTypeError msg
-    | TypeUtils.TypeDestructionError msg ->
-      if fail_on_ir_type_error then
-        failwith ("IR Type Error: " ^ msg)
-      else
+  let occurrence_string = string_of_occurrence occurrence in
+  raise (Errors.IRTypeError (msg ^ occurrence_string))
+
```
jstolarek: Why so many empty lines?

core/irCheck.ml
```diff
@@ -5,30 +5,85 @@ open Ir
...
+           Print the error and return the alternative value *)
         Debug.print
           ("\n--------------------------------------------------------------------\n" ^
-            "Continuing after IR Type Error: " ^
-            msg ^
-            "\n-------------------------------------------------------------------\n" ); alternative
-    |  e-> failwith ("Unknown exception during IR type-checking: " ^ Printexc.to_string e)
+           "Continuing after IR Type Error:\n" ^
+           msg ^
+           "\n-------------------------------------------------------------------\n" );
```
jstolarek: Why separate arguments to a function with empty lines?

jstolarek:
Hm... I just made some comments about too many empty lines, but now it looks like a github glitch. Let me check out your branch and see what it actually looks like. I'll update my comments in a few minutes.

jstolarek:
Ok, comments updated. Most were not a glitch.

core/typeUtils.ml
```diff
@@ -54,7 +54,7 @@ let extract_row t = match concrete_type t with
   | `Record row -> row
   | `Variant row -> row
   | t ->
-      failwith
+      error
```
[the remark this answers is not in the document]
frank-emrich: `error` is defined in `typeUtils` as a function creating `TypeDetructionError` exceptions

core/irCheck.ml
frank-emrich: Indeed, here it doesn't make much sense, removed them

core/irCheck.ml
frank-emrich: Many of the larger methods in `irCheck` now start with a big nested function, which is used at the end. I tried to make that visually more obvious, but I guess it's a matter of taste.

core/irCheck.ml
frank-emrich: Intentional, see other comment

core/irCheck.ml
frank-emrich: Intentional, see other comment

core/irCheck.ml
frank-emrich: Trying to separate the inner function visually from the other one

core/irCheck.ml
frank-emrich: Yes, I agree it's ugly. We have a little pretty-printier in `pp.ml`, I'll give it a try.

core/irCheck.ml
frank-emrich: Did we ever decide on using a 80 column limit? I agree that many lines in `irCheck` are too long, but I really hope we're not going for a 80 column limit.

core/irCheck.ml
jstolarek: Hm... I don't feel having 4-5 empty lines looks good or helpful. Wonder what others think.

core/irCheck.ml
jstolarek: No, we never did, so feel free to ignore :-) I personally don't go beyond 80th column.

core/irCheck.ml
frank-emrich: I just tend to keep scrolling past the end of very long functions/methods when the next one starts immediately afterwards. But I'm happy to remove those extra lines.

frank-emrich:
I removed the extra empty lines now

core/irCheck.ml
frank-emrich: How do you like the following:


```
let string_of_occurrence : ir_snippet -> string =
  let nl = "\n" in
  function
  | STC tc    ->
     nl ^
     "occurring in IR tail computation:" ^
     nl ^
     Ir.string_of_tail_computation tc
  | ...
```


I didn't use the `pp` thing, it turned out to make things more complicated, rather than less.

frank-emrich:
@dhil 
> I think, in the future we should define IR errors in a separate module from the front end errors.

The only issue here is that the current design is such that `driver.ml` wraps all of the actual processing with the error handling done in `errors.ml`. This is good, because it prevents unhandled errors to escape. On the other hand, it means that there is one location where all the error handling happens.

This is why as part of this PR, I had to do the opposite: moving `IRTypeError` from `irCheck.ml` to `errors.ml`.

dhil:
> @dhil
> 
> > I think, in the future we should define IR errors in a separate module from the front end errors.
> 
> The only issue here is that the current design is such that `driver.ml` wraps all of the actual processing with the error handling done in `errors.ml`. This is good, because it prevents unhandled errors to escape. On the other hand, it means that there is one location where all the error handling happens.
> 
> This is why as part of this PR, I had to do the opposite: moving `IRTypeError` from `irCheck.ml` to `errors.ml`.

Yes I understand. I think, again for the future, we should have a "driver" for the front end and another for the back end. Then we would have a third module that composes them.

jamescheney:
@jstolarek @dhil are your concerns addressed?  I think we should merge this and create other issue(s) about reorganizing the errors to separate frontend and backend error handling (which might happen naturally relatedly to #547).

PR #556, opened 2019-04-07
Parallelize execution of tests

frank-emrich:
Only the tests belonging to one group (i.e., .tests file) are executed in parallel. The .tests files are still processed sequentially. In particular, the bash scripts remain unchanged.

I've ensured that printing all results for one test is atomic, so the output will always look as if everything happened sequentially.

On my machine, this reduces the duration of the test suite from 5 to 2 minutes. I don't know if travis allows multi-(threading/processing), so the effect there may be less.


frank-emrich:
> I don't know if travis allows multi-(threading/processing), so the effect there may be less.

Seems like the answer to that is no. Travis took 13:45 min for this, whereas the last run on master took 14:30 min.

dhil:
I am no python expert, but I think Python is single-threaded, so you need to write your program in SPMP style and use physical processes. Since the test script is I/O-heavy it may suffice to refactor the program to be in lock-free style and use asynchronous programming (something along the lines of [this guide](https://www.machinelearningplus.com/python/parallel-processing-python/)).

frank-emrich:
AFAIK Python is multi-threaded, but has a weird locking model.

It doesn't really, though. The relevant work is happening in the Links processes that the python script spawns.
Thus, the threads in the thread pool spawn Links and wait for it to exit, then do a little bit of validation work. In any case, my change allows n Links instances to run in parallel, where n is the size of the thread pool. This value is set to be the number of CPUs on your system.

dhil:
> AFAIK Python is multi-threaded, but has a weird locking model.

A little bit of research reveals that Python has a global interpreter lock, meaning only a single thing can one at a time. So even if Python has genuine threads you would not get much of a performance benefit from using threads.

If we want to maximise performance of the test script, then we should write it in map-reduce style.

frank-emrich:
But the parallelism of the threads is irrelevant for our workload. Each thread spawns a Links process, which enjoys true parallelism, independent from the thread that spawned it. The threads spend most of their lifetime waiting for the Links process to finish. Thus, the fact that the threads don't run in parallel doesn't matter, because the Links processes do.

Check out the branch and you will immediately notice the difference. As I said, the run-time more than halved for me.

test-harness
```diff
@@ -3,14 +3,18 @@
 links = './links'
 
 import sys, re, time
+import concurrent.futures, threading, multiprocessing
 from subprocess import Popen, PIPE
 
 successes = failures = ignored = 0
 TIMEOUT = 15 # seconds before a test is timed out.
 
+the_lock = threading.Lock()
```
dhil: Don't use locks. That is unnecessary complicated. This problem is an **embarrassing parallel problem**. Use a structured programming technique like map-reduce.

jstolarek:
Testing time went down from ~3m50s to ~1m15s on my laptop. I very much approve if this.

test-harness
frank-emrich: If I understand Pyhton's GIL correctly, we only need locks for printing, as all changes to shared variables (like successes, failures, ignored) are atomic. 
I'm not sure if a single print("foo") is atomic under the GIL, but it seems so.

We then just need a lock for the three independent print calls in case of a failure such that those three prints don't interleave with another print.
Alternatively, we could just build a string for the overall message, followed by a single print.
This way, we would get rid of all the locking.

Of course, this yields the rather philosophical question of whether we should treat the GIL as an implementation detail or as an intrinsic part of the language. To me it feels wrong to write code with unsynchronized access to shared resources.


In any case, I don't see how using map-reduce would make this particular code less complicated. We would have n reducers running in parallel, working on a list of test cases each.


- If we use idiomatic MR, we would collect the results (e.g., successes, failures, ignored) locally in each reducer rather than globally (as we do now). We would then have to accumulate and return those separate results after all reducers finished.

- I would like to see the results for each test case as soons as they finish, not at the end after all reducers finished. This means that the reducers would have to print the result immediately after each test case. We then face the same problem as with my current solution: Either use a lock to make several print calls atomic or create a single string to be printed and only print once.

- Python doesn't have library support for parallel map-reduce. This means that we would still have to use a pool of threads or processes to parallelize the reducers.

dhil: It seems there is something like a parallel map, c.f. [https://docs.python.org/3/library/multiprocessing.html#multiprocessing.pool.multiprocessing.Pool.map](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.pool.multiprocessing.Pool.map).

dhil:
Please wait with merging this until next week.

frank-emrich:
After discussing this is person with @dhil we agreed to use this as it is

PR #559, opened 2019-04-10
Rename clientTime to clientTimeMilliseconds and add clientTime

### LINKED ISSUE #230: clientTime
There are three builtin functions to get the current time:

```
 clientTime       : () ~> Int
 serverTime       : () ~> Int
 serverTimeMilliseconds       : () ~> Int
```

However, `clientTime` returns the time in milliseconds, not seconds.  I suggest adding `clientTimeMilliseconds` with this behavior and changing `clientTime` to return the time in seconds, like `serverTime`.  This will break current code but will hopefully unbreak the hearts of future generations of Links programmers.

They all need to be added to the documentation too.

jamescheney:
Fixed #230.

### DIFF
--- a/core/lib.ml
+++ b/core/lib.ml
@@ -1048,6 +1048,11 @@ let env : (string * (located_primitive * Types.datatype * pure)) list = [
    datatype "() ~> Int",
    IMPURE);
 
+  "clientTimeMilliseconds",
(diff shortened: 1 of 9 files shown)


jamescheney:
Should change the documentation before merging.

Issue #563, opened 2019-04-12
InsertReturning cleanup


jamescheney:
While working on #558 I've noticed some dd things about InsertReturning (which allows us to retrieve the value of a field after insertion, which is especially useful for fields with default or automatically-populated values.)

We can't anymore write:
```
insert factorials values (f,i) [(f=32,i=43)] returning i;
```
because the parser rule that should handle this expects the list expression to be a `db_expression` instead of `exp`.  Easy to fix.

We can also write insert returning commands that insert zero or many rows, and get a runtime error.
```
links> insert factorials values (f,i) [] returning i;
*** Fatal error : InsertReturning: undefined for empty list of rows
links> insert factorials values (f,i) [(f=1,i=1),(f=2,i=2)] returning i;
*** Runtime error: Returned the wrong number of results executing insert into factorials("f","i") VALUES (1,1), (2,2) returning "i"
```
It might be better to either return a list of the inserted fields, or specialize the InsertReturning syntax/typing so that it is impossible to insert zero or multiple rows.  We don't appear to be making heavy (or any?) use of this and there are no tests for InsertReturning syntax or behavior, so this change seems low-impact.

- [x] fix syntax for InsertReturning
- [x] add test cases for InsertReturning

jamescheney:
Edited to remove the "possibly change syntax to only allow inserting one thing".  Created separate issue for that.  The other things are fixed in #558.

jamescheney:
Also, note that fixing the `db_expression` to `exp` problem caused shift-reduce conflicts.  The `exp` needs to be `typed_expression` instead, for future reference.

jamescheney:
Closed by #564

PR #564, opened 2019-04-12
[WIP] add insert to IR, superfluate desugarDbs

### LINKED ISSUE #558: Cleanup or eliminate desugarDbs
`desugarDbs` currently does two things:

- `desugar_dbs` pass to remove the `DBInsert` abstract syntax and replace with calls to library functions
- `has_no_dbs` predicate to test whether `DBInsert` is present.  Never called from outside this module.

The code contains comments indicating that previously this stage handled `DBUpdate` and `DBDelete` syntax, no longer does, and that `DBInsert` syntax should also be moved to the IR:

```
TODO:

move insert into the IR

    insert table values rows
  -->
    InsertRows (table, rows)

    insert table values rows returning e
  -->
    InsertReturning (table, rows, field)
```

I think we should do this, and remove `has_no_dbs`, thereby decreasing the number of desugaring passes by one.

To do:
- [x] add insertion operations to `evalir.ml` (adapting code from `lib.ml`)
- [x] add `Insert` IR forms
- [x] extend EvalIR and IrCheck and other IR utilities to handle these forms
- [x] add `sugartoir` cases to translate `DBInsert` forms to their IR equivalents
- [x] disable `desugarDbs` and debug everything
- [x] delete dead `desugarDbs` and library `insert` code

jamescheney:
Fixed #558 and part of #563.  Seems to work, but leaving as WIP for review prior to final cleanup of old `desugarDbs` and `InsertRows`/`InsertReturning` code from `lib.ml` which should happen before merging.


core/irCheck.ml
```diff
@@ -745,6 +745,14 @@ struct
 
               Query (range, e, t), t, o
 
+        | InsertRows (_source, _rows) ->
+	    (* TODO: check that the type of rows being inserted matches the table type *)
+	    assert false
```
jamescheney: @frank-emrich, I haven't attempted to add code to typecheck the insertion operators.  I can take a shot at this and if this is straightforward, it would be good to do before merging, if not, I'll create an issue.

core/parser.mly
```diff
@@ -883,7 +883,7 @@ database_expression:
 | INSERT exp VALUES LPAREN record_labels RPAREN exp            { db_insert ~ppos:$loc $2 $5 $7 None }
 | INSERT exp VALUES LBRACKET LPAREN loption(labeled_exps)
   RPAREN RBRACKET preceded(RETURNING, VARIABLE)?               { db_insert ~ppos:$loc $2 (labels $6) (db_exps ~ppos:$loc($6) $6) $9  }
-| INSERT exp VALUES LPAREN record_labels RPAREN db_expression
+| INSERT exp VALUES LPAREN record_labels RPAREN exp
```
jamescheney: This fixes part of #563.

core/evalir.ml
```diff
@@ -10,6 +10,24 @@ open Var
...
+
+let row_columns : Value.t -> string list = function
+  | `List ((`Record fields)::_) -> List.map fst fields
+  | r -> failwith ("Internal error: forming query from non-row (row_columns): "^ Value.string_of_value r)
+
+let row_values db = function
+  | `List records ->
+      (List.map (function
+        | `Record fields -> List.map (value_as_string db -<- snd) fields
+        | _ -> failwith "Internal error: forming query from non-row") records)
```
SimonJF: Could you use `Errors.internal_error` instead of `failwith` here?

core/evalir.ml
```diff
@@ -729,6 +747,33 @@ struct
...
+	begin
+          match value env source, value env rows with
+          | `Table _, `List [] ->  apply_cont cont env (`Record [])
+          | `Table ((db, _params), table_name, _, _), rows ->
+              let field_names = row_columns rows in
+              let vss = row_values db rows in
+              Debug.print ("RUNNING INSERT QUERY:\n" ^ (db#make_insert_query(table_name, field_names, vss)));
+              let () = ignore (Database.execute_insert (table_name, field_names, vss) db) in
+	      apply_cont cont env (`Record [])
+          | _ -> failwith "Internal error: insert row into non-database"
```
SimonJF: `failwith` here, too, and a few below

core/frontend.ml
```diff
@@ -66,7 +66,7 @@ struct
        ->- after_typing_ext session_exceptions ((DesugarSessionExceptions.insert_toplevel_handlers tyenv)#program ->- snd3)
        ->- after_typing_ext session_exceptions ((DesugarSessionExceptions.desugar_session_exceptions tyenv)#program ->- snd3)
        ->- after_typing ((DesugarProcesses.desugar_processes tyenv)#program ->- snd3)
-       ->- after_typing ((DesugarDbs.desugar_dbs tyenv)#program ->- snd3)
+       (*->- after_typing ((DesugarDbs.desugar_dbs tyenv)#program ->- snd3)*)
```
SimonJF: Can this just be deleted, along with the file itself?

jamescheney: Yes, that's the plan.

core/evalir.ml
jamescheney: Thanks. The `failwith`s were what was there in the original code (this is copy-pasted from `lib.ml`).  But this seems like a reasonable time to kill them off.

core/evalir.ml
jamescheney: +1

jamescheney:
The CI run is exercising the new code; the old code is dead but not deleted yet.  I wanted to pause for any objections or comments on this change (particularly w.r.t. IR typing) before deleting code.

core/irCheck.ml
frank-emrich: I'm happy to create code for what I think should be happening here, so we can have a look at it together. I'd just commit it to this branch.

core/parser.mly
frank-emrich: I'm now getting the following warning:
```
      menhir core/parser.{ml,mli}
Warning: 2 states have shift/reduce conflicts.
Warning: 2 shift/reduce conflicts were arbitrarily resolved
```

core/irCheck.ml
jamescheney: Thanks!

core/parser.mly
jamescheney: Yes, I'm not sure why that is.  What was there before was definitely wrong (`db_expression` only derives delete/update expressions).

jamescheney: The reason is that `db_expression` used to have a fallthrough case to `typed_expression`, this was to avoid shift-reduce conflicts.  At some point, the fallthrough case was removed.  The right thing to do is probably to allow these INSERTs to end with `typed_expression` since update and delete operations in this position will never typecheck anyway.

jstolarek: Sadly, Menhir is always right when it reports conflicts. I was about to explain why the conflicts were happening but I'm glad I no longer have to :-)

Also, be careful with trailing whitespaces. I wonder whether CI will catch them?

frank-emrich: >  I wonder whether CI will catch them?

Yes, the rule checking script rejects certain files with trailing whitespace, including the parser files.

jstolarek: Indeed:

```
$ make rule-check
Applying rule check
Running 'tools/rules/trailing_whitespace'...
  - core/parser.mly
make: *** [rule-check] Error 1
```

jamescheney: Correct, my emacs for some reason inserts trailing whitespace for no apparent reason.

core/value.ml
```diff
@@ -6,6 +6,9 @@ open Var
 
 let _ = ParseSettings.config_file
 
+let internal_error message =
+  raise (Errors.internal_error ~filename:"instantiate.ml" ~message)
```
SimonJF: I think this might be a C&P error (`instantiate` --> `value`)

core/irCheck.ml
frank-emrich: I've pushed a commit dealing with this. Do you have an idea which of ours test cases or example has fancy polymorphic inserts? I would like to test this.

If it's easier to discuss in person, I'm happy to come around.

core/value.ml
jamescheney: Yup, thanks for catching.

core/irCheck.ml
```diff
@@ -745,6 +745,63 @@ struct
 
               Query (range, e, t), t, o
 
+        | InsertRows (source, rows)
+	| InsertReturning (source, rows, _) ->
```
jamescheney: We *could* unify these into a single constructor, with the `returning` an option, which is what the source AST does.  I did it this way because the translation from the previous library functions to `evalir` cases is more direct.  But it might actually be better/cleaner to do it the other way.

core/irCheck.ml
jamescheney: Thanks. I'm heading out shortly.  The code looks reasonable to me though.

As far as I know the only tests that exercise `insert` are in `tests/database/factorials.links`.  So running this with ir typing temporarily enabled should exercise this code.  In this PR I added some tests that exercise different syntaxes for inserts as well as `returning` clauses, to make sure that my changes work at least in common cases.  I'll be happy if IR typing works on these cases for now; if you want to add some additional examples to `factorials.links` that exercise the full generality of the IR typing code that would also be great.

Since we are already generating plenty of ill-typed IR, I'd rather get this merged and create an issue to revisit this later if there are obstacles to getting IR typing working for the specific case of inserts.

jamescheney: OK. So actually there look like there are several problems with typechecking the insert/delete/update operations (in addition to lots of other IR checking issues).  For example:
```
--------------------------------------------------------------------
Continuing after IR Type Error:
Internal error: attempt to extract a row from a datatype that is not a record or a variant: [(f:Int,i:Int)]
noccurring in IR special tail computation:
(Ir.InsertRows ((Ir.Variable 2066),
   (Ir.TApp ((Ir.Variable 56), [(f:Int,i:Int)]))))
-------------------------------------------------------------------
```
I think this is because in the line 
```
            let rows_r = TypeUtils.extract_row rows_t in

```
you first need to get rid of the list constructor on rows_t since it is of type list of record.

You can see the errors by running with `typecheck_ir` and `debug` enabled, without any database configuration, the IR typing messages will get printed out before run-time database errors.

frank-emrich: Thanks for letting me know. I'll play around with this a bit and fix the bugs in this new fragment immediately. I'll fix the bugs in existing code independently from this PR, to untangle the two things.

jamescheney:
I fixed another minor issue with IR typechecking of updates.  I think this is otherwise ready to go.  @SimonJF ?

core/irCheck.ml
jamescheney: Thanks, seems to work now.

Issue #567, opened 2019-04-17
queries returning unit result types fail at run time


jamescheney:
With shredding disabled, using `factorials.links` database:
```
links> query {for (x <-- factorials) where (true) [()]};
*** Fatal error : Column dummy had no type info in query's type spec: 
```
With shredding enabled:
```
for (x <-- factorials) [(a=x.i, b=for(y <-- factorials) where (x.i == y.i) [()])]
```
I get a different error:
```
*** Error: "Assert_failure core/query/evalNestedQuery.ml:745:17"
```

Issue #490, opened 2019-02-20
Queries with polymorphic types fail at run-time


jamescheney:
Thanks to @SimonJF for running into this.

Consider the following (using the tutorial factorials database):
```
links> var db = database "links";
links> var factorials = table "factorials" with (i : Int, f : String) from db;
links> query { for (x <-- factorials) [] };
*** Error: "Assert_failure core/query.ml:156:62"
links> query { for (x <-- factorials) [(a=1)] };
[(a = 1), (a = 1), (a = 1), (a = 1), (a = 1), (a = 1), (a = 1), (a = 1),
 (a = 1), (a = 1), (a = 1), (a = 1), (a = 1), (a = 1), (a = 1), (a = 1),
 (a = 1), (a = 1), (a = 1), (a = 1)] : [(a:Int)]
```

If the query result is an empty list then the assertion failure results from `Query.type_of_expression` trying to guess the query result type and failing because the empty list can have any type.

The/a principled way to handle this would be to retain enough information about the result types of queries (possibly using run-time type information) to do the right thing.

A workaround for now is to avoid writing queries that don't have easy-to-determine types (according to `Query.type_of_expression`).  We should also change the error message to something more enlightening.

SimonJF:
Somewhat worryingly, I've found that this can also manifest itself at runtime, when the result of a query returns the empty list. I haven't properly shrunk this example yet, but consider the following fragment of a query:

```
 [..]
              for (pp <-- Tables.pathophysiology)
                where (pp.disease_id == diseaseId)
                for (o <-- Tables.object)
                  where (o.object_id == pp.object_id)
                    [(id=o.object_id, name=o.name, comments=pp.comments,
                      role=pp.role, drugComments=pp.drugs, sideEffects=pp.side_effects,
                      therapeuticUse=pp.use,
                      references =
                        for (pp_ref <-- Tables.pathophysiology_refs)
                          where (pp_ref.pathophysiology_id == pp.pathophysiology_id)
                          [pp_ref.reference_id])]
```

I've found that if the result of the query determining `references` returns an empty result set, then the whole program errors with the above assert failure. This seems like it only happens with nested queries (several levels deep, usually).

Putting the result in a record type doesn't seem to help, either.

SimonJF:
So I hacked around this:
https://github.com/SimonJF/links/commit/fe036509843a056053c6bd4c1a753b391caeae85

[...]

PR #568, opened 2019-04-17
resolve issue 490

jamescheney:
Seems to fix the problem with polymorphic flat queries, but shredding/nested queries still need work.

@wricciot, over to you?


jamescheney:
NB: The same approach (using a designated name for fields whose type is unit) also works for #567 at least in the case of flat queries.  For nested queries the errors encountered in #567 still occur so more work is needed.

jamescheney:
If this now fixed #490 for both flat and nested queries then we should add tests similar to `empty.links` to the `shredding` tests directory.  (Including one that deals with an empty-list subexpression arising as a result of substituting in a function definition and normalizing, which was the original situation that @SimonJF encountered relating to #490 though it took us a little time to realize it.)

jamescheney:
Wilmer and I just discussed and added some more interesting tests (for both flat and nested queries).  This should be ready to merge once the CI finishes.

jamescheney:
This also fixes #567.

jamescheney:
@slindley, could you take a quick look at this before merging?

core/query/evalNestedQuery.ml
```diff
@@ -732,17 +732,29 @@ Avoiding unnecessary static indexes, or multiplexing pairs (a,d) where a is usua
...
+      | _ -> assert false
+    in
+    let idx_and_val = function
+      | `Record fl ->
+           let w_tmpl = try List.assoc "2" fl with NotFound _ -> `Record [] in
+           (match List.assoc "1" fl with
+            | `Record fl' -> (match List.assoc "1" fl', List.assoc "2" fl' with
+              | `Primitive (_,a_idx), `Primitive (_,d_idx) -> (a_idx,d_idx,w_tmpl)
+              | _ -> assert false)
+            | _ -> assert false)
```
slindley: I find `begin ... end` easier to read for nested pattern matching than `(...)`

jamescheney:
I get what is going on now: in the shredding code we are assuming that all of the queries are returning (at least) some special index fields and Wilmer's changes workaround that.

I think this is symptomatic of the fact that the current interface boundary between ordinary evaluation and (shredded) query evaluation is at the wrong place: shredding returns a shredded package decorated with SQL query strings, and in some cases when we statically know the query result, but the result is not something that SQL can straightforwardly express, we should have the option of just calculating the (empty) result directly and not round-tripping to the database.  However, this workaround works for now and I think we should merge this and create an issue to record this requirement as part of the logner-term goal of refactoring the query subsystem to make this work better.

PR #569, opened 2019-04-22
Escape carriage returns in JSON-quoted strings

SimonJF:
Carriage returns ("\r") were not being correctly escaped, leading to runtime errors.

I tried using `String.escaped` (as with `irtojs`), but some intricacies seemed to be getting in the way there (I think `String.escaped` was doing more than we wanted it to / was safe to do).

### DIFF
--- a/core/json.ml
+++ b/core/json.ml
@@ -48,8 +48,9 @@ let js_dq_escape_string str =
   (* escape for placement in double-quoted string *)
   Str.global_replace (Str.regexp_string "\"") "\\\""
     (Str.global_replace (Str.regexp_string "\n") "\\n"
+      (Str.global_replace (Str.regexp_string "\r") "\\r"
        (Str.global_replace (Str.regexp_string "\\") "\\\\"
-          str))
+          str)))
 
 (** Escape the argument for inclusion in a double-quoted string. *)
 let js_dq_escape_char =


core/json.ml
```diff
@@ -48,8 +48,9 @@ let js_dq_escape_string str =
   (* escape for placement in double-quoted string *)
   Str.global_replace (Str.regexp_string "\"") "\\\""
     (Str.global_replace (Str.regexp_string "\n") "\\n"
+      (Str.global_replace (Str.regexp_string "\r") "\\r"
```
dhil: Maybe we should roll our own escape function. For example we could do something along the lines of
```ocaml
(* Handling some (ascii) escape sequences. *)
module String = struct
  include String

  let encode s =
    let rec loop s i len =
      if i < len
      then Char.code s.[i] :: loop s (i + 1) len
      else []
    in
    loop s 0 (String.length s)

  let decode bs =
    let len = List.length bs in
    let rec loop dest i = function
      | b :: bs ->
         Bytes.set dest i (Char.chr b);
         loop dest (i + 1) bs
      | [] -> dest
    in
    Bytes.to_string (loop (Bytes.create len) 0 bs)

  let escape_ascii s =
    let is_escapable = function
      (* tab, newline, carriage return, double quote, backslash *)
      | 9 | 10 | 13 | 34 | 92 -> true
      | _ -> false
    in
    let escape_char = function
      | 9 -> 116  (* tab -> t *)
      | 10 -> 110 (* newline -> n *)
      | 13 -> 114 (* carriage return -> r *)
      | c -> c
    in
    let rec esc = function
      | b :: bs when is_escapable b -> 92 :: (escape_char b) :: esc bs
      | b :: bs -> b :: esc bs
      | [] -> []
    in
    decode (esc (encode s))
end
```
It seems to work with backslash and newlines at least
```ocaml
# print_endline (String.escape_ascii (really_input_string stdin (String.length "Hello World\nfoobar")));;
Hello\World
foobar
Hello\\World\nfoobar
```
Incidentally, this method may also be more efficient as it performs all replacements in one linear pass. With a little more work, the supported escape sequences could be made parameters.

dhil: I meant to say, then we could reuse the escape function in other places. I seem to remember, that we have some string escaping in other modules as well. Maybe taking it a step further, if we want to be really robust, then we should use the same escape/unescape functions on the server-side and client-side, rather than relying on the built-in ones.

dhil:
Happy birthday, @SimonJF!

SimonJF:
Haha, thanks @dhil! We should definitely do it properly at some point though, as you suggest.

PR #570, opened 2019-04-24
[WIP] Some more errors easy wins

### LINKED ISSUE #525: Phase out `failwith` for error reporting
`failwith` is often used in parts of the frontend prior to typechecking to report error messages. It has the disadvantage of being rather ad-hoc, and often leads to lazy error message reporting.

Given all of the improvements to the position propagation recently, I wonder whether it might be a good idea to try and phase out `failwith` in favour of using the `Errors` module, with proper position reporting.

SimonJF:
More on my quest to fix #525; hoping to get all of the `failwith`s except the ones in the lens code before this is ready to merge.


core/utility.ml
```diff
@@ -982,6 +982,8 @@ let getenv : string -> string option =
 let safe_getenv s =
   try Sys.getenv s
   with NotFound _ ->
+    (* We need to retain this `failwith` since `errors.ml` depends on
```
jstolarek: Does OCaml have a way to break module cycles?

core/jsonparse.mly
```diff
@@ -150,7 +153,7 @@ object_:
                                             (Value.unbox_string (List.assoc "name" bs),
                                              Value.unbox_string (List.assoc "args" bs)) in
                                           `Database (Value.db_connect driver params)
-                                    | _ -> failwith ("jsonparse: database value must be a record")
+                                    | _ -> internal_error ("jsonparse: database value must be a record")
```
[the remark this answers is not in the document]
SimonJF: Ooh, no, good spot.

core/utility.ml
dhil: Yes. You can always break a cycle by introducing another layer of indirection, for example by functoring the module.

Morally, I don't think we should use the `errors` module here anyways, as errors originating from the `utility` module ought to be interpreted at the use sites, rather than directly by the definition of `utility`. What we really should do, is to define and export meaningful exceptions in the `utility` module that may be raised by utility functions.

core/errors.ml
```diff
@@ -70,17 +73,6 @@ let format_exception =
       let pos, expr = Position.resolve_start_expr pos in
         Printf.sprintf "%s:%d: Syntax error: %s\nIn expression: %s\n"
           pos.pos_fname pos.pos_lnum s expr
-  | Sugartypes.RedundantPatternMatch pos ->
```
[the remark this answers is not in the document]
SimonJF: They are unused in the source code (grepping for `RedundantPatternMatch` yields no results other than the definition). Additionally, it introduces a module cycle (Sugartypes relies on `errors`, and in order to print the error, `errors` relies on `Sugartypes`).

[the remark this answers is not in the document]
dhil: Currently, we have no support for exhaustiveness and redundancy checking of pattern (in the past we used to). I think it is fine to delete it, we can reintroduce it if we decide to support checking of patterns again in the future.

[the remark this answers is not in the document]
SimonJF: Indeed: I think we would need to redefine in `errors.ml` anyway.

core/errors.ml
```diff
@@ -105,6 +97,8 @@ let format_exception =
       let pos, expr = Position.resolve_start_expr pos in
       Printf.sprintf "%s:%d: Kind mismatch: Type argument %d for type constructor %s has kind %s, but an argument of kind %s was expected. \nIn:\n%s\n"
           pos.pos_fname pos.pos_lnum tyarg_number name provided expected expr
+  | SettingsError message ->
+      Printf.sprintf "*** Settings Error: %s" message
```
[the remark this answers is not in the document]
SimonJF: Excellent suggestion. Working on that now.

core/jsonparse.mly
```diff
@@ -10,10 +10,13 @@ struct
   module Value = Value
 end
 
+let internal_error message =
+  raise (Errors.internal_error ~filename:"jsonparse.mly" ~message)
+
 (* let unparse_label = function *)
```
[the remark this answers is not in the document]
SimonJF: I'd prefer to keep it, so we don't have to keep repeating the `~filename:"jsonparse.mly" ~message:...` boilerplate throughout the file.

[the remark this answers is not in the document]
jstolarek: No, no, I meant the commented-out piece of code.

jamescheney:
I'm not sure if this is the right place to address this, but Wilmer and I noticed (while looking at #490) that when you use `internal_error` defined as a separate function (with hardwired filename) you get a stack trace that shows the `raise` in the `internal_error` function, not at the `internal_error` call site.  This, combined with the fact that some `internal_error` messages aren't easily searchable, is unhelpful, though the previous `failwith` approach had the same problem.  It would be nice if there were some way of treating `internal_error` as a macro rather than a function call, so that what OCaml's exception stack tracing does by default would be the right thing.

SimonJF:
Hrm, well-spotted. I wonder whether using `[@@inline]` (c.f. https://caml.inria.fr/pub/docs/manual-ocaml/extn.html) would be helpful here.

jamescheney:
It depends on whether inlining is semantics-preserving (i.e. correct) or not...  or at least, whether the inlining happens before exception raise sites are tagged with line number information or after.

dhil:
> [..] when you use `internal_error` defined as a separate function (with hardwired filename) you get a stack trace that shows the `raise` in the `internal_error` function, not at the `internal_error` call site. 

I suppose you can get around this by making `internal_error` construct the exception, not raise it, then every use of `internal_error` should be prefixed by a `raise`.

SimonJF:
@dhil -- yes, that seems sensible, and less ad-hoc than inlining. Thanks.

jamescheney:
Oh yeah, exceptions are first class.  Good call @dhil !

SimonJF:
OK, this should be the "easy wins" done now. It eliminates most of the non-lens "failwith" cases, and the ones which don't need too much refactoring to get the right errors displaying. I'll have a look at the rest soon, but I think this can go in now (pending review of course).

I've also fixed @jamescheney and @wricciot's problems by implementing @dhil's suggestion and ensuring exceptions are raised at the correct places.

core/jsonparse.mly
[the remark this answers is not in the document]
SimonJF: Yep, gone now.

core/jsonparse.mly
```diff
@@ -11,12 +11,6 @@ struct
 end
 
 let internal_error message =
-  raise (Errors.internal_error ~filename:"jsonparse.mly" ~message)
```
jstolarek: I think this line should stay?

SimonJF: ...whoops, seems I was a little eager with my delete button. Thanks, fixed.

SimonJF:
That should be all comments so far addressed, and the CI's no longer complaining :)

SimonJF:
@jstolarek: Have the revisions satisfied your review? If so, are we set to merge?

jstolarek:
Sorry, this went off my radar. I'll review the patch later today.

SimonJF:
Thanks! Not a problem at all, whenever you get chance :)

core/errors.ml
```diff
@@ -39,74 +41,82 @@ exception TypeApplicationArityMismatch of
...
-      ("*** Parse error: " ^ s.filename ^ ":"
+      pos_prefix ("Parse error: " ^ s.filename ^ ":"
        ^ s.linespec ^ "\n"
        ^ s.message ^ "\n" ^ prefix_lines "  " s.linetext ^ "\n"
        ^ "   " ^ s.marker)
   | DesugaringError { pos; stage; message } ->
       let pos, expr = Position.resolve_start_expr pos in
-      Printf.sprintf "%s:%d: Error %s: %s\nIn expression: %s.\n"
-        pos.pos_fname pos.pos_lnum (string_of_stage stage) message expr
+      pos_prefix ~pos:(Some pos)
```
jstolarek: Just `pos_prefix ~pos`

core/errors.ml
```diff
@@ -39,74 +41,82 @@ exception TypeApplicationArityMismatch of
...
-  | IRTypeError msg -> Printf.sprintf "IR Type Error: %s" msg
+      pos_prefix ~pos:(Some pos)
+        (Printf.sprintf "Type error: %s\nIn expression: %s.\n" s expr)
+  | IRTypeError msg -> pos_prefix (Printf.sprintf "IR Type Error: %s" msg)
   | UnboundTyCon (pos, tycon) ->
       let pos, _ = Position.resolve_start_expr pos in
-      Printf.sprintf "%s:%d: Unbound type constructor %s\n"
-                    pos.pos_fname pos.pos_lnum tycon
-  | Runtime_error s -> "*** Runtime error: " ^ s
+      pos_prefix ~pos:(Some pos)
```
jstolarek: Just `pos_prefix ~pos`

core/errors.ml
```diff
@@ -39,74 +41,82 @@ exception TypeApplicationArityMismatch of
...
+        (Printf.sprintf "Error %s: %s\nIn expression: %s.\n"
+           (string_of_stage stage) message expr)
   | Getopt.Error s -> s
   | Type_error (pos, s) ->
       let pos, expr = Position.resolve_start_expr pos in
-        Printf.sprintf "%s:%d: Type error: %s\nIn expression: %s.\n"
-          pos.pos_fname pos.pos_lnum s expr
-  | IRTypeError msg -> Printf.sprintf "IR Type Error: %s" msg
+      pos_prefix ~pos:(Some pos)
```
jstolarek: Just `pos_prefix ~pos`

core/errors.ml
```diff
@@ -39,74 +41,82 @@ exception TypeApplicationArityMismatch of
...
-          pos.pos_fname pos.pos_lnum expr
-  | Sugartypes.PatternDuplicateNameError(pos, name) ->
-      (* BUG: this can't be right (the pattern and the expression are the same!) *)
-      let pos, expr = Position.resolve_start_expr pos in
-      let pattern = expr in
-        Printf.sprintf
-          "%s:%d: Syntax Error: Duplicate name `%s' in pattern\n  %s\nIn expression: %s"
-          pos.pos_fname pos.pos_lnum name (xml_escape pattern) (xml_escape expr)
-  | Failure msg -> "*** Fatal error : " ^ msg
+      pos_prefix ~pos:(Some pos)
```
jstolarek: Just `pos_prefix ~pos`

core/errors.ml
```diff
@@ -39,74 +41,82 @@ exception TypeApplicationArityMismatch of
...
+          StringMap.fold (fun name positions message ->
+                            message^" "^name^":\n  "^
+                  (mapstrcat "\n  " Position.show (List.rev positions)))
+          duplicates "")
   | InvalidMutualBinding pos ->
       let pos, expr = Position.resolve_start_expr pos in
-      Printf.sprintf
-        "%s:%d: Mutual blocks can only contain `fun` and `typename` bindings, but the block contained: %s.\n"
-        pos.pos_fname pos.pos_lnum expr
+      pos_prefix ~pos:(Some pos)
```
jstolarek: Just `pos_prefix ~pos`

core/errors.ml
```diff
@@ -39,74 +41,82 @@ exception TypeApplicationArityMismatch of
...
-        filename message
+      pos_prefix
+      (Printf.sprintf
+         "Internal Error in %s (Please report as a bug): %s\n"
+         filename message)
   | TypeApplicationArityMismatch { pos; name; expected; provided } ->
       let pos, expr = Position.resolve_start_expr pos in
-      Printf.sprintf ("%s:%d: Arity mismatch: Type %s expects %d type arguments, but %d arguments were provided. In: %s\n")
-          pos.pos_fname pos.pos_lnum name expected provided expr
+      pos_prefix ~pos:(Some pos)
```
jstolarek: Just `pos_prefix ~pos`

core/errors.ml
```diff
@@ -39,74 +41,82 @@ exception TypeApplicationArityMismatch of
...
+        (Printf.sprintf ("Arity mismatch: Type %s expects %d type arguments, but %d arguments were provided. In: %s\n")
+             name expected provided expr)
   | TypeApplicationKindMismatch { pos; name; tyarg_number; expected; provided } ->
       let pos, expr = Position.resolve_start_expr pos in
-      Printf.sprintf "%s:%d: Kind mismatch: Type argument %d for type constructor %s has kind %s, but an argument of kind %s was expected. \nIn:\n%s\n"
-          pos.pos_fname pos.pos_lnum tyarg_number name provided expected expr
+      pos_prefix ~pos:(Some pos)
```
jstolarek: Just `pos_prefix ~pos`

core/evalir.ml
```diff
@@ -277,7 +277,7 @@ struct
            with
                  UnknownProcessID _ ->
                    (* FIXME: printing out the message might be more useful. *)
-                   internal_error("Couldn't deliver message because destination process has no mailbox.")) >>= fun _ ->
+                   raise (internal_error("Couldn't deliver message because destination process has no mailbox."))) >>= fun _ ->
```
jstolarek: Would it be easy to do what the FIXME is saying?

core/query/query.ml
```diff
@@ -378,7 +390,8 @@ struct
       | `PrimitiveFunction (f,_) -> Primitive f
           (*     | `ClientFunction f ->  *)
```
jstolarek: Can we delete these commented-out cases?

core/desugarLAttributes.ml
```diff
@@ -15,7 +15,7 @@ open SourceCode.WithPos
 
 let desugaring_error pos message =
   let open Errors in
-  raise (desugaring_error ~pos ~stage:DesugarLAttributes ~message)
+  desugaring_error ~pos ~stage:DesugarLAttributes ~message
```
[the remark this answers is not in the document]
SimonJF: You raise a good point in that the exact raise location does not matter as much for user errors as for internal errors. 

That said, I don't think it hurts either way: the user won't see a stack trace by default. I've gone for uniformity, but if you would rather I factored out the `raise` for user errors, I'm happy to do so.

core/errors.ml
```diff
@@ -39,74 +41,82 @@ exception TypeApplicationArityMismatch of
...
     expected: string; provided: string }
+exception SettingsError of string
+exception DynlinkError of string
+exception ModuleError of string
 
 
 let prefix_lines prefix s =
   prefix ^ Str.global_replace (Str.regexp "\n") ("\n" ^ prefix) s
 
+let pos_prefix ?pos:(pos=None) line =
```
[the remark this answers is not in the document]
SimonJF: Neat! Thanks!

core/evalir.ml
SimonJF: Yep. Done so, thanks.

core/query/query.ml
SimonJF: Yep. I've also taken the opportunity to print the value, so we get a more informative error message.

core/desugarLAttributes.ml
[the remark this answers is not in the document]
jstolarek: I'd be all for uniformity. I complained because this seemed to me like a lack of uniformity, i.e. `desugaring_error` is a user error but unlike other such errors it does not include `raise` in its definition.

SimonJF:
Thanks for the review, @jstolarek! (Especially the optional arguments syntax; that's a really good one to know).

I've addressed everything except the raise-sites for non-user errors. I'm in favour of leaving them as is for the sake of uniformity, but if you have strong feelings, I'll hoist them out.

core/evalir.ml
```diff
@@ -275,9 +275,12 @@ struct
               | `ClientPid (client_id, process_id) ->
                   Mailbox.send_client_message msg client_id process_id
            with
-                 UnknownProcessID _ ->
+                 UnknownProcessID id ->
                    (* FIXME: printing out the message might be more useful. *)
```
jstolarek: Do we still need that FIXME comment?

I am also surprised to see an exception being replaced with a debug message. But perhaps that's fine - I don't really understand what's going on here.

core/desugarLAttributes.ml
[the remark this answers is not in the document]
SimonJF: Ah, I see! I missed that there are other such helper functions which do include raise in their definitions. Thanks for clarifying.

In that case, I will ensure that all of these exception generation functions do not `raise`, so everything is uniform.

jamescheney:
The CI is failing here due to the same problem I ran into last night in another PR, which I fixed in 045435b 

I am inclined to merge #568 now so that rebasing on master avoids this problem.  Or you can cherry-pick 045435b

SimonJF:
Thanks, @jamescheney. I'd be happy to wait until #568 is merged rather than cherrypicking.

core/evalir.ml
SimonJF: Oops, the FIXME can go.

I think the FIXME was saying that a failed message send to a mailbox (rather than a session channel) needn't be fatal, along with requiring a bit more information.

jstolarek:
I cherry-picked the CI-fixing commit onto master

jamescheney:
I think we agreed this can be merged after #568

jamescheney:
Or we can merge now, since the CI fix was cherrypicked by @jstolarek already.

SimonJF:
I'd be happy to have it merged now if others would be OK with it. I've updated the branch; let's see whether the CI passes now, and then we could?

jamescheney:
CI passed; merging

PR #577, opened 2019-04-29
Resolves issue #576

### LINKED ISSUE #576: Regression: Formlets generate unsound code
Clicking either "Checkout" or "Cancel" in the "A simple test formlet" application on the examples' page causes an evaluation error
```
Links Error

*** Error: Links_core.Evalir.Exceptions.EvaluationError("Error projecting label 4")
```
Here is a small reproduction
```
fun myForm() server {
  formlet
    <#>
     <span>{ input -> text }</span>
     <span>{ submitButton("Submit") -> submit }</span>
    </#>
  yields
    (text = text)
}

fun processForm(data) server {
  page
   <#>
     {stringToXml(data.text ^^ "")}
   </#>
}

fun formPage(_) server {
  page
    <#>
      {myForm() => processForm}
    </#>
}

fun main() {
  addRoute("/", formPage);
  servePages()
}

main()
```
Clicking "Submit" causes the following error
```
Links Error

*** Internal Error in value.ml (Please report as a bug): Type error unboxing string: (1 = "") 
```

dhil:
The "smart" constructor `tuple' has a rather hacky method for
preserving 1-tuples, which is not always correct. In particular, it
introduced a soundness regression for formlets.

This commit augments the interface of `tuple' with an optional boolean
flag to enable or disable the said hack. By default the hack is on.

Resolves #576.

### DIFF
--- a/core/desugarFormlets.ml
+++ b/core/desugarFormlets.ml
@@ -127,7 +127,7 @@ object (o : 'self_type)
                         let base : phrase =
                           fn_appl pure_str [`Type ft; mb]
                             [fun_lit ~args:(List.rev args) dl_unl (List.rev pss)
-                                     (tuple vs)] in
+                                     (tuple ~one_tuple_hack:false vs)] in
                         let p, et =
                           List.fold_right
                             (fun arg (base, ft) ->
--- a/core/sugarConstructors.ml
+++ b/core/sugarConstructors.ml
@@ -92,8 +92,8 @@ module SugarConstructors (Position : Pos)
   let record ?(ppos=dp) ?exp lbls = with_pos ppos (RecordLit (lbls, exp))
 
   (* Create a tuple.  Preserves 1-tuples. *)
-  let tuple ?(ppos=dp) = function
-    | [e] -> record ~ppos [("1", e)]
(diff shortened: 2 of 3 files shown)


jstolarek:
Hm.... that looks like it would be my fault, but I could swear I copied this verbatim from the parser. Anyway, the only nitpick is that `one_tuple_hack` is not a very informative name. Perhaps `one_tuple_as_record` or something along those lines?

slindley:
How would copying this from the parser help?

(The only place I see it used in the parser is for `orderby`.)

You changed a whole load of invocations of `TupleLit` in the desugaring code to invoke a smart constructor that behaves differently when supplied with a single value. Please can you check all of the other cases to make sure they do the right thing (i.e. what they did before). Probably easiest to grep `TupleLit` on the commit immediately before 73e5892b51823d9207759d680ec9332b53e132fd.

If you didn't explicitly refactor any of the other code, then every instance of `tuple` in desugaring should be operating without the 1-tuple hack, and perhaps you've implemented the wrong smart constructor.

In the case of desugarFormlets.ml many of the cases don't matter either way because we already know the argument is not a list of length 1.

PR #579, opened 2019-04-30
Sound types for the `addRoute'-family of functions.

dhil:
This patch refines the type of `unsafeAddRoute`, and its relatives, in
particular `addRoute`, which previously had the type

    addRoute : (String, (String) ~> Page) ~> ()

which elaborates to

    addRoute : (String, (String) ~e~> Page) ~f~> ()

where `e` and `f` are distinct effect variables. Consequently the
invariant that all user-definable effectful operations are handled
prior to reaching the toplevel is broken, for instance

    addRoute("/", fun(_) {do Fail})

would cause a "no handler installed for Fail" error to occur at
runtime when visiting the root page, even though one might have
wrapped the `addRoute` in a handler, i.e.

    handle(addRoute("/", fun(_) {do Fail})) {
      case Fail -> ()
    }

This does not handle the effects of the anonymous function, because the function
gets executed in a separate context where the handler is not
installed.

To ensure that all user-definable computational effects are handled,
this patch "corrects" the type of `unsafeAddRoute` and its associated
functions, as a result the type of `addRoute` is now

    addRoute : (String, (String) {hear{_}}~> Page) ~> ()

which elaborates to

    addRoute : (String, (String) {hear{p}}~> Page) ~f~> ()

In other words, a page generating function must handle all of its
effects -- save for the two built-in effects `hear` and `wild`.

### DIFF
--- a/core/lib.ml
+++ b/core/lib.ml
@@ -1530,7 +1530,11 @@ let env : (string * (located_primitive * Types.datatype * pure)) list = [
      IMPURE);
     "unsafeAddRoute",
     (`PFun (fun _ -> assert false),
-     datatype "(String, (String, Location) ~> a, (String, String, Location) ~> a) ~> ()",
+     (* The `hear' effects on the second argument (request handler)
+        and third argument (error handler) should have different
+        presence variables as they are executed in different
+        contexts. *)
+     datatype "(String, (String, Location) {hear{_}}~> a, (String, String, Location) {hear{_}}~> a) ~> ()",
      IMPURE);
     "servePages",
     (`PFun (fun _ -> assert false),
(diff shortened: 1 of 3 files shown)


dhil:
> I have a slightly bigger concern however. I tried your original example (but with servePages added afterwards, so I could try it):

I don't see how that example can type check with this patch, and indeed on my branch it fails to type check.

SimonJF:
Apologies; I was on the wrong branch (d'oh, more coffee needed). I've rechecked everything and it's fine. Thanks!

Please add those unit tests, as well as 

```
handle({ addRoute("/", fun(_) {do Fail}); servePages() }) {
  case Fail -> debug("failed")
}
```

and I think we're good to go.

SimonJF:
Well spotted, and thanks for the patch!

PR #582, opened 2019-04-30
remove accidental upload of polling script

rudihorn:
Didn't mean to upload this file during my last PR.


jstolarek:
If this is a script that you keep in your local copy of the Links repo, you can edit `.git/info/exclude` to make sure this file does not get committed accidentally. The said file works just like `.gitignore`, except that it is local to a particular copy of a repo.

Issue #571, opened 2019-04-26
About the semantics of `open`


dhil:
Automatic resolution and loading of external modules is part of the current Links semantics. For example consider two modules `foo.links` and `bar.links`
```links
# foo.links
var hello = "Hello World";
```
```links
# bar.links
open Foo;

fun main() {
  println(hello)
}
main()
```
Executing `./links -m --path=. bar.links` prints out `Hello World`, that is, the compiler has implicitly loaded the external module `foo.links` for us.

However, if you were to slightly rewrite `bar.links` to
```links
# bar.links
fun main() {
  println(Foo.hello)
}
main()
```
then executing `./links -m --path=. bar.links` yields a compile time error
```
bar.links:2: Type error: Unknown variable Foo.hello.
```
Thus the compiler did not load `foo.links` for us, which may be a little surprising if we think of ML languages. However, it is arguably OK, because a simple typo could potentially cause the compiler to load an arbitrary amount of modules.

We can instruct the compiler to load `foo.links` for us by adding back the `open`
```links
open Foo;

# bar.links
fun main() {
  println(Foo.hello)
}
main()
```
Note that `hello` is still prefixed by `Foo`. Running `./links -m --path=. bar.links` prints out `Hello World` just as the first program did. So `open` really acts as an "import statement" instructing the compiler to go figure out how to load the said module.

Now, we also have support for local module definitions. In particular, local definitions may shadow external file-based modules.
```links
open Foo;

module Foo {
  var hello = "Hello from Bar!";
}

open Foo;

fun main() {
  println(hello)
}
main()
```
Running `./links -m --path=. bar.links` prints `Hello from Bar!`. The `bar.links` program contains two `open Foo` expressions. The two are morally different. The first instance **imports a global module**, whilst the second instance **opens a local module** (or more generally, opens a module that has already been loaded [local modules are trivially loaded]).

An import entails resolving a module on disk and transitively load its dependencies, which is quite convenient, but as I argue it is different from simply "opening" a module to bring its contents into scope.

I would like the difference to be reflected in the language for two reasons

* It can lead to mildly confusing behaviour.
* The implementation of the loader can be vastly simplified, as currently the loader needs to understand the binding structure and semantics.

I propose to add `import` (or `using`) as a primitive, whose intended semantics is to load global modules, e.g.
```links
# bar.links
import Foo;

fun main() {
  println(hello)
}
main()
```
[...]

PR #599, opened 2019-05-07
Introduce syntax for imports

dhil:
This patch is a step towards the implementation of #571.

This patch introduces the syntax for imports, at the moment they are operationally equivalent to opens, however, that will change in a subsequent patch.

### DIFF
--- a/core/chaser.ml
+++ b/core/chaser.ml
@@ -39,7 +39,8 @@ object(self)
     {< shadow_table = shadow_table >}
 
   method! bindingnode = function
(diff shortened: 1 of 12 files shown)


dhil:
Hah, this patch has a great side effect. It lets me run the new loader and the old one simultaneously. I should be able to gradually move responsibility to the new loader.

jamescheney:
This seems ready to merge, to enable #602?

dhil:
@jamescheney yes! I shall merge it later today.

Issue #596, opened 2019-05-05
Illeity


dhil:
I have been working towards an implementation for #571. While trying to understand what the behaviour of the current chaser/loader is, and in particular how it deals with some edge cases, I discovered this gem.

It is possible for a top-level module to clone itself (as in code being emitted multiple times), consider the following program
```links
# foo.links
open Foo;

fun main() {
  println("Hello World")
}

Foo.main()
```
The file `foo.links` requests the module `Foo` to be loaded, and invokes the `main` function from that module. Exactly what the symbolic name `Foo` refers to depends on the runtime namespace (or path, which is the current term of arts in Links). By running `./links -m foo.links` I would expect to see an unbound module error, since I haven't provided a path. However, current behaviour suggests that in the absence of a path `.` is used, so at the very least I would expect a cyclic dependency error. Neither happens, instead Links generates code for `foo.links` twice:
```
# omitted the IR for prelude.links for brevity
([(Ir.Fun
     ((318, (() ~> (), "Foo.main", Var.Scope.Global)),
      ([a::Row], [],
       ([],
        (Ir.Apply ((Ir.TApp ((Ir.Variable 240), [{ |_ }])),
           [(Ir.Constant (CommonTypes.Constant.String "Hello World"))])))),
      None, CommonTypes.Location.Unknown));
   (Ir.Fun
      ((319, (() ~> (), "main", Var.Scope.Global)),
       ([a::Row], [],
        ([],
         (Ir.Apply ((Ir.TApp ((Ir.Variable 240), [{ |_ }])),
            [(Ir.Constant (CommonTypes.Constant.String "Hello World"))])))),
       None, CommonTypes.Location.Unknown))
   ],
 (Ir.Apply ((Ir.TApp ((Ir.Variable 318), [{  }])), [])))
```
As you can tell, the generated code contains two copies of the bindings of `foo.links`. Curiously, the tail computation `Foo.main()` is only emitted once. I suspect this is because the chaser/loader inlines bindings from dependencies, which is incorrect.

Consider the scenario, where we have `foo.links` and `examples/foo.links`, where the latter may be
```links
# examples/foo.links
open Foo;

fun main() {
  println("Hello Examples")
}

Foo.main()
```
Then running `./links -m --path=examples examples/foo.links` generates the following code
```
([(Ir.Fun
     ((318, (() ~> (), "Foo.main", Var.Scope.Global)),
      ([a::Row], [],
       ([],
[...]

Issue #337, opened 2018-05-31
A proper module system


dhil:
As discussed during the previous Links meeting, the current module system is causing some issues, and rather than treating it symptomatically, we ought to design and implement a proper module system.

I envisage implementing the module system in stages, such that each stage successively increases the richness of the system, e.g.

1. A simple namespace
2. Encapsulation and abstract types
3. Higher-order modules (i.e. functors)
4. First-class modules
5. Separate compilation (possibly relevant for some future compiler)

In principle each stage can be implemented by a different person. I am happy to lay the foundation, that is to implement basic namespaces properly.

I think Xavier Leroy's [Modular Module System](https://caml.inria.fr/pub/papers/xleroy-modular_modules-jfp.pdf) provides a good starting point for an initial implementation. As we start considering more advanced features, we may want to move towards Andreas Rossberg's [1ML](https://people.mpi-sws.org/~rossberg/papers/Rossberg%20-%201ML%20--%20Core%20and%20modules%20united%20[JFP%20Draft].pdf) approach.

The initial implementation is intended to fix the following open issues: #292 #278 #225.

SimonJF:
Thoroughly agreed :)

PR #602, opened 2019-05-14
Desugar modules properly

### LINKED ISSUE #292: Modules and FFI don't go well together.
Minimal example:
```
# a.links
open B
42
```
```
# b.links
alien javascript "/my/js/file.js" {
  foo: () ~> ();
}
```

```
links -m a.links
```
yields
```
*** Error: "Assert_failure refineBindings.ml:265:23"
```

### LINKED ISSUE #456: Typenames can escape their scope
Syntactically, it is possible to declare a typename in any binding context. Therefore it is possible to declare a typename in a local scope, and *that* typename can escape its scope!

Consider the following program.
```links
typename T = Bool;
fun f() {
  typename T = Int;
  0 : T
}

links> f();
0 : T
```
The invocation `f()` produces an integer value, whose type is `T`. But which `T`? Looking at the type constructor environment, it seems that `T = Bool`:
```
links> @tyconenv;
 T = Bool
```
If that is true, then `0 : T` would be ill-typed.

What has happened here, is that the local declaration of `T` within `f` has escaped its scope, and thereby gets conflated with the `T` in the outer scope.

I can think of a couple of potential solutions:

1. Only allow typename declarations in module contexts.
2. Unroll locally defined typenames in positive positions.
3. Disallow locally defined typenames in positive positions.

My preference is for 1, which I believe aligns with other functional programming languages such as `OCaml` and `Haskell`.

### LINKED ISSUE #571: About the semantics of `open`

### LINKED ISSUE #595: How to print type names from opened modules?
Consider the following program:

```
module A {
  typename Foo = Int;
}

open A;
sig x : Foo
var x = 3;
x
```

We currently have a test case assuming that the result of this program should be printed as `3 : A.Foo`, meaning that we fully expand the name to its original location. That sounds reasonable, but doing so systematically would mean that printing all types from the prelude would be prefixed similarly. I would prefer seeing `Maybe` as a type name rather than `Prelude.Maybe`. Note that this only affects the printing of types, not the types that the user writes (as seen in the example program, we can write `Foo` after opening `A`. Similarly, the prelude is implicitly opened in each program).

Here are a few solutions:
1. Do not expand the type names, meaning that the result of the program above would be   `3 : Foo` instead of  `3 : A.Foo`.
2. Expand all type names, meaing that we would see  `3 : A.Foo`, but also `Prelude.Mayne`.
3. Expand user-written type names (e.g., do  `3 : A.Foo`) but allow some minor inconsistency by suppressing `Lib` and `Prelude` when printing the type names from those modules.

### LINKED ISSUE #598: Boundary Peek
Modules break the lexical scope as exemplified by the following program
```links
# scope.links
module A {
  var x = "A1";
}

open A;

module A {
  var x = "A2";
}

println(x)
```
The first module `A` is opened, and hence `x` is brought into scope unqualified. The second definition of `A` shadows the first, but crucially its contents are not brought into the toplevel scope, hence the program should print `A1`, but running `./links -m scope.links` produces the output
```
A2
() : ()
```
Thus it is possible to peek across scope boundaries.

The problem has little to do with modules, it just happens that modules expose the problem. The real problem is that our handling of names in the front end is broken. We should really be resolving names early on in the front end to rule out a whole class of problems.

dhil:
This patch fixes all known scoping issues due to modules (except #596
which is related to the loader/chaser). It makes no changes to the
implementation strategy of the current module system, rather, it fixes
the current implementation such that it is hygienic.

It fixes some bugs due to misconceptions such as qualifying labels in
variants. Labels should never be qualified as they are structural.

Amongst other minor improvements are better error handling and
printing of error messages.

My intention is to use the current module system as "an executable
specification" for the new, proper module system.

#### Meta
Depends on PR #599.
Contributes toward resolving #337 and #603.
Resolves #292, resolves #456, resolves #571, resolves #595, and resolves #598.


core/desugarModules.ml
```diff
@@ -17,328 +24,497 @@
...
+       type checker. We produce a "best guess" of its name, which is
+       simply its qualified form. *)
+    let best_guess : name list -> name
+      = String.concat "."
+
+    let module' : name list -> t -> scope
+      = fun names scopes ->
+      try S.Resolve.module' names scopes.inner
+      with Notfound.NotFound _ ->
+        S.Resolve.module' names scopes.outer (* Allow any errors propagate. *)
```
SimonJF: minor: "allow any errors _to_ propagate"

core/desugarModules.ml
```diff
@@ -17,328 +24,497 @@
...
+
+let rec desugar_module : ?toplevel:bool -> Epithet.t -> Scope.t -> Sugartypes.binding -> binding list * Scope.t
+  = fun ?(toplevel=false) renamer scope binding ->
+  match binding.node with
+  | Module (name, bs) ->
+     let visitor = desugar ~toplevel (Epithet.remember ~escapes:(not toplevel) name renamer) (Scope.renew scope) in
+     let bs'    = visitor#bindings bs in
+     let scope' = visitor#get_scope in
+     let scope'' = Scope.Extend.module' name scope' scope in
+     (* Printf.printf "enter_module before:\n%s\n%!" (string_of_scopes scope);
```
SimonJF: Can these comments go before merging?
(Same for the other comments below; maybe a better way of doing this would be to have debug log levels...)

core/desugarModules.ml
```diff
@@ -17,328 +24,497 @@
...
+      * Printf.printf "enter_module after:\n%s\n%!" (string_of_scopes scope''); *)
+     (bs', scope'')
+  | _ -> assert false
+and desugar ?(toplevel=false) (renamer : Epithet.t) (scope : Scope.t) =
+  let open Sugartypes in
+  object(self : 'self_type)
+    inherit SugarTraversals.map as super
+
+    val scope : Scope.t ref = ref scope
```
SimonJF: Why not use `mutable val` here, instead of a combination of vals and references?

core/desugarModules.ml
```diff
@@ -17,328 +24,497 @@
...
+                    ~pos ~stage:Errors.DesugarModules
+                    ~message:("Modules are not enabled. To enable modules set the `modules' setting to true or use the flag `-m'."))
 
-    method! unary_op = function
-      | UnaryOp.Name n -> (self, UnaryOp.Name (resolve n term_shadow_table))
-      | uo -> super#unary_op uo
+    method bindings = function
+      | [] -> []
+      | { node = Import names; pos } :: bs ->
+         self#extension_guard pos;
```
SimonJF: This is _very_ pretty.

core/desugarModules.ml
```diff
@@ -17,328 +24,497 @@
  *
  *  --->
  *
- * val Foo.bobsleigh = ...;
- * fun Foo.x() { ...}
- * fun Foo.Bar.y() { ... }
- * val x = ...;
+ * val Foo0.bobsleigh0 = ...;
```
SimonJF: I'm assuming these are internal and aren't exposed like this -- could you modify the comment to make this clearer?

core/desugarModules.ml
dhil: Good point.

core/desugarModules.ml
dhil: Sure!

core/desugarModules.ml
dhil: Oh yes, I will delete them. Thanks!

core/desugarModules.ml
dhil: Cheers!

dhil:
@SimonJF I think this is ready now.

A few other highlights:

- Modules work in the REPL.
- Slightly improves the debugging experience of generated JavaScript as the translation of qualified names adds less noise.

SimonJF:
As we discussed earlier, I can't get this to run. I think this is something on my end, but I'm baffled as to what, since I've done a clean checkout, am on the right branch, have done `make clean && make nc && dune install`, and I've even used your config file: 

```
[simon@dazzle daniel-links]$ ./links --config=/home/simon/scraps/daniel-config --path=examples/webserver:examples/games:examples/dictionary:examples examples/webserver/examples.links 
examples/webserver/examples.links:35: Type error: The function
    `addRoute'
has type
    `(String, (String) {hear{a}}~> Page) ~c~> ()'
    `String'
and
    `(String) {|wild|hear{a}}~> ()'
    `|wild|d'
In expression: addRoute("/examples/draggable.links", fun(_) {Draggable.main()}).
```

Is this just me? Could someone else try?

frank-emrich:
I can reproduce Simon's problem using a default configuration, besides enabling modules with `-m`.

dhil:
That is strange. I cannot reproduce the problem on either of my machine with the default configuration with modules enables or my personal configuration. Both the `examples.links` and `examples-nodb.links` run as expected.

dhil:
@SimonJF @frank-emrich try: `./links -m --path=examples:examples/webserver:examples/games:examples/dictionary examples/webserver/examples.links`

Note that `examples` have moved to the front of the list. I think the problem you are experiencing is a problem with loader. There are multiple candidates for `draggable.links`, but the current loader/chaser does not warn you in case of an ambiguous choice (the new loader does).

PR #607, opened 2019-05-17
Switch examples to use import

dhil:
This patch changes the examples such that they use `import` rather than `open` to require external modules. Currently, `open` subsumes `import`, but not vice versa. In a subsequent patch I will change the semantics of `open` such that it can no longer be used to require an external module.

### DIFF
--- a/core/chaser.ml
+++ b/core/chaser.ml
@@ -50,7 +50,7 @@ object(self)
     | b -> self#bindingnode b.node
 
   method! bindingnode = function
-    | Import ns
+    | Import { path = ns; _ }
     | Open ns ->
         (* Try to resolve the import; if not, add to ICs list *)
         let lookup_ref = List.hd ns in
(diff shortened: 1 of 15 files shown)


dhil:
The patch also includes new syntactic sugar for simultaneous import and opening of compilation units. For example `import open Foo;` resolves the corresponding source file for `Foo` and pulls its contents into the current scope.

jamescheney:
I think I'd mildly prefer "open import Foo" to "import open Foo" if we only allow one form, since that seems more standard (i.e. like in Agda).  We're opening the imported Foo, not important the opened Foo. But this seems bikesheddy.

dhil:
> I think I'd mildly prefer "open import Foo" to "import open Foo" if we only allow one form, since that seems more standard (i.e. like in Agda). We're opening the imported Foo, not important the opened Foo.

That is fine by me. I will flip the order.

PR #609, opened 2019-05-18
Semantics of `open` is not `include`.

### LINKED ISSUE #606: Names shadowed by transitive dependencies
Just as I'd merged the last patch, I tried running my larger application (should have done that first -- will know for future), and ran into the following issue.

Essentially, `open` in one module affects name resolution in another module. As an example, consider the three following files:

`one.links`:

```
fun concat(xs) {
  fold_right((^^), "", xs)
}
```

`two.links`:
```
open One;

fun foo() {
  concat(["Hello", "World"])
}
```

`three.links`:
```
open Two;

fun main() {
  concat([[1], [2], []])
}

main()
```

In `one.links`, we define `concat` which shadows the `concat` from the prelude. It now has the type `([String]) ~> String` instead of `([[a]]) ~> [a]`.

Running `links three.links`, I now get:

```
[simon@dazzle mfe-shadowing]$ links three.links 
registering driver for null
compiling to IR
compiled IR
Finishing process MAIN
Unification error: Couldn't match ([String]) ~> String against ([[Int]]) -@ _::Any
three.links:4: Type error: The function
    `concat'
has type
    `([String]) ~a~> String'
    `[[Int]]'
    `|b::Any'
In expression: concat([[1], [2], []]).
```
where I would expect `[1, 2]`.

dhil:
The desugar modules patch inadvertently changed the semantics of
`open` to be that of `include`. This regression was spotted by
@SimonJF in #606. From reading the generated code I mistakenly
concluded that it was a loader problem. However, as pointed out by
@SimonJF, the chaser is supposed to lift dependencies into "synthetic"
modules in the combined program.

With that in mind, it became clear to that the desugar modules patch
inadvertently changed the semantics of `open`, because I heedlessly
added the bindings resulting from an `open` to the module scope that
was being built. As a result they would correctly shadow the current
bindings in scope, but they would also inadvertently become part of
the module.

This patch rectifies this mistake. The implementation still makes use
of two-level scope data structure, however, now it only keeps track of
the current scope and its "delta", where the delta consists of all the
bindings defined in the current immediate scope. Bindings resulting
from an `open` are not added to the delta, and thus, do not become
part of the module's implicit signature.

Fixes #606.

### DIFF
--- a/core/desugarModules.ml
+++ b/core/desugarModules.ml
@@ -35,15 +35,14 @@
  *)
 
 (* Resolution strategy: We make use of a two-level scope data
-   structure to build up static scopes. The "outer" level represents
-   the parent scope, whilst the "inner" represents the scope being
-   built. Upon entry to a module scope the "inner" scope becomes an
-   "outer" scope, and exploration of the module initiates with an
-   empty "inner" scope. Meaning that after the exploration, the
-   "inner" scope only contains the top-level bindings of the said
-   module. After exploration we restore the old "outer" and "inner"
-   context, and add the explored module to the "inner" context with
-   its scope. *)
+   structure to build up static scopes. The "visible" level contains
+   everything that is visible in the current scope, whilst the "delta"
+   level contains everything that is *defined* in the immediate
+   scope. The "delta" is emptied upon entry to a module scope. Meaning
+   that after the exploration, the "delta" scope only contains the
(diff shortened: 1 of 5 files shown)


SimonJF:
I have just tested this on my larger application, and can confirm that it solves the issues with my larger application. My larger application is now working smoothly with your rewritten system. Thanks!

PR #613, opened 2019-05-18
Change the semantics of open

dhil:
It is no longer possible to require an external compilation unit via
`open`. Meaning that `import` and `open` behave differently now.

Note this is a **breaking change**.


dhil:
> The scenario in which this error is raised should make adding a hint along the lines of
>
> "did you mean open import Import_via_open1?"
>
> easier, and will benefit people who have not been following the modules discussions.

I agree, that would probably be more helpful.

I think ultimately we should generate error messages (i.e. OCaml code) along with their hints from a structured description. I envisage having a small configuration language for Links to defining error messages. I further envisage having small configuration languages for describing compilation dependencies and description for dynamic plugins. But that is another matter.

SimonJF:
Sure. But given that (I believe) this error is easily added, and it describes how to update code broken by a breaking change, I think it would be good to add it before merging.

dhil:
> But given that (I believe) this error is easily added [...]

I am not convinced it is so *easily* done correctly. The suggested hint only makes sense, if the `open` occurs at the top-level of the compilation unit. I would need to plum a bit more state around to differentiate between the top-level of a compilation unit and a module.

SimonJF:
> The suggested hint only makes sense, if the open occurs at the top-level of the compilation unit.

Did the previous semantics allow `open` to include files when it occurred midway through a program before? (I believe this was possible before, but I may be wrong). If this was the case, then I see no issue with adding the hint. Even if not, I don't think there's anything wrong with having a hint which doesn't necessarily apply (c.f. the "did you mean X" errors in OCaml).

dhil:
> [...]  I see no issue with adding the hint.

Take the user's perspective for moment. The user writes
```links
fun main() {
  open A;
  f() # intention: f is defined in A.
}
``` 
and gets told `Unbound module A. Hint: Did you mean open import A?`. The user follows the hint and writes
```
fun main() {
  open import A;
  f()
}
```
Now the user gets `Syntax error: open import A;`. Essentially, we have suggested that the user replace one error by another error. The result is a terrible user experience.

SimonJF:
OK, good point well made. Thanks for the example.

In which case, happy to approve this!

PR #615, opened 2019-05-18
Removes `rewrite.ml` which is unused.

dhil:


dhil:
> No point keeping around stuff we don't use.

indeed, we can always restore it from the history, if needed.

PR #616, opened 2019-05-19
Bump the version number to prepare for the Burghmuirhead release.

dhil:


dhil:
> Will merge once CI passes.

Or you could be audacious and just force the merge... It would be amazing if it broke anything, given that #615 only removes an unused module.

SimonJF:
I won't tell if you won't.

PR #621, opened 2019-05-22
Fix #597 (IR Typing: Desugaring of session handlers)

### LINKED ISSUE #597: Desugaring session exceptions is not type-preserving
Desugaring session exceptions - specifically the part that inserts top-level handlers inside a `spawn`  breaks typing. The problem is that the translation to insert top-level handlers creates a `TryInOtherwise` whose body needn't have the special `_SessionFail` effect. (This breaks the typing of practically every single program that uses session exceptions as session typing with spawning is rather useless.)

A couple of possible fixes come to mind.

 * One option would be to change the types of the family of spawn functions such that the body always has the special `_SessionFail` effect whenever the `--session-exceptions` flag is set.

 * Another (probably better) option would be to insert an upcast on the body during the translation to enable top-level handlers.

SimonJF:
This PR fixes the IR typing of desugaring of session handlers.

- Makes `session_exceptions` a system setting
- If `session_exceptions` is true, then all the body of all spawned processes have the `SessionFail` effect
- Some massaging of the translation to make sure it's type-preserving (binding result to a wildcard binding)
- Fix a few things in `desugarProcesses`

A slight problem is that this approach means that the Prelude does not have type signatures for some functions using `spawn` anymore, since whether that process raises `SessionFail` depends on the value of `session_exceptions`.


SimonJF:
I neglected to mark this as a draft PR, so please don't review until I've fixed the CI.

SimonJF:
This has gone into @slindley's `ordered-quantification` branch, so there's little point having it as a separate PR.

Issue #540, opened 2019-03-27
split Query.Eval.eval into separate translation and normalization stages


jamescheney:
Factor the work done in `Query.Eval.eval` into two stages, a first pass to translate `Ir.t` to not-necessarily-normalized `Query.Lang.t` (which would perform any latent evaluation such as resolving `Database` and `Table` references) and a second, normalization stage that is an endofunction on `Query.Lang.t`.  This would hopefully:

- [x] allow simplifying `Query.Lang.t` for example to remove the `Database` case
- [x] allow for transformign and re-normalizing queries, a la @fehrenbach's provenance translations

Admittedly it may also be slower than translating and normalizing in one pass.  On the other hand, we could explore doing the translation even earlier, i.e. translating `query`-subexpressions to embedded `Query.Lang.t` expressions viewed as a quoted DSL.

jamescheney:
Closed by #540

Issue #619, opened 2019-05-21
XML literals in queries


jamescheney:
XML literals are allowed in queries, but actually trying to use them seems to cause strange outcomes, perhaps (in some cases) due to the punning of `Xml` as `[XmlItem]`.

```
links> query { for (x <- [(<a>asdf</a>)]) [(b=2)] };
***: Internal Error in query/query.ml (Please report as a bug): Runtime type error: failed to unbox list.
This should not happen if the type system / checker is correct. Please file a bug report.
 
links> query { for (x <- (<a>asdf</a>)) [(b=2)] };
***: Internal Error in query/query.ml (Please report as a bug): Runtime type error: failed to unbox list.
This should not happen if the type system / checker is correct. Please file a bug report.
 
links> query { for (x <- (<a>asdf</a>)) [(b=2)] };
***: Internal Error in query/query.ml (Please report as a bug): Runtime type error: failed to unbox list.
This should not happen if the type system / checker is correct. Please file a bug report.

links> query {var x = <a>asdf</a>; [(a=1)]};
***: Internal Error in query/query.ml (Please report as a bug): Runtime type error: failed to unbox list.
This should not happen if the type system / checker is correct. Please file a bug report.
```

PR #624, opened 2019-05-31
Factorize Query.Eval.eval into translation + normalization (#540)

wricciot:
This PR splits the query evaluation function `Query.Eval.eval` into two distinct phases of IR to query translation (`Query.Eval.xlate`) and query normalization (`Query.Eval.norm`). 
The changes to the code include extensions to the type `Query.Lang.t` to account for non-normalized queries.

This code appears to work but hasn't been tested thoroughly yet. We may also want to revise some implementation choices (e.g. eta expansion of record variables is presently done in `xlate`, but it makes more sense for it to be part of `norm`; this would require further changes to `Query.Lang.t`, however).

`xlate` is presently "shallow", in the sense that it doesn't perform translation recursively inside closures; for technical reasons, a deep translation is challenging to write...

Comments are welcome, but do not merge yet :-)


core/query/query.ml
```diff
@@ -715,13 +648,94 @@ struct
...
+      reduce_eq env (v, w)
+    | Q.Primitive f, args ->
+        Q.Apply (Q.Primitive f, args)
+    | Q.If (c, t, e), args ->
+        reduce_if_condition env (c, apply env (t, args), apply env (e, args))
+    | Q.Apply (f, args), args' ->
+        apply env (f, args @ args')
+    | t, _ -> eval_error "Application of non-function: %s" (string_of_t t)
+
+  and reduce_concat env vs =
```
jamescheney: The various reduce_ fucntions were previously not really mutually recursive with translation/normalization.  Now they are mutually recursive with `norm,` but only for the trivial reason that they first call `norm` on their arguments, which could easily be avoided by moving these calls back into `norm` itself where the `reduce_` functions are called.  There is also no other particular reason `env` needs to be threaded through the `reduce_` functions other than that `norm` is being called and requires `env`.  The `reduce_` functions could be moved earlier, defined before / separatelt from the mutual recursion with `norm` and `env` removed as an argument.   Indeed I think they could even live in `Query.Lang`.  I think that would be cleaner, even though there is then a bit more clutter when `reduce_` functions are called.

core/query/query.ml
```diff
@@ -865,27 +885,43 @@ struct
...
+      res
+    end
+	*)
 
   let eval env e =
 (*    Debug.print ("e: "^Ir.show_computation e); *)
-    computation (env_of_value_env env) e
+    norm_comp (env_of_value_env env) e
+
+  let reduce_where_then = reduce_where_then (Value.Env.empty, Env.Int.empty)
```
jamescheney: I'd avoid using the same name for the function result and function name, even though in this bounded / nonrecursive let it is technically fine.

core/query/query.mli
```diff
@@ -21,8 +21,9 @@ sig
       | Erase     of t * StringSet.t
       | Variant   of string * t
       | XML       of Value.xmlitem
-      | Apply     of string * t list
-      | Closure   of (Ir.var list * Ir.computation) * env
+      | Apply     of t * t list
+      | Closure   of (Var.var list * Ir.computation) * env
```
jamescheney: This seems to get us epsilon closer to translating `Ir` expressions to queries with no dangling `Ir` references.  If we're handing `Case` constructs, how much harder would it be to handle function bodies/closures the same way?  It seems like the main issue is making sure that the translation does the right thing when translating under a binder (when encountering variables that aren't bound to values in `env`), which is already something that has to work right for `For` and `Case`.

Not something to address here necessarily, just a thought.

core/query/query.ml
```diff
@@ -1163,7 +1199,14 @@ let compile_delete : Value.database -> Value.env ->
   ((Ir.var * string * Types.datatype StringMap.t) * Ir.computation option) -> string =
   fun db env ((x, table, field_types), where) ->
     let env = Eval.bind (Eval.env_of_value_env env) (x, Q.Var (x, field_types)) in
-    let where = opt_map (Eval.computation env) where in
+    let where = opt_map (Eval.norm_comp env) where in
     let q = delete db ((x, table), where) in
       Debug.print ("Generated update query: "^q);
       q
+
+let is_list = Q.is_list
```
jamescheney: Maybe we could add these to the public signature of `Query.Lang` instead?

core/query/evalNestedQuery.ml
```diff
@@ -487,9 +489,9 @@ struct
     function
       | Constant c    -> Constant c
       | Primitive p   -> Primitive p
-      | Apply ("Empty", [e]) -> Apply ("Empty", [flatten_inner_query e])
-      | Apply ("length", [e]) -> Apply ("length", [flatten_inner_query e])
-      | Apply (f, es) -> Apply (f, List.map flatten_inner es)
+      | Apply (Primitive "Empty", [e]) -> Apply (Primitive "Empty", [flatten_inner_query e])
+      | Apply (Primitive "length", [e]) -> Apply (Primitive "length", [flatten_inner_query e])
+      | Apply (Primitive f, es) -> Apply (Primitive f, List.map flatten_inner es)
```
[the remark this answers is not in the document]
wricciot: Assuming we operate on normalized queries, the first argument of `Apply` _must_ be a Primitive (if it's not, then there's a problem somewhere else). On the other hand, I don't really see why we need the extra cases for "Empty" and "length", though it could be a (minimal) optimization.

core/query/query.ml
wricciot: This seems plausible, I'll try it.

core/query/query.ml
wricciot: This is a case of encapsulation + shadowing of a recursive function, in which we want to prevent some technical details from being visible. In these cases I often use the same name for the implementation and the encapsulating function (rather than e.g. finding an alternate name for the recursive implementation which would be likely to be a bit silly). 
It is certainly a practice I don't want to abuse, but I believe it should be ok here?

core/query/evalNestedQuery.ml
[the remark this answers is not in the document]
jamescheney: OK.  It looks like here the first two, specialized, cases are now formally redundant, except perhaps that they check that empty and length have exactly one argument (something that should already have been checked earlier).

core/query/query.ml
jamescheney: Oh, I see, this is  due to `reduce_where_then` being exported, and you are rebinding them to sensible-to-call-remotely partially applied versions.  If my earlier suggestion (about decoupling `reduce_*` from normalization) works out then this would go away.  I guess this assumes that the places where they are called elsewhere are also normalizing the argument(s) before calling (or that normalization is unnecessary).  However, I think before the `reduce_` functions were not recursively calling translation/normalization anyway so perhaps this is further reason to revert and transplant them to `Query.Lang`.

jamescheney:
I think we agreed the next step for this (besides the possible code cleanups above) is for Simon to check it out and see if running GtoPDB with the changes turns up any problems.  If so, perhaps they can be turned into testcases.

(It might also be cool to extract the schema and queries for GtoPDB and use them as a stress test for Links - this need not mean having a local copy of GtoPDB itself...)

wricciot:
I have now merged the last changes in master (particularly James's cleanup) with this branch, so Simon should be able to test it more easily.

SimonJF:
Thanks! This is really useful. I'll separate out my Elm changes and get them up-to-date with `master`, and then I should be able to test this with GtoPdb.

SimonJF:
So I've got this working with my GtoPdb application, but it seems a lot slower (minutes to load a page, as opposed to maybe a second).

Is there anything you've done in this PR which may have affected performance?

wricciot:
@SimonJF a performance penalty due to the separation of translation and normalization must be expected, but this looks like a couple orders of magnitude. Can't think of a reason why it would be so inefficient, but it's clear we can't merge it like this.

@jamescheney made a point that the `reduce_*` functions shouldn't be mutually recursive with `norm`, and based on the running time I conjecture that there may be cases in which we are executing `norm` on the same arguments more than once. Let's see if that is the issue.

jamescheney:
Also, some of the reduce_ functions are called from outside query.ml, e.g.
from evalNestedQuery.ml, where normalization should no longer be
necessary.  This could also be causing problems.

I think previously the reduce_ functions satisfied "if the inputs are
normalized then so is the output".

It would be good to see if there is a noticeable time difference just on
the test queries, or if @SimonJF can identify a specific query exhibiting
the problem.

wricciot:
Ok, after these last commits, I have removed some redundant recursive normalization steps and cleaned up the code a little.
There are new debug prints for Query.Eval.eval that show the execution time of query evaluation: according to my tests with Simon's GtoPdb implementation, this is never more than a couple millisec, and is well dominated by db execution time.

But, it appears that this last change broke something, because several tests are failing now. With any luck, it's a very stupid bug. I'll keep working on this.

wricciot:
Alright, all the tests are succeeding now. Separating 'xlate' and 'norm' means that both functions need to be fixed in order for XML tags to be allowed in queries (this was a regession wrt #619).
Concretely, 'xlate' wants to unbox the list of children of an XML node, but fails when this is not a literal list, but an applied 'stringToXml', which will immediately reduce to a singleton list. The fix is to perform this single reduction step at translation time.

@jamescheney @SimonJF this might be ready for merging now.

core/query/evalNestedQuery.ml
```diff
@@ -30,8 +30,10 @@ let tag_query : QL.t -> QL.t =
         | Erase (e, fields) -> Erase (tag e, fields)
         | Variant (l, e) -> Variant (l, tag e)
         | XML v -> XML v
-        | Apply (f, es) -> Apply (f, List.map tag es)
+        | Apply (u, vs) -> Apply (tag u, List.map tag vs)
+(*        | ApplyPrim (f, es) -> ApplyPrim (f, List.map tag es) *)
```
jamescheney: Unsure why this comment is here.  Is this something we might want to add, or a reminder that we don't need this here?

core/query/evalNestedQuery.ml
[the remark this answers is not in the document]
jamescheney: On second thought, this and lines 385-389 should stay as is, since the Empty and length primitives are treated differently than other primitives.  (using *_inner_query instead of *_inner).

core/query/query.ml
```diff
@@ -399,7 +554,7 @@ struct
         match Value.Env.lookup var val_env, Env.Int.find exp_env var with
         | None, Some v -> v
         | Some v, None -> expression_of_value v
-        | Some _, Some v -> v (*eval_error "Variable %d bound twice" var*)
+        | Some _, Some v -> v (*query_error "Variable %d bound twice" var*)
```
jamescheney: Does this comment still need to exist?  It seems we are just using the exp_env version now shadowing the val_env occurrence.

SimonJF:
Delighted to report that this works absolutely fine with my GtoPdb clone, with no noticeable slowdowns.

core/query/query.ml
```diff
@@ -713,179 +810,112 @@ struct
...
+        Q.Apply (Q.Primitive f, args)
+    | Q.If (c, t, e), args ->
+        Q.reduce_if_condition (c, apply env (t, args), apply env (e, args))
+    | Q.Apply (f, args), args' ->
+        apply env (f, args @ args')
+    | t, _ -> query_error "Application of non-function: %s" (string_of_t t)
+
+  and norm_comp env c = norm env (computation env c)
+
+(*
```
jamescheney: Can this be cleaned up?

wricciot:
@jamescheney yes, that feels like a more robust way to address the issue, and a quick change to the code suggests that it works; my only concern is whether there could exist unapplied `stringToXml` that become applied during normalization

core/query/query.ml
```diff
@@ -713,179 +810,112 @@ struct
...
-        | (a, b) -> Apply ("==", [a; b])
+    | Q.Primitive "not", [v] ->
+      Q.reduce_not (v)
+    | Q.Primitive "&&", [v; w] ->
+      Q.reduce_and (v, w)
+    | Q.Primitive "||", [v; w] ->
+      Q.reduce_or (v, w)
+    | Q.Primitive "==", [v; w] ->
+      Q.reduce_eq (v, w)
+    | Q.Primitive "stringToXml", [v] ->
```
jamescheney: This seems like it should be handled just once by `xlate` and if any `stringToXml`s remain during normalization it should be an error

core/query/query.ml
```diff
@@ -713,179 +810,112 @@ struct
...
+      (* Debug.print ("Applying closure"); *)
+      (* Debug.print ("body: " ^ Ir.show_computation body); *)
+      (* Debug.print("Applying query closure: " ^ show_t (`Closure ((xs, body), closure_env))); *)
+      (* Debug.print("args: " ^ mapstrcat ", " show_t args); *)
+        let env = env ++ closure_env in
+        let env = List.fold_right2 (fun x arg env ->
+            bind env (x, arg)) xs args env in
+        (* Debug.print("Applied"); *)
+          norm_comp env body
+    | Q.Primitive "AsList", [xs] ->
```
jamescheney: Perhaps this could be removed by `xlate` also

core/query/query.ml
wricciot: TBH I don't know how this would happen and on what grounds we ignore the val_env version.
It looks like this was previously considered an error, but now it's not.
I updated the comment because the original `eval_error` has now become a more generic `query_error`. If the consensus is that the comment is now useless, I'll remove it.

core/query/query.ml
wricciot: I can inline it. Personally, though, I think the code is more readable this way. Does it even make any difference after compiling?

core/query/evalNestedQuery.ml
wricciot: Certainly not needed any more. Looks like a reminder of the previous state of affairs I left when changing the code.

core/query/query.ml
wricciot: Sounds ok.

core/query/query.ml
jamescheney: OK, let's leave as is for now unless someone chimes in.

core/query/query.ml
jamescheney: "This" in my question was about the comment starting on line 903, not a suggestion to inline `norm_comp`.

wricciot: Ahhh I see. Ok.

jamescheney:
I believe this is waiting (only) on:
- [ ] syncing with current master branch
- [ ] some benchmarking to measure the impact of the refactoring on query normalization performance as erquested by @slindley

wricciot:
@jamescheney Ok

wricciot:
About the slowdown of this factorization:
- The queries in the test suite take on average 41% more than the original code to be normalized: this time is however always in the order of some dozens of _micro_seconds
- I have also tested the code with general nested queries of a given depth and width. Here the new normalization algorithm has a slowdown of 27% for depth=2 and 20% for depth=3: so, it appears the slowdown becomes less relevant for larger queries.
However, I must also report that normalization of queries with depth=3 climbs to an execution time of about 40 _seconds_ with the old code and 47 seconds with the new one, which seems to be a lot.

SimonJF:
@wricciot: thanks for doing the performance analysis.

How did you get a query with depth 3 which took 40 *seconds* to evaluate with the old code? Is this a completely pathological case which is unlikely to arise in practice? If so, then maybe the increase isn't such a problem?

(Asking as it's a shame to have this PR still hanging around half a year later, so if we can merge it while there aren't any merge conflicts, it would be nice to).

slindley:
More to the point, why is there even a difference between the execution times?
Does execution time include normalisation time?
Are the generated SQL queries the same?

wricciot:
@SimonJF You can have a look at the generated test, but to run it you will have to patch desugarFors.ml and transformSugar.ml to fix the scope bug. 
[performance_links.txt](https://github.com/links-lang/links/files/3808959/performance_links.txt)

@slindley The time only includes query generation.

wricciot:
Additional files for debugging: a test query at depth 2:
[performance_links_2.txt](https://github.com/links-lang/links/files/3820091/performance_links_2.txt)

A bash script to generate test queries of arbitrary size (to change the size, adjust the parameters to the "query" function, towards the end of the script.
The test query is printed to stdout.
[linksperf_sh.txt](https://github.com/links-lang/links/files/3820092/linksperf_sh.txt)

SimonJF:
Thanks, @wricciot.

Having patched Links with the `desugarFors` and `transformSugar` fixes from this patch (thanks -- I wondered what was going on there; I'd run into it a few times with my other stuff too!), I tried the `performance_links.txt` file and it stack overflowed. But that's a different issue and shouldn't get in the way of this being merged.

I then tried the pathological case for depth 2 on both stock Links (with the fixes) and this PR. The differences in time were minimal.

Wilmer: thanks for confirming that your measurements are for normalisation only.  I am not worried by the given numbers: a slight slowdown is to be expected given that you have untangled two passes and therefore there is more work to do. Having tried the patch on my GtoPdb clone, I've not seen any noticeable slowdown there, and there are some hefty queries.

I am in favour of merging. @slindley: once @wricciot verifies that the SQL is the same, what are your merging criteria?

wricciot:
I can confirm that the SQL output of the old and new procedure is the same, up to alpha-renaming.

jamescheney:
One more tiny question:

I intend/hope that we can eventually perform the "translation"stage just once, when generating the IR.

I think Sam's concern regarding the slowdown would be addressed if it were clear that the slowdown is entirely due to the extra translation stage pass, so that (if we move that into the IR generation so that it only happens once), the run-time normalization cost is the same (or less) than in the current approach where both happen in one pass.

Wilmer, could you check how the time for the factored version breaks down into the time for translation (which should eventually be a one-time cost) and the time for normalization (which will continue to happen at run time)?

wricciot:
@jamescheney on the testsuite, the translation phase is responsible for 34.9% of the execution time, and the normalisation phase for the remaining 65.1%. This means the normalization time is about 0.918 times than the old Query.eval time, or 8.2% faster.

jamescheney:
OK, thanks.  That is roughty what I was hoping/expecting.

slindley:
I think this is OK to merge modulo some nonsense regarding continuous integration - supposedly the database shredding tests fail on "Job mac" on Azure.

wricciot:
This is now merged and compiling correctly, with the expected shredding test failure on Mac (#792).

wricciot:
It looks like we are ignoring the test failure until a decision is made on whether to disable the mac os azure CI. Merging while it's still up-to-date.

Issue #566, opened 2019-04-13
Fresh type variables should not be generated during parsing


slindley:
Links conveniently allows the programmer to omit the name of a type variable when it is irrelevant (e.g. because it occurs only once in a type). This syntactic sugar is particularly handy for first-order effect-polymorphic functions. For instance the type of `+` may be written as

    (Int, Int) -> Int

which actually means

    (Int, Int) -e-> Int

for some fresh effect variable `e`.

Currently the generation of fresh variables is performed during parsing, which is a mistake as it limits the sophistication of the algorithm for generating fresh names. (In general it is a bad practice to mix computation with parsing and we should strive to separate them.)

Currently Links generates fresh type names of the form `_XXX` where `XXX` is a globally unique integer. We should change the parser to just generate the special name `_` and then apply a desugaring phase immediately after parsing in order to generate fresh names.

The particular use-case I have in mind is to extend the syntactic sugar in order to allow effect variables to be omitted from most higher-order functions, inspired by [Frank](https://github.com/frank-lang/frank). Currently the type of map is

    ((a) -e-> b, [a]) -e-> [b]

whereas with the sugar I have in mind it could instead be written as

    ((a) -> b, [a]) -> [b]

The key idea (taken from Frank) that makes this possible is the observation that almost always all of the effect types on the arrows of a higher-order function match up. This is hardly surprising given that typically higher-order functions apply their argument functions, and in order to do so the effects of the arguments must match up with those of the higher-order function itself.

SquidDev:
Should have been resolved by #625.

PR #625, opened 2019-06-04
Syntactic sugar for effect variables in higher-order functions

SquidDev:
See #566 for a bit of a deeper explanation.

 - Remove fresh name generation from the parser, just generating a placeholder variable. Fresh variable names are now generated as part of `DesugarDatatypes`.
 - Add a `shared_effect_vars` flag. When enabled, functions with no explicit effect variables will use the same effect variable. For example, `map : ((a) -e-> b, [a]) -e-> [b]` can be written as `map : ((a) -> b, [a]) -> [b]`.
 - Start some work on allowing printing types as above. This requires a bit more thought before it's actually usable, so happy to revert it for now.

A couple of things worth mentioning/I'm not entirely happy with:

 - `shared_effect_vars` isn't a fantastic name, so any ideas are appreciated.
 - We need to disambiguate between "explicit but unnamed" effect variables (`-_->`) and "implicit" ones (`->`). I'm currently doing this by naming the placeholder variables `_` and `_anon`, but that's not especially pleasant.
  - I've removed most of the fresh name generation from `SugarConstructors`, as it's no longer used - happy to revert this if it may be useful again though.

### DIFF
--- a/core/basicsettings.ml
+++ b/core/basicsettings.ml
@@ -225,6 +225,7 @@ end
 module Types = struct
   let show_mailbox_annotations = Settings.add_bool("show_mailbox_annotations", true, `User)
   let show_raw_type_vars = Settings.add_bool("show_raw_type_vars", false, `User)
+  let shared_effect_vars = Settings.add_bool ("shared_effect_vars", false, `User)
   module Print = struct
     let show_quantifiers     = Settings.add_bool   ("show_quantifiers"    , false    , `User)
     let show_flavours        = Settings.add_bool   ("show_flavours"       , false    , `User)
--- a/core/basicsettings.ml
+++ b/core/basicsettings.ml
@@ -225,7 +225,7 @@ end
 module Types = struct
   let show_mailbox_annotations = Settings.add_bool("show_mailbox_annotations", true, `User)
   let show_raw_type_vars = Settings.add_bool("show_raw_type_vars", false, `User)
-  let shared_effect_vars = Settings.add_bool ("shared_effect_vars", false, `User)
+  let effect_sugar = Settings.add_bool ("effect_sugar", false, `User)
   module Print = struct
     let show_quantifiers     = Settings.add_bool   ("show_quantifiers"    , false    , `User)
     let show_flavours        = Settings.add_bool   ("show_flavours"       , false    , `User)
--- a/core/desugarDatatypes.ml
+++ b/core/desugarDatatypes.ml
@@ -48,7 +48,7 @@ class basic_freshener = object
   method! known_type_variable =
     let module SC = SugarConstructors.SugartypesPositions in
     function
-    | (("_" | "_anon"), None, freedom) ->
+    | (("$" | "$anon"), None, freedom) ->
        SC.fresh_known_type_variable freedom
        |> super#known_type_variable
     | v ->  super#known_type_variable v
@@ -59,7 +59,7 @@ let freshen_vars =
   (* Determine if this is an "anonymous" effect type variable, and so introduced
      within a simple arrow (->, ~>, -@, ~>) *)
   let is_anon_effect = function
-    | Datatype.Open ("_anon", None, `Rigid) -> true
+    | Datatype.Open ("$anon", None, `Rigid) -> true
     | _ -> false in
   (* If this function type is exclusively composed of anonymous effect type
      variables. Or rather, there are no explicitly mentioned effect variables. *)
--- a/core/desugarDatatypes.ml
+++ b/core/desugarDatatypes.ml
@@ -41,6 +41,21 @@ let concrete_subkind =
   | Some subkind -> subkind
   | None         -> default_subkind
 
+(** Replaces all placeholder variables with a fresh one. *)
+class basic_freshener = object
+  inherit SugarTraversals.map as super
+
+  method! known_type_variable =
+    let module SC = SugarConstructors.SugartypesPositions in
+    function
+    | ("_", None, freedom) ->
+       SC.fresh_known_type_variable freedom
+       |> super#known_type_variable
+    | v ->  super#known_type_variable v
+end
+
+let freshen_vars = new basic_freshener
+
 (* Find all unbound type variables in a term *)
 let typevars =
 object (self)
@@ -617,10 +632,12 @@ object (self)
 end
 
 let phrase alias_env p =
+  let p = freshen_vars#phrase p in
   let tvars = (typevars#phrase p)#tyvar_list in
     (desugar alias_env (snd (Desugar.generate_var_mapping tvars)))#phrase p
 
 let binding alias_env b =
+  let b = freshen_vars#binding b in
   let tvars = (typevars#binding b)#tyvar_list in
     (desugar alias_env (snd (Desugar.generate_var_mapping tvars)))#binding b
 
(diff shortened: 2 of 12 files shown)


core/basicsettings.ml
```diff
@@ -225,6 +225,7 @@ end
 module Types = struct
   let show_mailbox_annotations = Settings.add_bool("show_mailbox_annotations", true, `User)
   let show_raw_type_vars = Settings.add_bool("show_raw_type_vars", false, `User)
+  let shared_effect_vars = Settings.add_bool ("shared_effect_vars", false, `User)
```
dhil: I would simply call it `effect_sugar`.

core/parser.mly
```diff
@@ -171,6 +171,9 @@ let parseRegexFlags f =
               | 'g' -> RegexGlobal
               | _ -> assert false) (asList f 0 [])
 
+let fresh_typevar freedom = ("_", None, freedom)
+let fresh_effects = ([], Datatype.Open ("_anon", None, `Rigid))
```
dhil: I wonder whether the name `_anon` can ever leak out? I suppose it can't, because the name `_anon` is compiled away before the semantic type is constructed, and the type printer refreshes the type variable names anyway. And the error reporter for syntax errors extracts excerpts from the actual source code, in which `_anon` isn't present.

core/parser.mly
```diff
@@ -961,16 +964,16 @@ straight_arrow:
...
+| parenthesized_datatypes LOLLI datatype                       { Datatype.Lolli    ($1, fresh_effects, $3) }
 
 squiggly_arrow:
 | parenthesized_datatypes
   squig_arrow_prefix SQUIGRARROW datatype                      { Datatype.Function ($1, row_with_wp $2, $4) }
 | parenthesized_datatypes
   squig_arrow_prefix SQUIGLOLLI datatype                       { Datatype.Lolli    ($1, row_with_wp $2, $4) }
-| parenthesized_datatypes SQUIGRARROW datatype                 { Datatype.Function ($1, row_with_wp (fresh_row ()), $3) }
-| parenthesized_datatypes SQUIGLOLLI datatype                  { Datatype.Lolli    ($1, row_with_wp (fresh_row ()), $3) }
+| parenthesized_datatypes SQUIGRARROW datatype                 { Datatype.Function ($1, row_with_wp (fresh_effects), $3) }
```
dhil: Nit: You can drop the parentheses.

core/lib.ml
```diff
@@ -402,7 +402,7 @@ let env : (string * (located_primitive * Types.datatype * pure)) list = [
 
   "spawnWait",
   (`PFun (fun _ -> assert false),
-   datatype "(() ~> a) ~> a",
+   datatype "(() ~_~> a) ~> a",
```
dhil: I see. The point is that the two implicit effect variables are distinct, but with `shared_effect_vars` enabled they become the same variable. 

I would use the non-sugared form `() { |_}~> a`

core/desugarDatatypes.ml
```diff
@@ -41,6 +41,71 @@ let concrete_subkind =
...
+      | Function (args, (effects, eff_var), ret) when is_anon_effect eff_var
+        -> super#datatypenode (Function (args, (effects, var), ret))
+      | Lolli (args, (effects, eff_var), ret) when is_anon_effect eff_var
+        -> super#datatypenode (Lolli (args, (effects, var), ret))
+      | ty -> super#datatypenode ty
+  end in
+
+  object
+    inherit basic_freshener as super
+    method! datatypenode = let open Datatype in
```
dhil: I think you can implement this by overloading the `row` method. Then this rewrite should apply if the (sub)kind of the row variable is `Eff`. This might even make it possible to extend this syntactic sugar to user definable datatypes such as `Comp`.

core/desugarDatatypes.ml
```diff
@@ -41,6 +41,71 @@ let concrete_subkind =
...
+      function
+      | (Function _ | Lolli _) as dt ->
+         (* If shared_effect_vars is enabled, and all variables are fresh (and so
+            not explicitly named), then we will substitute the same variable
+            across all arrows. Otherwise every arrow gets its own row
+            variable, as normal. *)
+         if Settings.get_value Basicsettings.Types.shared_effect_vars
+            && (all_anon_effects#datatypenode dt)#all_anon then
```
dhil: One way to make this a linear time algorithm is to use non-linear control and apply the syntactic sugar optimistically. Use three objects: one to traverse the term structure, let's call simply call it `effect_var_desugarer`, and two objects for refreshing effect type variables akin to your `shared_refresh` and `basic_refresh`.

Whenever the `effect_var_desugarer` comes across a `datatypenode` it has to pick between the two visitors. It should optimistically select the `shared_refresh` visitor. The `shared_refresh` visitor should optimistically rewrite a given datatype, however, if it comes across an effect variable that violates the assumption, then it should throw an exception. This exception should be caught by the `effect_var_desugarer`, causing it to apply the `basic_refresh` visitor instead.

This algorithm should have linear time complexity. Let `n` be the size of the datatype, in the worst case we visit the whole datatype before discovering that there is a variable which violates the assumption. Moreover, the depth of call stack relative to the initial invocation of the visitor can at most be `n`. So it takes `2*n` time to discover and propagate the information back to the `effect_var_desugarer` visitor, which then applies the other visitor, which must necessarily visit the entire datatype and return again, requiring at most `2*n` time also. Summing up `2*n + 2*n = O(n)`.

tests/functions.tests
```diff
@@ -51,6 +51,20 @@ fun (x) {x:a} : a
...
+Printer shares type variables
+fun (f, x: Int) { f(x) + 0 }
+stdout : @fun : \(\(Int\) -> Int, Int\) -> Int
+args : --config=tests/functions.tests.shared_effect_vars.config
+
+Parser shares type variables
+fun (f, x: Int) { f(x) + 0 } : ((Int) -> Int, Int) -> Int
+stdout : @fun : \(\(Int\) -> Int, Int\) -> Int
+args : --config=tests/functions.tests.shared_effect_vars.config
+
```
frank-emrich: Should we have a test case for when the type variables are different, and we should therefore not print `->`?

tests/functions.tests
```diff
@@ -51,6 +51,20 @@ fun (x) {x:a} : a
 stderr : @.*
 exit : 1
 
+Infers effect variables
+fun (f, x: Int) { f(x) + 0 }
+stdout : @fun : \(\(Int\) -a-> Int, Int\) -a-> Int
```
frank-emrich: I'm not sure why we want to show the variable here. In combination with the next test, it seems that adding the type annotation (without the variable) suppresses the printing of the variable. This seems  inconsistent to me.

SquidDev: This test runs without the setting enabled, just to check that the previous behaviour is maintained - hence it displaying the type variables.

tests/functions.tests
SquidDev: I think the `pickleCont` test (lines 68-71) does that (although I'm not sure if the current behaviour is ideal).  

I don't know if it'd better just to have a standalone definition though, rather than pulling from the prelude?

tests/functions.tests
frank-emrich: Yes, I meant that we should have a test for the following: If the setting is enabled and the effects are not the same, the printer must not use plain `->`

tests/functions.tests
frank-emrich: Yes, we have functions in the prelude that exhibit the behavior I'm interested in for the test case. But in the prelude, we don't check that the printer handles such cases correctly.

frank-emrich: Urg, I misread your comment as "we have such a function in the prelude". What you did is fine.

PR #626, opened 2019-06-06
Cleanup of SQL generation in query.ml

jamescheney:
This is the result of a code read-through/review of query.ml, working toward the eventual aim of cleaning up the query and update code and packaging it up separately from the main evaluator.  In the process, I noticed some unused arguments and unnecessary mutual recursion.  I believe all of the changes are pure refactorings, i.e. have no effect on the actual behavior of the code.


jamescheney:
@wricciot Yup.  Now if only there were some way to get rid of the index parameter.  We could maybe at least hoist it out of the recursion since in a given call to sql_of_* it is always a constant.  

Anyway, CI and people both seem happy so merging.

Issue #437, opened 2018-12-05
Lens constructs only work properly at toplevel


jamescheney:
Here's a small example:

```
var db = database "links";

var tracksTable =
	table "tracks"
	with (track: String, date: Int, rating: Int, album: String)
	tablekeys [["track", "album"]]
	from db;

fun foo() {
var tracksLens = lens tracksTable with { track -> date rating };
var tracks = lensget tracksLens;
tracks
}

foo()
```

which yields

```
*** Error: "Assert_failure core/sugartypes.ml:391:27"
```

It works if you remove the function declaration / call.  Weirdly, this version also works:
```
...
var foo = fun () {
var tracksLens = lens tracksTable with { track -> date rating };
var tracks = lensget tracksLens;
tracks
};

foo()
```
so perhaps the problem is a bug in how named function declarations are typechecked/desugared that is not present in the case for anonymous functions.

This seems to be because the "LensFunDepsLit" construct persists to `sugartypes.ml` but it should have been eliminated earlier.  The desugaring happens insde `type_check` in `typeSugar.ml`.

Observed in stock Links 0.8.

jamescheney:
Has anyone (e.g. @rudihorn) looked at this any further?  Was it fixed by #454?

jamescheney:
Looks like it is fixed either by #434 or #627.

jamescheney:
Fixed by #627

PR #627, opened 2019-06-07
[WIP] Relational Lenses dynamic predicate support

rudihorn:
This patch allows predicates to be specified as arbitrary functions going from `r -> bool` where `r` is the record type of the lens. This allows the generation of predicates which depend on runtime values. The code works by introducing a conversion from links IR to `Lens.Phrase.t` using techniques similar to `query/query.ml`. Long term it might be nice to try and merge the code. 

The code also introduces another lens construct `lenscheck` which can currently be considered an acknowledgement of the "warning" that the lens is not statically checked and may yield runtime errors during construction. As discussed with @jamescheney it would probably be nice to implement these using proper exception handling in the future, possibly using algebraic effect handlers.

I should still write some tests to ensure that this does not break. 

Code sample:
```
var db = database "links";
var t1 = table "t1" with (album: String, quantity: Int, bla : Float, flag : Bool) tablekeys [["album"]] from db;
var l1 = lens t1 with { album -> quantity bla };

var bla = "hello";

fun fs1(x) { x.album == bla }
fun fs2(x) { negate (x.quantity) > 5 || negatef (x.bla) > 5.0 && x.flag }
fun fs(x) { fs1(x) && x.quantity == 5 && fs2(x) && not (x.flag) && not (not (x.flag)) && 4 - ( 5 + 10 ) > 2}

lenscheck (lensselect from l1 by fs)
```

yields:
```
(lensselect from (table t1) by (`album` = 'hello' AND `quantity` = 5 AND (- `quantity` > 5 OR - `bla` > 5.0) AND `flag` AND NOT `flag` AND NOT (NOT `flag`) AND 4 - (5 + 10) > 2)) : LensChecked((album : String, bla : Float, flag : Bool, quantity : Int), { album -> bla quantity })
```


rudihorn:
I've just decided to squash merge and force push this as somehow it was all a mess with past rebases etc.

jamescheney:
I think we agreed this can go in pending someone looking over it, provided that the added functionality is still guarded by the relational_lenses flag.

SimonJF:
As an update on this, I've begun using relational lenses for the curation part of my project. We found a few issues yesterday which should be fixed now (one was about aliases, one was about desugaring).

My plan is to use this for a bit, see if the core functionality works, and then delve more into the code. So far it's brilliant though!

jamescheney:
Great, looking forward to seeing this in action - it would be a shame to do all those proofs and then find out this isn't useful for anything...

While you're at it, can you and/or @rudihorn check whether #437 is definitely fixed now - it was pretty debilitating so if you are able to use lenses in a Web application at all, that probably means it's fixed.  Also please feel free to add issues to the "relational lenses" project for wishlist entries.

SimonJF:
Just checked and #437 is definitely fixed by this, yes, if not by a previous patch.

jamescheney:
OK, should close #434 once merged.

SimonJF:
I'm having a few more issues using dynamic predicates. I'm using the most up-to-date version from your branch.

Take the following example, where I'm trying to delete using a lens:

```
var db = database "links";
var t1 = table "t1" with (album: String, quantity: Int, bla : Float, flag : Bool) tablekeys [["album"]] from db;
var l1 = lens t1 with { album -> quantity bla };

fun foo(album) {
  var selectLens = lensselect from l1 by fun(x) { x.album == album };
  var selectLens = lenscheck selectLens;
  lensput selectLens with [];
}

foo("Ultralife")
```

I get the following error:
```
***: Fatal error : unsupported initial (Ir.Closure (2097, [],
                       (Ir.Extend ({2094 => (Ir.Variable 2094); }, None)))) 
```

If I then try hoisting it to an inner function definition, I get the following error:
```
var db = database "links";
var t1 = table "t1" with (album: String, quantity: Int, bla : Float, flag : Bool) tablekeys [["album"]] from db;
var l1 = lens t1 with { album -> quantity bla };

fun foo(album) {
  fun filterAlbum(x) { x.album == album }
  var selectLens = lensselect from l1 by filterAlbum;
  var selectLens = lenscheck selectLens;
  lensput selectLens with [];
}

foo("Ultralife")

***: Fatal error : Unsupported sugar phrase: (Sugartypes.TAppl ((Sugartypes.Var "filterAlbum"),
                             [{ |bla:Float,flag:Bool,quantity:Int }; {  }])) 
```

And hoisting it to an outer function gets me:
```
var db = database "links";
var t1 = table "t1" with (album: String, quantity: Int, bla : Float, flag : Bool) tablekeys [["album"]] from db;
var l1 = lens t1 with { album -> quantity bla };

fun pred(album) { fun (x) { x.album == album } }

fun foo(album) {
  var selectLens = lensselect from l1 by pred(album);
  var selectLens = lenscheck selectLens;
  lensput selectLens with [];
}

foo("Ultralife")


***: Fatal error : Unsupported sugar phrase: (Sugartypes.FnAppl ((Sugartypes.Var "pred"),
                             [(Sugartypes.Var "album")])) 
```

rudihorn:
Thanks for this @SimonJF will fix those bugs! for the moment you might be able to try option 2 with forcing it to specify that album is a string to remove the type application.

SimonJF:
Thanks, @rudihorn!


Alas, adding type annotations hasn't seemed to work:

```
var db = database "links";
typename TableRecord = (album: String, quantity: Int, bla : Float, flag : Bool);
var t1 = table "t1" with TableRecord tablekeys [["album"]] from db;
var l1 = lens t1 with { album -> quantity bla };

fun foo(album: String) {
  fun filterAlbum(x: TableRecord) { x.album == album }
  var selectLens = lensselect from l1 by filterAlbum;
  var selectLens = lenscheck selectLens;
  lensput selectLens with [];
}

foo("Ultralife")

***: Fatal error : Unsupported sugar phrase: (Sugartypes.TAppl ((Sugartypes.Var "filterAlbum"),
                             [{  }])) 

```

(I have plenty of other stuff to do, so don't drop everything just to fix this :) But just so you know)

rudihorn:
@SimonJF I think all of the bugs you found should be fixed in the latest patch. I don't think it will be complete, but it should be a good start.

SimonJF:
That seems to have solved that error! Thanks :)

SimonJF:
Thanks, Rudi -- I've managed to get a basic select / update going now. For deletion I need to worry about foreign key constraints but I think that's something for application logic (at least for now), and way outside of the scope of this PR.

I did run into a type soundness bug though. In short, I can pass things that aren't lists to a `lensput`:

```
var db = database "links";

var tracksTable =
	table "tracks"
	with (track: String, date: Int, rating: Int, album: String)
	tablekeys [["track", "album"]]
	from db;

fun foo() {
  var tracksLens = lens tracksTable with { track -> date rating };
  var tracks = lensget tracksLens;
  var newTracks = (track="High on Humans", date=0, rating=10, album="ultralife");
  ignore(lensput tracksLens with newTracks);
  newTracks
}

foo()
```

Consequently, we then hit an unboxing error (theoretically this amounts to preservation not holding) when trying to evaluate the `lensput` which is expecting a list, but instead gets a record.

Is this something we can expect lenses to be able to statically check?

SimonJF:
(I bet you're loving having a user, right? :) )

SimonJF:
Great, thanks! Could you add a regression test?

core/evalir.ml
```diff
@@ -629,17 +630,27 @@ struct
...
+            ~default:[default]
             ~key:(Alias.Set.singleton key)
+          |> Lens_errors.unpack_type_drop_lens_result ~die:(eval_error "%s")
         in
+
         apply_cont cont env (`Lens (Value.LensDrop { lens; drop; key; default; sort }))
     | LensSelect (lens, predicate, _sort) ->
         let open Lens in
-        let lens = value env lens |> get_lens in
+        let lens = value env lens |> get_lens in        let predicate =
```
SimonJF: Formatting nit -- the `let` should be on a different line

core/evalir.ml
```diff
@@ -669,12 +684,12 @@ struct
...
         let data = value env data |> Value.unbox_list in
         let data = List.map Lens_value_conv.lens_phrase_value_of_value data in
-        let classic = Settings.get_value Basicsettings.RelationalLenses.classic_lenses in
-        if classic then
-            Lens.Helpers.Classic.lens_put lens data
-        else
-            Lens.Helpers.Incremental.lens_put lens data;
-        apply_cont cont env (Value.box_unit ())
+        let behaviour =
+          match Settings.get_value Basicsettings.RelationalLenses.classic_lenses with
```
SimonJF: Why not use `if-then-else`?

core/ir.ml
```diff
@@ -71,10 +71,11 @@ and binding =
 and special =
   | Wrong      of Types.datatype
   | Database   of value
-  | Lens       of value * Lens.Sort.t
-  | LensDrop   of value * string * string * value * Lens.Sort.t
-  | LensSelect of value * Lens.Phrase.t * Lens.Sort.t
-  | LensJoin   of value * value * string list * Lens.Phrase.t * Lens.Phrase.t * Lens.Sort.t
+  | Lens       of value * Lens.Type.t
+  | LensDrop   of value * string * string * value * Lens.Type.t
+  | LensSelect of value * [`Static of Lens.Phrase.t | `Dynamic of value] * Lens.Type.t
```
SimonJF: Is there a reason these are polymorphic variants, rather than just using standard variants but namespaced properly?

core/ir.mli
```diff
@@ -72,10 +72,11 @@ and binding =
 and special =
   | Wrong      of Types.datatype
   | Database   of value
-  | Lens of value * Lens.Sort.t
-  | LensDrop of value * string * string * value* Lens.Sort.t
-  | LensSelect of value * Lens.Phrase.t * Lens.Sort.t
-  | LensJoin of value * value * string list * Lens.Phrase.t * Lens.Phrase.t * Lens.Sort.t
+  | Lens of value * Lens.Type.t
+  | LensDrop of value * string * string * value* Lens.Type.t
+  | LensSelect of value * [`Static of Lens.Phrase.t | `Dynamic of value] * Lens.Type.t
```
SimonJF: as above

core/ir.ml
```diff
@@ -71,10 +71,11 @@ and binding =
...
+  | LensJoin   of value * value * string list * Lens.Phrase.t * Lens.Phrase.t * Lens.Type.t
```
SimonJF: What is the `string list` referring to? I'd recommend either adding a comment, adding a type alias, or (better) making this long payload a record with descriptive names

lens/eval_incremental.ml
```diff
@@ -0,0 +1,217 @@
...
+      in
+      let sort2 = Value.sort right in
+      let proj2 =
+        Sorted.project_onto delt ~columns:(Sort.cols_present_aliases sort2)
+      in
+      let delta_m0 = delta_merge_affected left proj1 in
+      let delta_n0 = delta_merge_affected right proj2 in
+      let on' = List.map ~f:(fun a -> (a, a, a)) cols_simp in
+      let delta_l =
+        Sorted.merge
```
SimonJF: Why not do `let open Sorted` here, to avoid all of the prefixes?

(In fact, there's an argument for doing so before the `match`)

lens/fun_dep.ml
```diff
@@ -98,25 +173,142 @@ module Tree = struct
...
+  let is_acyclic fds =
+    let nextfds node =
+      Set.filter (fun fd -> Alias.Set.equal (left fd) node) fds
+    in
+    let rec find_cycle node bt =
+      match bt with
+      | x :: xs ->
+          if Alias.Set.equal x node then List.append bt [node]
+          else find_cycle node xs
+      | [] -> [Alias.Set.singleton "ERROR RECONSTRUCTING CYCLE"]
```
SimonJF: This seems a bit of a hack -- could it lead to "ERROR RECONSTRUCTING CYCLE" ending up in a user program? I'd expect an exception would be a better way to handle this

lens/lens_format.ml
```diff
@@ -12,5 +14,7 @@ let pp_newline_list f b v = pp_print_list ~pp_sep:pp_print_newline f b v
 
 let pp_padded ~length fmt f v =
   let str = Format.asprintf "%a" fmt v in
-  let count = Pervasives.max 0 @@ length - String.length str in
+  let count = Pervasives.max 0 @@ (length - String.length str) in
```
SimonJF: I bet this one was fun to debug :)

lens/lens_result.ml
```diff
@@ -9,25 +9,41 @@ let bind v ~f =
...
   | Result.Error _ as err -> err
 
-let map v ~f =
-  bind v ~f:(fun v -> f v |> return)
+let map v ~f = bind v ~f:(fun v -> f v |> return)
+
+let ok_exn v =
+  match v with
+  | Result.Ok v -> v
+  | _ -> failwith "Unexpected error unpacking result."
```
SimonJF: Please don't introduce new `failwith`s if you can help it

lens/lens_result.ml
```diff
@@ -9,25 +9,41 @@ let bind v ~f =
...
   match r with
   | Error e -> f e |> error
   | Ok r -> Ok r
 
+let of_bool cond ~error = if cond then return () else Result.Error error
+
+let unpack_error_exn t =
+  match t with
+  | Error e -> e
+  | _ -> failwith "Unexpected error unpacking result to error."
```
SimonJF: as above

lens/lens_seq.ml
```diff
@@ -12,32 +11,27 @@ let return v () = Cons (v, empty)
...
+  | Cons (v, r) -> if f v then Cons (v, filter f r) else filter f r ()
 
 let rec filter_map f v () =
   match v () with
   | Nil -> Nil
-  | Cons (v, r) ->
-    (match f v with
-     | None -> filter_map f r ()
-     | Some v -> Cons (v, filter_map f r))
+  | Cons (v, r) -> (
```
SimonJF: Consider using `begin..end` instead of brackets

lens/operators.mli
```diff
@@ -27,7 +38,10 @@ module Binary : sig
 
   val to_string : t -> string
 
-  val of_string : string -> t
+  val of_string : string -> (t, Operator_not_found_exception.t) result
+
+  (** Do not use. *)
+  val of_string_exn : string -> t
```
SimonJF: Why is this here if we're not allowed to use it?

lens/phrase.ml
```diff
@@ -122,61 +123,96 @@ let[@inline always] val_numeric op_i op_f v1 v2 =
...
+  match expr with
+  | Constant c -> Constant c
+  | Var v ->
+      lookup v
+      |> Option.map ~f:(fun v -> Constant v)
+      |> Option.value ~default:(Var v)
+  | InfixAppl (op, a1, a2) -> (
+      let a1 = partial_eval a1 ~lookup in
+      let a2 = partial_eval a2 ~lookup in
+      match (op, a1, a2) with
```
SimonJF: Consider `begin...end` instead of brackets again (it took me a while to spot the brackets and I was confused). Also, the spacing seems very inconsistent below

lens/phrase_typesugar.ml
```diff
@@ -23,17 +23,24 @@ let rec tc_infix ~env ~data ~op p q =
   tc ~env q
   >>= fun q ->
   match op with
-  | Binary.LogicalAnd | Binary.LogicalOr -> (
+  | Binary.LogicalAnd
```
SimonJF: Spacing is strange here?

lens/sorted_records.ml
```diff
@@ -1,126 +1,133 @@
...
+        let r = compare rs.(m) record in
         match r with
         | 0 -> Some r
-        | a when a > 0 -> pivot s (m-1)
-        | a when a < 0 -> pivot (m+1) e
-        | _ -> failwith "impossible" in
-    let r = pivot 0 ((Array.length rs) - 1) in
+        | a when a > 0 -> pivot s (m - 1)
+        | a when a < 0 -> pivot (m + 1) e
+        | _ -> failwith "impossible"
```
SimonJF: `failwith`

lens/sorted_records.ml
```diff
@@ -1,126 +1,133 @@
...
-        | a when a > 0 -> pivot s (m-1) b
-        | a when a < 0 -> pivot (m+1) e b
-        | _ -> failwith "impossible" in
-    let b = pivot 0 ((Array.length rs) - 1) false in
-    let e = pivot 0 ((Array.length rs) - 1) true in
-    (b,e)
+        | 0 -> if b then pivot (m + 1) e b else pivot s (m - 1) b
+        | a when a > 0 -> pivot s (m - 1) b
+        | a when a < 0 -> pivot (m + 1) e b
+        | _ -> failwith "impossible"
```
SimonJF: failwith

lens/sorted_records.ml
```diff
@@ -164,75 +176,74 @@ let find rs ~record =
...
-    | None -> failwith ("Column " ^ col ^ " not in record set.") in
-  let filter rows = Array.of_list (List.filter (
-      fun r -> Phrase.eval predicate (get_col_val r) = Value.box_bool true)
-      (Array.to_list rows)) in
-  {
-    columns = rs.columns;
-    plus_rows = filter rs.plus_rows;
-    neg_rows = filter rs.neg_rows;
-  }
+    | None -> failwith ("Column " ^ col ^ " not in record set.")
```
SimonJF: failwith

lens/sorted_records.ml
```diff
@@ -164,75 +176,74 @@ let find rs ~record =
...
-        | a when a > 0 -> (* x > y, so take y and try find x *)
-          let (left, right) = do_next left ys in
+      | a when a > 0 ->
+          (* x > y, so take y and try find x *)
+          let left, right = do_next left ys in
           (left, y :: right)
-        | _ -> failwith "impossible"
-      end
-    | _ -> (left, right) (* one of them is empty so return rest *)
+      | _ -> failwith "impossible" )
```
SimonJF: failwith

lens/sorted_records.ml
```diff
@@ -241,35 +252,47 @@ let project_onto rs ~columns =
...
+      @@ Array.to_list rs1.plus_rows
+      |> Array.of_list
+    in
+    {rs1 with plus_rows}
 
 let reorder_cols cols ~first =
   if not (cols_contain cols first) then
-    failwith "Columns do not contain all reorder keys.";
-  let rest = List.filter (fun a -> not (List.mem ~equal:String.equal first a)) cols in
+    failwith "Columns do not contain all reorder keys." ;
```
SimonJF: failwith

SimonJF:
CI failure seems to be transient -- restarted by merging latest `master` in

core/ir.ml
dhil: I think polymorphic variants is the right thing to use here, however as we have decided not to have them in the syntax tree or ir, you should probably use monomorphic variants.

jamescheney: The terminology `Static`/`Dynamic` is slightly misleading here anyway.  The first case means that the convenient syntax for lens predicates is used, whereas the second case allows an arbitrary Links function instead (which allows constructing "dynamic" predicates).  I believe @rudihorn plans to get rid of the first case for the time being anyway, at least from the IR; the convenient syntax using `Lens.Phrase.t` can be desugared to the second case if needed.

rudihorn: > 
> 
> The terminology `Static`/`Dynamic` is slightly misleading here anyway. The first case means that the convenient syntax for lens predicates is used, whereas the second case allows an arbitrary Links function instead (which allows constructing "dynamic" predicates). I believe @rudihorn plans to get rid of the first case for the time being anyway, at least from the IR; the convenient syntax using `Lens.Phrase.t` can be desugared to the second case if needed.

This is now changed. Static and Dynamic now both use the same syntax, but static predicates need to be inline functions that don't refer to any external variables (i.e. completely known statically). This allows a conversion from the sugar directly to `Lens.Phrase.t`, whereas dynamic lenses are first evaluated to IR and then converted to `Lens.Phrase.t`. I have only dropped the convenient syntax, but the static/dynamic differentiation still exists.

I used polymorphic variants, as this seems to be the way it is done for inline types in other projects, but I can convert it to a monomorphic variant.

lens/eval_incremental.ml
rudihorn: I do this on purpose as I find it makes the code more readable as one can directly see which function comes from which module and applies to which type just as I usually leave the module `List` etc. It also works better with autocompletion, as when looking for a specific function typing in `Sorted.` brings up a relevant list: 

image

lens/fun_dep.ml
rudihorn: If a cycle is found, the function `find_cycle` tries to reconstruct the cycle. Hypothetically this should always correctly terminate as it is used to produce an exception. I don't really want exception handling code to produce more exceptions.

lens/lens_format.ml
rudihorn: the `@@` is equivalent to the Haskell $, and should precisely be there to avoid the brackets. The brackets were added by ocamlformat, and I'm slightly confused why... I've removed the @@ for now.

lens/lens_seq.ml
rudihorn: This is an ocamlformat setting which I have left to the default value as I personally prefer brackets over begin end. I find begin end quite quirky and usually results in an additional line, and auto formatting ensures correct / readable indentation.

lens/operators.mli
rudihorn: `_exn` functions are library functions useful for writing quick code / unit tests but as the name suggests they can throw exceptions which is not how most of the code I have been writing is supposed to work. This is why most are implemented with `failwith` as I try to (not 100% succesfully, but working on it) to not use any `_exn` functions in the code.

lens/phrase.ml
rudihorn: >--indicate-nested-or-patterns={space|unsafe-no}
           Control whether or not to indicate nested or-pattern using indentation. space starts lines of nested or-patterns with " |" rather than "| ".
           unsafe-no does not indicate nested or-patterns. Warning: this can produce confusing code where a short body of a match case is visually hidden
           by surrounding long patterns, leading to misassociation between patterns and body expressions. The default value is space.

The "inconsistent spacing" is an indication of nested or patterns, see manual extract above. I didn't want to set it to `unsafe-no`. If the match was missing a bracket it would be indented differently.

lens/phrase_typesugar.ml
rudihorn: See comment above.

lens/operators.mli
jamescheney: Maybe the fact that your intention is to avoid and eventually remove these could be documented by using  `_failwith` as a suffix instead of `_exn`?

lens/lens_seq.ml
jamescheney: I personally grew up with SML's `case...of` where the only option is parentheses since SML didn't have `begin...end`.  I think using `begin...end` in OCaml is preferable since the extra space can be helpful in helping visually find the beginning or ending of a larger (like multi-screen) pattern match.  It isn't as big of an issue here.

Mental note: all future functional languages should have a built-in "end of match cases" syntax delimiter.

lens/fun_dep.ml
SimonJF: OK - in which case, could you make this either an empty set or assert false, depending on the logic?

lens/eval_incremental.ml
SimonJF: OK, matter of taste -- fine by me.

SimonJF:
After discussion with @rudihorn, I will merge this once the branch is updated and the CI passes again.

jamescheney:
@rudihorn can you (or @SimonJF) update the "Relational Lenses" wiki page to account for any changes since it was first written?

rudihorn:
Thanks @SimonJF!

@jamescheney I can look into doing this in the coming days.

PR #628, opened 2019-06-09
[Azure Pipelines] Extend the test matrix

dhil:
This patch extends the Azure Pipelines test matrix to include macOS (version 10.13, I think) in addition to Ubuntu 16.04 LTS.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -44,7 +44,6 @@ steps:
 - script: |
     ./opam2_local switch create ocaml-base-compiler.4.06.1
     eval `./opam2_local config env`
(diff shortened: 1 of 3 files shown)


dhil:
After a lot of trial-and-error I finally managed to implement a configuration that builds and tests Links on Linux and macOS simultaneously. This should help us avoid regressions on the macOS platform (@jamescheney).

There is still room for improvement, for example I haven't made use of Azure Pipelines' notion of jobs yet to enable parallel processing of steps.

I think it is worthwhile to merge this patch, test it for awhile, and then at the next meeting decide whether we want to switch our continuous integration to Azure Pipelines.

jamescheney:
> macOS

Yay!

PR #630, opened 2019-06-10
Upgrade Websocket dependency

dhil:
This patch replaces `websocket-lwt` with `websocket-lwt-unix` to
enable Links to work with the latest versions of its dependencies.

Links depends on [OCaml Websocket library](https://github.com/vbmithr/ocaml-websocket) and its
sublibrary `websocket-lwt`. However, `websocket-lwt` has become stale,
meaning that packages which share common dependencies with
`websocket-lwt` no longer can be upgraded.

It seems that `websocket-lwt` has been superseeded by
`websocket-lwt-unix`, which happens to provide a slightly different
interface.

### DIFF
--- a/core/dune
+++ b/core/dune
@@ -12,7 +12,7 @@
   (synopsis "Links compiler library")
   (modes native)
   (libraries str yojson ppx_deriving_yojson.runtime unix safepass base64
-             ANSITerminal linenoise cohttp lwt websocket websocket-lwt.cohttp
+             ANSITerminal linenoise cohttp lwt websocket websocket-lwt-unix.cohttp
              findlib menhirLib links.lens)
   (preprocess (pps ppx_deriving.std ppx_deriving_yojson)))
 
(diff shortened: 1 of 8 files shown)


dhil:
@SimonJF FYI see [s.ml](https://github.com/mirage/ocaml-cohttp/blob/master/cohttp-lwt/src/s.ml#L109) for the documentation of the new result type `response_action` of `upgrade_connection`.

SimonJF:
Thanks for spotting the issue and doing the update.

I didn't manage to get `examples/sessions/chatServer.links` working with this new patch -- I got
```
jslib.js:394 WebSocket connection to 'ws://localhost:8080/ws/cid_8' failed: Error during WebSocket handshake: Unexpected response code: 500
```

That said, I think this is my responsibility to debug, so I'll take a look.

SimonJF:
OK. There we go. Let's see how the CI goes, but I think this can go in after that (@dhil would you mind taking a look?)

dhil:
@SimonJF LGTM. As we discussed, the whole webserver and processes infrastructure is due a refactor. At least with this change we get to use more recent versions of `cohttp`.

A note to all @jamescheney, @rudihorn, @slindley, @frank-emrich, @jstolarek, @wricciot: this change will break your local builds. Once merged, you must install the `websocket-lwt-unix` package to be able to build Links, e.g.
```
opam update && opam install websocket-lwt-unix
```

dhil:
Let's merge!

Issue #491, opened 2019-02-20
Kind context of type/row/etc. variables should be handled correctly/checked in desugaring


jamescheney:
This works:

```
typename PolyPingPong =
  [+|
    Go: forall a. !a.?a.PolyPingPong,
    Stop: EndQuery
  |+];
```
but this doesn't work:
```
typename PolyRowPingPong =
  [+|
    Go: forall a::Row. !(a).?(a).PolyRowPingPong,
    Stop: EndQuery
  |+];
```
it results in an error `*** Fatal error : Free variable (a) in definition of typename FPolyPingPong
`
I think that the second and third examples should both work or both not work (and if not, better error message).  

Also, the session types above are a red herring: the same kind of behavior using variants also works:
```
typename FPolyPingPong =
  [| Go: forall a. (a) {}-> (a,FPolyPingPong)
   | Stop: ()
  |];
```
but locally quantifying over a row variable does not work:
```
typename FPolyRowPingPong =
  [| Go: forall a::Row. ((a)) {}-> ((a),FPolyRowPingPong)
   | Stop: ()
  |];
```

jamescheney:
Ah wait, but this works:

```
typename FPolyRowPingPong2 =
  [| Go: forall a. ((a)) {}-> ((a),FPolyRowPingPong2)
   | Stop: ()
  |];
```
because `a` is used as a `Type` above rather than a `Row` (the `(a)`s are treated as 1-tuples / ignored).  

This also works:
```
typename FPolyRowPingPong =
  [| Go: forall a::Row. ((|a)) {}-> ((|a),FPolyRowPingPong)
   | Stop: ()
  |];
```
because `a` is used in a place where a row variable is expected, which I forgot to do before.

So the problem is really that ill-kinded type variable occurrences are not handled correctly during datatype desugaring, leading to a very mysterious error message.  (When we see `a` in a place where we expect a `Type` but it was actually quantified as a `Row`, the error message just saying that `a` is unbound is misleading; it just wasn't bound in the context we were expecting to find it.)

I also wonder if the right thing happens in the presence of shadowing:
```
typename FPolyRowPingPong3 =
  [| Go: forall a:: Type. forall a::Row. ((|a)) {}-> ((a),FPolyRowPingPong3)
   | Stop: ()
  |];
```
Spoiler alert: the above is apparently just fine!

[...]

Issue #344, opened 2018-06-13
Type variable names are shared amongst definitions of mutually recursive functions


dhil:
Somehow type variables become shared across mutually recursive functions, e.g.
```
sig f : () {hear{a} |_}~> ()
fun f() { g() }

sig g : () {hear{_} |a}~> ()
fun g() { f() }
```
In `f` the type variable `a` is intended to have kind `Presence`, while the type variable `a` in `g` should be of kind `Row`. That is, both variables are supposed have different implicit quantifiers. However the above program produces the following kinding error
```
*** Fatal error : kind mismatch in type variable: ("a", (`Presence, None), `Rigid) vs: ("a", (`Row, None), `Rigid)
```
Introducing explicit quantifiers resolves the problem; the following program type checks
```
sig f : forall a::Presence . () {hear{a} |_}~> ()
fun f() { g() }

sig g : forall a::Row . () {hear{_} |a}~> ()
fun g() { f() }
```
Omitting either of the quantifiers results in one of two errors.

1. Omitting `forall a::Presence` in the signature for `f` yields the error ``*** Fatal error : kind mismatch in type variable: ("a", (`Presence, None), `Rigid) vs: ("a", (`Row, None), `Rigid)``
2. Omitting `forall a::Row` in the signature for `g` yields another error ``*** Error: DesugarDatatypes.UnexpectedFreeVar("a")``

In the absence of recursion the problem does not arise as the follow program demonstrates
```
sig f : () {hear{a} |_}~> ()
fun f() { g() }

sig g : () {hear{_} |a}~> ()
fun g() { () }
```
And for completeness, the problem does not arise when the names are different, e.g.
```
sig f : () {hear{a} |_}~> ()
fun f() { g() }

sig g : () {hear{_} |b}~> ()
fun g() { f() }
```

jamescheney:
This seems potentially related to #42, #377, #368.

Issue #327, opened 2018-02-27
Duplicate type parameters


dhil:
It is possible to define type names which have duplicate type parameters, e.g.
```
links> typename T(a::Type, a::Type) = a;
T = a,b.a
```
This kind of code should arguably be rejected by the compiler. To fix this issue we need to check for duplicate names (just as we do for function parameters) in type names.

PR #631, opened 2019-06-11
Desugar datatype cleanup

SquidDev:
This is about the third iteration of this PR, which is somewhat embarrassing. The previous iterations were more of an overhaul to the implementation, but neither ended up especially nice, so I'm going with this for now:

- Store all type variables in one scope. This avoids problems where different variables with different kinds did not shadow each other (meaning `forall a::Eff. forall a::Type -> () ~a~> ()` and the like was valid).
 - Improve error messages when we encounter unexpected problems (kind mismatches, unexpected free type variables, etc...) (see #491).
 - Error on duplicate type variables within quantifiers (forall, type aliases) (see #327).
 - Prevent mutually recursive top-level functions sharing type variables (see #344).
 - A couple of minor cleanups to the general processes.

I've think I've added test cases for the above issues, but let me know if there's anything I missed!


SimonJF:
> This is about the third iteration of this PR, which is somewhat embarrassing. 

Never be embarrassed about iterating a PR! It shows that you care about its quality :)

core/desugarDatatypes.ml
```diff
@@ -140,103 +185,128 @@ object (self)
...
   method! fieldspec = let open Datatype in function
     | Absent -> self
     | Present t -> self#datatype t
     | Var (x, k, freedom) -> self#add (x, (pk_presence, k), freedom)
 end
 
-type var_env = { tenv : Types.meta_type_var StringMap.t;
-                 renv : Types.meta_row_var StringMap.t;
-                 penv : Types.meta_presence_var StringMap.t }
+type var_env = { tyvars : meta_var StringMap.t }
```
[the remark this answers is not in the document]
dhil: I think it is fine to leave it as a record. Possibly add a no boxing annotation, e.g.
```ocaml
type var_env = { tyvars : meta_var StringMap.t }
               [@@unboxed]
```

jamescheney:
@SimonJF Just to clarify, yes, the `FPolyPingPong3` example definitely should not work!

core/desugarDatatypes.ml
```diff
@@ -140,103 +177,128 @@ object (self)
...
       if tv <> tv' then
-        raise (internal_error
-          ("kind mismatch in type variable: " ^
-           Sugartypes.show_type_variable tv ^ " vs: " ^
-           Sugartypes.show_type_variable tv'));
+        raise (typevar_mismatch Position.dummy tv tv');
       self#register tv
-    else
-      (self#register tv)#add_name name
+    end else (self#register tv)#add_name name
```
jamescheney: Nitpick: put "else" on the next line?

core/desugarDatatypes.ml
```diff
@@ -123,12 +156,16 @@ object (self)
...
+
+  method register ((name, _, _) as tv) = {<tyvars = StringMap.add name tv tyvars>}
+
+  method replace name map = {<tyvars = StringMap.update name (fun _ -> StringMap.find_opt name map) tyvars>}
+
   method bind tv = self#register tv
 
   method add ((name, (pk, sk), freedom) as tv) =
-    if StringMap.mem name tyvars then
+    if StringMap.mem name tyvars then begin
```
jamescheney: Nitpick: put "begin" on next line (and indent)?

tests/desugar_datatypes.tests
```diff
@@ -0,0 +1,72 @@
...
+./tests/desugar_datatypes/nestedTypename.links
+filemode : true
+exit : 1
+stderr : @.*Unbound type variable `a'.*
+
+Rows can be promoted to types
+./tests/desugar_datatypes/rowPromote.links
+filemode : true
+stdout : () : ()
+ignore : Not sure if a good idea yet.
```
jamescheney: The original example worked only because the `a` appearing in the session type was assumed to be a type variable.  We don't allow implicit conversions from row (variables) to types elsewhere and should not do so here either.  I think this test case should fail.

slindley:
@jamescheney that's definitely something we want to do, but it won't be feasible to properly implement until we have a uniform representation of different kinds of types - which is the right way to go, but will require quite a big refactoring. (The issue is that, e.g., `Alias is not a constructor for rows.)

Doing the right thing will require moving to a single (OCaml) data type for all kinds of (Links) data type both for the frontend (Sugartypes) and the backend (Types) representations.

tests/desugar_datatypes/rowPromote.links
```diff
@@ -0,0 +1,5 @@
+typename PolyPingPong =
+  [+|
+    Go: forall a::Row. !a.?a.PolyPingPong,
```
jamescheney: Or, this example should be fixed to be `!(a).?(a).PolyPingPong` that is to actually convert the row variables into types. (as in #491).  But we should not be promoting row variables to type variables no matter what.

jamescheney:
@slindley Ah, OK.

tests/desugar_datatypes/rowPromote.links
SquidDev: Hrmr - from what I can tell, there shouldn't be a difference between `!(a).X` and `!a.X` any more, so this should fail in both cases (you'd need an explicit `(a|)`). Maybe getting the wrong end of the stick though!

I've just deleted this test case for the time being. It was ignored, and I don't think it really checks anything which wasn't done elsewhere in the suite.

jamescheney: I think you're right, I hadn't reread the later part of the issue where I (re)learned the harsh lesson that `(a)` does not turn a row into a type like I thought.  (I think you mean `(|a)` though.)

As long as no Row->Type implicit coercion was happening (or intended) I'm happy now!

SquidDev:
I've just pushed a small patch for desugar datatypes not checking subkinds annotations were consistent on type aliases (`typename A(a::Any) = a::Base` was considered fine). Subkinds are currently (somewhat) checked when finding free type variables - as type names don't have free variables, they were skipped and so the check was never done.

The current fix is just to include type aliases within the free variable gathering pass, but exclude them from the final result. It works fine, but it feels a little ugly mixing the two things together - I don't know if there's a nicer solution?

dhil:
> Edit: Put together an alternative implementation in SquidDev/links@31057db, which infers kinds for unannotated variables and ensures subkinds are consistent for each type alias/forall. Don't really know if that's a desirable change or not though...

I think that is great. We just need to hide it under a flag/settings for the time-being, e.g. `infer_kinds`.

dhil:
Thanks @SquidDev. Let's merge!

SquidDev:
@dhil I assume I'm OK to close all the linked issues?

jamescheney:
Please do!  (This can happen automatically on merge if you say "closes #xyz")

Issue #584, opened 2019-04-30
Abandon Travis in favour of Azure Pipelines?


dhil:
Travis offers only dated solutions for projects like ours, meanwhile Microsoft [Azure Pipelines](https://github.com/marketplace/azure-pipelines) offers parallel jobs and a more recent version of Ubuntu, for free.

frank-emrich:
I'm in favor of this, because 
1) It gives us Ubuntu 16.04 rather than the ancient 14.04
2) Sometimes, when multiple people have open PRs they push to and/or a branch in the main Links repo used for active development, it can take some time until your CI job is executed. This is because Travis doesn't give us parallel jobs.

dhil:
... also since Microsoft has acquired GitHub it may be a safe bet to use their CI solution.

dhil:
As an experiment commit [c1b4de49](https://github.com/links-lang/links/commit/c1b4de491f89ab327e9295b1e31cbc199780f9b8) enables Azure Pipelines. The current config is essentially just a transcription of the Travis config, and as such, it does not take advantage of parallel jobs.

dhil:
Patch #628 adds macOS to the test matrix.

dhil:
Patch #633 fixes a regression in the macOS image. Suddenly overnight `brew services start postgres` began to fail due to wrong permissions. The patch ditches the `brew services` interface and instead uses `pg_ctl` to interact directly with the postgres service.

This incident may indicate that using `brew` in the CI image is a poor choice.

jstolarek:
These random failures would be tolerable if there was a sane way to restart a build. Creating a dummy commit does not count as sane. For me that's a blocker for transitioning to pipelines.

dhil:
> These random failures would be tolerable if there was a sane way to restart a build.

Ah, you can restart a build from the web interface. It probably doesn't show for you, because I haven't given you access! I only gave the students access, because I didn't want to spam your inboxes. I should have sent you an invite now.

jstolarek:
What credentials should I use for login?

dhil:
Your university (Microsoft) account. You should have received an invite on that e-mail -- at least I was able to look you up in the Office365 directory.

dhil:
I had to push two changes to the Azure Pipelines configuration (#779 and #781) due to changes in either `brew` or the macOS image. I wonder whether there exists a better alternative to `brew` or a way to configure it such that the CI is more resilient against third party changes.

frank-emrich:
[...]

PR #633, opened 2019-06-12
[CI] macOS database installation regression

dhil:
It seems that the `brew` package manager might be somewhat unstable. An update to `brew` overnight has introduced a regression on the CI image for macOS, which means the postgresql service no longer starts.

What follows is a series of ridiculous attempts to fix the macOS database installation regression.

Related #584.


dhil:
The problem was the `brew services` command. Rather than using `brew` to interface with the `postgresql` service, the script now interacts directly with `pg_ctl` to start the service.

PR #635, opened 2019-06-12
Publish test reports to Azure pipelines

SquidDev:
I'm not sure how useful/desirable this actually is - so feel free to close this! Just felt like experimenting a little bit with it all...

This changes a couple of the test runners to generate JUnit-style XML files, which can then be processed by Azure Pipelines. Effectively means you have a [fancy interface](https://dev.azure.com/links-lang/links/_build/results?buildId=64&view=ms.vss-test-web.build-test-results-tab) <sub>([image if the link doesn't work](https://user-images.githubusercontent.com/4346137/59389066-5774da80-8d65-11e9-8ccd-037165815e95.png))</sub> to see which tests passed/failed/were ignored as part of the CI.

Some caveats with the current implementation:
 - This only works for tests run from OUnit or the `test-harness` script (so none of the database ones, nor `*.testscript` files).
 - We generate XML files within `test-harness`, which is a bit icky.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -90,7 +90,7 @@ steps:
 
   - script: |
       eval `./opam2_local env`
-      make tests
+      LINKS_TEST_EXPORT=1 make tests
     displayName: 'Server-side tests'
 
   - script: |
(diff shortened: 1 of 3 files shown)


dhil:
I think we can stash this PR for now and revisit it later.

PR #641, opened 2019-06-13
Implement `partition` and `span` in prelude

SimonJF:
This patch incorporates some prelude tweaks I had in my GtoPdb development branch. It implements the `span` and `partition` functions, and defines `takeWhile` and `dropWhile` in terms of `span`.

### DIFF
--- a/prelude.links
+++ b/prelude.links
@@ -257,30 +257,31 @@ fun join(glue,list) {
 # elemBy :: ((a) -> (a) -> Bool) -> (a) -> ([a]) -> Bool
 # notElemBy :: ((a) -> (a) -> Bool) -> (a) -> ([a]) -> Bool
 # span :: ((a) -> Bool) -> ([a]) -> ([a],[a])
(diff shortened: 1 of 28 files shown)


prelude.links
```diff
@@ -257,30 +257,31 @@ fun join(glue,list) {
 # elemBy :: ((a) -> (a) -> Bool) -> (a) -> ([a]) -> Bool
 # notElemBy :: ((a) -> (a) -> Bool) -> (a) -> ([a]) -> Bool
 # span :: ((a) -> Bool) -> ([a]) -> ([a],[a])
+sig span : ((a) ~b~> Bool, [a]) ~b~> ([a], [a])
+fun span(p, xs) {
+  fun go(acc, xs) {
```
dhil: We should really have an optimisation pass for this; you can eliminate the `go` closure by adding `p` to its parameter list.

SimonJF: (very) well-spotted -- done.

PR #642, opened 2019-06-13
Allow primes/single-quotes in identifier names

dhil:
This patch extends the definition of `def_id` such that identifier names may contain primes/single-quotes. However, the initial character of an identifier may not be a prime. Thus one may write `var x' = 42;` but not `var 'x = 42;`.

There is one exception: identifiers for alien things may not contain primes, this restriction is due to our current compilation scheme for aliens, where we expect their Links source names to be the same as their JavaScript source names, but sadly JavaScript does not permit primes in identifier names (yet).

### DIFF
--- a/core/errors.ml
+++ b/core/errors.ml
@@ -160,8 +160,11 @@ let format_exception =
      | None -> pos_prefix message
      end
   | PrimeAlien pos ->
-     let pos, _ = Position.resolve_start_expr pos in
-     pos_prefix ~pos "Foreign binders cannot contain single quotes `'`."
+     let pos, expr = Position.resolve_start_expr pos in
+     let message =
+       Printf.sprintf "Syntax error: Foreign binders cannot contain single quotes `'`.\nIn expression: %s." expr
+     in
+     pos_prefix ~pos message
   | Sys.Break -> "Caught interrupt"
   | exn -> pos_prefix ("Error: " ^ Printexc.to_string exn)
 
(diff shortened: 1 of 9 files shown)


dhil:
> Have you checked how this interacts with the Emacs syntax mode? 

Good call, I have updated the highlight patterns.

> Also, have you tried out some client programs with primed variables?

Yes, I tried this
```
var x' = 42;

fun f'() client {
  x'
}

fun mainPage(_) {
  page
   <#>
      <button l:onclick="{println(intToString(f'()))}">Click me!</button>
      {stringToXml(intToString(x'))}
   </#>
}

fun main() {
  addRoute("/", mainPage);
  servePages();
}

main()
```

SimonJF:
LGTM technically. I think this is a nice enhancement and am happy for it to go in.

I imagine, as per Wadler's law, that this is something that people will have opinions about, however :) So let's see what others say before merging.

dhil:
I think the controversy-factor of this patch is somewhat low. We have discussed this at a meeting a couple of years ago, for whatever reason my internal cache fetched it from the persistent memory store yesterday.

jamescheney:
I like using primes in identifiers too, the main complication is some languages use them for string literal quotes, including JavaScript.  So things like foo'bar' should be checked to make sure that they're escaped correctly (and don't translate to something JavaScript interprets "identifier foo followed by string literal bar").  I guess this is the point of the example above, if so great.

dhil:
> [..] the main complication is some languages use them for string literal quotes, including JavaScript. So things like foo'bar' should be checked to make sure that they're escaped correctly [..]

Agreed. The identifier `foo'bar'` currently translates to `fooprimebarprime_<unique id>` on the client. This is following the scheme we have for translating other symbols like `@`, `.`, etc.

PR #643, opened 2019-06-14
Optional support for coercing null DB integers

SimonJF:
Currently, trying to read from a table which contains `null` integers in the database results in a runtime `int_of_string` error.

This patch allows a rather hacky workaround until we have proper support for this, in that we can define a default value to use if a null integer is detected.

In particular, we have two settings: `coerce_null_integers` and `null_integer`. If `coerce_null_integers` is set to true, then `null_integer` (default -1) will be used instead of crashing at runtime.


jamescheney:
This looks reasonable to me, with the caveat that the code should document that this is a "temporary workaround" just in case it turns out not to be so temporary...

I do have a list of ideas somewhere of better ways to do this.

SimonJF:
Caveat added in `database.ml`.

PR #644, opened 2019-06-14
[CI] Travis: Upgrade Trusty -> Xenial

dhil:
Supposedly Travis supports Ubuntu 16.04 LTS now, let's see...


dhil:
Indeed it does, let's upgrade!

PR #645, opened 2019-06-14
Playing with Azure

rudihorn:
This is just a PR for me to experiment with azure. Ignore.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -21,7 +21,7 @@ steps:
     displayName: Environment
 
   - script: |
-      sudo apt-get install m4
+      sudo apt-get install m4 -y
     displayName: 'System dependencies'
 
   - script: |
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -26,7 +26,7 @@ steps:
       sudo usermod -a -G opam `whoami`
       ls -lah ~/.opam/4.07/.opam-switch/lock
       # sudo chown `whoami`:`whoami` ~/.opam/4.07/.opam-switch/lock
-      newgrp -
+      newgrp `id -g opam`
       groups
     displayName: 'Setup environment'
 
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -24,7 +24,11 @@ steps:
       cat >/tmp/opam  <<-EOF
         #!/bin/bash
         
-        sudo -u opam /usr/bin/opam $@
+        if [ "$1" -eq "env ]; then
+          /usr/bin/opam env
+        else
+          sudo -u opam /usr/bin/opam $@
+        fi
       EOF
       sudo chmod +x /tmp/opam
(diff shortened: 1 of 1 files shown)


rudihorn:
This is an experimental branch where I have been trying to switch from downloading and setting up ocaml to using the docker hub container for opam. The advantage of this is that it is quicker as there is a lot less to download (takes 1m 49 seconds, then 1min to install ocamlformat dependencies, links dependencies unknown), with possibly more potential for improvement as/if Microsoft has better container caching. In comparison the current setup requires 4m 36s to setup ocaml, 2m 5s to setup ocaml and 3m 41s for the links dependencies.

The current issue is that docker is a bit quirky, and the opam docker image is setup to login with the user `opam`, but azure has some  dodgy setup for setting up users:
```bash
##[command]/usr/bin/docker exec  0c1681c9740ac7169e652ae67bf147b6019e6917a11bc3cbb0257e73c3f5a53f bash -c "cat /etc/group"
##[command]/usr/bin/docker exec  0c1681c9740ac7169e652ae67bf147b6019e6917a11bc3cbb0257e73c3f5a53f groupadd -g 117 azure_pipelines_docker
##[command]/usr/bin/docker exec  0c1681c9740ac7169e652ae67bf147b6019e6917a11bc3cbb0257e73c3f5a53f usermod -a -G azure_pipelines_docker vsts_azpcontainer
```

As a workaround I add the option to force using root (`-u 0`), but this means that all commands in the azure file seemingly need to be prepended to use the correct user.

Bottom of the line: are people interested in pursuing improving this further? I think it would be possible to perform the same improvements using travis as well, and having faster CI checking seems quite helpful.

rudihorn:
I think it is also possible for us to have our own machine added as a worker machine to the azure pipelines, which could have the cached docker images. This would probably having the 1m 40 sec task be run almost instantly.

rudihorn:
With a bit more playing I was able to get this to fully work and to use my local uni machine, completely cutting down on the container initialization time yielding:

image

The remaining large bottlelnecks are now install the database server (can be fixed by running a seperate service container), installing ocamlformat and links dependencies (maybe build a docker image on top of opam).

rudihorn:
image

By switching to a database service container I have been able to reduce the database setup time from just under 4 mins to 22 seconds, and somehow the links dependencies time has sped  up by a lot to.

rudihorn:
For reference the config file used is: https://github.com/rudihorn/links/blob/azure_hack_serv/azure-pipelines.yml

dhil:
I think using our own machine (or providing additional infrastructure) is sub-optimal. Nevertheless, I suppose the time you list is only for the Linux pipeline, thus I suppose the total time would still be bounded by the MacOS pipeline.

I will stash this for now. I think we should first discuss whether we want to adopt Azure Pipelines as our CI provider. If we agree to do so, then we can start thinking about optimising time spent on CI.

Issue #603, opened 2019-05-15
Proper modules migration plan


dhil:
This is a meta-issue to track the progress of upstreaming the new
module system.

My intention is that the current and new system should live
side-by-side for a while to avoid inadvertently breaking (correct)
code that works today. The reasoning is that the new module system
should be an improvement, not a regression. But there may be bugs
lurking in there. Initially the new module system will be an opt-in
option (in other words, disabled by default). After a couple of
"stable" releases, I envisage removing the old module system and
enable the new one by default.

I have spent some time getting an overview of the implementation of
the new module system. Since it is a rather big and invasive change, I
would like to upstream it in batches as outlined below.

## Phase 0

- [ ] Introduce the syntax and semantics of `import` (Patches #599 and #602).
- [ ] Patch up the current module system (Patch #602).
- [ ] Switch examples to use `import` rather than `open` at the top-level.

## Phase 1

- [ ] Introduce the notion of compilation units.
- [ ] Introduce a flag for toggling between loaders, e.g. `--legacy-loader` (defaults to on).
- [ ] Upstream an alternative loader (currently exists in a [separate branch](https://github.com/dhil/links/tree/master-loader)) .

## Phase 2

- [ ] Consolidate qualified and unqualified names internally.
- [ ] Update typing environment to work with qualified names.
- [ ] Introduce a flag for toggling between module systems, e.g. `--legacy-modules` (defaults to on).
- [ ] Type check modules.
- [ ] Compile modules away during the sugar-to-ir pass.

The timescale, that I have in mind, spans a couple of weeks. I would
like to resolve this issue by June 6. Meaning that the new module
system would effectively be upstreamed by the beginning of June.

Closing this issue should involve transitively closing #394.

dhil:
I am obviously lagging behind my own plan; I will try to land the loader next week.

jamescheney:
Snoozing until post-0.9.0

PR #648, opened 2019-06-15
Removes the `dumpTypes` function from the Links standard library.

dhil:
This patch removes the built-in function `dumpTypes`. Judging from the
comment in `dumpTypes.ml` it seems that its sole purpose was to
provide some interface for displaying type information in an IDE. The
comment even states that it is a hack. Hack or not, the function does
not work any longer as it triggers an internal failure on use. My
guess is that this has been the case for a while.

I would like to remove this function as it specifically requires
access to the `prelude` typing environment. Access was provided via a
global hack which is at odds with how I would like to structure the
new modular loader (c.f. #603).

If we want to provide support for any kind of inspection, then I
firmly believe we should do the proper thing and generate object files
with adequate meta information.


frank-emrich:
I'm always in favour of removing unused, unmaintained code :)

slindley:
This was far too fast. It's probably fine to remove this code in this case, but in future please don't do the "oh I don't know what this is, it doesn't seem to be needed, I'll just delete it" thing without asking me first. I believe this was used by an undergraduate student of mine about 11 or 12 years ago for a project in which he wrote a nice web-based IDE for Links:

  http://links-lang.org/papers/undergrads/tom.pdf

We don't have a working version of Tom's IDE, though perhaps this kind of functionality could be useful in future - e.g. for an improved emacs mode.

Next time someone does this, I will revert pending a proper chance to assess what is actually being deleted.

jamescheney:
Sorry, I was aware that this was used in a project, but didn't mention it (I became confused as to which project, since I wasn't applying the rewrite "Carl" -> "Tom").  It looks like the code for LODE still exists in branch `tom`, where `dumpTypes` still exists, and if we want to resurrect it then the branch is still there.

In any case, seems like this should have been an issue for discussion before taking action.

slindley:
Yeah - it should have been an issue for discussion before taking action.

dhil:
> This was far too fast. It's probably fine to remove this code in this case, but in future please don't do the "oh I don't know what this is, it doesn't seem to be needed, I'll just delete it" thing without asking **me** first.

I disagree. If nobody is willing to fix the broken code, then the right thing to do is to just get rid of it. We should not collect a insurmountable broken code base. I don't think it is feasible to have you assess every decision before it is made. Instead we should make much more use of the power of revision control, you even say it yourself

> Next time someone does this, I will revert pending a proper chance to assess what is actually being deleted.

We can just restore it again, no harm done.

I don't believe in the argument that we should keep a feature around because it was part of a project/phd/master/undergraduate thesis. It is definitely nice to be able to read the project/thesis and then run the code in Links. However, the right thing to do is to **tag** the master branch whenever a project or thesis is completed. That way one can always restore the exact state from the time of the project. We should be much happier getting rid of code that nobody wants to maintain.

slindley:
You're missing the point. It should be established what is being removed before removing it. Absolutely, we can and should remove code that is no longer of interest. The problem is that recently people have been a bit too gung-ho about removing features they don't understand without proper discussion. In this case the pull-request you created was fine, but it should not have been approved without discussing with me or someone else who was familiar with the feature first.

SimonJF:
Holding my hand up here: I was too keen with the merge button. My reasoning is that it seemed like an "obvious" win: the code was a hack (and stated as much); it was only used by ancient code that's no longer around; there was general consensus on the PR that removing it was a good idea; and it was actively blocking a new feature.

That said, it highlights a weakness in our previous process. Lessons learnt -- let's make sure we get that right going forward.

PR #649, opened 2019-06-17
Type "unsafe" signatures

SquidDev:
There are several functions within the prelude (`map`, `concatMap`, etc...) for which we override their type definition to convert them from wild functions to non-wild.

However, this modification is done _after_ the prelude is loaded. As a result, desugaring list comprehensions which occur within the prelude results in an inconsistent type - we say it is pure, but `map`/`concatMap`, etc... are wild. As a result, intermediate desugaring stages do not type check.

This PR allows overwriting the type at the definition point, so the "correct" type is available immediately, and so desugared prelude functions will type-check.
```
sig sortByBase : ((a) ~b~> (| _::Base), [a]) ~b~> [a]
unsafe sig sortByBase : ((a) -b-> (| _::Base), [a]) -b-> [a]
fun sortByBase(f, l) {sortBy(f, l)}
```

There is a small form of sanity checking here: we attempt to unify the inferred signature with the "unsafe" one, so wildly different signatures will be rejected.

That said, it's not perfect - you can specify a more general type on the unsafe signature (e.g. `unsafe sig x : a -> a sig x : Int -> Int`) - probably something to improve, but wasn't sure of a nice way of achieving this.


SimonJF:
Sorry, without being involved in the discussion around this PR I'm unsure of what it's trying to do. Could you please edit the PR description with a clarification? Thanks!

jamescheney:
@SimonJF If I understand right, what this is doing is getting rid of the workaround that lets us define `concatMap`, `map`, and `sortBy` in the prelude as recursive functions (for host-side execution) yet also give them types without `wild` effects (so that they can be used on the database side).

Getting rid of this workaround in favor of a more generic workaround seems like a reasonable step, but IMHO the design here could / should be improved long-term so that `unsafe` isn't needed.  But I don't object to this especially if it decreases technical debt in other places.  I'm not totally certain it does; for example the number of lines of code actually increases slightly (though a few are tests).

frank-emrich:
One major advantage of adding this feature is that it allows us to get rid of the special handling for the prelude. It can then be processed as any other file.

jamescheney:
@frank-emrich Yes, good point - and/or the currently specially treated prelude functions could be moved into another stdlib module, e.g. `List`.

SimonJF:
I talked to @SquidDev earlier about this and am in agreement (although I think the PR needs a better description for the commit record).

I do wonder how `unsafe` type definitions work with polymorphic type variables.

slindley:
This is exactly the design that @frank-emrich, @SquidDev, and I converged on. It is a quick hack that will allow us to run the type checker after desugaring for comprehensions. Indeed we may want something more general in future, but the current priority is to get into a state in which desugaring phases correctly preserve type correctness first.

An interesting question is how best to do something similar with #621.

SquidDev:
Sorry for the slow reply, and apologies for terrible PR description - it all sounded much clearer in my head! I've updated the description to something I hope is clearer, but do shout if there's anything missing.

> Would an unsafe signature consisting of only a type variable `a` unify with a "proper" signature?

Sadly, yes. I wasn't sure of a nice way of avoiding this - ideally we'd only unify effects (or even just remove the wild effect and check the types are exactly the same).

core/typeSugar.ml
```diff
@@ -3803,7 +3832,7 @@ and type_binding : context -> binding -> binding * context * usagemap =
                          let _, fti = Instantiate.typ ft in
                          let () = unify pos ~handle:Gripers.bind_rec_annotation (no_pos shape, no_pos fti) in
                            ft
-                     | Some _ -> assert false
+                     | Some _ -> failwith "Sugartypes.datatype without Types.typ"
```
SimonJF: Can you make this `raise Errors.internal_error` rather than `failwith`?

jamescheney:
>  I wasn't sure of a nice way of avoiding this - ideally we'd only unify effects (or even just remove the wild effect and check the types are exactly the same).

I'm not sure I understand the rationale for the current design.  

Why not just infer a type as normal, and (if `unsafe` is present) just traverse the type and remove all `wild` effects?  That would be slightly less general, but would cover the current use cases, and I don't think we necessarily want to use this feature any more than currently (the only situations I can think of are adding new query language features, e.g. `distinct`, that would be handled the same way).

Was this alternative considered and discarded?

SquidDev:
> Why not just infer a type as normal, and (if unsafe is present) just traverse the type and remove all wild effects? [...] Was this alternative considered and discarded?

Honestly, this just didn't occur to me - it feels to me like a cleaner solution, so go that route if @slindley agrees?

slindley:
@jamescheney are you suggesting that we'd write something like:

    unsafe sig sortByBase : ((a) ~b~> (| _::Base), [a]) ~b~> [a]
    fun sortByBase(f, l) {sortBy(f, l)}

?

jamescheney:
Ah, I didn't have any change to the surface syntax in mind, just the check that the inferred type is "correct" modulo wild.  Unifying the two types seems way too permissive.  I think what I was suggesting was instead of unifying the inferred type with the expected type, just take the inferred type, remove all occurrences of `wild` (or equivalently, make `wild`'s presence polymorphic rather than Present) and check that the expected type and erased inferred type are exactly equal.  This seems more conservative while not requiring writing a separate "type equality" function just for this.  But maybe writing the "erase wild" pass is almost as much work?

slindley:
Ah OK. I had something like that in mind initially too. But at this stage I don't think it really matters. If it's a really easy change then let's do it; if not, then let's stick with the current implementation.

SquidDev:
I've made some of the modifications suggested by James, which has made things much stricter, but I'm sure it's quite what was suggested.

I'm not entirely confident how type equality would work here, as the rigid variables will be different across the inferred and expected types. Sorry - still not familiar with all the ins-and-outs of the type checker.

jamescheney:
Concretely, my intuition for what should be happening is as follows:
1.  We infer a type for the function (or group of mutually recursive functions; not sure if "unsafe" is allowed for them.)
2. We erase `wild` effects as you are doing.
3. Finally, we check the inferred type against the ascribed one.  (By default the inferred type should be more general than the ascribed type, they don't have to be equal as I mistakenly suggested before.)

Links already does steps 1 and 3 for ordinary type signatures, so my intuition is that just removing the `wild` effects and then doing what we did before should be fine; at least, it shouldn't introduce any new bugs that we didn't have before (aside from new possibly unsound uses of `unsafe`, of course.)

I won't insist on doing it this way (given that it seems not to simplify things any more), but this looks reasonable.  Thanks!

jamescheney:
Or wait, maybe my understanding of what is needed is too simplistic: 
I would have expected "unsafe" to be a simple boolean annotation indicating whether or not to sanitize away any `wild` effects from the inferred type, and then check against the ascribed type.

It looks like you actually are allowing both normal and unsafe signatures on the same declaration?  Are we actually using this?  Skimming through the code diff for `unsafe_signature` I am not sure I see anywhere this is actually being used, other than as an argument to `type_unsafe`.


Likewise it looks like in typeSugar:3796 where you are calling `type_unsafe` you are retaining both the original signature t and the unsafe signature ft. 

Are we using the two signatures later in a context-dependent way?  If so maybe what I had in mind is too simple-minded to work.

dhil:
Some alternatives to the `unsafe` syntax

- `trustme`
- `iknowwhatiamdoing`
- `isolemnlyswearnevertoabusemypower`
- `ishallnowgowashmyhandsforifeeldirty`

jamescheney:
I think I follow/approve.  One minor refinement could be to reorganize 
```
    fun_signature: datatype' option;
    fun_unsafe_signature: bool;
```
to 
```
    fun_signature: (datatype' * bool) option;
```
thus avoiding the impossible corner case of an `unsafe=true` when `fun_signature=None`.  But this doesn't seem like enough of an issue to justify reverting all those changes, and might create headaches where `fun_signature` is already used.

jamescheney:
Is there any obstacle to merging this, other than the need to bring up to date with master?

SquidDev:
Ahhh, sorry! This somewhat got subsumed into Sam's ordered-quantification branch. I don't know if it's better to merge this now, or just close it and wait for that branch to be merged instead?

slindley:
Let's close this and merge it as part of the ordered-quantification branch - which I'd like to get merged soon.

PR #653, opened 2019-06-20
Disable dune implicit-transitive-deps.

dhil:
It looks like the next release of dune may [flip the default value of
`implicit_transitive_deps`](https://dune.readthedocs.io/en/latest/advanced-topics.html#implicit-transitive-deps) such that it becomes `false`. The flag
controls whether transitive dependencies are implicitly brought into
the project scope. We can either explicitly set it to `true` or
prepare for the new default.

In this patch I have opted for the latter. The advantage of setting it
to `false` is that we have to be explicit about libraries, which
should make our code base more resilient against upstream
changes. Furthermore, flipping the value revealed that we have been
somewhat sloppy with regard to dependencies. For example the `lens`
package implicitly depended on the `result` package. The `repl` module
relied on implicit dependencies from the `core` package.


dhil:
Interesting, Travis seems to run to completion, but fails to notify the GitHub UI of the success. I'll try the "reboot" equivalent of GitHub PRs, namely, closing and opening...

PR #654, opened 2019-06-21
Fix CI for MVU and address bug in harness

SimonJF:
There was a small error in one of the MVU files which slipped through the CI because of a combination of a misconfiguration in one of the tests files, and a bug which meant that a failure was not reported.

This patch fixes the error in the MVU example, fixes the misconfiguration, and also fixes the issue in the `test-harness` file which allowed it to slip through undetected.

(Arguably, the `config` check shouldn't be there, since it would be nice to override the default config. But that's for a different patch.)

### DIFF
--- a/tests/typecheck_examples.tests
+++ b/tests/typecheck_examples.tests
@@ -726,7 +726,7 @@ filemode : true
 Typecheck example file examples/mvu/pong.links
 examples/mvu/pong.links
 filemode : true
-args : --config=foo
+args : --path=examples/mvu/todomvc
 
 Typecheck example file examples/mvu/todomvc/todoMVC.links
 examples/mvu/todomvc/todoMVC.links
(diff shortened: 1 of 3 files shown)


tests/typecheck_examples.tests
```diff
@@ -717,25 +717,22 @@ args : --path=examples/sessions/mind
...
-filemode : args
-args : --config=tests/typecheck_examples.tests.config
+filemode : true
 
 Typecheck example file examples/mvu/pong.links
 examples/mvu/pong.links
-filemode : args
-args : --config=tests/typecheck_examples.tests.config
+filemode : true
+args : --config=foo
```
dhil: WAT.

Do you have a config file named `foo`? In that case it seemingly hasn't been added to the revision control. I suspect it is a bogus name... but why?

SimonJF: bollocks

SimonJF: This was me testing for whether my fix actually worked. I'm going to pretend that I left it there just to be sure the CI broke.

:)

PR #656, opened 2019-06-23
[WIP] attempt to fix newlines configuration

jamescheney:
I ran into a bizarre problem with `tests/dos-newlines.links` showing changes in my local repository, I observed this on Mac OSX and Linux even after doing new clones.  After discussion with @slindley and looking at #503 it seems that the current configuration is too aggressive and may be leading to inconsistent states.  This changes the configuration so that by default all files are treated as text, with automatic newline replacement (this should mean that whatever is native is used for local files, but translated to CR on the repository).  The exception is tests/dos-newlines.text, which is specified to use CRLF.  (The version in the repository still is stored using CR, but the local file copy is stored as CRLF).

This worked for me, but  needs to pass the CI and be tested by people using different operating systems before it is safe to merge.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,10 +1,2 @@
-# Checkout files in repository using native line endings by default
-* text=auto
-
-# Some file extensions are binary and should not be modified in any way
-*.pdf binary
-*.zip binary
-
-# An important exception is a file for testing Links' support for DOS line
-# endings
+# Ensure the local copy of the DOS line endings test uses CRLF
 tests/dos-newlines.links text eol=crlf
--- a/.gitattributes
+++ b/.gitattributes
@@ -7,4 +7,4 @@
 
 # An important exception is a file for testing Links' support for DOS line
 # endings
-tests/dos-newlines.links text=crlf
+tests/dos-newlines.links text eol=crlf
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,2 @@
-# Ensure the local copy of the DOS line endings test uses CRLF
-tests/dos-newlines.links text eol=crlf
+# Ensure the DOS line endings test is preserved exactly
+tests/dos-newlines.links binary
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,5 @@
-# Checkout files in repository using UNIX line endings
-* text eol=lf
+# Checkout files in repository using native line endings by default
+* text=auto
 
 # Some file extensions are binary and should not be modified in any way
 *.pdf binary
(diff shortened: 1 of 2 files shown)


rudihorn:
I've been trying to understand the man pages and I'm not sure I can make sense of them. Especially when I consider the bug I used to have. Maybe the correct solution is to set both tags in the .gitignore file, as the settings you propose should allow explicit checking in of CRLF files, which would then continue to be checked in as CRLF files. The manpage states:

>This setting forces Git to normalize line endings to LF on checkin and prevents conversion
               to CRLF when the file is checked out.

Which seems to be the behaviour we want, and possibly git isn't correctly identifying it as a text file or something which is the issue. I also don't see an explicit differentiation between the check in and check out behaviour of git, and the sentence "It enables end-of-line conversion without any content checks, effectively setting the text attribute." attached to the eol property makes it seems like the settings are somewhat related. Was this issue related to text files or non text files?

jamescheney:
@rudihorn See discussion in the links slack channel.  I was getting phantom diff's for the tests/dos-newlines.links test (which has to be preserved with CRLFs to work as intended).

I believe what is happening is:

- if the file is tagged `binary` then the file is preserved exactly - this may be the best option for the dos-newlines.links test.
- if it is treated as text then CRLFs are always standardized to LFs in the repository
- if text=auto then on checkout the host OS's preferred newline treatment is used
- if text=lf or text=crlf then on checkout the host OS's preference is overridden so the local file has LFs or CRLFs respectively no matter what.

Nevertheless, what would be most helpful would be you (or other Windows users) to do a fresh clone to see whether the right thing happens - doing "git diff" from within the clone should not show any differences.  The CI has already validated that this works on the Mac and Linux configurations it uses.

jamescheney:
I already changed the first line to correctly explain that `text=auto` means use native line endings.

I think we can do without this line.

I think the problem was with the line 
```
tests/dos-newlines.links text=auto eol=crlf
```
which contains contradictory settings: both use native newlines and ensure the local copy has `CRLF`s.

We still need something here but I think what we had originally
```
tests/dos-newlines.links text=auto eol=crlf
```
was correct.  It may have interfered somehow with the `* text=lf` setting that we don't need / want (certainly if no one is using Windows, which I thought was the problem).

If we don't have the `* text=???` line then I think the binary attributes become unnecessary too since Git will usually figure this out using `file` if we don't override it.

jamescheney:
OK, I'm not sure that worked - when I did a local fresh clone, the dos-newlines file was showing changes again.  I'm going to see if using `binary` fixes this - it's a blunt instrument but I don't want to waste any more time on this.

jamescheney:
OK, that seems better to me.  Let's see what the CI says...

jamescheney:
@slindley I believe I have made changes compatible with what you requested.  Have you (or anyone) encountered a problem with the current proposed change?  I'd like to get this merged because it is stopping me doing anything else.

PR #660, opened 2019-07-09
Refactor subkind checking code, and add the Mono restriction

SquidDev:
This PR introduces the `Mono` subkind, which is used to constrain the type of function arguments when visiting their bodies. Consider the following examples:

```links
sig poly : forall e::Row. (forall a, e::Row. (a) -e-> a) -e-> (Int, Bool)

fun(f) { (poly(~f), f(42) + 1) } # Type checks, as poly is visited first
fun(f) { (f(42) + 1, poly(~f)) } # Fails, as f is visited first.
```

With this patch, `f` is given the type `%::(Any, Mono)`, which forces the type to be monomorphic. Thus, unifying with `(forall a, e::Row. (a) -e-> a)` causes a type error, meaning both examples fail, and so removing the order-dependence on our inference of polymorphic types. `Mono` subkinds are replaced with `Any` when quantifying, so this should have no observable effect outside of the function's body.

Supplying a type annotation to `f` will cause `f` to not have the Mono subkind (unless the type has flexible type variables within it).

---

As part of this change, I've tried to refactor the subkind constraint code to be a little more modular, as there was a lot of near-identical code blocks both in `Types` and `Unify`. I've explained the design in full detail within 5fc3f2b, but it effectively boils down to:

 - Every subkind which imposes a constraint (`Any` linearity, `Base`, `Mono`, `Session`, restriction) defines a module of type `Constraint`, which has predicates for checking if a type satisfies that constraint, and functions to try to force that type to satisfy it.
 - `Types` define several helper classes which can be subclasses to allow constraints to write predicates/iterators over types in a "simple" manner.
 - `Unify` will find the appropriate constraint(s) depending on what the linearity/restriction, and try to force the type to satisfy that constraint, failing if it cannot.

There's definitely some elements of this design I'm not especially happy with (lots of weird usages of first class modules, some weird inheritance stuff), but I wasn't quite sure of a nicer way without duplicating a lot more code. If people have suggestions it'd be very appreciated as there's _definite_ room for improvement.


core/types.ml
```diff
@@ -649,282 +539,305 @@ let primary_kind_of_type_arg : type_arg -> PrimaryKind.t =
...
-      end
+  val make_type : typ -> unit
+  val make_row : row -> unit
+end
 
-let rec type_can_be_unl : var_set * var_set -> typ -> bool =
-  fun ((rec_vars, quant_vars) as vars) ->
-    let tcu t = type_can_be_unl vars t in
-    function
+type predicate_context = StringSet.t * var_set * var_set
```
SimonJF: Could you please add a comment clarifying what a type predicate is, and why we need a predicate context?

core/generalise.ml
```diff
@@ -223,6 +223,19 @@ let rigidify_quantifier : quantifier -> unit =
     | (_, _, `Row point)      -> rigidify_point point
     | (_, _, `Presence point) -> rigidify_point point
 
+let mono_type_args : type_arg -> unit =
```
SimonJF: Please add a comment describing what happens here, since `mono_type_args` and the type signature aren't descriptive enough to see what's going on.

If I'm understanding correctly, the function is called to change all `Mono` type variable subkinds to `Any` after typechecking a function body is complete?

core/types.ml
```diff
@@ -649,282 +539,305 @@ let primary_kind_of_type_arg : type_arg -> PrimaryKind.t =
...
-  fun f (rec_vars, quant_vars) point ->
-    begin
-      match Unionfind.find point with
-      | `Closed -> true
-      | `Var (var, (lin, _), _) -> IntSet.mem var quant_vars || Linearity.is_nonlinear lin
-      | `Body t -> f (rec_vars, quant_vars) t
-      | `Recursive (var, t) ->
-        check_rec var rec_vars true (fun rec_vars' -> f (rec_vars', quant_vars) t)
-    end
+module type Constraint = sig
```
SimonJF: In the context of reviewing this PR, it's obvious, but would you mind adding a comment describing what a `Constraint` abstracts over?

core/types.ml
```diff
@@ -649,282 +539,305 @@ let primary_kind_of_type_arg : type_arg -> PrimaryKind.t =
...
-      fields
-      true in
-  unl_row_var && unl_fields
-
-let is_unl_type = is_unl_type (IntSet.empty, IntSet.empty)
-let is_unl_row = is_unl_row (IntSet.empty, IntSet.empty)
+    | `Alias (_, t) -> self#is_type vars t
+    | `MetaTypeVar point -> self#is_point self#is_type vars point
+    | `ForAll (qs, t) -> self#is_type (rec_appl, rec_vars, add_quantified_vars !qs quant_vars) t
+    | `Application _ | `RecursiveApplication _ -> true (* For now, we assume these are acceptable! *)
```
SimonJF: This seems a bit sketchy -- does it give the same behaviour as before? Why do you not need to inspect the type arguments?

core/types.ml
```diff
@@ -649,282 +539,305 @@ let primary_kind_of_type_arg : type_arg -> PrimaryKind.t =
...
+
+  method is_point : 'a 'c . (predicate_context -> 'a -> bool) -> predicate_context -> ([< 'a meta_max_basis] as 'c) point -> bool
+    = fun f ((rec_appl, rec_vars, quant_vars) as vars) point ->
+    match Unionfind.find point with
+    | `Closed -> true
+    | `Var ((id, _, _) as var) -> IntSet.mem id quant_vars || self#is_var var
+    | `Body t -> f vars t
+    | `Recursive (var, t) -> check_rec var rec_vars true (fun rec_vars' -> f (rec_appl, rec_vars', quant_vars) t)
+
+  method is_type ((rec_appl, rec_vars, quant_vars) as vars) : typ -> bool = function
```
SimonJF: Checking a `typ` to see whether it is a type seems counterintuitive. Could you clarify when this would return false (and add as a comment?)

core/types.ml
```diff
@@ -649,282 +539,305 @@ let primary_kind_of_type_arg : type_arg -> PrimaryKind.t =
...
-  FieldEnv.iter (fun _name -> make_field_unl vars) fields
+    | `Primitive _ -> ()
+    | `Function (a, e, r) | `Lolli (a, e, r) -> self#visit_type vars a; self#visit_row vars e; self#visit_type vars r
+    | `Record r | `Effect r | `Variant r -> self#visit_row vars r
+    | `Table _ -> ()
+    | `Lens _ -> ()
+    | `Alias (_, t) -> self#visit_type vars t
+    | `MetaTypeVar point -> self#visit_point self#visit_type vars point
+    | `ForAll (qs, t) -> self#visit_type (rec_appl, rec_vars, add_quantified_vars !qs quant_vars) t
+    | `Application _ | `RecursiveApplication _ -> () (* For now, we assume these are acceptable! *)
```
SimonJF: Again -- I'm surprised that you have a visitor where the type arguments aren't visited -- why is this?

core/types.ml
SquidDev: Would it be better to call these methods `x_satisfies` instead of `is_x` (so `type_satisfies`, etc...)? Might be a little clearer what we're referring to.

core/types.ml
SquidDev: So it _technically_ gives the same behaviour as before, but that's largely because we're either rejecting (recursive) type applications, or handling them specially within the specific constraint.

That said, this behaviour is definitely not correct for `Mono`, so it's probably best to handle it :).

core/types.ml
SimonJF: Oh definitely -- that makes it a lot clearer.

SquidDev:
Thanks for the looking over this Simon!

PR #667, opened 2019-07-16
Record field punning

SimonJF:
A feature I've been missing from OCaml is record field punning. In
OCaml, we can write:

```
let hello = "hello!" in
{ hello; world = 5 }
```

instead of

```
let hello = "hello!" in
{ hello = hello; world = 5 }
```

This patch is a simple parser hack to allow the same sort of thing in
Links. We can't do field punning directly since we get a reduce/reduce
with a parenthesised expression, but we can delineate punned fields with
an operator (I chose a colon), allowing us to write:

```
var hello = "hello!";
(:hello, world = 5)
```

which is parsed in as

```
var hello = "hello!";
(hello = hello, world = 5)
```

### DIFF
--- a/tests/records.tests
+++ b/tests/records.tests
@@ -103,3 +103,7 @@ stdout : (hello = "hellooo") : (hello:String)
 Record field punning (2)
 var hello = "hellooo"; (:hello, world = 5)
 stdout : (hello = "hellooo", world = 5) : (hello:String,world:Int)
+
+Record field punning (3)
+switch((hello=5, world=10)) { case (:hello, world=y) -> hello + y }
+stdout : 15 : Int
--- a/tests/records.tests
+++ b/tests/records.tests
@@ -107,3 +107,19 @@ stdout : (hello = "hellooo", world = 5) : (hello:String,world:Int)
 Record field punning (3)
 switch((hello=5, world=10)) { case (=hello, world=y) -> hello + y }
 stdout : 15 : Int
+
+Record field punning (4)
+var y = 2; (x=1, =y)
+stdout : (x = 1, y = 2) : (x:Int,y:Int)
+
+Record field punning (5)
+var y = 2; (x=1, =y, z=3)
+stdout : (x = 1, y = 2, z = 3) : (x:Int,y:Int,z:Int)
+
+Record field punning (6)
+var z = 3; (x=1, y=2, =z)
+stdout : (x = 1, y = 2, z = 3) : (x:Int,y:Int,z:Int)
+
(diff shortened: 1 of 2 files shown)


SimonJF:
Good point. I don't tend to pattern match on records too often, so hadn't thought of this. I'll have a play.

SimonJF:
Yep, deconstruction can be trivially handled with precisely the same parser trick. PR updated.

dhil:
I don't want to invoke Wadler's Law*, but using a colon is a bit strange, they appear only at the type-level for records otherwise. Why not just use `=`, such that `(=hello, world = y)`? Alternatively you could choose to be heroic and refactor the necessary parts of the parser to allow just `(hello, world = y)`.

\* That is, I'm not doing to spend time debating this, if you disagree I'll back off.

SimonJF:
I knew the Wadler's Law discussion was coming ;)

The colon was a placeholder. I liked it because it was non-invasive, but your objection regarding the possible conflation of term- and type-level syntax is reasonable. Maybe `=` is better. I'm open to suggestions of other characters, too.

> Alternatively you could choose to be heroic and refactor the necessary parts of the parser to allow just (hello, world = y).

I don't think it's so much of a parser refactoring problem as an actual grammar error. Suppose we removed the marker: then, `(hello, world)` could mean either `(hello = hello, world = world)` or the tuple `(hello, world)`. This comes about in our setting and not OCaml's because we use parentheses for both tuples and records, whereas OCaml uses parentheses for tuples (with a `*` separator rather than a `,`) and braces for records.

SimonJF:
OK - @dhil, I think you're right regarding the marker -- good suggestion. I've changed it to `=`.

dhil:
> [...] good suggestion. I've changed it to `=`.

Phew, an infinitude of time spent discussing minor syntax has been averted!

tests/records.tests
```diff
@@ -95,3 +95,15 @@ exit : 1
 Possibly absent label in a closed row
 fun (x : (a{%b}|%c)) {x : ()}
 stdout : fun : (()) -> ()
+
```
dhil: Possibly add the following tests too
```
var y = 2; (x=1, =y)
var y = 2; (x=1, =y, z=3)
var z = 3; (x=1, y=2, =z)
var x = 1; var y = 2; var z = 3; (=x, =y, =z)
```

SimonJF:
Thanks for the test suggestions -- I've added them in the latest commit. Since we now have two approvals and all comments addressed, let's merge after the CI passes.

Issue #314, opened 2017-12-18
Side effecting modules


dhil:
In our code base we have several modules which play effects at the top level. Some modules, including `evalir.ml`, `value.ml`, `irtojs.ml`, and `links.ml`, rely on these side-effects to be played in a specific order. Common for side-effecting modules are that the depend on the handling of settings and command line arguments. In PR #227 the settings and their handling were centralised to ensure that their side-effects were played before initialisation of the module `Value` as it depends on the presence or absence of flags like `--enable-handlers` and `--sessions-exceptions`.

Previously the `Makefile` enforced a linking order of modules which happened to be doing the right thing, perhaps by sheer luck. However since PR #254 we no longer control the linking order as a result it introduced artificial dependencies amongst `Value`, `Irtojs`, `Evalir`, and `ParseSettings` to ensure that the side-effecting modules are loaded in the correct order.

This is obviously bad, very bad. The handling of settings needs to be decentralised again, and made pure.

jamescheney:
Doesn't seem critical to resolve before 0.7.3 so happy to kick to next version.

PR #668, opened 2019-07-17
Some minor cleanup for OCaml >=4.08

### LINKED ISSUE #661: links build broken on OCaml 4.08.0
The links `dune` file enables deprecation-warnings-as-errors, and OCaml 4.08 deprecates things, so the build is broken. See:

https://ci.ocaml.org/log/saved/docker-run-0cf7e556a8be306698e84594758c3278/3a5dfdecaf926c973f1eba32e2b0776f7d533c17
https://github.com/ocaml/opam-repository/pull/14465/commits/19031f5aca46cd101873b2e0a38df02c469dc951

SquidDev:
- Rename the `development` Dune profile to `dev`. Dune uses the `dev` environment by default (except when installing the package), so this saves us having to manually specify it on the command line.

   There may be a historic reason for having a separate profile, so happy to revert this if preferred.

 - Do not error on compile warnings when using `--profile=release`. This prevents us from failing on deprecation warnings (fixes #661). Note that this is the default Dune behaviour too, so it's probably worth following.

 - Remove most of our remaining uses of `Pervasives` - I believe these were a remnant of when `Utility.(|>)` had a different signature to `Pervasives.(|>)`. There's still a couple of remaining usages, largely for clarity's sake (such as `let compare = Pervasives.compare`), but happy to remove these too if preferred.

This allows us to compile and run on OCaml 4.08.0 _using the release profile_ - there's still some deprecation warnings due to our use of stringly-typed formatting tags and remaining usages of `Pervasives`.


dune
```diff
@@ -1,19 +1,18 @@
...
+          -w A-4-44-45-60  ;; Ignores warnings 4, 44, 45, and 60.
           -g               ;; Adds debugging information to the resulting executable / library.
          )))
  (release
   (flags (:standard
           ;; The following flags are the same as for the "development" profile.
           -strict-formats -strict-sequence -safe-string -annot
-          ;; The following flags are unique to the "release" profile.
-          -w @a-4-44-45-60 ;; Treats all warnings as errors, except that 4, 44, 45, and 60 are ignored altogether.
+          -w A-4-44-45-60 ;; Ignore warnings 4, 44, 45, and 60.
```
dhil: I think we should still error on warnings when using the `release` profile. However, we can allow a bit more slack. I think deprecation warnings have the number `3`, thus if you add that to the list those warnings should be ignored.

dune
```diff
@@ -1,19 +1,18 @@
...
           -strict-sequence ;; Enforces the lhs of a sequence to have type `unit'.
           -safe-string     ;; Immutable strings.
           -annot           ;; Dumps information per (file) module about types, bindings, tail-calls, etc. Used by some tools.
-          -w Ae-44-45-60   ;; Ignores warnings 44, 45, and 60.
+          -w A-4-44-45-60  ;; Ignores warnings 4, 44, 45, and 60.
           -g               ;; Adds debugging information to the resulting executable / library.
          )))
  (release
   (flags (:standard
           ;; The following flags are the same as for the "development" profile.
```
dhil: s/development/dev/

core/processTypes.ml
```diff
@@ -75,7 +75,6 @@ struct
   let make_name_with_id id = NameInfo.prefix ^ (string_of_int id)
 
   let create () =
-    let open Pervasives in
     fresh_name ()
     |> make_name_with_id
```
dhil: This is an instance of what I'd call pointless point-free style.

dhil:
> [...] there's still some deprecation warnings due to our use of stringly-typed formatting tags and remaining usages of Pervasives.

I'd like to see some of those, do you mind pointing some of them out to me?

SquidDev:
> I'd like to see some of those, do you mind pointing some of them out to me?

I think these are the remaining usages (aside from that other one I missed in `processTypes`).

https://github.com/links-lang/links/blob/67e6a25930ffa7ef9adf13e95a0d499b0375ae07/core/types.ml#L123

https://github.com/links-lang/links/blob/67e6a25930ffa7ef9adf13e95a0d499b0375ae07/core/processTypes.ml#L82

https://github.com/links-lang/links/blob/67e6a25930ffa7ef9adf13e95a0d499b0375ae07/core/getopt.ml#L103

dhil:
> > I'd like to see some of those, do you mind pointing some of them out to me?
> 
> I think these are the remaining usages:
> 
> https://github.com/links-lang/links/blob/67e6a25930ffa7ef9adf13e95a0d499b0375ae07/core/types.ml#L123
> 
> https://github.com/links-lang/links/blob/67e6a25930ffa7ef9adf13e95a0d499b0375ae07/core/processTypes.ml#L82
> 
> https://github.com/links-lang/links/blob/67e6a25930ffa7ef9adf13e95a0d499b0375ae07/core/getopt.ml#L103
> 
> I'm kinda unwilling to remove them as, while they're redundant, they do add a bit of clarity (`let compare = compare` is a little odd otherwise).

Thanks. I think it is fine to write `let compare = compare`. This kind of boilerplate is unfortunately quite typical whenever you create a new instance of `Map` or `Set`.

Similarly for the `getopt` just drop the prefix (for #314 I want to replace `getopt` anyways).

dune
SquidDev: Part of the issue here is that some other warnings have become more aggressive in OCaml 4.08.0. The following now warns, as we're popping from a stack of functions, and so it considers the function partially applied (due to warning no. 5).

https://github.com/links-lang/links/blob/145dc45c389e8790779c31676a21fe48d91823ad/core/lexer.mll#L149

It should be OK to fix it, but I feel it's safer to leave it off to avoid similar breaking changes in the future.

dhil: Rewriting it as `ignore (Stack.pop lexers)` may eliminate the warning.

SquidDev: Sadly [it turns out it isn't](https://github.com/ocaml/ocaml/blob/658233568930c9b1c291f78d8ac2177e67aa9981/typing/typecore.ml#L4149)! Resorting to `[@warning "-5"]` works though.

dhil:
> Rename the development Dune profile to dev. Dune uses the dev environment by default (except when installing the package), so this saves us having to manually specify it on the command line.
>
> There may be a historic reason for having a separate profile, so happy to revert this if preferred.

Just for the history books. It used to be the case that Links could not compile with `dune`s standard `dev` profile (and I think it used to be the case that you couldn't overload its definition), because it was too strict (implying our codebase was too sloppy).

SquidDev:
> Just for the history books. [...]

Ahh, that's interesting to know. Wonder what they changed... On a semi-related note, [the Menhir wrapped workaround](https://github.com/links-lang/links/blob/master/core/parser.mly#L23-L34) appears to be redundant now - I don't know if this PR is a good a time/place to remove it?

dhil:
> Ahh, that's interesting to know. Wonder what they changed... 

I think it implies `-strict-format` and `-strict-sequence` which we used to violate.

> On a semi-related note, the Menhir wrapped workaround appears to be redundant now - I don't know if this PR is a good a time/place to remove it?

Sure, I'm all in for eliminating redundancy. Let's give it a try.

PR #669, opened 2019-07-17
Expanding the existing effect sugar

SquidDev:
An extension of #625, this extends the previously-added `effect_sugar` flag:

 - The effect sugar is now applied universally, rather than just in types with no explicit type variable. Similarly, explicitly unnamed type variables (`_`) will now also be mapped to the "shared effect variable".

 - Type names can now be written with an implicit effect variable.
   ```
   typename Comp(a) = () ~> a # Instead of Comp(a, e::Eff) = () ~e~> a

   sig forever : Comp(()) ~> () # Instead of Comp((), e) ~e~> ()
   fun forever(f) { f(); forever(f) }
   ```

   Note that the effect variable has been moved to the right hand side of the definition, to be consistent with Frank's implementation.

 - Presence variables can now be left off effects:
   ```
   sig evalState : (s) ->
                   (Comp(a, {Get:s,Put:(s) {}-> () |_})) ->
                    Comp(a, {                      |_}) 
       # Instead of Comp(a, {Get{_},Put{_}         |_})
   ```
   Note this also works with the previous sugar, so you can write `Comp(a)` on its own.

 - Make the type of `fork` and the like stricter.

   Fork previously had the type `sig fork : {SessionFail:[||] | e}~> ()) ~e~> ~s`, which does not work with our effect sugar (as we tried to rewrite `~e~>` to `{ SessionFail{p} | e}~>`. We now give it the (more morally correct) type `((s) {SessionFail:[||]}~> ()) ~> ~s` (note the closed row).

   Ideally we'd also update the type of the `spawn*` methods, but that's a more invasive change, as it requires touching the typing of `spawn` blocks.


SimonJF:
Thanks for the patch. It seems like the handler examples test no longer works with this -- could you take a look?

slindley:
Ideally the pretty-printer should match the behaviour of the parser (so `Comp(a) -> ()` should print as `Comp(a) -> ()`).

dhil:
> Ideally the pretty-printer should match the behaviour of the parser [...]

I agree. But I think it is an orthogonal issue.

slindley:
@dhil, I was responding to @SquidDev's question about pretty-printing. I'm sure the issues with the tests are easy to fix, but that's not what I was referring to.

dhil:
@slindley, I understand that. I simply meant to point out that the pretty-printing can be fixed in a subsequent patch. I have no idea about the issues with tests that you are talking about. But maybe that comment wasn't meant for me.

slindley:
@dhil Oh... I see now! What you meant was that this can be merged before resolving the caveats. That seems reasonable to me.

The first caveat should also be fixed at some point. I can't remember precisely how Frank does it, but a conservative approach is to determine in advance whether a collection of mutually recursive types requires an extra effect type variable by checking whether any functions or type applications with implicit effect variables appear in the body of any of the mutually recursive types.

In Frank, once a data type with an implicit effect parameter has been defined with n explicit type parameters it may be applied with either n or n+1 explicit arguments. Confusingly, "do be do be do" says the implicit parameter goes first, but actually in the implementation it goes last (I'll make sure to fix this in "doo bee doo bee doo"). Either works, but I think there's a reasonable argument to be made that what the Frank implementation does is slightly less bug-prone than what the paper describes. We should consider whether to follow suit with Links.

examples/handlers/lambda.links
```diff
@@ -60,7 +60,7 @@ fun lit(x)() { x }
 
 # Due to the (current) lack of multi-handlers in Links, we are going
 # simulate them by reifying operations.
-typename Expr(a) = Comp(a, {Var:(Var) -> Expr(a), Lam:(Var) -> ()});
+typename Expr(a) = Comp(a, {Var:(Var) {}-> Expr(a), Lam:(Var) {}-> ()});
```
dhil: Why are the operation arrows explicitly closed? The existing sugar should rewrite `(Var) -> Expr(a)` as `(Var) {}-> Expr(a)`. 

Ideally, one should not be able to write `{}->` at all, since the arrow in question is not a function arrow, but that is an entirely different concern.

SquidDev: The trouble is that our check for "does this typename require an implicit effect variable" occurs before the desugaring has run, and so it thinks this is a normal arrow.

It's a bug (hence me adding the test case), I just wanted to get something working - it's on my list to fix tomorrow, I just need to think about how best to do it.

PR #671, opened 2019-07-18
Implement unlines and unwords.

ezrakilty:
I wanted `unwords` in particular because I wanted to calculate a set of CSS classes and then blat them into a `@class` attribute, which is exactly what `unwords` does. `unlines` wasn't something I needed but it was easy enough to do.

`words` and `lines` are perhaps more ambiguous: What do you do with double-blanks and leading/trailing blanks?

I wonder how long these comments have been sitting in prelude.links unimplemented... I'm going to say since about 2005 :-D


SimonJF:
Thanks for the patch, Ezra! I think the CI is failing because of the syntax of signatures: in the patch, they're down as
```
sig unlines :: [String] -> String
```

whereas the syntax now is

```
```

dhil:
I like how Github says "@ezrakilty first-time contributor". It has no clue, really.

ezrakilty:
@dhil Haha yes, indeed! I'm rusty enough I feel like a first-timer anyway.

So originally I "tested" this by just running some links programs, and I didn't get any errors, where I was supposing that the prelude.links would get loaded each time, and so errors would get reported. The prelude must be somehow cached—how can I force it to reload that?

dhil:
> The prelude must be somehow cached—how can I force it to reload that?

The prelude should not be cached -- we removed that "feature" some time ago because it was buggy (we have been discussing introducing a proper caching mechanism though). The only scenario I can think of that would "cache" the Prelude would be in the REPL. If you are in a REPL session and make changes to `prelude.links` then those changes wont be reflected until you restart the session or issue `@load "prelude.links"`.

SquidDev:
I believe `./links` loads from the Dune `_build` directory rather than the root one, which may be the problem here - I've definitely encountered this before. `dune build lib/prelude.links` should update it.

Not sure if this is what is going on here though...

dhil:
> I believe ./links loads the prelude from the Dune _build directory rather than using the one in the the project root, which may be the problem here - I've definitely hit this before. dune build lib/prelude.links should update it.
>
> Not sure if this is what is going on here though...

Ah yes, that is the default behaviour. You can change this by overriding the value of `prelude` in a configuration file, for example I have
```
prelude=$LINKSHOME/prelude.links
```
where `$LINKSHOME` is set to the root directory of the Links repository.

slindley:
Yep. I think this behaviour is a misfeature. I find myself typing `make` each time I edit `prelude.links`.

ezrakilty:
The type signatures I inherited were wrong in two ways. I'm remembering, I think, that we just took the Haskell prelude type signatures and pasted them in as a big comment, as a to-do list of things to implement. And never changed the signatures to match Links syntax :^)

ezrakilty:
I don't understand why these changes produced the test failures there. Does anyone?

ezrakilty:
BTW, to your point @dhil, github does know something about my contributions. Some of these lines have my long-time avatar (it's a man in a long coat walking in front of a nice brick wall—you can see the picture on my Twitter account, http://twitter.com/ezrakilty):

Screen Shot 2019-07-19 at 9 57 11 PM

PR #672, opened 2019-07-18
Fix warnings being fatal in development mode

SquidDev:
This is a small amendum to #668 - sorry, I should have picked this up earlier.

The dev profile by will error on most warnings by default. While additional strictness may be useful, it can be some little frustrating as non-fatal issues cause the build to fail (and make editor error highlighting a little more confusing).

We use `-warn-error -a+31` (the default setting for ocamlc) to prevent these warnings from erroring.

This change also allows us to build Links in development mode on 4.08 (albeit with it warning on `bin/repl.ml`).

### DIFF
--- a/dune
+++ b/dune
@@ -5,6 +5,7 @@
           -strict-sequence ;; Enforces the lhs of a sequence to have type `unit'.
           -safe-string     ;; Immutable strings.
           -annot           ;; Dumps information per (file) module about types, bindings, tail-calls, etc. Used by some tools.
+          -warn-error -a+31 ;; Only error on warning 31.
           -w A-4-44-45-60  ;; Ignores warnings 4, 44, 45, and 60.
           -g               ;; Adds debugging information to the resulting executable / library.
          )))


SimonJF:
What is the warning on `bin/repl.ml`? If it's easily fixable, could you include it in this PR?

SquidDev:
@SimonJF `Format.pp_set_formatter_tag_functions` is now deprecated. The solution is to use the `stag` equivalent instead, however that is only available on OCaml 4.08.0+ so we can't swap it out quite yet.

It should be fairly simple to suppress the warning in this particular case, but I feel it might be better to keep it around so we remember to fix it when dropping 4.06.1 support. Happy to do so if preferred though!

SimonJF:
OK, thanks for the clarification. In that case, let's merge.

dhil:
You can discuss this during the next Links meeting. Now that we are compatible with OCaml 4.08, I think we should drop 4.06.1 support and move to 4.07.

PR #673, opened 2019-07-22
MVU: Unescaped text nodes

SimonJF:
Minor patch: currently the MVU library only supports escaped text nodes (for example, showing raw HTML and "&"-coded symbols). Sometimes this is too restrictive (like in my GtoPdb clone).

This patch allows you to do

```
unescapedTextNode("&amp;")
```

for example, which will correctly come out as an ampersand. The trick is to have a `span` tag and set its `innerHTML` attribute.

The patch also contains a fix to the `for` attribute (the library works using properties rather than attributes generally, so it must be `htmlFor` rather than just `for`).

### DIFF
--- a/lib/js/vdom.js
+++ b/lib/js/vdom.js
@@ -538,14 +538,9 @@ const VDom = (function() {
     } else if (lbl =="HTMLAppend") {
       return evalHTML(val["1"]).concat(evalHTML(val["2"]));
     } else if (lbl == "HTMLText") {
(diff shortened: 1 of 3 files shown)


SquidDev:
There's a bit of me which feels it might be nicer to have a separate constructor case (`HTMLRaw` or something), rather than a boolean flag which changes behaviour.

SimonJF:
Ooh, that is nicer. Thanks.

Issue #518, opened 2019-03-01
Rationalize/reuse parsers?


jamescheney:
#517 is a good idea but seems like an occasion to rethink/rationalize the various lexers/parsers we currently support (Links, json, XML).

The XML lexer/parser was added a while ago and is not used internally in Links but is exposed as a library function `parseXml`.  I don't know if we have any active use cases where we need this.

@SimonJF has a use case for an HTML parser (i.e. to handle auto-closed tags and non-quoted attributes), which is currently implemented in plain Links using parser combinators, and is probably a lot slower than an ocamllex/yacc parser would be.  We could in principle do the same thing to parse plain XML.

The JSON lexer/parser exists to allow parsing JSON messages but is not accessible from Links.  Might it be useful to support a `parseJson` library function - especially if we also were to support sending RPC-calls to other REST-style services (or serving such calls from Links)?

Alternatively, if we want to support efficient parsing of all three either internally or exposed as Links library functions, I can also imagine that there might be standard JSON, XML or HTML parsing libraries available for OCaml that we could simply reuse here. From a quick google of OPAM, `yojson` appears to support JSON parsing and `markup` appears to support HTML and XML parsing, and we already depend on both libraries.

PR #674, opened 2019-07-22
Allow hyphenated HTML attributes

SimonJF:
Modern HTML allows hyphens in attributes. Take this Bootstrap code:

```
      <button class="navbar-toggler" type="button" data-toggle="collapse"
        data-target="#navbars" aria-controls="navbars"
        aria-expanded="false" aria-label="Toggle navigation">
        <span class="navbar-toggler-icon"></span>
      </button>
```

This small patch allows hyphenated attributes in Links.

### DIFF
--- a/core/lexer.mll
+++ b/core/lexer.mll
@@ -242,10 +242,12 @@ exception LexicalError of (string * Lexing.position)
 }
 
 let def_id = (['a'-'z' 'A'-'Z'] ['a'-'z' 'A'-'Z' '_' '0'-'9' '\'']*)
+let def_attr_id = (['a'-'z' 'A'-'Z'] ['a'-'z' 'A'-'Z' '_' '-' '0'-'9']*)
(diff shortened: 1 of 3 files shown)


jamescheney:
What about `xmlParser`/`xmlLexer`?  Do they have the same problem?

SimonJF:
Yep, `xmlLexer` does (having never had to touch `xmlLexer`, I missed this one). Should be fixed in latest commit.

jamescheney:
Cool. As per #518 I kind of think we should turn things like `xmlLexer`/`xmlParser` into Links code eventually (if there are any volunteers to port ocamllex/ocamlyacc to Links... :cricket: :cricket:)

SimonJF:
I was _very_ confused by that comment until I saw that `:cricket:` rendered as the insect in Slack :)

jamescheney:
Yeah, my browser doesn't render it correctly in Github either.

Issue #578, opened 2019-04-30
Broken example: Draggable Cropping Frame


dhil:
The "Draggable Cropping Frame" example is currently broken because the image [http://groups.inf.ed.ac.uk/links/data/17-Heriot-Row.jpeg](http://groups.inf.ed.ac.uk/links/data/17-Heriot-Row.jpeg) is unavailable. The server produces a 503 error.
```
$ curl -v http://groups.inf.ed.ac.uk/links/data/17-Heriot-Row.jpeg
*   Trying 129.215.202.26...
* TCP_NODELAY set
* Connected to groups.inf.ed.ac.uk (129.215.202.26) port 80 (#0)
> GET /links/data/17-Heriot-Row.jpeg HTTP/1.1
> Host: groups.inf.ed.ac.uk
> User-Agent: curl/7.58.0
> Accept: */*
> 
< HTTP/1.1 503 Service Unavailable
< Date: Tue, 30 Apr 2019 09:11:02 GMT
< Server: Apache/2.4
< X-Frame-Options: SAMEORIGIN
< Content-Length: 299
< Connection: close
< Content-Type: text/html; charset=iso-8859-1
< 
<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>503 Service Unavailable</title>
</head><body>
<h1>Service Unavailable</h1>
<p>The server is temporarily unable to service your
request due to maintenance downtime or capacity
problems. Please try again later.</p>
</body></html>
* Closing connection 0
```
I am not sure who controls the "links group", @slindley, @jamescheney?

jamescheney:
It is not just this one URL but also http://groups.inf.ed.ac.uk/links/ returns the same 403.  Other groups.inf.ed.ac.uk pages appear to be fine so it is a problem with the links group, possibly something computing support should handle?  

I have never had any involvement with the setup of groups.inf.ed.ac.uk/links.  I suspect it has bitrotted since the shift to using www.links-lang.org.  Probably we should just put the image file on the examples server (or even make it a relative URL and not have a fixed global URL hardwired into the code).

dhil:
> Probably we should just put the image file on the examples server (or even make it a relative URL and not have a fixed global URL hardwired into the code).

Yes, it would be convenient to keep the resources together in a place where everybody has access. I do not have access to the Links group on the inf-network.

> I have never had any involvement with the setup of groups.inf.ed.ac.uk/links.

[...]

PR #679, opened 2019-07-23
Fix crop example

SimonJF:
Fix draggable handles example.

It was nothing insidious (although took a bit of debugging): turns out we just needed to have `px` added when setting the styles. My guess is that the behaviour was removed from modern browsers, or this was a result of adding in the `<!doctype html>` tag.


SimonJF:
Fixes #578 .

slindley:
Ah. Good. There appears to still be a bug - but maybe it was always there. The initial display doesn't show the view inside the initial cropping box but rather the entire image. It only does the right thing when we drag one of the corners for the first time.

SimonJF:
Good spot - it was the same issue, just in static HTML.

slindley:
That's better, but it's still not right. The initial view is of the top-left corner, but the cropping frame is not placed there.

SimonJF:
I have no idea why this has brought 52 files' worth of diff in -- but I'll close and open another PR.

Issue #229, opened 2017-07-14
Abstract JavaScript Backend


dhil:
Pull request #227 abstracts the structure of the client side continuation. The CPS compiler is parameterised by the continuation structure such that it can generate code using both the standard Links continuation structure and the enriched structure required to support handlers. This is clearly an advantage as multiple CPS translations can coexist and reuse the same code generator.

I suggest we go even further and abstract the JavaScript compiler. This would allow us to have multiple different compilers, e.g. a direct-style compiler, a cps compiler, etc.

jamescheney:
@dhil is this done / something that is likely to be done as part of your work on effects/handlers soon?

dhil:
Well, it is partially done. In the sense, that I have a separate branch which implements several alternative JavaScript backends. However, it is not yet mature enough to be included in the master branch. I am hoping to find time after the upcoming POPL deadline to improve my implementation.

jamescheney:
No problem, just curious.

PR #690, opened 2019-07-28
Fixes compilation of `show`.

### LINKED ISSUE #689: Function called "show" is broken
The following program crashes when I click the word "Test." It seems that functions called `show` are given incorrect JS compilation.

```
fun show(x) {
  intToString(x)
}

fun core() {
  println(show(7));
  <div id="me" l:onclick="{replaceNode(core(), getNodeById("me"))}" { [("style", "color:#ff8080")] }>
  Test
  </div>
}

fun thepage() {
  page
    <html>
    { core() }
    </html>
}

fun main() {
  addRoute("/", fun(_) {thepage()});
  servePages()
}

main()
```

The error is
```
jslib.js:3465 Uncaught TypeError: k is not a function
    at _yieldCont_Default (jslib.js:3465)
    at show_2104 ((index):2439)
    at (index):2447
    at _yield (jslib.js:3450)
    at core_2105 ((index):2447)
    at Object._2105 ((index):363)
    at (index):2442
    at _yield (jslib.js:3450)
    at _fun__g132_2110 ((index):2442)
    at _2110 ((index):362)
```
And the incorrect JS code, I think, is this:

```
function show_2104(_2103, __kappa) { var x_2102 = _2103; _yieldCont(__kappa, _intToString(x_2102)) }

function core_2105(__kappa) {
_yield(function () { println_239(show_2104(7), function (_2112) {
  _applyCont(function (_2111) {
    _yieldCont(__kappa, _addAttributes(LINKS.XML('div', {'key':_2111, 'id':'me'}, {"_head": _stringToXml('\n  Test\n  '),
      "_tail": null}), _Cons({'2':'color:#ff8080', '1':'style'}, Nil)))
    }, _registerEventHandlers(_Cons({'2':partialApply(_fun__g132_2110, {'2105':core_2105}), '1':'click'}, Nil)))
  }) })
}
```

Note that `show_2104` is called without a continuation, though it expects one.

Changing the function name to say `shower`, it works fine.

dhil:
The builtin function `show` has an impure effect signature (`(a) ~> String`), yet it was marked as `PURE` in `lib.ml`, meaning that the JavaScript compiler would not CPS transform the function.

This patch marks the function as `IMPURE` in `lib.ml`.

Resolves #689.

### DIFF
--- a/core/lib.ml
+++ b/core/lib.ml
@@ -307,7 +307,7 @@ let env : (string * (located_primitive * Types.datatype * pure)) list = [
   "show",
   (p1 (fun v -> Value.box_string (Value.string_of_value v)),
    datatype "(a) ~> String",
-   PURE);
+   IMPURE);
 
   "exit",
   (`Continuation Value.Continuation.empty,


ezrakilty:
I suppose this is a good change, but is it right that my user-defined `show` somehow inherited its effect signature from the builtin one? What happens if I now write an impure `show` that shadows this builtin pure one, perhaps inheriting its effect signature again?

dhil:
> but is it right that my user-defined show somehow inherited its effect signature from the builtin one?

I suspect this behaviour is due to our horrendous handling of builtin names. Currently, the compiler relies on the source name of a variable to determine whether it is a primitive. Since `show` is a builtin function, and it was marked `PURE`, the compiler will incorrectly recognise your implementation of `show` as the builtin and then not CPS transform it.

I'm working on a rather long patch which will improve this situation, it is part of the effort to have a proper module system (issue #603).

> What happens if I now write an impure show that shadows this builtin pure one, perhaps inheriting its effect signature again?

I think it would just happen to do the right thing -- but not for the right reasons. Our CPS compiler is only selective on the hard-wired primitives in `lib.ml`;  the `PURE` and `IMPURE` markers instructs the CPS compiler as to whether it should transform a function. Meaning a user-defined pure function is always transformed, which is somewhat displeasing.

I have another patch in the works which revamps the CPS compiler (issue #229). One of the things I would like to do is to take advantage of our effect system to determine whether any function should be CPS transformed.

slindley:
I'm pretty sure the poor name hygiene dates back to the early days of Links, so @ezrakilty probably gets to share part of the blame along with me and @jyallop :smiley:

ezrakilty:
I have no doubt of that, @slindley! Perhaps the bug should be assigned to me...

Issue #694, opened 2019-07-29
Compiling query causes "unbound variable" IR typing error


jstolarek:
code:
```
fun () {query {[]}};
```
error:
```
***: IR Type Error: Type variable 354 is unbound
occurring in IR special tail computation:
(Ir.Query (None,
   ([],
    (Ir.Return
       (Ir.TApp ((Ir.Variable 56),
          [`Type (`Record (({},
                            Point @ 0x7ffa8d85e430 `Var ((354,
                                                          (CommonTypes.Linearity.Any,
                                                           CommonTypes.Restriction.Base),
                                                          `Rigid)),
                            false)))
            ]
          )))),
   `Application (({ Types.Abstype.id = "List"; name = "List";
                    arity =
                    [(CommonTypes.PrimaryKind.Type,
                      (CommonTypes.Linearity.Unl, CommonTypes.Restriction.Any))
                      ]
                    },
                  [`Type (`Record (({},
                                    Point @ 0x7ffa8d85e430 `Var ((354,
                                                                  (CommonTypes.Linearity.Any,
                                                                   CommonTypes.Restriction.Base),
                                                                  `Rigid)),
                                    false)))
                    ]))
   ))
```
config:
```
typecheck_ir=true
prelude=empty.links
print_types_pretty=false
optimise=false
debug=true
hide_fresh_type_vars=false
show_flavours=true
fail_on_ir_type_error=true
check_frontend_transformations=true
show_quantifiers=true
```
Possibly an instance of #693 ?

jstolarek:
Fixed in d72e68d156e439caa6440bb07ff67782492b4432

Issue #693, opened 2019-07-29
Trivial anonymous functions cause "unbound type variable" IR typing errors


jstolarek:
code:
```
fun (x) {1}
```
error:
```
***: IR Type Error: Type variable 350 is unbound
occurring in IR binding:
(Ir.Let (
   (229,
    (`MetaTypeVar (Point @ 0x7f8165a5ba58 `Var ((350,
                                                 (CommonTypes.Linearity.Unl,
                                                  CommonTypes.Restriction.Any),
                                                 `Rigid))),
     "x", Var.Scope.Local)),
   ([], (Ir.Return (Ir.Variable 230)))))
```
config:
```
typecheck_ir=true
prelude=empty.links
print_types_pretty=false
optimise=false
debug=true
hide_fresh_type_vars=false
show_flavours=true
check_frontend_transformations=true
show_quantifiers=true
```
Other test cases that report same errors:
```
fun (x) {x}
fun (f) {fun (x){f (x)}}
fun (x) {<a>{x}</a>}
fun () {query {[]}}
```

jstolarek:
Also, anonymous composition function causes the same error:

```
fun (f,g) {fun (x){f (g(x))}};
```

jstolarek:
Also, disabling `fail_on_ir_type_error` reveals multiple unbound variable errors, e.g. for the anonymous composition function above we get:

```
Continuing after IR Type Error:
Type variable 353 is unbound
occurring in IR binding:
(Ir.Let ((231, (a::Any, "x", Var.Scope.Local)),
   ([], (Ir.Return (Ir.Variable 233)))))
backtrace:

-------------------------------------------------------------------


--------------------------------------------------------------------
Continuing after IR Type Error:
Type variable 354 is unbound
occurring in IR tail computation:
(Ir.Apply ((Ir.Project ("229", (Ir.Variable 239))), [(Ir.Variable 231)]))
backtrace:

-------------------------------------------------------------------


--------------------------------------------------------------------
Continuing after IR Type Error:
Type variable 354 is unbound
occurring in IR tail computation:
(Ir.Apply ((Ir.Project ("230", (Ir.Variable 239))), [(Ir.Variable 232)]))
backtrace:

-------------------------------------------------------------------

                                                                                                                                          [17/1848]
--------------------------------------------------------------------
Continuing after IR Type Error:
Type variable 353 is unbound
occurring in IR binding:
[...]

PR #695, opened 2019-07-30
Keep quantifiers on IR let-bound values

### LINKED ISSUE #620: IR Typing and effect instantiation
Say I have the following program:

```
sig foo : () -> ()
fun foo() {
  ()
}

sig bar : () ~> ()
var bar = foo;
```

That is, I have a pure function, and I want to be able to treat it as wild. This types fine, but we get an IR type error:

```
Continuing after IR Type Error:
Type mismatch:
 Expected:
  () ~> ()
 Actual:
  () ~> ()
occurring in IR binding:
(Ir.Let ((2123, (() ~> (), "bar", Var.Scope.Global)),
   ([], (Ir.Return (Ir.TApp ((Ir.Variable 2122), [{ |wild|_ }]))))))
backtrace:
```
I have also tried this on the `ordered-quantification` branch, which seems to have various fixes.

(Forgive me if I'm missing something obvious here.)

jstolarek:
Resolves #620 by keeping quantifiers in the IR on let-bound values. I'm not sure whether both `Val` cases of `eval_bindings` in `sugartoir.ml` should be amended. All the affected programs we identified so far are fixed by this patch.

### DIFF
--- a/core/ir.mli
+++ b/core/ir.mli
@@ -107,7 +107,7 @@ val binder_of_fun_def : fun_def -> binder
 
 val tapp : value * tyarg list -> value
 
-val letm : binder * tail_computation -> binding
+val letm : ?tyvars:tyvar list -> binder * tail_computation -> binding
 val letmv : binder * value -> binding
 (*val letv : tybinder * value -> binding*)
 
--- a/tests/typed_ir/T620.links
+++ b/tests/typed_ir/T620.links
@@ -1,0 +1,7 @@
+sig foo : () -> ()
+fun foo() {
+  ()
+}
+
+sig bar : () ~> ()
+var bar = foo;
--- a/core/ir.ml
+++ b/core/ir.ml
@@ -117,7 +117,7 @@ let tapp (v, tyargs) =
     | [] -> v
     | _ -> TApp (v, tyargs)
 
-let letm (b, tc) = Let (b, ([], tc))
+let letm ?(tyvars=[]) (b, tc) = Let (b, (tyvars, tc))
 let letmv (b, v) = letm (b, Return v)
 
 let rec is_atom =
--- a/core/sugartoir.ml
+++ b/core/sugartoir.ml
@@ -117,7 +117,8 @@ sig
   val condition : (value sem * tail_computation sem * tail_computation sem) -> tail_computation sem
 
   val comp : env -> (CompilePatterns.Pattern.t * value sem * tail_computation sem) -> tail_computation sem
-  val letvar : (var_info * tail_computation sem * (var -> tail_computation sem)) -> tail_computation sem
+  val letvar : (var_info * tail_computation sem * tyvar list *
+               (var -> tail_computation sem)) -> tail_computation sem
 
   val xml : value sem * string * (name * (value sem) list) list * (value sem) list -> value sem
   val record : (name * value sem) list * (value sem) option -> value sem
@@ -267,7 +268,7 @@ struct
 
     val lift_alist : ('a*'b sem) list -> (('a*'b) list) M.sem
 
-    val comp_binding : var_info * tail_computation -> var M.sem
+    val comp_binding : ?tyvars:tyvar list -> var_info * tail_computation -> var M.sem
 
     val fun_binding :
       Var.var_info * (tyvar list * binder list * computation) * location ->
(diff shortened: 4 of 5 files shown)


core/ir.mli
```diff
@@ -107,7 +107,7 @@ val binder_of_fun_def : fun_def -> binder
 
 val tapp : value * tyarg list -> value
 
-val letm : binder * tail_computation -> binding
+val letm : ?tyvars:tyvar list -> binder * tail_computation -> binding
```
frank-emrich: It feels like the only place other than `sugartoir.ml` that calls `letm` is `compilePatterns.ml`. We should not make the `tyvars` optional here, but change the calls  in `compilePatterns.ml` (and verify that creating a let with empty `tyvars` there is actually the right things to do).

jstolarek:
Ugh, how do I un-fail Azure build? It's tripping over some nonsense.

tests/typed_ir/T620.links
```diff
@@ -0,0 +1,7 @@
+sig foo : () -> ()
```
frank-emrich: Looks like the only polymorphism here are the effect variables hidden by using -> and ~>. Maybe we should make them explicit here, as their handling is what this test actually checks.

jstolarek: I don't mind adding more tests, if there are any cases to be tested, but I definitely want to include the program that was originally reported in the bug report.

Overall, I'm not too worried about this test. Once we fix all the IR typing error we can enable IR checking in Travis/Azure and that will provide an extensive testing of IR correctness of programs.

core/ir.mli
jstolarek: > verify that creating a let with empty tyvars there is actually the right things to do

I am almost certain it is not the right thing to do - it is what most likely causes all the "unbound type variable" failures in #693 and #694. That being said, I did not yet investigate `compilePatterns`. Indeed, it might turn out that what you're proposing is the next step in fixing the above-mentioned bugs.

frank-emrich: If investigating `compilePatterns.ml` turns out to be a can of worms, we might just change it to explicitly pass `[]` for now, maintaining the existing behavior and possibly changing that later, if we fix those typing bugs in a dedicated PR.

frank-emrich:
> Ugh, how do I un-fail Azure build? It's tripping over some nonsense.

If you go to the "Checks" tab in this PR, you can click to re-run the Azure tests. Unfortunately, nothing happened when I did that. I then tried to click on the details of the Azure run, and login (using the "use Github" button on the bottom of the login page). That gave me an error. I was hoping to trigger a re-run this way, because that was one way of getting a "please re-run tests" button in Travis.

Could you try these two things?

jstolarek:
> Could you try these two things?

I tried and got the same results: nothing happens when I want to re-run the tests and I'm getting an error when attempting to log into Azure :-/

core/ir.mli
jstolarek: For now I'm inclined not to change `compilePatterns.ml` if there's no good reason to do so. So far it looks like the cause of #693 is somewhere else - when we get to compilation of function into let binding the list of type variables is empty, which clearly is wrong. Once that is fixed there shouldn't be a need to change pattern compilation, I think. But I'd like to move the discussion on that to #693.

SquidDev:
Sorry to get involved - I encountered this problem in #663, and so wanted to check if this change resolved those issues.

One case this still fails is with `var f = fun(x) {};`. This compiles to:

```
[(Ir.Fun
    ((2151, ((6088) -6089-> (), "_fun__g134", Var.Scope.Local)),
     ([], [(2150, (6088, "", Var.Scope.Local))],
      ([(Ir.Let ((2149, (6088, "x", Var.Scope.Local)),
           ([], (Ir.Return (Ir.Variable 2150)))))
         ],
       (Ir.Return (Ir.Extend ({}, None))))),
     None, CommonTypes.Location.Unknown));
  (Ir.Let ((2152, ((6088) -6089-> (), "f", Var.Scope.Global)),
     ([6088; 6089::Row], (Ir.Return (Ir.Variable 2151)))))
  ]
```

As the function is emitted outside the let binding, the type variables are not actually in scope when they're required. I don't know if this is possible to fix now, or just should be filed away for later. Sorry!

jstolarek:
@SquidDev, thanks for a test case! I just tried it. Without a patch in this PR IR typechecking returns:

```
Type mismatch:
 Expected:
  forall a,b::Row.(a) -b-> ()
 Actual:
  (a) -b-> ()
occurring in IR binding:
(Ir.Let ((232, (forall a,b::Row.(a) -b-> (), "f", Var.Scope.Global)),
   ([], (Ir.Return (Ir.Variable 231)))))
```
With this PR that error goes away just as expected. There are also two other "unbound type variable" errors, that don't go away, also as expected, since this PR does not address this (yet?).

dhil:
> 
> 
> Ugh, how do I un-fail Azure build? It's tripping over some nonsense.

Try an empty commit.
```
git commit --allow-empty
git push origin js-T620
```

jstolarek:
> Try an empty commit. 

Ugh. If there's no sane way of restarting Azure builds then it's not very useful.

jstolarek:
Merged as d7489b7.

PR #696, opened 2019-07-31
Allow quoted field names in lens expressions

SimonJF:
Record definitions allow fields like the following:
  (foo: Int, "returning": Bool)

for field names like "returning" which are keywords. However, these
quoted field names were not allowed in relational lens expressions.


SimonJF:
As this is a straightforward change, I will merge unless there are any objections.

PR #700, opened 2019-08-05
Fix default jsliburl handling

SimonJF:
The handling of `jsliburl` was buggy in that slashes weren't properly handled in the case of a non-empty base URL. This manifested itself in lots of errors for new users who did not have `jsliburl` explicitly set to `/lib/`.

This patch fixes the handling (and the default value of the property) so the config need not explicitly contain `jsliburl` anymore.

### DIFF
--- a/core/webserver.ml
+++ b/core/webserver.ml
@@ -257,7 +257,7 @@ struct
         let prefixed_lib_url =
           let base_url = Settings.get_value Basicsettings.Appserver.internal_base_url in
           let js_url = Settings.get_value Basicsettings.Js.lib_url in
-          if base_url = "" then js_url else
+          if base_url = "" then "/" ^ (Utility.strip_slashes js_url) ^ "/" else
           "/" ^
           (base_url |> Utility.strip_slashes) ^ "/" ^
           (js_url |> Utility.strip_slashes) ^ "/" in
--- a/core/basicsettings.ml
+++ b/core/basicsettings.ml
@@ -115,7 +115,7 @@ module Js =
 struct
   let optimise = Settings.add_bool("optimise_javascript", true, `User)
   let elim_dead_defs = Settings.add_bool("elim_dead_defs", false, `User)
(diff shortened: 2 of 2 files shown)


core/webserver.ml
```diff
@@ -257,7 +257,7 @@ struct
         let prefixed_lib_url =
           let base_url = Settings.get_value Basicsettings.Appserver.internal_base_url in
           let js_url = Settings.get_value Basicsettings.Js.lib_url in
-          if base_url = "" then js_url else
+          if base_url = "" then "/" ^ (Utility.strip_slashes js_url) ^ "/" else
```
dhil: I think you can use the `Filename` module.
```ocaml
if String.equal base_url "" then Filename.concat "/" js_url
else Filename.(concat "/" (concat base_url (concat js_url "")))
```
This would only work correctly on Linux and Mac, as the I believe `Filename` will insert `\\` on Windows.

We would really be using a small "url" building API instead of the manuel string glueing. It ought to be much less prone to error.

dhil: I have implemented a subset of the path specification of RFC 3986 (c.f. https://github.com/dhil/links/blob/url/core/utility.ml#L1391). I only need to implement the normalisation procedure, then that ought to provide a more robust interface for constructing URI paths.

SimonJF: This is great, thanks! A full URI DSL seems superfluous for the purposes of this patch though (which really does just add slashes at the beginning and end of a URL if they're not there already). I'd like to get this in before 0.9 drops later this week because it's a really irritating bug -- so would you be OK with this being merged as is at the moment, and incorporating the URI DSL later?

dhil: Yes. My comment was more of a general nature. We have like 15-20 places where we manually strip and glue slashes back on.

SimonJF: Agreed. OK -- looking forward to your PR, it looks like a great idea!

PR #702, opened 2019-08-05
Fix addAttributes

### LINKED ISSUE #670: HTML attributes broken in client code?
The following program doesn't seem to behave correctly. In the initial (server-side) rendering, it works as expected, with the text in red, but after clicking, it becomes black. It seems that the JS code that evaluates attribute list doesn't actually know about the list data structure The attributes are replaced with _head="[Object object]" _tail="[Object object]", which the browser doesn't know what to do with.

```
fun core() {
  <div id="me" l:onclick="{replaceNode(core(), getNodeById("me"))}" { [("style", "color:#ff8080")] }>
  Test
  </div>
}

fun thepage() {
  page
    <html>
    { core() }
    </html>
}

fun main() {
  addRoute("/", fun(_) {thepage()});
  servePages()
}

main()

```

SimonJF:
Fixes #670. 

The issue was that the `addAttributes` function was not correctly implemented. Since I don't think any of us were aware of the sugar for allowing attribute lists, nobody noticed!

This patch correctly implements the logic needed for `addAttributes` to work.

### DIFF
--- a/lib/js/jslib.js
+++ b/lib/js/jslib.js
@@ -2172,10 +2172,9 @@ function _addAttributes(xml, attrs) {
 
   const xmlItem = JSON.parse(JSON.stringify(first));
 
-  for (let attr in attrs) {
-    xmlItem.attrs[attr] = attrs[attr];
-  }
-
+  LINKEDLIST.forEach(attrs, function(a) {
+    xmlItem.attrs[a[1]] = a[2];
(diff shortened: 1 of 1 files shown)


ezrakilty:
Cool, thanks!

SimonJF:
Thanks for the report!

PR #704, opened 2019-08-05
Fix type variable name generation in error messages

jstolarek:
This PR fixes a bug which made it non-straightforward to share variable names between two calls to `string_of_datatype`. I think the reason is that when I originally wrote name generation code I only focused on its use by grippers and didn't think that it might be used more widely. But now that we have the IR typechecker we get useless messages like:

```
Type mismatch:
 Expected:
  (229:(a::Any, b::Any) -c-> d::Any,230:a::Any)
 Actual:
  (229:(a::Any, b::Any) -c-> d::Any,230:a::Any)
```
Here both types are identical and that's because `string_of_datatype` generates a fresh set of variables for each call. With this PR we have:

```
Type mismatch:
 Expected:
  (229:(a::Any, b::Any) -c-> d::Any,230:a::Any)
 Actual:
  (229:(a::Any, b::Any) -e-> d::Any,230:a::Any)
```
which makes a lot more sense.

Also, I allowed myself to drop an unused `include_aliases` optional argument. It wasn't used (i.e. it was always `true`) and the possibility of having it set to `false` was marked as a bug in a comment.

### DIFF
--- a/core/irCheck.ml
+++ b/core/irCheck.ml
@@ -398,8 +398,10 @@ let eq_types occurrence : type_eq_context -> (Types.datatype * Types.datatype) -
 
 let check_eq_types (ctx : type_eq_context) et at occurrence =
   if not (eq_types occurrence ctx (et, at)) then
+    let exp_str = Types.string_of_datatype et in
+    let act_str = Types.string_of_datatype ~refresh_tyvar_names:false at in
     raise_ir_type_error
-      ("Type mismatch:\n Expected:\n  " ^ Types.string_of_datatype et ^ "\n Actual:\n  " ^ Types.string_of_datatype at)
+      ("Type mismatch:\n Expected:\n  " ^ exp_str ^ "\n Actual:\n  " ^ act_str)
       occurrence
 
 let check_eq_type_lists = fun (ctx : type_eq_context) exptl actl occurrence ->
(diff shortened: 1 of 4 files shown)


jstolarek:
I added this to 0.9.0 milestone - I think it's a useful bugfix to have.

core/irCheck.ml
```diff
@@ -422,7 +424,11 @@ let ensure_effect_present_in_row ctx allowed_effects required_effect_name requir
 let ensure_effect_rows_compatible ctx allowed_effects imposed_effects_row occurrence =
   ensure
     (eq_types occurrence ctx (`Record allowed_effects, `Record imposed_effects_row))
-    ("Incompatible effects; Allowed:\n" ^ (Types.string_of_row allowed_effects) ^ "\nactual effects:\n" ^  (Types.string_of_row imposed_effects_row))
+    (let allowed_str = Types.string_of_row allowed_effects in
+     let actual_str  = Types.string_of_row ~refresh_tyvar_names:true
```
jstolarek: Ugh. This is bogus. Should be `~refresh_tyvar_names:false`. Will fix and merge into master

core/irCheck.ml
```diff
@@ -619,8 +625,13 @@ struct
               end
             else
               begin
-              ensure (eq_types (SVal orig) (o#extract_type_equality_context ()) (vt, t) || is_sub_type (vt, t)) (Printf.sprintf "coercion error: %s is not a subtype of %s"
-                                         (string_of_datatype vt) (string_of_datatype t)) (SVal orig);
+              ensure (eq_types (SVal orig) (o#extract_type_equality_context ())
+                               (vt, t) || is_sub_type (vt, t))
+                (let vt_str = string_of_datatype vt in
+                 let t_str  = string_of_datatype ~refresh_tyvar_names:true t in
```
jstolarek: And same mistake here.

jstolarek:
Merged as aa262ff12e9390112e5c57938c693f18fe7d825f

PR #705, opened 2019-08-05
Dynamic examples page

SimonJF:
This patch removes the static examples page, reimplementing it in Links.
Most functionality was straightforward to port over, however we needed
two additional JS library functions: domGetAnchor and jsSetWindowEvent.


SimonJF:
I've checked this, but I'd like someone to give this a try before I put it in, if that's OK.

SimonJF:
I've checked this in detail on a fresh VM, so I'm confident that it works. Merging.

PR #707, opened 2019-08-07
Fix #706.

### LINKED ISSUE #706: ***: Error: Links_core.Utility.Disk.AccessError("") when using DB driver from OPAM install on 0.9
Using the installed version of Links (try pinning it) will result in the above error. I have observed this error on Wilmer's and Sid's machines, necessitating source installations. I have also reproduced it on a fresh VM.

This is something to do with the interplay between OPAM and the dynamic linking code, and is of course a blocker for 0.9.

SimonJF:
The issue was with an edge case which was the result of an empty
database drivers setting string.

This patch ensures that an AccessError is correctly handled when looking
up database drivers. It also special-cases the (common) case of an empty
string.

### DIFF
--- a/core/databaseDriver.ml
+++ b/core/databaseDriver.ml
@@ -79,7 +79,10 @@ module Locator = struct
          loop false pattern (opam_share driver_name)
       | [] -> raise Not_found
       | path :: paths ->
-         match Glob.files path pattern with
+         let files =
+           try Glob.files path pattern with
+             | Disk.AccessError _ -> [] in
+         match files with
          | [] -> loop fallback pattern paths
          | fs -> fs
     in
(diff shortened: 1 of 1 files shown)


core/databaseDriver.ml
```diff
@@ -79,7 +79,10 @@ module Locator = struct
          loop false pattern (opam_share driver_name)
       | [] -> raise Not_found
       | path :: paths ->
-         match Glob.files path pattern with
+         let files =
```
dhil: Since 4.08 we can use exception patterns in a nested match, so
```ocaml
match Glob.files path pattern with
| [] -> loop fallback pattern paths
| fs -> fs
| exception Disk.AccessError _ -> []
```

PR #708, opened 2019-08-07
Updates to release toolchain

SimonJF:
This patch puts in place the various changes we need in order to release 0.9 using `dune-release`.


Makefile
```diff
@@ -130,17 +130,23 @@ rule-check: tools/rule-check
 	@echo "Applying rule check"
 	@tools/rule-check
 
-# The below machinery is used to prepare a release using topkg.
+# The below machinery is used to prepare a release using dune-release.
 REPO=../opam-repository
 PACKAGES=$(REPO)/packages
 
+TAG_NAME := $(shell git describe --tags)
```
dhil: I think you should use lazy expansion here, i.e.
```
TAG_NAME = $(shell git describe --tags)
```

SimonJF:
Thanks, will merge after CI passes.

PR #709, opened 2019-08-07
Display an alert if `jslib` fails to load.

### LINKED ISSUE #590: Take user-friendly measures in the event of jslib load failure
I recently had a case where Sam and me thought that the examples were broken, when actually I had just entered a path for jslib in my config file that was not accessible. The webserver should simply not start in this case.

dhil:
This patch instruments the Links web runtime with a "readiness check"
before calling `_startRealPage`. If the readiness check fails it emits
an error message (in HTML) and discontinues the start up process. As a
consequence Links will no longer render the user-defined webpage in
the event of the runtime dependency `jslib` failing to load. Instead the following error message is displayed:

```
<h1>Startup error: Runtime dependency `jslib.js' is not loaded.</h1>
```

Currently, the readiness check only checks whether `jslib` is
loaded. It could be extended to check whether alien JavaScript sources
have been loaded too. However, I think the page compiler is due a
refactor before we go down that path.

Resolves #590.

### DIFF
--- a/core/page.ml
+++ b/core/page.ml
@@ -87,6 +87,18 @@ module Make_RealPage (C : JS_PAGE_COMPILER) (G : JS_CODEGEN) = struct
                         ^ env
                         ^ head
                         ^ script_tag (String.concat "\n" defs)
+                        ^ "<script type=\"text/javascript\">
+                             'use strict';
+                             function _isRuntimeReady() {
+                                if (window._JSLIB === void 0 || window._JSLIB !== true) {
+                                   const msg = \"<h1>Fatal error: Runtime dependency `jslib.js' is not loaded.</h1>\";
+                                   document.body.innerHTML = msg;
+                                   document.head.innerHTML = msg;
+                                   return false;
+                                }
+                                return true;
+                             }
+                           </script>"
                      )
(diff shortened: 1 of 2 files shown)


core/page.ml
```diff
@@ -87,6 +87,18 @@ module Make_RealPage (C : JS_PAGE_COMPILER) (G : JS_CODEGEN) = struct
                         ^ env
                         ^ head
                         ^ script_tag (String.concat "\n" defs)
+                        ^ "<script type=\"text/javascript\">
+                             'use strict';
+                             function _isRuntimeReady() {
+                                if (window._JSLIB === void 0 || window._JSLIB !== true) {
+                                   const msg = \"<h1>Startup error: Runtime dependency `jslib.js' is not loaded.</h1>\";
+                                   document.body.innerHTML = msg;
```
SimonJF: Will this mean that the message is displayed twice if the page has a `<head>` and a `<body>`? If so, would it be worth just doing `document.innerHTML = <html><body>...</body></html>`?

SimonJF: Actually, testing this, it's fine. Let's merge.

dhil: No. It resets the head and body. The body displays the message in text, whilst the header sets the page title to be the message.

PR #710, opened 2019-08-08
Fix #678 (Quantifiers escape their scopes)

### LINKED ISSUE #678: Quantifiers can escape their scope
Consider:

```
var f = id(id);

sig g : forall a. (a) -> a
fun g(x) { f(x) }
```

Initially f is given the type `(%t) -> %t` (as it cannot be generalised). When then visiting the body of `g`, we unify `%t` with `g's` `a` quantifier. As a result, `f` now has the (unquantified) type `(a) -> a`, meaning the quantifier has leaked outside `a`'s definition.

The fix here is to check a function's quantifiers do not leak outside its body (i.e. do not appear in the outer environment) after typing the function itself.

SimonJF:
This patch does a fairly simple fix which ensures that given an explicitly-quantified signature, that the quantifiers do not appear in the typing environment after typechecking the body. The check is only performed when a function has explicit quantifiers.

As an example, the program from #678:
```
var f = id(id);

sig g : forall a. (a) -> a
fun g(x) { f(x) }
```

Now gives the error:
```
squid-bug.links:4: Type error: The quantifiers in the type of function
    g: forall a.(a) -b-> a
escape their scope, as they are present in the types:
    f: (a) -b-> a
In expression: fun g(x) { f(x) }.
```


core/typeSugar.ml
```diff
@@ -4070,6 +4093,31 @@ and type_binding : context -> binding -> binding * context * usagemap =
...
+            (* Note that `type_predicate` returns true iff *all* child nodes of
+             * the type satisfy the predicate. Thus the checker returns true if
+             * *all* type variables are *not* in quantifier_set *)
+            let checker = object(_self)
+              inherit Types.type_predicate
+              method! var_satisfies (i, _, _) = not (IntSet.mem i quantifier_set)
+            end in
+            let (is_safe, _) = checker#predicates in
+            let escapees =
+              Env.fold (fun name dt acc ->
```
jstolarek: Random thought: having `Env.filter` and `Env.filterMap` would be useful, though at the moment it seems that `filter` would only be used here, and `filterMap` in one place in `types.ml`

SimonJF:
Thanks, @jstolarek -- I was unsure about the type variable printing API and your comments are very helpful.

I have (I hope) fixed the type variable printing issue, and have implemented `filter` and `filterMap` for environments.

(Previous CI failure on Azure looked transient).

core/typeSugar.ml
```diff
@@ -1515,6 +1522,20 @@ end
                "but the currently allowed effects are" ^ nli () ^
                 code ppr_lt)
 
+    let escaped_quantifier ~pos ~var ~annotation ~escapees =
+      let escaped_tys = List.map snd escapees in
+      build_tyvar_names (annotation :: escaped_tys);
+      let policy () = { (error_policy ()) with Types.Print.quantifiers = true } in
+      let display_ty (var, ty) =
+        let ppr_ty = Types.string_of_datatype ~policy ty in
```
jstolarek: Again, not `Types.string_of_datatype` but `string_of_datatype`. `Types.string_of_datatype` will refresh type variables on each call, `string_of_datatype` will not.

jstolarek: Also, might be worth inlining `ppr_ty`.

core/typeSugar.ml
```diff
@@ -4070,6 +4091,27 @@ and type_binding : context -> binding -> binding * context * usagemap =
...
+             * *all* type variables are *not* in quantifier_set *)
+            let checker = object(_self)
+              inherit Types.type_predicate
+              method! var_satisfies (i, _, _) = not (IntSet.mem i quantifier_set)
+            end in
+            let (is_safe, _) = checker#predicates in
+            let escapees =
+              Env.filter (fun _ dt -> not (is_safe dt)) env
+              |> Env.bindings in
+            if not (escapees = []) then
```
jstolarek: `ListUtils.empty` ?

core/typeSugar.ml
SimonJF: Gotcha -- I think you mean `show_type`? But that doesn't allow me to pass in a policy (and I require one here in order to show the quantifiers in the error message).

I think my best bet would therefore to be explicit and say
```
Types.string_of_datatype ~policy ~refresh_tyvar_names=false
```

jstolarek: Why not:

```OCaml
Printf.sprintf "%s: %s" var (string_of_datatype ~policy ty) in
```
Am I missing something obvious here?

jstolarek: Ah, yes. I am missing something obvious - just realized what you mean by `show_type.` Give me a moment to look at the code.

SimonJF: `string_of_datatype` doesn't seem to exist in the scope unqualified:
```
File "core/typeSugar.ml", line 1530, characters 37-55:
Error: Unbound value string_of_datatype
```

jstolarek: Yes, you're absolutely right. Apparently the API is a bit confusing and easily leads to bugs (e.g. on line 1501 in `typeSugar.ml`). This probably needs rethinking.

I wonder of it would make sense to display quantifiers by default in error messages?

jstolarek: But either way, this is beyond the scope of this PR.

SimonJF:
Thanks for the review! I'll merge after CI passes.

PR #715, opened 2019-08-13
Decouple IR type structure simplifications from IR type checking.

### LINKED ISSUE #561: Decouple show_compiled_ir_after_backend_transformations from typecheck_ir
Currently `show_compiled_ir_after_backend_transformations` setting only takes effect when `typecheck_ir` is enabled. I don't this needs to be the case. I'd like to be able to see the generated IR even if it isn't getting typechecked.

dhil:
This patch resolves #561 by separating the type structure
simplification pass from the IR type checker. A new flag
`simplify_types` is introduced which dictates whether the
simplification pass is run -- well that is, it runs if the flag is set
or `typecheck_ir` is set as I am not sure whether the IR type checker
depends on the simplification pass.

The patch also fixes a couple of issues with side effects in the IR
backend pipeline infrastructure. As settings `typecheck_ir` and
`show_compiled_ir_after_backend_transformations` had no effect if
Links was not booted with those settings set to true, meaning that
they could not be toggled from within the REPL. This is rectified by
replaying the side effects whenever a backend pipeline is invoked.

### DIFF
--- a/core/backend.ml
+++ b/core/backend.ml
@@ -23,6 +23,16 @@ let only_if predicate transformer =
 let only_if_set setting =
              only_if (Settings.get_value setting)
 
+let only_if_any_set settings transformer =
+  if List.mem true (List.map Settings.get_value settings)
+  then transformer
+  else (fun _ x -> x)
+
+let debug_tell msg =
+  only_if_set
+    Basicsettings.debugging_enabled
+    (fun _tyenv prog ->
+      Debug.print msg; prog)
 
 let print_program _ p =
   Debug.print (Ir.string_of_program p);p
@@ -32,7 +42,7 @@ let print_bindings _ bs =
 
 
 let run pipeline tyenv p =
-  List.fold_left (fun p transformer -> transformer tyenv p) p pipeline
+  List.fold_left (fun p transformer -> transformer tyenv p) p (pipeline ())
 
 let measure name func tyenv p = Performance.measure name (uncurry func) (tyenv, p)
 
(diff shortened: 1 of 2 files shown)


core/backend.ml
```diff
@@ -42,21 +52,41 @@ let perform_for_side_effects side_effecting_transformer tyenv p =
...
 
-    let typechecking_pipeline = [
+    let simplify_type_structure_program () = [
+        debug_tell "simplifying types";
         (*IrTraversals.NormaliseTypes.program;
         IrTraversals.ElimRecursiveTypeCycles.program;*)
         IrTraversals.ElimTypeAliases.program;
         IrTraversals.ElimBodiesFromMetaTypeVars.program;
         (only_if_set Basicsettings.Ir.show_compiled_ir_after_backend_transformations print_program);
+        debug_tell "simplified types"
```
jstolarek: I would move this debug output before `(only_if_set ...` line. Right now the IR is displayed between `"simplifying types"` and `"simplified types"` messages so it's not clear whether this is IR before or after transformations.

core/backend.ml
```diff
@@ -42,21 +52,41 @@ let perform_for_side_effects side_effecting_transformer tyenv p =
...
+      ]
+
+    let simplify_type_structure_bindings () = [
+        debug_tell "simplifying types";
+        (*IrTraversals.NormaliseTypes.bindings;
+        IrTraversals.ElimRecursiveTypeCycles.bindings;*)
+        IrTraversals.ElimTypeAliases.bindings;
+        IrTraversals.ElimBodiesFromMetaTypeVars.bindings;
+        (only_if_set Basicsettings.Ir.show_compiled_ir_after_backend_transformations print_bindings);
```
jstolarek: Same as above.

core/backend.ml
```diff
@@ -66,22 +96,35 @@ struct
...
       ]
 
-    let prelude_pipeline = [
+    let prelude_pipeline () = [
         (* May perform some optimisations here that are safe to do on the prelude *)
         (fun tenv globals -> Closures.bindings tenv Lib.primitive_vars globals);
         (fun tenv globals -> BuildTables.bindings tenv Lib.primitive_vars globals; globals);
-        only_if_set Basicsettings.Ir.typecheck_ir (run prelude_typechecking_pipeline);
+        only_if_any_set
+          [Basicsettings.Ir.typecheck_ir; Basicsettings.Ir.simplify_types]
```
jstolarek: @frank-emrich, can you clarify why simplifying pass happens only when IR typechecking is enabled?

core/backend.ml
```diff
@@ -66,22 +96,35 @@ struct
...
+        only_if
+          perform_optimisations
+          (measure "optimise" (run optimisation_pipeline));
         Closures.program Lib.primitive_vars;
-        perform_for_side_effects (BuildTables.program Lib.primitive_vars);
-
-        only_if_set Basicsettings.Ir.typecheck_ir (run typechecking_pipeline);
+        perform_for_side_effects
+          (BuildTables.program Lib.primitive_vars);
+        only_if_any_set
```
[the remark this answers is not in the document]
dhil: I am not sure what fragment you mean is identical to what other fragment? The lists may look the same, but they invoke different functions. the `main_pipeline` invokes `program` and `prelude_pipeline` invokes `bindings`.

core/backend.ml
dhil: I disagree, this follows the scheme for `show_compiled_ir`.

core/backend.ml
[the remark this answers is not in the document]
jstolarek: Ah, didn't notice that. Sorry.

PR #718, opened 2019-08-14
Decouple IR type structure simplifications from IR type checking.

### LINKED ISSUE #561: Decouple show_compiled_ir_after_backend_transformations from typecheck_ir

dhil:
This patch resolves #561 by separating the type structure
simplification pass from the IR type checker. A new flag
`simplify_types` is introduced which dictates whether the
simplification pass is run -- because the IR type checker depends on
the `ElimBodiesFromMetaVars` pass the simplification pass will also,
as a temporary workaround, run if `typecheck_ir` is set to `true`.

The patch also fixes a couple of issues with side effects in the IR
backend pipeline infrastructure. As settings `typecheck_ir` and
`show_compiled_ir_after_backend_transformations` had no effect if
Links was not booted with those settings set to true, meaning that
they could not be toggled from within the REPL. This is rectified by
replaying the side effects whenever a backend pipeline is invoked.


dhil:
Here it is again, @jstolarek.

frank-emrich:
> as I am not sure whether the IR type checker depends on the simplification pass.

Originally, that was the plan. However, it then turned out that just supporting certain features in the type-checker (e.g., aliases) is easier than removing them beforehand (and really making sure that they are not re-introduced).

I'm not sure if it's worthwhile keeping these two transformations. They make IR dumps shorter and easier to read, but are not and should not be necessary for the IR type-checker to work.

dhil:
Thanks @frank-emrich for sharing your wisdom. I have updated the PR accordingly.

dhil:
@frank-emrich the CI begs to differ. It seems the IR type checker does depend on either of the simplification passes.

dhil:
This is mind-boggling. The issue does indeed seems to be with the pretty printer for types.
```
expected:
fun : (_::(Unl,Mono)) -> Int
actual:
fun : (_) -> Int

expected: 
fun : (a::(Any,Mono)) -> a::(Any,Mono)
actual:
fun : (a::Any) -> a::Any

expected:
fun : ((a::(Any,Mono)) -b-> c::Any, (d::(Any,Mono)) -b-> a::(Any,Mono)) -> (d::(Any,Mono)) -b-> c::(Any,Mono)
actual:
fun : ((a::Any) -b-> c::Any, (d::Any) -b-> a::Any) -> (d::Any) -b-> c::Any
```

Note that the CI passed on commit [c160d02](https://github.com/links-lang/links/pull/718/commits/c160d0227f709ecefc728ca240240d36cef6e389), but if I roll back, rebuild, and rerun the `typed_ir.tests` I get the same expectation mismatch errors. It even happens on top of the master branch on either of my machine.

This is sort of odd behaviour exciting and worrying at the same time.

dhil:
The problem seems to be that `show_kinds` is not set to `full` when running the test suite...

dhil:
So the composition test fails if the `ElimBodiesFromMetaTypeVars` isn't run:
```
!FAILURE: Anonymous composition function
  error: Anonymous composition function
      (Unexpected return code:
expected `0';
     got `1')
  error: Anonymous composition function
      (Unexpected stdout:
expected `fun : ((a::(Any,Mono)) -b-> c::(Any,Mono), (d::(Any,Mono)) -b-> a::(Any,Mono)) -> (d::(Any,Mono)) -b-> c::(Any,Mono)';
     got `')
  error: Anonymous composition function
      (Unexpected stderr:
expected `';
     got `***: IR Type Error: Type mismatch:
 Expected:
  (_::(Any,Mono)) -> _::(Any,Mono)
 Actual:
  (a::(Any,Mono)) -b-> c::(Any,Mono)
occurring in IR binding:
(Ir.Let ((231, ((_::(Any,Mono)) -> _::(Any,Mono), "g", Var.Scope.Local)),
   ([], (Ir.Return (Ir.Variable 238)))))')
```
So the typechecker is sensitive to at least one of the transformations.

dhil:
> Note that the CI passed on commit c160d02, but if I roll back, rebuild, and rerun the typed_ir.tests I get the same expectation mismatch errors. It even happens on top of the master branch on either of my machine.
>
> This is sort of odd behaviour exciting and worrying at the same time.

Oh my.... I understand this behaviour now. It is because our test-runner-script is **broken**! It is no longer true that `./test-harness` can be run independently of `./run-tests`. Hence why it was failing locally on master for me. Sad.

So the real reason why the test suite was suddenly failing is that the IR typechecker does depend on the transformation `ElimBodiesFromMetaTypeVars`.

jstolarek:
@frank-emrich, it indeed seems that IR typechecker depends on these two transformations. Consider `fun (x) {<a>{x}</a>}` as a test program with config:

```
typecheck_ir=true
prelude=empty.links
print_types_pretty=false
debug=true
hide_fresh_type_vars=false
show_flavours=true
show_quantifiers=true
generalise_toplevel=false
```

Disabling both `ElimTypeAliases` and `ElimBodiesFromMetaTypeVars` transformations leads to IR typechecking error:

```
Type mismatch:
 Expected:
  `MetaTypeVar (Point @ 0x7f8ccedb5d60 `Body (`Alias ((("Xml", [], []),
         `Application (({ Types.Abstype.id = "List"; name = "List";
                           arity =
                           [(CommonTypes.PrimaryKind.Type,
                             (CommonTypes.Linearity.Unl,
                              CommonTypes.Restriction.Any))
                             ]
                           },
                         [`Type (`Primitive (CommonTypes.Primitive.XmlItem))]))))))
 Actual:
  `Alias ((("Xml", [], []),
         `Application (({ Types.Abstype.id = "List"; name = "List";
                          arity =
                          [(CommonTypes.PrimaryKind.Type,
                            (CommonTypes.Linearity.Unl,
                             CommonTypes.Restriction.Any))
                            ]
                          },
                        [`Type (`Primitive (CommonTypes.Primitive.XmlItem))]))))

```
The difference is only in the `MetaTypeVar (Point @ 0x7f8ccedb5d60 Body` part. The code responsible for checking equality is the `eqt` function (around line 200 in `irCheck.ml`).  Note that:

1. `unalias` function there will only remove top-level aliases but not the recursive ones, which is potentially a problem.

2. `unalias` will convert second type to `Application ...`, which is then compared against a `MetaTypeVar`:

```
| `MetaTypeVar lpoint ->
    begin match Unionfind.find lpoint with
      | `Recursive _ -> Debug.print "IR typechecker encountered recursive type"; true
      | lpoint_cont ->
        begin match t2 with
          `MetaTypeVar rpoint ->
          begin match lpoint_cont, Unionfind.find rpoint with
          | `Var lv, `Var rv -> handle_variable pk_type lv rv context
          | `Body _, `Body _ -> raise (internal_error "Should have removed `Body by now")
          | _ -> false
          end
          | _                   -> false
        end
    end
```
The result is obviously `false`.

I suppose it is possible to inline the logic from `ElimTypeAliases` and `ElimBodiesFromMetaTypeVars` into `irCheck` but I don't think this will be good design. I'd opt for a compositional approach with several passes, each responsible for doing only one thing.

Issue #720, opened 2019-08-14
The test suite runners coherence invariant is broken


dhil:
We have two entry points to our main test suite, `test-harness` and `run-tests`. The former runs any individual `.tests`-file, whilst the latter runs all the `.tests`-files in the directory `tests/` by invoking `test-harness` on each file.

Now, it used to be the case that `./run-tests no-db` would only succeed if and only if `./test-harness` would succeed on every `.tests`-file in `tests/`. This is no longer true as witnessed by running the following commands on the master branch:
```bash
$ ./run-tests no-db && echo $?
0
$ ./test-harness tests/typed_ir.tests || echo $?
1
```
I suspect the problem is that `run-tests` attempts to discover test configuration files, and pass any as an additional argument to `test-harness`.

This problem led to a fair amount of confusion in #718 when attempting to debug a failing CI run.

frank-emrich:
I made that change originally, mostly due to a slightly different interpretation of `run-tests` and `test-harness`: In my opinion, the former is a driver for tests (doing stuff like discovering config files), the latter just executes test.
I feel like that is a clearer separation of responsibilities, but of course this breaks the property you describe:
> Now, it used to be the case that ./run-tests no-db would only succeed if and only if ./test-harness would succeed on every .tests-file in tests/

I don't object to this change per se, since Simon seems to share your interpretation of things (cf. his comment in #722 ). However, the change made by that PR evokes that the .config file found in the directory takes precedence over the one supplied as a command-line argument. That sounds like a bad idea: I had cases where I deliberately wanted to run the same `.tests` file using different `.config` files.

slindley:
The prior behaviour didn't work and made it difficult to configure the test suite properly - in particular it is important that we be able to set different configuration operations for different tests in the same test file. The property Daniel describes is important. The configuration options are part of a test.

It might be useful to supplement the now fixed behaviour with additional functionality, but that should not break what we have now.

dhil:
It is wrong to automatically discover configuration files. They should have to be explicitly listed in the test script files.

PR #724, opened 2019-08-16
Endow testscripts with an optional header section.

dhil:
This patch aims to remove some of the "magic" in the test suite
runners. Specifically, this patch augments our ad-hoc testscript
language with an optional header section, which contains key value
pairs (currently only "config: path/to/file.config" is supported) that
can be used to fix parameters globally for the tests in the respective
script. So if a script uses a particular config file it has to be
listed explicitly in the script header.

Syntactically a header is surrounded by a pair of `---` followed by a
blank line. A header may only contain key-value pairs of the form
`key: value`. This syntax is adopted directly from the Markdown header
format. Example:

```
---
config: tests/typed_ir.tests.config
---

Anonymous constant function
fun(x) {1}
stdout : fun : (_::(Unl,Mono)) -> Int
```

It is possible to override the value of the `config` key by explicitly
passing the name of a configuration file as a command line
argument. For example the following command

```
$ ./test-harness tests/typed_ir.tests tests/typed_ir.tests.config-alternative
```

would run `typed_ir.tests` with the alternative configuration file.

### DIFF
--- a/test-harness
+++ b/test-harness
@@ -27,7 +27,7 @@ def parse(stream):
     are auxiliary options"""
     def tests():
         tests = []
-        for line in stream.readlines():
+        for line in stream:
             if not str.strip(line) and tests[-1]:
                 yield tests
                 tests = []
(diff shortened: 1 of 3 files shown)


dhil:
Sort of a first approximation to resolving the discussion in #720.

frank-emrich:
I like the idea, this feels like the right mechanism for solving the magic discovery of global config files.

How do people feel about just making the header mandatory? We could easily add
```
---
---
```

to all other existing test files to make the script simpler (no more peeking).

dhil:
I am kind of indifferent to as whether the header is mandatory. We can possibly add that later.

frank-emrich:
Okay, let's just merge

Issue #721, opened 2019-08-14
The IR type checker depends on type structure simplification


dhil:
The IR type checker depends on the `ElimBodiesFromMetaVars` simplification pass for types. If it isn't enabled then the IR type checker fails on the composition example in `tests/typed_ir.tests`:
```
fun (f,g) {fun (x){f (g(x))}}
```

frank-emrich:
Solved by #725

PR #725, opened 2019-08-19
Improve support for `Body and `Alias in IR type equality checks

frank-emrich:
Solves #721 

There is a bit of a design question about whether or not we want a design where we rely on certain simplifications having taken place before the IR is type-checked or not. We have no mechanism in place checking that certain settings (e.g., `typecheck_ir`) may only be enabled if another setting is set (`simplify_types`).
 
Until then, this patch fixes the issue that the IR type-checker was failing unless all `Body` constructors were removed from all types beforehand. In particular, it is not necessary anymore to enable `simplify_types` for the IR type-checker to work.

### DIFF
--- a/core/irCheck.ml
+++ b/core/irCheck.ml
@@ -127,6 +127,8 @@ type type_eq_context = {
   tyenv: Types.kind Env.t (* track kinds of bound typevars *)
 }
 
+
+(* Detects recursion at any level *)
 module RecursionDetector =
 struct
   class visitor =
@@ -157,6 +159,26 @@ end
 
 
 
+(* For both rows and types, detect recursion at top-level or immeiately under
+   a `Body or `Alias *)
+let rec is_toplevel_rec_type = function
+  | `MetaTypeVar mtv ->
+     begin match Unionfind.find mtv with
+       | `Recursive _ -> true
+       | `Body b -> is_toplevel_rec_type b
+       | _ -> false
+     end
+  | `Alias (_, t') -> is_toplevel_rec_type t'
+  | _ -> false
+
+let rec is_toplevel_rec_row (_, row_var, _) =
+  match Unionfind.find row_var with
+       | `Recursive _ -> true
+       | `Body r -> is_toplevel_rec_row r
+       | _ -> false
+
+
+
 let eq_types occurrence : type_eq_context -> (Types.datatype * Types.datatype) -> bool =
   fun context (t1, t2) ->
     let lookupVar lvar map =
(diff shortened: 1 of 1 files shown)


core/irCheck.ml
```diff
@@ -353,13 +347,17 @@ let eq_types occurrence : type_eq_context -> (Types.datatype * Types.datatype) -
       | `End, `End -> true
       | _, _ -> false
     and eq_rows (context, r1, r2) =
-      let (lfield_env, lrow_var, ldual) = remove_absent_fields_if_closed (Types.flatten_row r1) in
-      let (rfield_env, rrow_var, rdual) = remove_absent_fields_if_closed (Types.flatten_row r2) in
-      let r1 = eq_field_envs (context, lfield_env, rfield_env) in
-      let r2 = eq_row_vars (context, lrow_var, rrow_var) in
-        r1 && r2 && ldual=rdual
+      if is_toplevel_rec_row r1 || is_toplevel_rec_row r2 then
```
dhil: I take this as the IR typechecker cannot handle recursive rows at the moment.

dhil:
> We have no mechanism in place checking that certain settings (e.g., `typecheck_ir`) may only be enabled if another setting is set (`simplify_types`).

I think the IR type checker should be able to handle the full type structure.

>  In particular, it is not necessary anymore to enable `simplify_types` for the IR type-checker to work.

Is there any point in retaining the type structure simplification pass? Should we just get rid of it?

jstolarek:
I actually suggested something quite opposite to Frank, i.e. lets have passes that simplify types as much as possible so that the IR typechecker can be simpler. This would mean that the simplification passes perhaps should not be optional, so that we don't end up in a situation where we now that some given program is type incorrect but when we turn on IR typechecking we are actually checking a different program since the original program gets simplified first.

jstolarek:
Rationale for the above is that I find it more elegant to have several simple passes, each with a clearly defined purpose, rather than having all the logic inlined in a single pass. It's just like we have multiple desugaring passes instead of just one that does all the transformations.

dhil:
I think it is an important property to be able to run the type checker on all forms of the IR (or AST), such that it can be used a means for lightweight verification between different transformation passes.

jstolarek:
Ok, that's a good point. I'm thinking your approach might indeed a better one.

core/irCheck.ml
frank-emrich: Indeed, it doesn't handle recursive rows or types at the moment.

frank-emrich:
> Is there any point in retaining the type structure simplification pass? Should we just get rid of it?

Hard to say. It can de-clutter the types when printed using the generated pp_* functions (rather than our own pretty printer, which does its own simplifications).

PR #727, opened 2019-08-20
Remove performance documentation

SimonJF:
While the performance documentation in the `documentation/` directory seems like a lot of work, I don't think it belongs in the main repository anymore. Having a `documentation` and a `doc` directory is quite confusing.

Instead, I have moved it here: https://github.com/links-lang/misc


jamescheney:
+1 I think this was stored in a branch somewhere documenting improvements made in an internship project.  I think this was acted on long ago and is at best of historical interest.

SimonJF:
@slindley do we have the go-ahead to delete this from the main repo?

slindley:
This has nothing to do with me and never really belonged in the Links repository. As you should be able to see from the logs and the PDF file it comes from James's intern Dariusz Jędrzejczak's work on speeding up JavaScript code in 2014.

SimonJF:
Right. Thought so, but (as with any large removal) thought I'd check before removing. In that case, let's merge.

PR #728, opened 2019-08-20
Fix #619

### LINKED ISSUE #619: XML literals in queries

SimonJF:
A bug which has been nagging me for a while because soundness errors make me sad. 
One of those bugs where the fix is straightforward, but diagnosing and
debugging is less so.

Turns out that we need to implement the `stringToXml` function during
query normalisation, otherwise evaluation gets stuck, and we end up with what looks like a soundness bug (but in fact is due to an unevaluated application of a primitive).

### DIFF
--- a/core/query/query.ml
+++ b/core/query/query.ml
@@ -58,8 +58,8 @@ let unbox_pair =
           x, y
     | _ -> raise (runtime_type_error "failed to unbox pair")
 
-let rec unbox_list x =
-  match x with
+let rec unbox_list =
+  function
     | Q.Concat vs -> concat_map unbox_list vs
     | Q.Singleton v -> [v]
     | _ -> raise (runtime_type_error "failed to unbox list")
--- a/core/query/query.ml
+++ b/core/query/query.ml
@@ -58,8 +58,8 @@ let unbox_pair =
           x, y
     | _ -> raise (runtime_type_error "failed to unbox pair")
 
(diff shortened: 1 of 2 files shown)


SimonJF:
I will merge this today if there are no objections.

(Re-running CI).

PR #729, opened 2019-08-20
Relational Lens serial column support

rudihorn:
This is a patch to support postgresql serial columns. Example:

Create a postgresql table:

```SQL
CREATE TABLE test (
  id SERIAL,
  name TEXT);
```

Then run the following example:
```
var db = database "links";

var testTable =
    table "test"
    with (id: Int, name: String)
    tablekeys [["id"]]
    from db;

var testLens = lens testTable with { id -> name };
var testLens = lensserial testLens with id;

var entries = lensget testLens;
var entries = entries ++ [( id = NewKey, name = "test" )];

var r = lensput testLens with entries;
var entries = lensget testLens;
entries
```

### Implementation
It is implemented by adding a new variant type `Serial` to the prelude which supports one of three values: `Key(Int)`, `NewKey` and `NewKeyMapped(Int)`. `Key(Int)` indicates a known key value retrieved from the database. `NewKey` indicates that a value should be chosen by the database server and `NewKeyMapped(Int)` behaves similarly, except that it allows multiple entries to refer to the same key (e.g. in the case of a table join, where two new entries are inserted referring to the same right table).

Equality for serial values is handled in a way where two `NewKeyMapped` values are identical if they have the identical integer value, while any two `NewKey` values are considered distinct. `Key` behaves as any integer does.

The `lensserial` expression is used to change existing column values into serial columns on primitive lenses. This is probably not the nicest syntax, so may be good to think of something else.

@SimonJF perhaps you could test this and see what breaks?


SimonJF:
This is great, well done and thanks! I will have an in-depth look later in the week.

core/errors.ml
```diff
@@ -183,6 +184,7 @@ let format_exception =
      pos_prefix (Printf.sprintf "Error: Cannot load plugin '%s' (link error: %s)\n" file (Dynlink.error_message err))
   | ClientCallOutsideWebMode fn ->
      pos_prefix (Printf.sprintf "Error: Cannot call client side function '%s' outside of web mode\n" fn)
+  | MissingBuiltinType alias -> Printf.sprintf "Error: Missing builtin type with alias '%s'. Is it defined in the prelude?" alias
```
dhil: Is this some expectation mismatch error? Do note that in the (near) future most/all type aliases will be declared and implemented in a Links source file like `prelude.links`.

core/parser.mly
```diff
@@ -906,6 +906,7 @@ lens_expression:
 | LENS exp DEFAULT                                             { with_pos $loc (LensLit ($2, None))}
 | LENS exp TABLEKEYS exp                                       { with_pos $loc (LensKeysLit ($2, $4, None))}
 | LENS exp WITH LBRACE fn_deps RBRACE                          { with_pos $loc (LensFunDepsLit ($2, $5, None))}
+| LENSSERIAL exp WITH VARIABLE+                                               { with_pos $loc (LensSerialLit ($2, $4, None))}
```
dhil: Been punching the spacebar too many times?

core/errors.ml
rudihorn: This is the error thrown when the lens code tries to find the `Serial` type (which is already defined in the prelude). See https://github.com/rudihorn/links/blob/65ed306e12cfeee43384a8862deaccd038ca9f02/core/lens_type_conv.ml#L24 and https://github.com/rudihorn/links/blob/65ed306e12cfeee43384a8862deaccd038ca9f02/core/lens_type_conv.ml#L15.

dhil:
I assume you expect the database backend to postgresql. So what happens if I run your example with the sqlite3 or mysql backend? How do you handle the expectation mismatch?

core/parser.mly
rudihorn: Fixed this formatting error.

rudihorn:
> 
> 
> I assume you expect the database backend to postgresql. So what happens if I run your example with the sqlite3 or mysql backend? How do you handle the expectation mismatch?

I haven't actually tested it but it may or may not work. Currently it translates links database connections into lens database connections, which support custom `quote_field` and `escape_field` functions. AFAIK the subset of SQL I do use works on all three of them.

dhil:
> I haven't actually tested it but it may or may not work.

Do you mind testing it locally? So that we can at least document the "expected" behaviour.

core/errors.ml
dhil: Thank you for the clarification.

PR #731, opened 2019-08-27
Enable modules by default

SimonJF:
Since our webserver examples all use modules, and since they've been in for a few releases already, I'm of the opinion that modules should be enabled by default. Having them not on by default at this point now seems to only confuse people if they forget the flag or the configuration option.

What do others think?


jamescheney:
I had in mind that this was waiting for "proper modules" (#337 #603).  However as of 0.9 I think we have made all of the breaking changes to existing module syntax/behavior we had in mind (as opposed to adding new syntax/features).  The only exception that comes to mind is a possible redesign of the behavior when searching for modules but this would mostly affect configurations rather than code and I think @dhil had in mind to support both loading behaviors for a transition period.

I don't see particular harm in this given that the use of the current module system is fairly well-established.  If we do do this then we should also remove the uses of "-m", or perhaps just remove the `modules` flag entirely.  Should probably discuss tomorrow.

SimonJF:
As decided at the meeting on Wednesday, I have removed the setting and flag altogether.

SimonJF:
Something funky has gone on with this PR, as it seems to be pulling in all of the other changes. Closing and trying again.

PR #737, opened 2019-08-30
Initial stab at Sphinx documentation

SimonJF:
This PR adds my initial work on a [Sphinx site](http://www.sphinx-doc.org/en/master/) for Links documentation.

I have filled out the installation and getting started guides, and ported much of the old `quick-help.pod` over. I have also written other bits and bobs, and have proposed a skeleton structure with other pages to put in.

To build the documentation, you will need Sphinx installed: http://www.sphinx-doc.org/en/master/usage/installation.html

You can then build the documentation by doing: `make doc`

If you have `xdg-open` installed (sorry, Mac users), you can do `make open-doc` to open the documentation; otherwise, you can find the compiled output in `doc/_build/html/index.html`.


jamescheney:
Any chance of automating the Sphinx install/documentation build to test that things work in the CI?

jstolarek:
@SimonJF, thanks again for doing this! I'll try to find some time next week to sit down and actually build and read through the generated docs. Some quick comments:

1. My understanding is that not everything was ported from the `.pod` file? If so, I would suggest to delete the parts that are now in the new documentation and only keep the ones that are not yet ported. That will give us a clear idea of what is already in the new docs and what still requires porting.

2. I'd love to have a section on compiler flags. Not asking you to write it, but I will appreciate creating an empty stub.

SimonJF:
@jamescheney:
> Any chance of automating the Sphinx install/documentation build to test that things work in the CI?

Good idea! Yes, we should test that the documentation builds correctly in the CI. I will add this in.

@jstolarek:
> My understanding is that not everything was ported from the .pod file? 
Correct. Wading through it, but it needs copy-editing and updating for the new versions. 

> I would suggest to delete the parts that are now in the new documentation and only keep the ones that are not yet ported. 

Good idea! Yes, I will do this. Maybe a roadmap of what still needs to be done is a good idea.

> I'd love to have a section on compiler flags. Not asking you to write it, but I will appreciate creating an empty stub.

Yes, another good idea. Not only compiler flags, but also settings would be very useful indeed. I'll add the stub in.

SimonJF:
Let's see how the CI goes with this one, as I've added some checks which may or may not work...

I've decided to leave `quick-help.pod` intact, but have added `quick-help-todo.pod` to keep track of what needs to be added to the new documentation.

azure-pipelines.yml
```diff
@@ -35,11 +35,11 @@ steps:
 
   - bash: |
       case $AGENT_OS in
-        "Linux") sudo apt-get install ocaml ;;
-        "Darwin") brew install ocaml ;;
+        "Linux") sudo apt-get install ocaml python3-sphinx;;
```
dhil: I suggest making it a separate job, say  "Install sphinx". Small one-purpose jobs make it faster and easier to locate and understand CI failures.  Although, the installation is probably of sphinx is probably unlikely to fail..

SimonJF: Sure, I can do that. Let's see if the current jobs work, and I'll add it in after.

jamescheney:
I would be ok with just checking building the documentation on Linux since doing it on MacOS is evidently a pain.

SimonJF:
With some agony (and a force-push to conceal the absurd number of attempts -- sorry to anyone who gets emails for each commit on a PR), I've managed to get the CI going.

SimonJF:
OK, added into CI, added a settings stub (I also took the opportunity to document the flags), and added "Install Sphinx" as a separate CI job. Can this be merged?

jstolarek:
Sorry, I should have sat down to reviewing this one earlier. So I just checked out your branch and tried building the docs. I am getting some warnings about code-block directive. Here's a full build log:


```
[killy@wau : /dane/projekty/links] make doc
cd doc && make html
make[1]: Entering directory '/dane/projekty/links/doc'
Running Sphinx v1.8.4
making output directory...
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 22 source files that are out of date
updating environment: 22 added, 0 changed, 0 removed
reading sources... [100%] settings
/dane/projekty/links/doc/getting-started.rst:27: WARNING: Error in "code-block" directive:
1 argument(s) required, 0 supplied.

.. code-block::

  links> 1 + 2 + 3;;
  6 : Int
/dane/projekty/links/doc/lang-ref/expressions.rst:10: WARNING: Error in "code-block" directive:
1 argument(s) required, 0 supplied.

.. code-block::

  if (x == y) {
    expr1
  } else {
    expr2
  }
/dane/projekty/links/doc/lang-ref/expressions.rst:467: WARNING: Error in "code-block" directive:
1 argument(s) required, 0 supplied.

.. code-block::

  links> take(2,[1,2,3]);
  [1, 2] : [Int]
  links> drop(2,[1,2,3]);
  [3] : [Int]
looking for now-outdated files... none found
pickling environment... done
checking consistency... done
preparing documents... done
writing output... [100%] settings
generating indices... genindex
writing additional pages... search
copying static files... WARNING: html_static_path entry '/dane/projekty/links/doc/_static' does not exist
done
copying extra files... done
dumping search index in English (code: en) ... done
dumping object inventory... done
build succeeded, 4 warnings.

The HTML pages are in _build/html.
make[1]: Leaving directory '/dane/projekty/links/doc'
```

Code blocks mentioned in the warnings are not rendered in the output html:

links-doc

Also, after changing `getting-started.rst`, running `make doc` does not cause a rebuild.

Makefile
```diff
@@ -123,13 +123,28 @@ clean:
...
 .PHONY: rule-check
 rule-check: tools/rule-check
 	@echo "Applying rule check"
 	@tools/rule-check
 
+doc/_build/html:
+	cd doc && make html
+
+.PHONY: doc
+doc : doc/_build/html
```
jstolarek: Replace this target with:

```
.PHONY: doc
doc :
        make -C doc html
```
This ensures that rebuilds are done correctly.

SimonJF:
Thanks, @jstolarek! Hopefully that should do the trick.

Makefile
```diff
@@ -139,7 +139,8 @@ doc/_build/html:
 	cd doc && make html
```
jstolarek: Do we still need `doc/_build/html` target?

SimonJF: Yes -- so that we can ensure that it is present when doing `make open-doc`. That said, I streamlined it a bit so there's less duplication and we now don't have the manual directory changes. (Thanks for the Makefile hint!)

SimonJF:
The warning was due to the `_static` directory not being included. Apparently Git doesn't include empty directories by default, but thanks to a hint found [here](https://steindom.com/articles/add-empty-directory-git-repository), this warning should no longer occur.

jstolarek:
> Apparently Git doesn't include empty directories by default

It's not that git doesn't include empty directories "by default" - it doesn't include them at all. The reason is that git tracks contents of files and directories are only an artefact of reconstructing files. If there are no files in a directory there's nothing to track from git's perspective. I think a proper way to deal with this - if there is a "proper" way here, since it's a design feature - would be to create a post-checkout hook that creates an empty directory. I might sit down and do it as an exercise one day. For now I'm perfectly happy with the `.gitignore` trick that you used.

frank-emrich:
The alternative convention seems to be to put an empty file called `.gitkeep` into the empty folder

SimonJF:
Is there anything stopping this from being merged?

jstolarek:
No, I don't think so. I thought this already went in.

SimonJF:
Wonderful! Thanks for taking the time to try this out and for the reviews, @jstolarek. I'll merge.

PR #738, opened 2019-08-30
Fix #383

### LINKED ISSUE #383: Errors during XHR processing are returned as raw HTML
If an error is encountered during XHR processing, then the "Links Fatal Error" response is sent as raw HTML, rather than a base64-encoded response that tells the client-side code what to do next.

One can experience this by starting the Links factorial tutorial program and typing something nonsensical (such as "a") into the textbox and clicking the "Submit Query" button.  But a smaller example can probably be constructed.

Depending on the browser, the result is either (Firefox) silent failure or (Chrome) an error such as 

```
jslib.js:981 Uncaught DOMException: Failed to execute 'atob' on 'Window': The string to be decoded is not correctly encoded.
    at Object.base64decode (http://localhost:8081/lib/jslib.js:981:40)
    at XMLHttpRequest.<anonymous> (http://localhost:8081/lib/jslib.js:869:48)
```

SimonJF:
The issue was that Lwt.catch was not discriminating between request
types, and was thus responding with a plain HTML page when it should
have been responding with a b64-encoded JSON string.

This patch ensures that the correct error is thrown in the console.

### DIFF
--- a/core/webif.ml
+++ b/core/webif.ml
@@ -163,33 +163,37 @@ struct
     (* We need to be a bit careful about what we respond here. If we are evaluating
      * a ServerCont or EvalMain, that is fine -- but we need to construct a b64-encoded
      * JSON object if we're responding to a ClientReturn or RemoteCall. *)
+
+    let handle_ajax_error = function
+      | Aborted r -> Lwt.return r
+      | e ->
+         let formatted_exn =
+           Errors.format_exception e
+           |> Json.js_dq_escape_string in
+         let error_json =
+           "{ \"error\": \"" ^ formatted_exn ^ "\"}" in
+         Lwt.return
+           ("text/plain", Utility.base64encode (error_json)) in
+
+    let handle_html_error e =
+      let mime_type = "text/html; charset=utf-8" in
+      match e with
+       | Aborted r -> Lwt.return r
+       | Failure msg as e ->
+          prerr_endline msg;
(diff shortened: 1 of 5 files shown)


core/webif.ml
```diff
@@ -160,14 +160,36 @@ struct
...
-      (function
-       | Aborted r -> Lwt.return r
-       | Failure msg as e ->
-          prerr_endline msg;
-          Lwt.return ("text/html; charset=utf-8", error_page (Errors.format_exception_html e))
-       | exc -> Lwt.return ("text/html; charset=utf-8", error_page (Errors.format_exception_html exc)))
-    >>= fun (content_type, content) ->
-    response_printer [("Content-type", content_type)] content
+      (fun e ->
+        if (is_ajax_call cgi_args) then
```
jamescheney: This function/ if-then-else in the middle of the >>= pipeline makes it more difficult to follow what's going on 0 roughly, do X, (if something then do Y1 else Y2); do Z.

Morally I feel like the middle (error handling) part should be two functions (that could maybe be hoisted earlier and called things like `handle_ajax_error` and `handle_regular_error`) and we choose which one based on is_ajax_call which does not depend on e.

SimonJF:
Good suggestion, was this what you had in mind?

jamescheney:
Yes, that looks fine.  (I would have been OK with leaving the if-then-else in the pipeline, but that's just yak shaving.)

Issue #330, opened 2018-03-08
Hear and hardcoded arrow syntax


dhil:
Any function which has the hear effect must also have the wild effect.

However, the wild arrow syntax is hard-coded together with the syntactic sugar for the hear effect in the parser, meaning that the following correct type (and program) is rejected
```
> sig f : () {:a,wild}-> () fun f() { () }
*** Parse error: <stdin>:1

  sig f : () {:a,wild}-> () fun f() { () };
```
In order to make it parse successfully one must use the wild arrow syntax, i.e.
```
> sig f : () {:a}~> () fun f() { () }
f = fun : () {:_}~> ()
```
... or spell out the hear effect, i.e.
```
> sig f : () {hear:a,wild}-> () fun f() { () };
f = fun : () {:_}~> ()
```

jamescheney:
Closed by #739

PR #739, opened 2019-08-30
one-line fix to #330

jamescheney:
I was just reading #330 randomly, and wondered whether this one-line fix suffices.  It does allow Daniel's example to be parsed:

```
links> sig f : () {:a,wild}-> () fun f() { () };
f = fun : () {:_}~> ()
```
but perhaps there are other negative consequences which are the reason this issue has been hanging around for over a year.  @dhil?


jamescheney:
This seems unproblematic, at least as far as existing tests are concerned.  (the MacOS tests failed the first time for random-looking reasons.)  Will merge Monday if there are no objections by then.

dhil:
> but perhaps there are other negative consequences which are the reason this issue has been hanging around for over a year.  @dhil?

I don't think there are any negative consequences.

jamescheney:
Closes #330

Issue #734, opened 2019-08-29
Should type variables be unique?


frank-emrich:
A question that has been bothering me for a while (and came up again in #693) is whether or not we require that each type variable may only be bound once in a program?

As a further refinement, we would then have to decide whether type variables bound by \forall may overlap with those bound by \Lambda. More concretely, in
```let f : forall a. a -> a = /\ a. fun (x:a) -> x```
do we enforce that the type variable `a` bound by forall coincides with the one bound by \Lambda, do we accept it (without requiring it), or do we forbid it?
In the the second and third case, the following program would be a valid alternative:
```let f : forall a. a -> a = /\ c. fun (x:c) -> x```

I'm using some explicitly typed pseudo-syntax here, but the question arises both for the frontend and the backend.

Currently, I frequently see cases like 
```let f : forall a. a -> a = /\ a. fun (x:a) -> x```,
where the quantifiers in the annotations coincide with the `tyvars` in the function, so the third option would probably require some work.

On the other hand, I have seen several cases where IR type-checking failed because in the IR program, the type variables bound in annotations were mixed up with those used in the `tyvars`.

Finally, looking at `instantiate.ml`, it seems like this code was written with the assumption in mind that all type variables are unique. Applying the substitution [b/a] to
(forall a. a -> a) -> Int would yield (forall a. b -> b) -> Int, which is bad.

I simply feel like clarifying this will stop people from making different assumptions about this and we could then encode it as a checker pass between transformations.

**Comment by slindley** [comment]
We currently make no assumption that different bound type variables have distinct names. What you describe sounds like a bug in `Instantiate` that should be fixed... but I don't know if it ever manifests in practice.

**Comment by jstolarek** [comment]
> We currently make no assumption that different bound type variables have distinct names. 

I think we sort of do. In the process of fixing #693 I discovered that checking whether quantifiers escape their scope (#678) might not work correctly if a same type variable is bound multiple times. Consider this lengthy snippet:

```
(Sugartypes.Fun
   { Sugartypes.fun_binder = ("and", forall 380::Row.([Bool]) ~380~> Bool);
     fun_linearity = CommonTypes.DeclaredLinearity.Unl;
     fun_definition =
     ([380::Row],
      ([[(Sugartypes.Pattern.Variable ("l", [Bool]))]],
       (Sugartypes.Block
          ([],
           (Sugartypes.FnAppl (
              (Sugartypes.TAppl ((Sugartypes.FreezeVar "all"),
[...]

PR #740, opened 2019-08-30
Support shadowing of type variables in instantiate.ml

frank-emrich:
Following the discussion in #734, this is a step towards supporting shadowing of type variables better.
Concretely, `instantiate.ml` would previously behave erroneously:
The instantiation [Int / a] on 
``` (forall a. a) -> a```
would result in 
``` (forall a. Int) -> Int```
rather than
``` (forall a. a) -> Int```

because  `instantiate.ml` ignored the fact that `ForAll binds type variables.

The diff looks scarier than it should just because the indentation level of most lines changed, make sure to hide whitespace-only changes.

### DIFF
--- a/core/instantiate.ml
+++ b/core/instantiate.ml
@@ -28,147 +28,173 @@ type instantiation_maps = (datatype IntMap.t * row IntMap.t * field_spec IntMap.
 *)
 
 let instantiates : instantiation_maps -> (datatype -> datatype) * (row -> row) * (field_spec -> field_spec) =
-  fun (tenv, renv, penv) ->
-    let rec inst : inst_env -> datatype -> datatype = fun rec_env datatype ->
-      let rec_type_env, rec_row_env = rec_env in
-        match datatype with
-          | `Not_typed -> raise (internal_error "`Not_typed' passed to `instantiate'")
-          | `Primitive _  -> datatype
-          | `MetaTypeVar point ->
-              let t = Unionfind.find point in
-                begin
-                  match t with
-                    | `Var (var, _, _) ->
-                        if IntMap.mem var tenv then
-                          IntMap.find var tenv
-                        else
-                          datatype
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: core/instantiate.ml [hunk 1]
```diff
@@ -28,147 +28,173 @@ type instantiation_maps = (datatype IntMap.t * row IntMap.t * field_spec IntMap.
...
+                        datatype
+                  | `Recursive (var, t) ->
+                      Debug.if_set (show_recursion) (fun () -> "rec (instantiate)1: " ^(string_of_int var));
+
+                      if IntMap.mem var rec_type_env then
+                        (`MetaTypeVar (IntMap.find var rec_type_env))
+                      else
+                        begin
+                          let var' = Types.fresh_raw_variable () in
+                          let point' = Unionfind.fresh (`Var (var', (lin_any, res_any), `Flexible)) in
```
jstolarek: `Types.fresh_row_variable (lin_any, res_any)`

frank-emrich: Interesting, that doesn't quite work. `Types.fresh_row_variable` creates a `row meta_row_var_basis`, whereas we need a `typ meta_type_var_basis` here. They differ in that the former also has `Closed` as a constructor. And `Types.fresh_type_variable` wraps the resulting `Var` in `MetaTypeVar`, so you get a type, not just `typ meta_type_var_basis`, so that doesn't quite work, either.

I'll just leave it as it was, rather than adding another helper for creating fresh variables... there are already too many of them anyway.

jstolarek: Ugh. I should sit down and clean up this mess one day. Leave it as is for now.

frank-emrich: I imagine that in most cases, that behavior is actually what we want. So it's just the naming that's slightly inconsistent, but not a big deal. And after the big refactoring of types, this distinction between row and type vars will be gone anyways.

PR #741, opened 2019-09-02
Move printing and free variable collection out of types.ml

frank-emrich:
This PR attempts to reduce the size of `types.ml` by ~900 lines by moving the pretty printing and free type variable collection out of it.

EDIT: See my comment https://github.com/links-lang/links/pull/741#issuecomment-533875410 for an updated description of the changes


dhil:
I would flavour the following layout:
- `printTypes.ml`: Pretty-printing of types, depends on `types.ml`
- `types.ml`: Syntax of types.

I think utility functions should be moved to `typeUtils.ml` (I think that would be a big, mundane change, so maybe it should be done gradually).

These are core modules which we frequently have look-up or change things in, therefore I think it is useful to keep them together. I don't mind a monolithic module like `typeUtils.ml` if its content are coherent.

frank-emrich:
> I would flavour the following layout:

Unfortunately, that isn't quite possible. Whichever module exports our definitions of dataypes must also export corresponding show_* and pp_* functions for those types. This is because the auto-generated printing functions (e.g., for the IR datatypes) rely on those functions being accessible in the module defining the corresponding types. The difficulty now lies in the fact that we currently overwrite all those show_ functions in order to respect a setting indicating whether or not types should be pretty-printed or not.
Those overwritten versions of the show_* functions must therefore
a) be defined in the same module that exports the dataype definitions and
b) depends on the pretty-printing code
this is why in my current proposal, there is `TypesBase` (containing the actual type definitions) and `Types` (exporting `TypesBase` and the printing stuff, yielding a module with a signature very similar to the original `types.ml`)

The only alternative would be to annotate all type definitions (e.g., of the IR types) to delegate directly to custom printing functions. However, that's very tedious, because we would have to provide a custom printer for each constructor that somwhere contains a type from the Types module.


> I think utility functions should be moved to typeUtils.ml (I think that would be a big, mundane change, so maybe it should be done gradually).

I agree that we can collect many of those construction and deconstruction functions in one module, and `TypeUtils` is a good candidate. If we try to split what is currently `types.ml` into too many smaller files, people never know where to look for something.
I'm just interested in identifying large chunks of independent code that can be moved out of either `Types`. The only candidate other than the pretty printing and collection of free type variables I have in mind are the type visitor classes.

dhil:
> Unfortunately, that isn't quite possible. Whichever module exports our definitions of dataypes must also export corresponding show_* and pp_* functions for those types. This is because the auto-generated printing functions (e.g., for the IR datatypes) rely on those functions being accessible in the module defining the corresponding types.

I think it is fine that auto-generated / derived functions co-exists with the syntax. Instead of overriding the `show` or `pp` functions, I think we should define separate `pretty` functions for pretty-printing.

> [...] I have in mind are the type visitor classes.

It may make sense to move them into their `traversals` module in the same spirit of `sugarTraversals`.

frank-emrich:
> Instead of overriding the show or pp functions, I think we should define separate pretty functions for pretty-printing.

The problem is that this interacts badly with generating other printers. Essentially, in every OCaml type definition we would have to annotate every constructor using one of the types from Types with an attribute saying "please use this custom printing function". For example in case of the IR, this yields quite a lot of noise.

frank-emrich:
I've now found a way to get closer to a design where we don't have to distinguish between `Types` and `TypesBase`

In the new design, `Types` is similar to the original version, minus the printing and free variable detection code. Also, we do not generate the pretty printers for types in `Types` anymore. Instead, the module `TypePrinter` does the following:

- It contains the original, hand-written pretty printing code
- Using `ppx_import`, it re-states the type definitions of the types defined in `Types` and generates pretty printers for them, followed by adding the necessary de-cycling for recursive types
- The first two printing options are exported using a common interface, in addition to a third version that respects the setting `print_types_pretty`
- Finally, for each of the 2+1 printing methods above, `TypePrinter` provides modules that are identical to `Types`, but include the necessary printing functions. This is handy in situations where we define large datatypes that use types from `Types`. In order to generate pretty-printers for these newly defined types, we would normally have to annotate each constructor using a type from `Types` with a special attribute, pointing out where to find a printing function for it (namely, in `TypePrinter`). However, these three re-exports of `Types` allow us to just temporarily the normal `Types` module with a version of Types with printers built-in. See `Ir` and `Sugartypes` for modules where we do this.

core/types/types.ml
```diff
@@ -1207,7 +1116,7 @@ let rec dual_type : var_map -> datatype -> datatype =
       (* Still, we might hope to find a way of preserving 'dual
          aliases' in order to simplify the pretty-printing of types... *)
       | `Alias (_, t) -> dt t
-      | t -> raise (Invalid_argument ("Attempt to dualise non-session type: " ^ show_datatype t))
+      | _t -> raise (Invalid_argument ("Attempt to dualise non-session type: " (*^  show_datatype t *)))
```
jamescheney: Maybe this function should go into typeUtils since it needs to print a datatype?

core/var.ml
```diff
@@ -2,6 +2,8 @@
 
 open Utility
 
+(* module Types = TypePrinter.BySettingPrintableTypes *)
```
jamescheney: ??

core/types/types.ml
frank-emrich: Following Daniel's comment before: Yes, that's where we should eventually collect most of the functions that are currently in `Types`. However, the dualization functions are part of a big blob of recursive functions, including the normalization code, too. That blob is then used by other stuff in `Types`, like subtyping. Thus, trying to move that particular function involves moving a big chunk of `Types` over to `TypeUtils`, which should probably not be part of this PR.

core/var.ml
frank-emrich: thanks, removed that

core/types/typePrinter.ml
```diff
@@ -0,0 +1,928 @@
...
+
+  let string_of_datatype = Pretty.pol_string_of_datatype
+  let string_of_row = Pretty.pol_string_of_row
+  let string_of_field_spec = Pretty.pol_string_of_presence
+  let string_of_type_arg = Pretty.pol_string_of_type_arg
+  let string_of_tycon_spec = Pretty.pol_string_of_tycon_spec
+  let string_of_row_var = Pretty.pol_string_of_row_var
+end
+
+module BySetting =
```
frank-emrich: I just realized that the previous definition (using first-class modules) for defining `BySetting` was too eager: The old version would not have updated the behavior if the user changes `print_types_pretty` via the CLI.

However, the new version of `BySetting` is very repetitive. Any ideas for a nicer solution?

core/types/types.ml
jamescheney: OK, makes sense.  (I was looking at this on a phone so didn't see the context.)

core/types/types.mli
```diff
@@ -1,6 +1,11 @@
+
+
```
jstolarek: Starting a file with empty lines feels weird.

core/unify.ml
```diff
@@ -3,6 +3,9 @@ open Utility
 open Types
 open Typevarcheck
 
+
+module TP = TypePrinter.Pretty
```
jstolarek: Why not `TypePrinter.BySetting`?

core/generalise.ml
```diff
@@ -2,6 +2,8 @@ open Utility
 open Types
 open CommonTypes
 
+module TP = TypePrinter.Pretty
```
jstolarek: Why not `TypePrinter.BySetting`?

core/instantiate.ml
```diff
@@ -4,6 +4,9 @@ open Types
 
 exception ArityMismatch of (int * int)
 
+module TP = TypePrinter.Pretty
```
jstolarek: Why not `TypePrinter.BySetting`?

core/typeUtils.ml
```diff
@@ -2,6 +2,8 @@ open CommonTypes
 open Utility
 open Types
 
+module TP = TypePrinter.Pretty
```
jstolarek: Why not `TypePrinter.BySetting`?

core/types/typePrinter.mli
```diff
@@ -0,0 +1,92 @@
...
+    -> tycon_spec
+    -> string
+end
+
+(* The modules below provide modules equivalent to the vanilla
+  Types module, but with added printing functions.
+  You can use them if you want to define a dataype that uses
+  types from Types and you want printing functions derived for
+  your new types.
+  Just shadow Types with one of the following modules then.
```
jstolarek: I had to read the more detailed description in https://github.com/links-lang/links/pull/741#issuecomment-533875410 to actually understand what this means. Perhaps give a bot more details here or provide an example?

frank-emrich:
Let's close this for now and do #762 first. `ppx_import` causes very weird error messages once something goes wrong, which is a bit of a maintainabilty nightmare. Once #762 is through I may have another look at whether or not splitting `Types` into some smaller modules makes sense.

PR #742, opened 2019-09-04
Add log10 function

SimonJF:
This simple patch adds the `log10` function, found in both the OCaml and JS libraries, to Links.


SimonJF:
CI failure is transient.

dhil:
Well, we could also define `log10` in the prelude using the change of base formula for logarithms.

SimonJF:
I considered this, but given that `log10` is defined in both standard libraries, I don't think there's any harm in patching through.

PR #743, opened 2019-09-09
Use Yojson for JSON serialisation, fix #650

### LINKED ISSUE #650: Mid-sized lists in `jsonState` cause a stack overflow on Chrome
Say we have a web page which tries to send a list mid-sized list (a quick binary search on sizes places it at around 1200 elements) to the client:

```
fun mainPage() {
  var xs = [1..1200];
  var _ = spawnClient {
    print(intToString(hd(xs)))
  };
  page
    <html><h1>Yup</h1></html>
}

fun main() {
  addRoute("/", fun(_) { mainPage() });
  serveWebsockets();
  servePages()
}
main()
```

then the page will crash on Chrome (my version: Chromium Version 74.0.3729.108 (Official Build) Arch Linux (64-bit)) with

```
Uncaught RangeError: Maximum call stack size exceeded
```

Recall that Links lists are represented as JS objects with `head` and `tail` fields on the client, rather than a native JS array for reasons of efficiency.

Note: I don't think it's anything to do with _traversing_ the JSON state (indeed, commenting out the functions doing something with the JSON state causes the same error) -- merely that Chrome doesn't seem to be able to handle objects with that depth.

SimonJF:
#650 was happening because the JavaScript parser was unable to cope with deeply-nested JS object literals. There was some discussion in #650 about JSON parsers being better -- they are, and solve this issue. So, this patch switches us to dumping the initial JSON as a string rather than an object.

Now, this isn't as easy as you might think. In fact, I got so depressed fighting with quoting / non-quoting rules that I tore out our JSON serialisation code and decided to use Yojson instead.

Examples all tested and working.


jamescheney:
Any chance of being able to get rid of jsonparse/jsonlex while you are at it?  It looks like jsonparse could be factored into generic JSON parsing followed by decoding the encoded Links values into (or websocket requests).

(On the other hand, I have also had idle thoughts about adding JSON quasiquotes/literals, but that would require adding to parser.mly)

SimonJF:
Sure, that should be possible. I was thinking of doing it as a separate patch, but I think you're right that it would make sense to do it all in one go.

core/page.ml
```diff
@@ -160,7 +161,7 @@ module Make_RealPage (C : JS_PAGE_COMPILER) (G : JS_CODEGEN) = struct
       ~cgi_env
       ~body:printed_code
       ~html:(Value.string_of_xml ~close_tags:true bs)
-      ~head:(script_tag welcome_msg ^ "\n" ^ script_tag (C.primitive_bindings) ^ "\n" ^ script_tag("  var _jsonState = " ^ state_string ^ "\n" ^ init_vars)
+      ~head:(script_tag welcome_msg ^ "\n" ^ script_tag (C.primitive_bindings) ^ "\n" ^ script_tag("  var _jsonState = " ^ escaped_state_string ^ "\n" ^ init_vars)
```
dhil: I think we should make `_jsonState` a function, because global variables never get garbage collected. I suspect the state can grow quite large.

lib/js/jslib.js
```diff
@@ -2371,12 +2371,13 @@ function _start(page) {
 }
 
 function _startRealPage() {
-  var state = LINKS.resolveJsonState(_jsonState);
+  const parsedState = JSON.parse(_jsonState);
```
dhil: Call `_jsonState()` here instead.

core/page.ml
dhil: On second thought. I may not matter in this case, as it we are only going to have one global variable. The big string would need to be allocated one way or another anyways.

lib/js/jslib.js
dhil: Nevermind.

SimonJF:
Thanks, everyone, for the reviews and comments. I'm nearly there with ripping out `jsonlex` and `jsonparse`, so let's hold off on merging until then.

SimonJF:
OK, that should be `jsonlex` and `jsonparse` removed now. I've tried this out on the examples (including MVU, sessions, and GtoPdb) and all seems in order -- so this is ready for people to take a look if they wish.

SimonJF:
(To think -- I thought fixing #650 would be a one-line fix :) )

PR #744, opened 2019-09-09
Compile session exceptions using deep handlers

SimonJF:
There is an arbitrary choice as to whether to compile session exceptions
using deep or shallow handlers. I switched this to Shallow at some
point, and since then it seems that shallow handlers no longer compile
on the client.

(Unfortunately, this means that session exceptions are broken in 0.9).
This patch switches them back to deep.

### DIFF
--- a/core/desugarSessionExceptions.ml
+++ b/core/desugarSessionExceptions.ml
@@ -122,7 +122,7 @@ object (o : 'self_type)
           (inner_effects, try_dt, outer_effects, otherwise_dt) in
 
         let hndl_desc = {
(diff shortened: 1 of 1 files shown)


dhil:
> [...] it seems that shallow handlers no longer compile on the client.

I don't think the compilation of shallow handlers ever made it into master. Thus I believe it worked previously by sheer accident. The compilation of shallow handlers exist in a branch on my fork of Links. I think the better choice would be to compile session exception handlers as shallow handlers, but for now it is fine to use deep handlers.

PR #748, opened 2019-09-11
Print an error message if database driver loading fails

### LINKED ISSUE #746: Friendlier error messages for driver location failures
I just ran Links without the `postgres` driver installed and got:
 
```
***: Error: Links_core.DatabaseDriver.LocateFailure("postgresql") 
```

This should be friendlier.

dhil:
This patch slightly improves error messages when Links fails to load a
database driver. Beforehand a raw exception would be printed to
stderr. With this patch a standard error message, conforming with the
style of messages in `errors.ml`, is printed.

For example, before Links would print `LocateFailure "postgresql"`
when it failed to locate the `postgresql` driver, now it will print

```
***: Error: Cannot locate database driver 'postgresql'
```

Resolves #746

### DIFF
--- a/core/databaseDriver.ml
+++ b/core/databaseDriver.ml
@@ -1,10 +1,5 @@
 open Utility
 
-exception LocateFailure of string
-exception IllformedPluginDescription of string
-exception DependencyLoadFailure of string * Dynlink.error
-exception LoadFailure of string * Dynlink.error
-
 let driver
   = Settings.(option "database_driver"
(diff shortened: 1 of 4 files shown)


SimonJF:
Didn't get to say this before it was merged -- but thanks for picking this up, it's appreciated!

dhil:
My pleasure! 

I thought the patch was uncontroversial, so I just went ahead with it.

PR #752, opened 2019-09-16
Iterative resolveServerValue in jslib.js

SimonJF:
Before, resolveServerValue was a non-tail-recursive function. This was
problematic for larger JSON values.

Since the ordering of the resolution does not seem to matter, this patch
changes the function to an iterative queue-based algorithm. This means
we can have much better scalability (and indeed this makes large initial
state dumps for my GtoPdb application work).


SimonJF:
I don't think the traversal order matters here. However, when I got tripped up by a bug (and before I realised what it was), I did write an order-respecting algorithm: https://github.com/SimonJF/links/commit/93123e997952c32588ac8ee6d52ca4a48b637f41.

lib/js/jslib.js
```diff
@@ -787,32 +787,39 @@ const LINKS = new (function() {
...
+          // Queue child object
+          queue.unshift(obj[i]);
 
-          obj[i] = f;
-        } else if (obj[i].key) {
-          obj[i].key = _lookupMobileKey(state, obj[i].key)
+          // Resolve this object
+          if (obj[i].func) {
+            const f = (!TYPES.isObject(obj[i].environment)) ?
+              eval(obj[i].func) :
```
dhil: I have always been curious about this particular code fragment. Since closure conversion, I would expect every function to be available at the top-level. So I would believe something less powerful than `eval` ought to do it, say: `window[function_name](arguments)`.

Issue #756, opened 2019-09-17
Numerical ranges ignored in `for` comprehensions


SimonJF:
Numerical ranges are ignored in `for` comprehensions:

```
links> query [4] { for (x <- [1,2,3,4,5,6]) [(num=x)] };;
compiling to IR
compiled IR
Finishing process MAIN
[(num = 1), (num = 2), (num = 3), (num = 4), (num = 5), (num = 6)] : [(num:Int)]
```

jamescheney:
This is probably because `range` is only ever used to generate `LIMIT` clauses in the SQL query. Should be easy to fix but overlaps with #757.

jamescheney:
We are correctly detecting if ranges appear on nested queries, but only at runtime.  We still fail to perform them on non-nested queries that can be executed in-memory like the example above.

We should be detecting when this happens and manually calling take/drop in that case.  We could also do this to support ranges for nested queries though this could be very inefficient.

jamescheney:
(See this comment https://github.com/links-lang/links/pull/757/files#r325809521)

PR #757, opened 2019-09-17
Support flat and shredded queries simultaneously (Fix #538)

### LINKED ISSUE #538: Support flat and shredded queries simultaneously
Right now we support either flat queries and nested queries, depending on settings.  In the REPL one can switch between these options by enabling/disabling some settings.  But within a given expression (or non REPL program) we can only access one style of query.

Since the flat and nested query handlers each support some features that the other one doesn't, and it is reasonable to want to use (e.g.) both ordering/limit/count and nested query results in different parts of the same program, I propose that we enable both via (for now) two different `query` keywords (or by adding some attributes/parameters to the existing `query` keyword). One suggested design is as follows:

- `query [params] { for ... }`  will use the flat query compiler, and the results are statically required to be flat.   I.e. existing Links programs that don't use shredding continue to work as usual.
- `nested_query { for ...}` will use shredding, the results are not statically checked to be flat (and so there may be runtime failures if the return type is not a nested relation).  Existing Links programs that do use shredding would need to be changed to use `nested_query` but this should be a simple search-and-replace, and there aren't many test programs that use shredding.

Though longer-term, I would prefer to abstract over different possible ways of handling queries (so that both `query` and `nested_query` are definable constructs rather than builtin keywords) I don't have a design for this yet.  For now, I think this should be straightforward by adding a keyword and refactoring existing code.  We would also have to change the CI configuration to handle shredding in the new way.  This might make life slightly easier in that we could use the same configuration for shredding and plain queries, unlike now.

SimonJF:
This patch allows the specification of whether a query should be treated as flat or shredded on a per-query basis. The patch extends query syntax as follows:

```
query [range] policy {
  ...
}
```
where `policy` is either `plain`, `nested`, or omitted. If `plain` is used, then the query will evaluated using the default evaluator, and if `nested` is used, then the query will be shredded. If the policy is omitted, then the `shredding` setting is used to decide, as before.

This allows us to have applications which use shredded queries by default, but also have queries which use ranges, ordering, or regexes.

### DIFF
--- a/core/evalir.ml
+++ b/core/evalir.ml
@@ -702,10 +702,10 @@ struct
        let evaluator =
          let open QueryPolicy in
          match policy with
-           | Plain -> `Plain
+           | Flat -> `Flat
            | Nested -> `Nested
            | Default ->
-               if Settings.get Database.shredding then `Nested else `Plain in
+               if Settings.get Database.shredding then `Nested else `Flat in
 
        let evaluate_standard () =
          match EvalQuery.compile env (range, e) with
@@ -757,7 +757,7 @@ struct
 
        begin
          match evaluator with
-           | `Plain -> evaluate_standard ()
+           | `Flat -> evaluate_standard ()
            | `Nested -> evaluate_nested ()
        end
     | InsertRows (source, rows) ->
--- a/core/evalir.ml
+++ b/core/evalir.ml
@@ -693,15 +693,41 @@ struct
             in apply_cont cont env (`Table ((db, params), Value.unbox_string name, unboxed_keys, row))
           | _ -> eval_error "Error evaluating table handle"
       end
-    | Query (range, e, _t) ->
+    | Query (range, policy, e, _t) ->
        let range =
          match range with
          | None -> None
          | Some (limit, offset) ->
             Some (Value.unbox_int (value env limit), Value.unbox_int (value env offset)) in
-       if Settings.get Database.shredding then
-         begin
-           if range != None then eval_error "Range is not supported for nested queries";
+       let evaluator =
+         let open QueryPolicy in
+         match policy with
+           | Plain -> `Plain
+           | Nested -> `Nested
+           | Default ->
+               if Settings.get Database.shredding then `Nested else `Plain in
+
+       let evaluate_standard () =
+         match EvalQuery.compile env (range, e) with
+           | None -> computation env cont e
+           | Some (db, q, t) ->
+               let q = Sql.string_of_query db range q in
+               let (fieldMap, _, _), _ =
+               Types.unwrap_row(TypeUtils.extract_row t) in
+               let fields =
+               StringMap.fold
+                   (fun name t fields ->
+                     match t with
+                     | `Present t -> (name, t)::fields
+                     | `Absent -> assert false
+                     | `Var _ -> assert false)
+                   fieldMap
+                   []
+               in
+               apply_cont cont env (Database.execute_select fields q db) in
+
+       let evaluate_nested () =
+         if range != None then eval_error "Range is not supported for nested queries";
            match EvalNestedQuery.compile_shredded env e with
            | None -> computation env cont e
            | Some (db, p) ->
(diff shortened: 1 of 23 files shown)


core/lexer.mll
```diff
@@ -198,6 +198,7 @@ let keywords = [
  "mu"       , MU;
  "mutual"   , MUTUAL;
  "native"   , NATIVE;
+ "nested"   , NESTED;
```
jamescheney: It's slightly unfortunate that we need to add new keywords.  I wonder whether it would be sensible to (eventually) move to these parameters (and the range parameters) being first-class values (which could in principle be provided dynamically).  On the other hand, if we were checking that `plain` queries have flat result types (which we currently aren't if nesting is enabled) then having the query policy information be static would be simpler.

core/typeSugar.ml
```diff
@@ -2950,10 +2951,21 @@ let rec type_check : context -> phrase -> phrase * Types.datatype * usagemap =
...
+            let () =
+              match evaluator with
+                | `Nested -> ()
+                | `Plain  ->
+                     let shape =
+                       Types.make_list_type
+                         (`Record (StringMap.empty,
+                            Types.fresh_row_variable (lin_any, res_base), false)) in
+                     unify ~handle:Gripers.query_base_row (pos_and_typ p, no_pos shape) in
+            Query (range, policy, erase p, Some (typ p)), typ p, merge_usages [range_usages; usages p]
```
jamescheney: There are a couple of dark corners here that could potentially be improved:

* Plain queries allow `orderby` and correctly handle converting regex matching to `LIKE`, whereas shredding doesn't try to handle the former and fails to handle the latter

* Plain queries also allow ranges.  For nested queries, these are just ignored.

* The query keyword does double duty as "here is a query; run it" and "this code is expected to be tame".  It's unclear to me what happens, or what should happen, if `query plain {...}` is nested inside `query nested {...}` or vice versa.  If this should never be possible then I don't see how to stop this using the current type system.

I'm not sure it makes sense to try to handle these all statically, but it would be good to at least check these corner cases and see if the resulting error messages are comprehensible, and if not improve them.

core/lexer.mll
dhil: To avoid new keywords in the short term, we can exploit the fact that `nested` and `plain` are valid variable names, then in the three `QUERY` rules we should expect an optional `VAR` rather than a `query_policy`. In the semantic action we can coerce the name into a `QueryPolicy` or throw a syntax error.

It is a hack, but it may be acceptable since there are only three, simple, and closely located instances of it.

SimonJF: I think @dhil's suggestion is a good one -- and that gives us the variable names back. I think `flat` is also a better descriptor, so I'll do the associated renaming.

core/typeSugar.ml
SimonJF: Thanks, @jamescheney: 

> Plain queries allow orderby and correctly handle converting regex matching to LIKE, whereas shredding doesn't try to handle the former and fails to handle the latter

> Plain queries also allow ranges. For nested queries, these are just ignored.

I think that given this additional static information, it's perfectly possible to make these forms either a parse or a type error. I will do this.

> The query keyword does double duty as "here is a query; run it" and "this code is expected to be tame". It's unclear to me what happens, or what should happen, if query plain {...} is nested inside query nested {...} or vice versa. If this should never be possible then I don't see how to stop this using the current type system.

Yes, I was wondering about this. I had some thoughts about tracking it in the effect system, but it felt fiddly. I think the best way forward for this patch is to pass the top-level policy into the normaliser and error out if a block with a different policy is detected -- this will rule out possible strange behaviours or bad error messages.

jamescheney: At the moment, I think everything allowed in a nested query is also allowed in a flat/ordered query, so adding an effect that flags use of orderby might work, but would entail many changes.

Longer term, I have some hope that we can unify the two approaches so that we don't need two policies. But I've been saying that for years now, and having a workable approach that supports different query backends would be a good idea anyway. (For example to handle queries for different data models).

Another strategy could be to distinguish between the "run" behavior of the query keyword and the "quote" behavior, which might be useful for other things. Then either dynamically error if we try to run a query with unhandled features (like in F#) or index the quoted code type to keep track of what language subset is used.  This feels a bit related to algebraic effects and handlers.

SimonJF:
Thanks for the review, @jamescheney. I have addressed your concerns as follows:

  1. `plain` and `new` are no longer keywords. Like with linearity and subkinds in the parser, I take @dhil's suggestion and parse the query policy as a variable. An invalid policy results in an error, while we can still use `flat` and `nested` elsewhere.

  2. Mismatching policy annotations are now checked dynamically. So, for example, `query nested { query plain { 0} }` is now an error.

jamescheney:
OK, thanks!

SimonJF:
Will merge tomorrow unless there are any objections.

core/evalir.ml
```diff
@@ -693,15 +693,41 @@ struct
...
+         let open QueryPolicy in
+         match policy with
+           | Flat -> `Flat
+           | Nested -> `Nested
+           | Default ->
+               if Settings.get Database.shredding then `Nested else `Flat in
+
+       let evaluate_standard () =
+         match EvalQuery.compile env (range, e) with
+           | None -> computation env cont e
```
jamescheney: To fix #756 this line needs to be changed to drop the first `offset` results and take the next `limit` results, where `range = (limit,offset)`.  One way to do this could be to explicitly insert calls to Links's take and drop functions, alternatively, adjust the continuation `cont` to inspect the returned (list) value and run `Utility.take` and `Utility.drop` on it.

SimonJF:
Thanks for having a look at this. I'll add the required functionality in to fix #756.

SimonJF:
CI's glitched on the Mac build, but everything is fine, so merging.

Issue #637, opened 2019-06-13
Types refactor


jamescheney:
Collapse the type structure into a single monolithic OCaml data type.

dhil:
I have written up an initial draft proposal for the new type structure, c.f. https://gist.github.com/dhil/97711b8c42f1a2fed91b3b6a1398783d, intended to serve as a starting point for discussion.

dhil:
For this work I have created a new branch on this repository. It is called types-refactor.

jamescheney:
Was this closed by #859?  If not can someone add a description of what remains to be done?

slindley:
The first phase of this was done and merged in #859 , but I've left the issue open because there's more to do. What remains includes tidying up (e.g. by being more consistent about the use of unboxed rows in certain places) and removing the need for the invariant that `Var`, `Recursive`, and `Closed` constructors can only occur inside `Meta` constructors.

jamescheney:
At today's meeting we agreed @slindley would add a checklist for remaining requirements for completion of this issue for 0.9.2.

jamescheney:
This is milestoned 0.9.2, which we hoped to get released around now.  Is it likely to be resolved in the next week or so, or should we postpone it?

(0.9.2 is needed as a bugfix release due to #807 )

dhil:
It is not crucial for the bugfix release. Postponed.

Issue #533, opened 2019-03-26
Use inline records instead of nested tuples in AST nodes


SimonJF:
In much of the Links AST, we use long tuples to describe information attached to nodes. Take `TableLit` as a (random) example:

```
  | TableLit         of phrase * (Datatype.with_pos * (Types.datatype *
                           Types.datatype * Types.datatype) option) *
                          (name * fieldconstraint list) list * phrase * phrase
```

It would be more descriptive to have these as named inline records:
var accessory_protein = table "accessory_protein" with (object_id: Int, full_name: String) from db;

```
  | TableLit         of { tlit_table_name: phrase
                             ; tlit_record: (Datatype.with_pos * (Types.datatype *
                           Types.datatype * Types.datatype) option)
                             ; tlit_field_constraints: (name * fieldconstraint list) list
                             ; tlit_table_keys: phrase
                             ; tlit_database: phrase
                             }
```

jstolarek:
One random thought regarding this. I have mentioned in one of the comments how nicely inline records work in GHC source code. However, Haskell allows to match record fields based on their positions, e.g.:

```Haskell
data F = A { a :: Int, b :: String    }
       | B { a :: Int, c :: Maybe Int }

foo :: F -> Int
foo (A x _) = x
foo (B _ (Just y)) = y
foo (B z Nothing ) = z
```
As far as I know OCaml does not have such a feature. We can rely on field punning to some extent, but some pattern matches are still likely to become very verbose, e.g.:

```OCaml
foo (TableLit { tlit_table_name; tlit_record = Some (pos, ty1, ty2, ty3)
              ; tlit_field_constraints = []; tlit_table_keys; _ }) = ...
```
instead of:
```OCaml
foo (TableLit ( name; Some (pos, ty1, ty2, ty3); []; keys; _ )) = ...
```

jstolarek:
I wonder whether we could make a first step towards this by having someone who actually understand the whole AST to write comments in the code detailing what each of the fields does, e.g.:

```OCaml
  | Conditional      of phrase * (* if condition *)
                        phrase * (* then branch  *)
                        phrase   (* else branch  *)
```

PR #758, opened 2019-09-18
Various datatype refactorings

jstolarek:
This is preliminary work for #637. I've taken various small datatypes and placed them in their own modules in `CommonTypes`. Notably:

  * there is now a `Name` module with `Name.t = string`, rather than having several modules defining `type name = string`.
  * I separated `quantifier` from large mutually recursive definition of types in `Types` and placed it, together with corresponding projection functions, in `CommonTypes.Quantifier` module. I am somewhat tempted to have that module renamed since what it really represents are kinded type variables, not just quantifiers. OTOH in most places where this module is used it really does handle quantifiers, so the name `Quantifier` is more informative than just `TyVar`. Thoughts?
  * I implemented @dhil's earlier suggestion to have `add_quantifiers` as part of `TypeVarSet` module.
  * sugartypes could use a similar cleanup, but that's a story for another day. Similarly for IR definitions. In particular I am not a huge fan of having a dozen of aliases to already existing types:
```
type scope = Var.Scope.t
  [@@deriving show]
(* term variables *)
type var = Var.var
  [@@deriving show,eq,yojson]
type var_info = Var.var_info
  [@@deriving show]
type binder = Var.binder
  [@@deriving show]

(* type variables *)
type tyvar = Quantifier.t
  [@@deriving show]
type tyarg = Types.type_arg
  [@@deriving show]

type name_set = Utility.stringset
  [@@deriving show]
type 'a name_map = 'a Utility.stringmap
  [@@deriving show]

type 'a var_map = 'a Utility.intmap
  [@@deriving show]

type language = string
  [@@deriving show]

type location = CommonTypes.Location.t
  [@@deriving show]
```
I understand this makes code shorter but it also makes it more obscure. But again, this is a separate story.

### DIFF
--- a/core/commonTypes.ml
+++ b/core/commonTypes.ml
@@ -134,6 +134,13 @@ module Kind = struct
     [@@deriving eq,show]
 end
 
+module Quantifier = struct
+  type t = int * Kind.t
+    [@@deriving show]
+
+  let to_string = Format.asprintf "%a" pp
+end
+
 module Location = struct
   type t = Client | Server | Native | Unknown
     [@@deriving show]
--- a/core/commonTypes.ml
+++ b/core/commonTypes.ml
@@ -138,7 +138,22 @@ module Quantifier = struct
   type t = int * Kind.t
     [@@deriving show]
 
+  let to_var = function
+    | (var, _) -> var
+
+  let to_kind : t -> Kind.t = function
+    | (_, k) -> k
+
+  let to_primary_kind : t -> PrimaryKind.t = function
+    | (_, (pk, _)) -> pk
+
+  let to_subkind : t -> Subkind.t = function
+    | (_, (_, sk)) -> sk
+
   let to_string = Format.asprintf "%a" pp
+
+  let eq : t -> t -> bool = fun lvar rvar ->
+    to_var lvar = to_var rvar
 end
 
 module Location = struct
--- a/core/commonTypes.ml
+++ b/core/commonTypes.ml
@@ -102,16 +102,14 @@ let res_mono    = Restriction.Mono
 let res_session = Restriction.Session
 let res_effect  = Restriction.Effect
 
-type subkind = Linearity.t * Restriction.t
+module Subkind = struct
+  type t = Linearity.t * Restriction.t
     [@@deriving eq,show]
 
-let min_subkind (ll, lr) (rl, rr) =
-  match Restriction.min lr rr with
-  | Some r -> Some (Linearity.min ll rl, r)
-  | None -> None
-
-let string_of_subkind (lin, res) =
-  Printf.sprintf "(%s,%s)" (Linearity.to_string lin) (Restriction.to_string res)
+  let to_string (lin, res) =
+    Printf.sprintf "(%s,%s)" (Linearity.to_string   lin)
+                             (Restriction.to_string res)
+end
 
 module PrimaryKind = struct
   type t =
--- a/core/commonTypes.ml
+++ b/core/commonTypes.ml
@@ -167,8 +167,10 @@ let loc_server  = Location.Server
 let loc_native  = Location.Native
 let loc_unknown = Location.Unknown
 
-type freedom = [`Flexible | `Rigid]
+module Freedom = struct
+  type t = [`Flexible | `Rigid]
     [@@deriving show]
+end
 
 module Name = struct
   type t = string
--- a/core/commonTypes.ml
+++ b/core/commonTypes.ml
@@ -167,6 +167,11 @@ let loc_unknown = Location.Unknown
 type freedom = [`Flexible | `Rigid]
     [@@deriving show]
 
+module Name = struct
+  type t = string
+    [@@deriving show]
+end
+
 module Primitive = struct
   type t = Bool | Int | Char | Float | XmlItem | DB | String
     [@@deriving show]
--- a/core/commonTypes.ml
+++ b/core/commonTypes.ml
@@ -129,6 +129,11 @@ let pk_type     = PrimaryKind.Type
 let pk_row      = PrimaryKind.Row
 let pk_presence = PrimaryKind.Presence
 
+module Kind = struct
+  type t = PrimaryKind.t * Subkind.t
+    [@@deriving eq,show]
+end
+
 module Location = struct
   type t = Client | Server | Native | Unknown
     [@@deriving show]
--- a/core/closures.ml
+++ b/core/closures.ml
@@ -55,7 +55,7 @@ struct
               zs in
             List.fold_left
               (fun o q ->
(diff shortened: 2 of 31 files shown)


core/commonTypes.ml
```diff
@@ -131,6 +129,33 @@ let pk_type     = PrimaryKind.Type
...
+module Kind = struct
+  type t = PrimaryKind.t * Subkind.t
+    [@@deriving eq,show]
+end
+
+module Quantifier = struct
+  type t = int * Kind.t
+    [@@deriving show]
+
+  let to_var = function
```
jamescheney: I'd be inclined to call these `get_var`, `get_kind`, ... or just `var` or `kind`, since to me `to_*` connotes translation to an equivalent (e.g. serialized) format rather than projecting/getting a part of a data structure.  Alternatively, it seems like `Quantifier.t` is mostly being used as a record `{var:int, kind: Kind.t}`, where `Kind.t` itself is a record `{primary:PrimaryKind.t, subkind: Subkind.t}`.  Defining them this way would obviate the need for all of these projection functions, and allow factoring `to_primary_kind` as `.kind.primary` etc.

jstolarek: Indeed, good points.

>  to me to_* connotes translation to an equivalent (e.g. serialized) format rather than projecting/getting a part of a data structure.

I feel similarly. However, in the previous refactorings we have followed the `to_` convention (c.f. `Binder.to_name`, `Binder.to_type`) and I wanted to stay consistent.

jamescheney: OK, I hadn't seen that.  Maybe agreeing on recordifying more things would be good (cf. #533 )

I still think `to_*` is potentially confusing since it suggests an information-preserving conversion or serialization (dual to `from_*`) rather than a projection.  But not a blocker to merging.

jstolarek:
Continuing discussion from in-line comments:

> I still think to_* is potentially confusing since it suggests an information-preserving conversion or serialization (dual to from_*) rather than a projection.

And I still agree. [Here](https://github.com/links-lang/links/pull/498/files/0581656581a9282edb93289edb55825ea593efcc#r260238304) is an earlier discussion on this. @dhil, what's your opinion here? Similarly to @jamescheney I find names like `to_kind`, `to_var`, etc. confusing in this case and would prefer `get_kind`, `get_var`.

Also, I'll try with the record approach and so how that looks.

jstolarek:
Ok, so I did a quick experiment with turning:

```
module Kind = struct
  type t = PrimaryKind.t * Subkind.t
    [@@deriving eq,show]
end
```
into:
```
module Kind = struct
  type t = { primary : PrimaryKind.t
           ; subkind : Subkind.t
           } [@@deriving eq,show]

  let make (primary : PrimaryKind.t) (subkind : Subkind.t) : t =
    { primary; subkind }
end
```
Sadly, this results in tons of extra verbosity and I really don't like the result. For example pattern match of the form `Some (PrimaryKind.Row, (_, Restriction.Effect))` now becomes `Some {Kind.primary=PrimaryKind.Row; Kind.subkind=(_, Restriction.Effect)}` (and something even worse if we turn subkind into a record as well), which I do not see as an improvement.

dhil:
> 
> 
> Continuing discussion from in-line comments:
> 
> > I still think to_* is potentially confusing since it suggests an information-preserving conversion or serialization (dual to from_*) rather than a projection.
> 
> And I still agree. [Here](https://github.com/links-lang/links/pull/498/files/0581656581a9282edb93289edb55825ea593efcc#r260238304) is an earlier discussion on this. @dhil, what's your opinion here? Similarly to @jamescheney I find names like `to_kind`, `to_var`, etc. confusing in this case and would prefer `get_kind`, `get_var`.

I'm fine with that. But I think it deviates from the "standard" OCaml style, which is usually to have `x_of_y`, `M.to_y`, `M.of_x`, etc.

jamescheney:
Ok. I'm not familiar with this convention being used for "getters". (I.e. "to_fst", "to_snd" etc.) But I guess I'm not a real ocaml programmer.

Given the widespread use of pattern matching on pairs I'm persuaded that recordifying is not a good idea.

jstolarek:
I am in favour of keeping to conventions, so if it is customary to use these kind of names in OCaml then I would keep them. Even if they feel weird to me.

BTW @dhil, when you say "standard OCaml style" do you mean style that is used in OCaml standard libraries? I wonder whether creators of large libraries (e.g. Jane Street) also follow these conventions or create their own?

jstolarek:
Merged now.

dhil:
> BTW @dhil, when you say "standard OCaml style" do you mean style that is used in OCaml standard libraries? 

Yes, exactly!

> I wonder whether creators of large libraries (e.g. Jane Street) also follow these conventions or create their own?

I'm not sure. Maybe @rudihorn can weigh in on this?

rudihorn:
To me I think this sums it up quite well:
> I still think to_* is potentially confusing since it suggests an information-preserving conversion or serialization (dual to from_*) rather than a projection. 

The list for example defines the functions `hd` and `tl`, which are neither called `to_hd` / `to_tl` nor `to_a` and `to_alist`. I don't recall explicitly having to use naming conventions `of_x` / `to_x` for every function going from one type to another and this seems quite unhelpful to me since your function name is losing "domain knowledge" about what it is explicitly doing. To me the `of_x` / `to_x` notation also indicates a type `x`, and so `to_fst` does not make much sense unless `fst` is a type. I also would just suggest using the name `kind` for a function that returns the kind rather than `get_kind` as "kind" is a noun and not a verb, implying you aren't doing anything and just "getting something" in any case. The `get_` prefix ends up being verbose and I don't really think I've seen this in use anywhere. (Some exceptions apply: The ocaml `Result` module uses `get_ok`, while the ocaml Core library uses `ok`, but the Ocaml `Result` module defines `ok` as a constructor, which Core calls `return` (and `error` becomes `fail`). Arguably `get_ok` is not simply a projection however as it can also fail. The constructor issue I find is mainly a sad consequence of data constructors not implicitly being treated as functions: `5 |> Result.Ok` does not work but should IMO (and does actually work in F#).

At the end of the day, I don't think this is strictly enforced anywhere and the only clear cases of using `to_x` and `of_x` seem to be cases of information-preserving conversions: `to_seq`, `of_alist`, `to_param` (https://ocaml.janestreet.com/ocaml-core/latest/doc/core_kernel/Core_kernel/Command/Spec/index.html) etc.

I would suggest: If you are doing something, use a verb. If you are just getting something (i.e. projecting), just name it by what you are getting. If you are taking something and changes its representation to something containing the same information (i.e. I have X but need this to be a Y) use `to_x` / `of_y`. If you have a constructor / projection clash, find some sensible resolve on a case-by-case basis, e.g. `box_int` / `unbox_int`.

PR #759, opened 2019-09-18
Fix small error in PID serialisation

SimonJF:


jamescheney:
Hmmm.  It's too bad OCaml doesn't have some kind of constant literal inlining, i.e. one can't do 

```
let x = "foo"
match y with (x,z) -> ...
```
but instead seemingly have to do
```
match y with ("foo",z) -->
```
or 
```
match y with (x',z) when x' = x -> ...
```

Then these constants could be hoisted and defined in one place.

PR #760, opened 2019-09-18
fix 755

### LINKED ISSUE #755: Range notation in `for` comprehensions results in nontermination
Range notation (`..`) in a query block causes nontermination:

```
links> query { for (x <- [1..6]) [(num=x)] };;
compiling to IR
compiled IR
```

jamescheney:
Makes it an error to do 
```
links> query {for (x <- [1..3] ) [(num=x)]};
<stdin>:1: Type error: Ranges are wild
    `(wild|a)'
but the currently allowed effects are
    `()'
In expression: [1..3].
```

Let's see if this breaks anything, now back to what I should have been doing.

If successful, closes #755.

### DIFF
--- a/core/typeSugar.ml
+++ b/core/typeSugar.ml
@@ -907,6 +907,7 @@ end
 
     let range_bound ~pos ~t1:_l ~t2:(_, _t) ~error:_ =
       die pos "Range bounds must be integers."
+
     let range_wild ~pos ~t1:(_, lt) ~t2:(_, rt) ~error:_ =
       build_tyvar_names [lt; rt];
       let ppr_rt = show_type rt in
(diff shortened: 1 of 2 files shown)


jamescheney:
Incidentally, it looks like the following code appears 11 times verbatim now, when checking that a given effect is currently allowed.
```
Types.make_singleton_open_row ("wild", `Present Types.unit_type) default_effect_subkind
```
Copy-pasting this made me feel dirty.  Making this a utility function somewhere might a good idea.  (It needs to be a function so that we get a fresh row variable per call to `make_singleton_open_row`.)  If no one stops me I might refactor it.

jamescheney:
Thanks, sorry I forgot the tests...

PR #763, opened 2019-09-25
ounit fix

frank-emrich:
ounit has switched to dune, which slightly changes how it is packaged.

Therefore, we need to use `ounit` and `ounit.advanced` rathern than `oUnit` in our dune files for *the unit tests*. Unfortunately, this breaks the unit tests for people using older versions of ounit. However, I assume that the unit tests are rarely run manually by people, as it is not part of `make tests`.

### DIFF
--- a/tests/unit/dune
+++ b/tests/unit/dune
@@ -2,5 +2,5 @@
   (name unitTests)
   (modes native)
   (flags :standard -safe-string -dtypes -w Ae-44-45-60 -g)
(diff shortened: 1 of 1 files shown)


SimonJF:
I think that even though it's not in `make tests`, the unit tests are run by the CI.

Thanks for doing this: let's merge if CI passes.

frank-emrich:
> the unit tests are run by the CI.

Indeed! That's why this is something we should fix quickly.

PR #764, opened 2019-09-27
MVU: Commands

SimonJF:
Commands in Elm allow side-effecting computations to be performed inside the MVU event loop. A particularly important side-effect is to spawn a computation which evaluates asynchronously, returning a message when it is complete.

This patch adds commands to the Links MVU library. It's existed in various branches of mine for a while, but I'm finding it necessary for the GtoPdb curation tool so thought it might be nice to have it in master. I'd say the primary use right now is to allow RPCs and MVU to coexist nicely.

Key changes: 
  - add a new type, `Command(Message)` which describes a computation which will produce a message of type `Message`
  - revise most general type of `updt` function from `(Message, Model) ~> Model` to `(Message, Model) ~> (Model, Command(Message))`

You can see the gist of the new functionality in the `examples/mvu/commands.links` example, which spawns an expensive computation asynchronously and awaits the result.

### DIFF
--- a/examples/mvu/commands.links
+++ b/examples/mvu/commands.links
@@ -1,0 +1,65 @@
+import MvuAttrs;
+import MvuHTML;
+import MvuCommands;
+import Mvu;
+
+typename Model = [| NotStarted | Waiting | Result:Int |];
+
+typename Message = [| StartComputation | DeliverResult:Int |];
+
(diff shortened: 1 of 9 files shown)


SimonJF:
As this will only really affect me at the moment and the CI is fine, are people OK with this being merged?

jamescheney:
Sounds good to me.

PR #769, opened 2019-10-16
[Patch 1/2] Code serialisation refactor

### LINKED ISSUE #766: Implement handler serialisation
Currently handlers are non-serialisable. However, they should be serialisable. I think handler serialisation will require every handler to be explicitly named in the IR. 

One potential problem with regard to serialisation might be the dynamic creation of resumptions.

dhil:
This patch refactors the serialisation logic for runtime values. Notably it moves from an intensional approach to an extensional approach which works by endowing `Continuation`, `Continuation.Frame`, `Value`, and `Value.Env` with a compression scheme that only makes use of the interface of the said modules.

The primary motivation for this change is actually to break the `xmlParser -> Value -> DesugarDatatypes` dependency because my subsequent patch introduces a new indirect dependency `DesugarDatatypes -> Value`, thus without this change there would be a cyclic dependency.

This patch does not fix #766 which remains entirely broken. I will leave that for a future patch (hoping to do some research with @SimonJF related to this issue).

The patch does also introduce a "regression" in `Webserver` as it now using some rather low-level knowledge to perform JSON serialisation. I have made no attempt to fix this, as I think that compilation unit is due a substantial refactoring.

(By transitivity this patch is prerequisite for #603)

**Merge strategy**: Please merge with rebase.


core/value.mli
```diff
@@ -55,16 +55,20 @@ and xml = xmlitem list
...
-| `Table of table
-| `Float of float
-| `Int of int
-| `XML of xmlitem
-| `String of string ]
-[@@deriving show]
+  | primitive_value_basis
+  | `Database of (database * string)
+  | `Table of table ]
+  [@@deriving show]
```
dhil: @jstolarek expressed concern regarding this change. I will argue it changes very little or nothing, because this is exactly the implementation in `value.ml`. This minor change reduces the amount of necessary boilerplate code in the `Compressible_Value` module in `serialisation.ml`.

frank-emrich:
I like the fact that this untangles serialisation from values. However, I don't understand the finer details and it's hard to say what was just moved from `Value` to `Serialisation` and what are more fundamental changes. I'm fine with this going in unless someone else volunteers to have an in-depth look at it.

Issue #629, opened 2019-06-10
What should be `bin`, and what should be `lib`?


SimonJF:
Tom Davey and I are working on a notebook interface (see Jupyter notebooks, for example) for Links. The current design is to write a wrapper application which uses Links as a library, and we can get quite far using `Links_core`.

However, right now, a fair chunk of functionality that would be quite useful is in `bin`, in particular `driver.ml`. Duplicating this code is fine, but a pity.

I wonder whether a better design might be to refactor somewhat in order to have nice things like Prelude loading callable as a library, rather than hardwired into the executable.

jstolarek:
> I wonder whether a better design might be to refactor somewhat in order to have nice things like Prelude loading callable as a library, rather than hardwired into the executable.

That sounds like a right thing to do.

Aside: I always find `bin` name confusing since typically this is a directory where build artefacts are placed. `src` would be a better name IMO.

SimonJF:
> Aside: I always find bin name confusing since typically this is a directory where build artefacts are placed. src would be a better name IMO.

I agree, this confused me in exactly the same way. I think it's a `dune` convention, however.

dhil:
> I wonder whether a better design might be to refactor somewhat in order to have nice things like Prelude loading callable as a library, rather than hardwired into the executable.

For #603 I have implemented a new loader, which contains a bootstrapping phase that loads the Prelude and the built-in `lib.ml` units. Ultimately the loader returns a compilation environment containing the loaded/processed compilation units. I was planning on keeping the wiring in `driver.ml`, but I can move that into `core` instead if need be.

Another thing is, that `core` would really be named `lib`, and our current lib should be named `runtime`.

SimonJF:
@dhil: 

Okay, that seems sensible; a lot of this does seem tied up with the new module system.

Some of the wiring does make sense to be executable-specific -- but I think it needs to be the absolute minimum that is specific to the Links executable. Looking at `driver.ml`, it would seem that a fair bit (if not all of it, in fact) could be moved to `core`. I think it's really lovely to be able to use the compiler as a library and this is a huge advantage of your previous work porting the system to `dune`. What's missing right now is the high-level glue allowing someone to take a raw string and get back a value and some updated environments.

Some of this will depend on how `prelude` is handled in the new module system. I'm imagining your aim is to have it as just a plain old module, as opposed to handled using the current specialised logic?

[...]

Issue #415, opened 2018-11-06
Isolae REPL code from links.ml and reorganize file processing


jamescheney:
As per discussion today relating to modules (#394)

jamescheney:
Is this closed by #427?

frank-emrich:
The "reorganize file processing" part is not. That will come as part of the module system

jamescheney:
- [X] Isolate REPL code from links.ml (DONE #427)
- [ ] Reorganize file processing

PR #770, opened 2019-10-16
[Patch 2/2] Structured transforms with compilation state passing

dhil:
This patch is an overhaul of the whole frontend and backend transformation pipelines and it brings structure to our ad-hoc transformation framework. 

Frontend transformations now state explicitly whether they are "untyped" or "typeability preserving". It also fixes a crucial bug in the frontend pipeline which would always return the initial inferred type of its given program rather than the type of the transformed program. 

The most substantial change is that the compilation state (known as `Context`) is now threaded through all of the frontend and backend transforms. The motivation for this change is to pave the way for the introduction of compilation units, which need to be accessible in all of the various transformation passes. Currently, the `Context` assumes whole program compilation and exposes fairly low-level details such as the various environments. This is will change with the upcoming compilation unit patch, where querying will happen through an interface rather than directly on the environments.

I have also managed to eliminate many of the distinct handling of `sentence` and `program`. The REPL and whole program compiler now uses the same pipeline, which I hope will help towards making the REPL more robust (past changes have often ignored the REPL).

I have also simplified `Loader`, `Links`, `Repl`, and `Driver` compilation units. In particular, the latter unit now explicitly identifies the various phases of the Links compiler, making it easy to run each phase separately. Hopefully this will help toward #415 and #629.

This patch depends on patch #769. This patch is a prerequisite for #603. 

**Merge strategy**: Please use rebase when merging this PR.

### DIFF
--- a/bin/driver.ml
+++ b/bin/driver.ml
@@ -181,12 +181,6 @@ module Phases = struct
     Webserver.init (valenv, nenv, tenv) globals ffi_files;
     Evaluate.run result
 
-  (* Error handling wrapper for [whole_program] above. *)
-  let whole_program context filename =
-    Errors.display
-      ~default:(fun _ -> exit 1)
-      (lazy (whole_program context filename))
-
   let evaluate_string : Context.t -> string -> (Context.t * Types.datatype * Value.t)
     = fun initial_context source_code ->
     Parse.string initial_context source_code
@@ -195,10 +189,4 @@ module Phases = struct
     |> Compile.IR.run
     |> Transform.run
     |> Evaluate.run
-
-  (* Error handling wrapper for [evaluate_string] above. *)
-  let evaluate_string context source_code =
-    Errors.display
-      ~default:(fun _ -> exit 1)
-      (lazy (evaluate_string context source_code))
 end
--- a/bin/driver.ml
+++ b/bin/driver.ml
@@ -3,20 +3,6 @@ open Webserver
 open Performance
 open Utility
 
-module Eval = Evalir.Eval(Webserver)
-module Webif = Webif.WebIf(Webserver)
-
-type evaluation_env =   Value.env (* maps int identifiers to their values *)
-                      * Ir.var Env.String.t (* map string identifiers to int identifiers *)
-                      * Types.typing_environment (* typing info, using string identifiers *)
-
-type evaluation_result =
-  {
-    result_env : evaluation_env;
-    result_value : Value.t;
-    result_type : Types.datatype
-  }
-
 (** Name of the file containing the prelude code. *)
 let prelude_file =
   let prelude_dir =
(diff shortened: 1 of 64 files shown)


bin/driver.ml
```diff
@@ -36,140 +22,183 @@ let typecheck_only
...
-   die_on_exception load_prelude_inner ()
+      (Context.({ context with value_environment = valenv' }), datatype, v)
+  end
+
+  module Interactive = struct
+    type result =
+      { context: Context.t;
+        sentence: [ `Definitions of Ir.binding list
+                  | `Expression of Types.datatype * Value.t
+                  | `Directive of string * string list ] }
```
jstolarek: Can we avoid a polymorphic variant here?

core/context.ml
```diff
@@ -0,0 +1,26 @@
+(* Compiler (functional) context/state. *)
+(* TODO localise environment within their associated compilation units. *)
```
jstolarek: Maybe it is worth creating a ticket and pointing at it in the comment instead of having multiple "TODO remove" comments?

core/frontend.ml
```diff
@@ -32,231 +40,267 @@ let show_post_frontend_ast
...
+       ; "inners", (module DesugarInners)
+       ; "session_execeptions", only_if
+                                  Basicsettings.Sessions.exceptions_enabled
+                                  (module DesugarSessionExceptions)
+       ; "processes", (module DesugarProcesses)
+       ; "fors", (module DesugarFors)
+       ; "regexes", (module DesugarRegexes)
+       ; "formlets", (module DesugarFormlets)
+       ; "pages", (module DesugarPages)
+       ; "funs", (module DesugarFuns) |]
```
jstolarek: I wonder if it would be possible to have modules define their own name/description, rather than having them defined independently?

core/irCheck.ml
```diff
@@ -1291,10 +1291,19 @@ struct
...
   end
 
-  let program tyenv p =
-    let lazy_check =
-      lazy (let p, _, _ = (checker tyenv)#computation p in p) in
-   handle_ir_type_error lazy_check p (SProg p)
+  (* let program tyenv p =
+   *   let lazy_check =
+   *     lazy (let p, _, _ = (checker tyenv)#computation p in p) in
+   *  handle_ir_type_error lazy_check p (SProg p) *)
```
jstolarek: Any reason to keep this around?

core/irTraversals.ml
```diff
@@ -897,19 +890,19 @@ module CheckForCycles =
...
-    let program tyenv p =
-      let p, _, _ = (ir_type_mod_visitor tyenv check_cycles)#program p in
-      p
-
-    let bindings tyenv bs =
-      let bs, _ = (ir_type_mod_visitor tyenv check_cycles)#bindings bs in
-      bs
+    (* let program tyenv p =
+     *   let p, _, _ = (ir_type_mod_visitor tyenv check_cycles)#program p in
+     *   p *)
```
jstolarek: Any reason to keep this around?

core/irTraversals.ml
```diff
@@ -927,19 +920,18 @@ module ElimBodiesFromMetaTypeVars =
...
-      p
-
-    let bindings tyenv bs =
-      let bs, _ = (ir_type_mod_visitor tyenv elim_bodies)#bindings bs in
-      bs
-
-  end
+    (* let program tyenv p =
+     *   let p, _, _ = (ir_type_mod_visitor tyenv elim_bodies)#program p in
+     *   p *)
```
jstolarek: Any reason to keep this around?

core/value.mli
```diff
@@ -55,16 +55,20 @@ and xml = xmlitem list
...
-| `Table of table
-| `Float of float
-| `Int of int
-| `XML of xmlitem
-| `String of string ]
-[@@deriving show]
+  | primitive_value_basis
+  | `Database of (database * string)
+  | `Table of table ]
+  [@@deriving show]
```
jstolarek: I don't like this. Can we keep it as a single data type? When we proceed with removing polymorphic invariants - and I hope this will happen at some point - this refactoring will only complicate things.

core/context.ml
dhil: There is alright a ticket for this one. That's #603. This is mostly a comment for me.

bin/driver.ml
dhil: Yes, easily. I mostly left whatever I could unchanged.

core/frontend.ml
dhil: Yes, that is a good idea, and should be easy to implement.

core/irCheck.ml
dhil: No. Good catch.

core/irTraversals.ml
dhil: No.

core/irTraversals.ml
dhil: No.

core/value.mli
dhil: This is introduced by #769.

core/env.mli
```diff
@@ -53,6 +53,8 @@ sig
   val filter : (name -> 'a -> bool) -> 'a t -> 'a t
 
   val filter_map : (name -> 'a -> 'b option) -> 'a t -> 'b t
+
+  val complement : 'a t -> 'a t -> 'a t
```
frank-emrich: Comment please, because the types don't tell you which parameter does what

bin/driver.ml
```diff
@@ -36,140 +22,171 @@ let typecheck_only
...
+           in
+           let (context', datatype, value) = Evaluate.run result in
+           context', `Expression (datatype, value)
+        | Sugartypes.Directive d ->
+           context, `Directive d
+      in
+      { context; sentence }
+  end
+
+  let dump_lib : out_channel -> unit
```
frank-emrich: It seems like `dump_lib`, `initialise`, `whole_program`, and `evaluate_string` shouldn't be part of `Phases`, but go into `Driver` directly

core/frontend.ml
```diff
@@ -32,231 +40,275 @@ let show_post_frontend_ast
...
+        then T.Untyped.sentence state sentence
+        else Identity.Untyped.sentence state sentence
+    end
+  end
+
+  let only_if : bool Settings.setting -> (module Untyped.S) -> transformer
+    = fun setting (module T) ->
+    (module Conditional(struct include T let condition () = Settings.get setting end))
+
+  module Collect_FFI_Files : Untyped.S = struct
```
frank-emrich: I know that some of this logic was present in `Frontend` before, but maybe this should go somewhere else now

core/transform.ml
```diff
@@ -0,0 +1,195 @@
...
+      include INTERFACE with type state := state and type 'a result := 'a result
+    end
+  end
+end
+
+module Untyped : UNTYPED = struct
+  type 'a transformer = 'a -> 'a
+
+  type state = Context.t
+  type 'a result = Result of { program: 'a;
```
frank-emrich: Any particular reason for this being variant type rather than just a record? It seems to work without the data constructor.

core/frontend.ml
dhil: Perhaps. I suggest we keep it there for now. In the future we might want to make the compilation pipeline more dynamic, allowing transforms to be plugged in.

core/transform.ml
dhil: In an earlier iteration it was relevant to allow multiple fields named `program` and `context`. However, I suppose they are unique in the current iteration. I suppose we can get rid of it.

PR #771, opened 2019-10-22
Type variable refreshing and subsitution in types and terms 

frank-emrich:
This PR implements two things:

1. Extend the `fold_map` visitor in `core/sugartraversals.ml` to visit semantics datatypes (i.e., those from `Types`) properly. The `map` visitor already did this, but `fold_map` didn't. This step is a prerequisite for the next step.

2. Implements `RenameTypeVariables` which substitutes type variables in sugar ASTs (namely, in all types found anywhere in an AST). Based on a flag, it can also refresh all quantifiers it comes across in the process.

This is a step towards @jstolarek and my efforts to improve `TypeSugar`, but the functionality of this PR is not used, yet.

### DIFF
--- a/core/sugarTraversals.ml
+++ b/core/sugarTraversals.ml
@@ -2335,7 +2335,7 @@ class fold_map =
       | `Type t -> let o,t = o#typ t in o, `Type t
       | `Row r -> let o, r = o#type_row r in o, `Row r
       | `Presence p -> let o, p =o#type_field_spec p in o, `Presence p
-    method tyvar : Types.quantifier -> ('self_type * Types.quantifier) =
+    method tyvar : Quantifier.t -> ('self_type * Quantifier.t) =
       o#unknown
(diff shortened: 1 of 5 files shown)


core/sugarTraversals.ml
```diff
@@ -1506,6 +1506,9 @@ class fold_map =
     method bool : bool -> ('self_type * bool) =
       function | false -> (o, false) | true -> (o, true)
 
+
+
+
```
jstolarek: Bogus empty lines.

jstolarek:
@frank-emrich, I know I'm the one who insisted on having a PR but now I think it would be easier if I could grab your code, merge it into may branch, and then include it in my own PR for #693. There are some things I need to tweak in your code and juggling code between two branches is inconvenient. So my suggestion would be to close this PR and then we can discuss this code once I'm done with my changes. Does that sound ok?

frank-emrich:
No problem, that sounds reasonable

PR #772, opened 2019-10-22
Distinction between nominal and structural 'names'.

dhil:
The module `Name` in `commonTypes.ml` is used to name both nominal and
structural things. It is conceptually wrong to mix the two. This patch
introduces a new module `Label` in `commonTypes.ml` which is used to
name structural things, reserving `Name` to be used for nominal
things. Although this change is only cosmetic, it becomes important as
we move towards hygienic (nominal) name handling in the frontend.

Addendum: I have added `Typename` in `commonTypes.ml` for type-level names and `Typevar` for type variable names.

**Merge strategy**: Please use rebase when merging this PR.

### DIFF
--- a/core/commonTypes.ml
+++ b/core/commonTypes.ml
@@ -199,6 +199,11 @@ module Name = struct
     [@@deriving show]
 end
 
+module Typename = struct
+  type t = string
+    [@@deriving show]
+end
+
 module Label = struct
   type t = string
     [@@deriving show]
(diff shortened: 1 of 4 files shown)


dhil:
The misconception is prevalent throughout the codebase. There are many more instances than the ones I have fixed here. I am discarding this patch in favour of another that I am working on, in which I have fixed these instances as I have encountered them.

PR #774, opened 2019-10-24
Fix "Unbound type variable" IR errors

### LINKED ISSUE #692: Prelude `and` causes IR typing error
code:
```
fun all(p, l) {
   switch (l) {
     case []    -> true
     case x::xs -> p(x) && all(p, xs)
   }
}

fun and(l) {
  all(fun(x) { x }, l)
}
```
error:
```
***: IR Type Error: Type variable 378 is unbound
occurring in IR binding:
(Ir.Fun
   ((243,
     (`Function ((`Record (({1 => `Present (`Primitive (CommonTypes.Primitive.Bool));
                             },
                            Point @ 0x7f31604dacc8 `Closed, false)),
                  ({},
                   Point @ 0x7f31607c3ba0 `Body (({wild => `Present (
                                                   `Record (({},
                                                             Point @ 0x7f31604dacc8 `Closed,
                                                             false)));
                                                   },
                                                  Point @ 0x7f315fe37cd0 `Var (
                                                  (378,
                                                   (CommonTypes.Linearity.Unl,
                                                    CommonTypes.Restriction.Any),
                                                   `Rigid)),
                                                  false)),
                   false),
                  `Primitive (CommonTypes.Primitive.Bool))),
      "_fun__g1", Var.Scope.Global)),
    ([],
     [(242, (`Primitive (CommonTypes.Primitive.Bool), "", Var.Scope.Local))],
     ([(Ir.Let (
          (241,
           (`Primitive (CommonTypes.Primitive.Bool), "x", Var.Scope.Local)),
          ([], (Ir.Return (Ir.Variable 242)))))
        ],
      (Ir.Return (Ir.Variable 241)))),
    None, CommonTypes.Location.Unknown))
```
config:
```
typecheck_ir=true
prelude=empty.links
print_types_pretty=false
optimise=false
debug=true
hide_fresh_type_vars=false
show_flavours=true
fail_on_ir_type_error=true
```
`all` IR-typechecks fine, it's `and` that's causing trouble.

jstolarek:
Primary goal of this patch is to fix "Unbound type variable" errors in the IR and also restore fix for #534, which for some reason got reverted during FreezeML implementation. This PR is marked as draft - see below for details.

In addition, this patch includes a renamer for type variables in terms (+ fallout changes in `SugarTraversals`) implemented by @frank-emirch (originally as #771). This renamer allows to fix type variable escapes check (#678), which didn't work as expected in some settings.

@frank-emrich: I made various modifications to the interfaces of the renamer but the core renaming code remains unchanged. In particular:
 
  * the renamer is now in `renamer.ml`

  * `replace_quantifiers` is now `build_instantiation_maps` and uses an optional argument for an initial map

  * `freshen_quantifiers` is now `freshen_and_build_instantiation_maps` and relies on `build_instantiation_maps` as well as `initialize.ml` (the latter was slightly refactored)

  * **Question**: previously `unknown` method in refresher would raise an exception instead of being a no-op (default implementation in `super`). Why? This exception really gets raised when using the renamer but using the default `super` implementation behaves as expected.

  * **Question**: instead of `rename_phrase` there is now `rename_phrase`, which is essentially the same except that it calls `funlit` method instead of `phrase` method.  Is there a way to write a polymorphic function that would invoke a method from the `renamer` depending on the type of the argument? Otheriwse we might be forced to write a dozen of entry points, each for a different type of node.

@slindley, during our discussion this week you insisted that renaming of quantifiers and corresponding type variables happens during typechecking. I couldn't figure out how to do that and consequently this patch implements renaming during `DesugarFuns` transformations. The problem that I ran into was that although I can easily rename variables in the definitions I can't rename the use sites and this leads to mismatching effect variables. I'm not sure why that happens, I would expect that these effect variables are instantiated in a normal way and thus their names at call sites don't matter. I need to investigate this further, but if you can offer any insight here I would appreciate it.
  
Fixes #692,#693,#694.

### DIFF
--- a/core/desugarFuns.ml
+++ b/core/desugarFuns.ml
@@ -98,15 +98,23 @@ object (o : 'self_type)
                  then `Lolli (args, mb, rt)
                  else `Function (args, mb, rt))
                argss rt in
+
     let f = gensym ~prefix:"_fun_" () in
-    let (bndr, lin, def, loc) =
+    let (bndr, lin, (_, def), loc) =
       unwrap_def (binder ~ty:ft f, lin, ([], lam), location) in
-    let e = block_node ([with_dummy_pos (Fun { fun_binder = bndr; fun_linearity = lin;
-                                               fun_definition = def; fun_location = loc;
-                                               fun_signature = None;
-                                               fun_frozen = true;
-                                               fun_unsafe_signature = false; })],
-                         with_dummy_pos (FreezeVar f))
+    let (tvs, tyargs), ft =
+      Generalise.generalise (o#get_var_env ()) (Binder.to_type bndr) in
+    let bndr = Binder.set_type bndr ft in
+    let o = o#bind_binder bndr in
+    let e = block_node ([with_dummy_pos
+                           (Fun { fun_binder           = bndr
+                                ; fun_linearity        = lin
+                                ; fun_definition       = (tvs, def)
+                                ; fun_location         = loc
+                                ; fun_signature        = None
+                                ; fun_frozen           = true
+                                ; fun_unsafe_signature = false })],
+                         with_dummy_pos (tappl (FreezeVar f, tyargs)))
     in (o, e, ft)
 
   method! phrasenode : Sugartypes.phrasenode -> ('self_type * Sugartypes.phrasenode * Types.datatype) = function
--- a/core/desugarFuns.ml
+++ b/core/desugarFuns.ml
@@ -102,9 +102,9 @@ object (o : 'self_type)
     let f = gensym ~prefix:"_fun_" () in
     let (bndr, lin, (_, def), loc) =
       unwrap_def (binder ~ty:ft f, lin, ([], lam), location) in
-    let (tvs, tyargs), ft =
+    let (tvs, tyargs), gen_ft =
       Generalise.generalise (o#get_var_env ()) (Binder.to_type bndr) in
-    let bndr = Binder.set_type bndr ft in
+    let bndr = Binder.set_type bndr gen_ft in
     let o = o#bind_binder bndr in
     let e = block_node ([with_dummy_pos
                            (Fun { fun_binder           = bndr
--- a/core/instantiate.ml
+++ b/core/instantiate.ml
@@ -391,6 +391,21 @@ let apply_type : Types.datatype -> Types.type_arg list -> Types.datatype = fun p
   let (t, instantiation_maps) = instantiation_maps_of_type_arguments false pt tyargs in
   instantiate_datatype instantiation_maps t
 
+let build_fresh_quantifiers :
+      Quantifier.t list -> Quantifier.t list * Types.type_arg list = fun qs ->
+  let open PrimaryKind in
+  List.split (List.map (function
+                  | (_, (Type, subkind)) ->
+                     let q, t = Types.fresh_type_quantifier subkind in
+                     q, `Type t
+                  | (_, (Row, subkind)) ->
+                     let q, r = Types.fresh_row_quantifier subkind in
+                     q, `Row r
+                  | (_, (Presence, subkind)) ->
+                     let q, f = Types.fresh_presence_quantifier subkind in
+                     q, `Presence f)
+                qs)
+
 (*
   ensure that t has fresh quantifiers
 *)
@@ -401,23 +416,8 @@ let freshen_quantifiers t =
           match qs with
             | [] -> body
             | qs ->
-               let open PrimaryKind in
-               let qs, tyargs =
-                 List.split
-                   (List.map
-                      (function
-                       | (_, (Type, subkind)) ->
-                          let q, t = Types.fresh_type_quantifier subkind in
-                          q, `Type t
-                       | (_, (Row, subkind)) ->
-                          let q, r = Types.fresh_row_quantifier subkind in
-                          q, `Row r
-                       | (_, (Presence, subkind)) ->
-                          let q, f = Types.fresh_presence_quantifier subkind in
-                          q, `Presence f)
-                      qs)
-                in
-                  `ForAll (qs, apply_type t tyargs)
+               let qs, tyargs = build_fresh_quantifiers qs in
+               `ForAll (qs, apply_type t tyargs)
         end
     | t -> t
 
--- a/core/instantiate.mli
+++ b/core/instantiate.mli
@@ -20,6 +20,7 @@ val recursive_application : string -> Quantifier.t list -> Types.type_arg list -
 (* Given a quantified type and a list of type arguments, create the corresponding instantiation maps *)
 val instantiation_maps_of_type_arguments : bool -> Types.datatype -> Types.type_arg list -> (Types.datatype * instantiation_maps)
 
+val build_fresh_quantifiers : Quantifier.t list -> Quantifier.t list * Types.type_arg list
 val apply_type : Types.datatype -> Types.type_arg list -> Types.datatype
 val freshen_quantifiers : Types.datatype -> Types.datatype
 val replace_quantifiers : Types.datatype -> Quantifier.t list -> Types.datatype
--- a/core/renamer.ml
+++ b/core/renamer.ml
@@ -1,0 +1,298 @@
+(* Rename for type variables in terms *)
+
+open Utility
+open CommonTypes
+open Sugartypes
+
+let internal_error message =
+  Errors.internal_error ~filename:"renamer.ml" ~message
+
+type instantiation_maps =
+  (Types.meta_type_var     IntMap.t *
+   Types.meta_row_var      IntMap.t *
+   Types.meta_presence_var IntMap.t)
+
+let fresh_instantiation_maps : instantiation_maps =
+  (IntMap.empty, IntMap.empty, IntMap.empty)
+
+let add_type_var var t (ty_map, r_map, p_map) =
+  (IntMap.add var t ty_map), r_map, p_map
+
+let add_row_var var r (ty_map, r_map, p_map) =
+  ty_map, (IntMap.add var r r_map), p_map
+
+let add_presence_var var p (ty_map, r_map, p_map) =
+  ty_map, r_map, (IntMap.add var p p_map)
+
+
+let remove_type_var var (ty_map, r_map, p_map) =
+  (IntMap.remove var ty_map), r_map, p_map
+
+let remove_row_var var (ty_map, r_map, p_map) =
+  ty_map, (IntMap.remove var r_map), p_map
+
+let remove_presence_var var (ty_map, r_map, p_map) =
+  ty_map, r_map, (IntMap.remove var p_map)
+
+
+let unwrap_type_point = function
+    | `MetaTypeVar point -> point
+    | _ -> assert false
+
+let unwrap_presence_point = function
+    | `Var point -> point
+    | _ -> assert false
+
+
+(** Build instantiation maps between two lists of quantifiers *)
+let build_instantiation_maps
+     ?(instantiation_maps=fresh_instantiation_maps)
+      (qs_old : Quantifier.t list)
+      (qs_new : Quantifier.t list) : instantiation_maps =
+  assert (List.length qs_old == List.length qs_new);
+  List.fold_right2
+    (fun q_old q_new imaps ->
+      let (var_old, (pk_old, sk_old)) = q_old in
+      let (var_new, (pk_new, sk_new)) = q_new in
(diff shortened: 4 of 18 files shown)


slindley:
The crucial point is that the current goal is to proceed on the basis that there is no guarantee that bound type variables have unique names. A key implication of non-uniqueness is that in the escape check we must freshen the bound variables (in the body of the function) as the escape check relies on checking that these variable do not appear in the environment. So what needs to happen is if a function has a non-empty collection of type variables, then those type variables must be freshened immediately before checking the body of the function.

I presume that the only reason to rename bound variables in DesugarFuns would be to try to ensure uniqueneses of bound type variables - an invariant that we are currently **not** enforcing. In future we may want to revisit the possibility of trying to enforce such an invariant, but that would require analysing the whole code base. For now, that is **not** what we are doing.

jstolarek:
> So what needs to happen is if a function has a non-empty collection of type variables, then those type variables must be freshened immediately before checking the body of the function.

Yes, exactly. The problem is figuring out how to do that correctly - simply renaming tyvars does not seem to work in the typechecker, though it does work when done inside desugarFuns. I'm suspecting this might have something to do with the presence of signatures.

Also, I realize that doing the renaming inside desugarFuns is not what we want. The reason I've put it there for now was to be able to test whether the renamer actually does the job and iron out possible issues.

jstolarek:
I fixed things that didn't work and broke some that did work. Now I need help.

Firstly, the renamer operates on the assumption that if we have a function binding then forall quantifiers in the corresponding type signature (if present) and type variables in a big lambda are identical. This assumption is already made implicitly in other places of the compiler so I don't feel guilty about making it a bit more explicit here.
 
@frank-emrich in his original renamer patch implemented optional freshening on nested quantifiers. I now came to believe that this is a misfeature because it would break the above assumption, i.e. if we came across nested function binding then it would freshen type variables in a forall but it wouldn't rename the big lambda accordingly. This could be implemented correctly if we really need to but I don't think we do. Hence I removed this feature.

Secondly, the original renamer did not handle variable shadowing. For example given a map `[a -> b]` it would rename `forall a. a -> a` to `forall a. b -> b`. I believe that my current implementation handles shadowing correctly, or at least it makes a better attempt at doing so.

At the moment there are two entry points to the renamer: one for renaming non-recursive function definitions and one for renaming recursive ones. Both entry points ensure that type variables in the renamed definition are freshened while all the type variables introduced by nested definitions remain unchanged - cf. shadowing of variables when renaming foralls (`renaming_type_visitor#typ`), value bindings (`renamer#bindingnode`), and function definitions (`renamer#handle_function`). @frank-emrich, does all of the above sound reasonable?

With these changes all the `typed_ir` tests pass - no more "Unbound type variable" errors. At the same time I have managed to spectacularly break code that uses session types. Run `./test-harness tests/typecheck_examples.tests` to see what I mean. For example, running `examples/AIPL14/LectureExamples/calc.links` produces a unification error:

```
Unification error: Couldn't match [+|Add:?(Int).?(Int).!(Int).EndBang,Mul:?(Int).?(Int).!(Int).EndBang|+] against [&||Add:?(Int).?(Int).!(Int).%399::Session,Mul:?(Int).?(Int).!(Int).%399::Session|&]
examples/AIPL14/LectureExamples/calc.links:5: Type error: The cases of an offer should match the type of its patterns, but the pattern
    `s'
has type
    `Calc'
while the subsequent patterns have type
    `[&||Add:?(Int).?(Int).!(Int).%399::Session,Mul:?(Int).?(Int).!(Int).%399::Session|&]'
In expression: offer (s) {
    case Add(s) ->
      var (x,s) = receive(s);
      var (y,s) = receive(s);
      send(x+y,s)
    case Mul(s) ->
      var (x,s) = receive(s);
      var (y,s) = receive(s);
      send(x*y,s)
  }.
```

I have nil experience with session types. Halp!

Also, what would be a canonical way of writing an assertion that two lists are equal?

frank-emrich:
> previously unknown method in refresher would raise an exception instead of being a no-op (default implementation in super). Why? This exception really gets raised when using the renamer but using the default super implementation behaves as expected.

I was under the assumption that before I patched `SugarTraversals.fold_map`, the only thing applied to `unknown` were semantic types (i.e., things from `Types`). As you've noticed that assumption is absolutely wrong, `unknown` is called on all kinds of stuff. The original idea was to detect if there were types somewhere in the AST that I missed when patching  `SugarTraversals.fold_map`. Given the observation above we won't be able to do that.

>  instead of rename_phrase there is now rename_phrase, which is essentially the same except that it calls funlit method instead of phrase method. Is there a way to write a polymorphic function that would invoke a method from the renamer depending on the type of the argument? Otheriwse we might be forced to write a dozen of entry points, each for a different type of node.

Typo here? you mentioned the same function/method name twice.
In general, I don't think there's a nice way of exposing different functions using the visitor easily. But that would only effect the public interface, where a few functions should be sufficient, right? Besides for `phrase` and `[recursive]_function_definition`, are there many forms we would like to refreshen from outside of `Renamer`? 

> Firstly, the renamer operates on the assumption that if we have a function binding then forall quantifiers in the corresponding type signature (if present) and type variables in a big lambda are identical. This assumption is already made implicitly in other places of the compiler so I don't feel guilty about making it a bit more explicit here.

I'm not sure if we should make this assumption. At the moment,  `TypeSugar` seems to establish this property, but I guess the idea of having tyvars in the first place is so that they can differ from the quantifiers in an annotation. Could you elaborate on in which sense the renamer breaks if this assumption is violated?

> Secondly, the original renamer did not handle variable shadowing. For example given a map [a -> b] it would rename forall a. a -> a to forall a. b -> b. I believe that my current implementation handles shadowing correctly, or at least it makes a better attempt at doing so. 

Yes, sorry, I realized this only after creating  #771 but then forgot about it.

> does all of the above sound reasonable?

I'll have a more detailed look at the code tomorrow or Wednesday. I may have an idea what is currently wrong w.r.t. to signatures. I'll also have a look at the session errors you're getting.

jstolarek:
> Typo here? you mentioned the same function/method name twice.

Yes, stupid GitHub did not allow me to edit my post to fix that :-/ Anyway, both entry points are now gone.

> But that would only effect the public interface, where a few functions should be sufficient, right?

Correct. There are two functions at the moment and hopefully we won't need more.

> Could you elaborate on in which sense the renamer breaks if this assumption is violated?

Shadowing will not work correctly. Let's say I am doing a renaming `[a -> b]` and encounter a function definition:
```
let f : forall a. (a) -> b = /\ a. body
```
To shadow the variables I take the tyvars introduced in the big lambda and subtract them from the domain of a substitution, which is then used to rename body, patterns and signature. So in this example I will have an empty substitution and nothing gets renamed. But if the big-lambda binder has a different name than a forall binder:
```
let f : forall a. (a) -> b = /\ c. body
```
then I will have a substitution `[a -> b]` and that will result in an incorrect type signature `forall a. (b) -> b`. In order to make this work correctly we need to shadow variables separately when renaming the body, the signature, and patterns. This is conceptually trivial but requires lots of boilerplate.

At this point I wonder whether a) OCaml is a horrible language or b) I don't know how to write good OCaml code. I mean, in Haskell it would be very easy to write concise code that does this thanks to monads and ability to pass monadic computations as functions:

```Haskell
let sig_ty' <- with_maps (shadow_vars maps forall_vars) (rename_type sig_ty)
```
In OCaml it seems I have to manually save the state and restore it afterwards as well as explicitly pass the self object around:
```Ocaml
let o, old_maps = o#with_vars (shadow_vars maps forall_vars) in
let o, sig_ty' = o#datatype sig_ty in
let o = o#set_maps old_maps
```
Am I doing something wrong or is this the way it has to be in OCaml?

jamescheney:
Doing things in abstraction-leaky-monad style is pretty idiomatic for OCaml (or at least I recall naturally starting to program in this style before I knew Haskell) though it is possible to approximate do notation as is done in some of the Links code using a monad module with an infix bind operator.  Definitely not as nice though.

The other idiomatic OCaml way to do stateful traversals is to use the "OCaml monad", i.e. use references and side effects to collect whatever state.  Then you can write things that look like do-notation (or C) but you lose any benefit of things being pure / explicit.

jstolarek:
Good catch! Thanks! It looks like we have a working code.

@slindley - any comments?

slindley:
> Doing things in abstraction-leaky-monad style is pretty idiomatic for OCaml (or at least I recall naturally starting to program in this style before I knew Haskell) though it is possible to approximate do notation as is done in some of the Links code using a monad module with an infix bind operator. Definitely not as nice though.
> 
> The other idiomatic OCaml way to do stateful traversals is to use the "OCaml monad", i.e. use references and side effects to collect whatever state. Then you can write things that look like do-notation (or C) but you lose any benefit of things being pure / explicit.

This isn't really a language issue, but rather to do with our somewhat crude use of the visitor pattern for generic traversals. I suspect we could make things more robust by implementing more refined traversals that are more aware of variable bindings. I also suspect that François Pottier's visitors library includes facilities to do this kind of thing:

  http://gallium.inria.fr/~fpottier/publis/fpottier-visitors-unchained.pdf

jstolarek:
> I suspect we could make things more robust by implementing more refined traversals that are more aware of variable bindings.

I see. But that would definitely be beyond the scope of this PR. That being said, are there any objections to merging? If not then I'd like to do that early next week.

jstolarek:
Fun fact: the patch is broken again after rebasing onto current master. One of the FreezeML tests breaks because expression `$(fun(x)(y) { y })` is assigned the type ` forall 354,357::Row,355::(Any,Any),356::Row,354,357::Row,355::(Any,Any),356::Row.(354) -357-> (355::Any) -356-> 355::Any` (each quantifier is repeated twice). I believe the reason is the following. In typeSugar we desugar `$` and turn it into a type abstraction, so we roughly have `/\ a b c d. fun(x)(y) {y}` with the type `forall a b c d. t`. Then in desugarFuns we desugar `fun(x)(y) {y}` and compute the type for it to be `forall a b c d. ...`, and that type is nested inside previously computed type - hence repeated quantifiers. I am not surprised to see this. What actually puzzles me more is why this didn't happen before a rebase. I don't see any obvious changes in master that should affect behaviour of these fragments of code.

jstolarek:
> I don't see any obvious changes in master that should affect behaviour of these fragments of code.

I do however see some less obvious candidates for these breakages - namely refactorings of frontend (2c7eb25b91e4aa88b73b6c659f9e2c04d802bee2) and backend (3f15461202ef8a2b7564e2fc7c7414cdab579b75) pipelines that @dhil implemented. If there's an error somewhere with passing the environments it could manifest in incorrect generalisation of type variables. That being said, if there is such an error it would be really surprising if it didn't manifest before.

jstolarek:
After bisecting I have found 2c7eb25b to be the offending commit. I'll continue the investigation tomorrow. @dhil, is there anything obvious that comes to your mind as a possible cause here?

dhil:
> After bisecting I have found 2c7eb25 to be the offending commit. I'll continue the investigation tomorrow. @dhil, is there anything obvious that comes to your mind as a possible cause here?

My guess is that my patch has unveiled a broken transform. There was a bug in the transformation pipeline previously, where the returned type would be the original inferred type rather than the transformed type. Now it returns the transformed type (as it should).

The patch shouldn't have changed the implementations of any of the transforms.

> I do however see some less obvious candidates for these breakages - namely refactorings of frontend (2c7eb25) and backend (3f15461) pipelines that @dhil implemented. If there's an error somewhere with passing the environments it could manifest in incorrect generalisation of type variables.

The environments are manipulated locally now, the places should be `sugartoir.ml` (updates name environment), `frontend.ml` (updates type environment after type checking), `repl.ml` (should only read the environments), and `driver.ml` (updates the "varified" environment).

jstolarek:
Thanks. Hopefully this will help in resolving this problem.

dhil:
On second thought, I did make one change that affects types. The default implementation of `program` normalises the result type (c.f. https://github.com/links-lang/links/blob/master/core/transformSugar.ml#L244).

I suppose this operation could be deferred until printing of types.

jstolarek:
I will be surprised if that is what makes the difference. What should be happening is that when we traverse body of a type abstraction we bind in the environment type variables bound by the abstraction so that we don't generalize over them. Currently this isn't happening (there is even no facility for doing this) and I am not surprised that the tests break. What surprises me is the fact that things worked in the first place.

jstolarek:
Merged as d72e68d156e439caa6440bb07ff67782492b4432

PR #778, opened 2019-11-05
Disable ocamlformat check

dhil:
It seems that the recent upgrade of OCamlformat 0.11 to 0.12 every
file under scrutinty fails.breaks our rule check script.

I can reproduce this problem locally on my machine. This commit is
currently empty to reproduce it on the CI.

### DIFF
--- a/tools/rule-check
+++ b/tools/rule-check
@@ -5,6 +5,19 @@
 #
 
 RULEDIR="tools/rules"
+BLACKLIST=("ocamlformat_check")
+
+function is_blacklisted () {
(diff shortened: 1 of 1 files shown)


dhil:
If the CI fails on this commit, as I expect it will, then I am going to disable the OCamlformat check as it is preventing me from committing to master. I am not willing to spend time debugging it.

dhil:
As expected the CI failed on an empty commit due to the OCamlformat upgrade. I am now disabling the check.

PR #780, opened 2019-11-06
An efficient interface for `Env`.

dhil:
This patch changes the signature of `Env.S` to allow for a more
efficient implementation of environments. For performance reasons it
changes the signature of `bind` from

```ocaml
bind : 'a t -> (name * 'a) -> 'a t
```

to

```ocaml
bind : name -> 'a -> 'a t -> 'a t
```

That is the `bind` now takes all its arguments in curried form. This
reduces the amount of necessary memory allocations by one per
invocation of `bind` (which is called a lot). For aesthetics and
coherence the data container (i.e. the environment of type `'a t`) is
now the last parameter. This aligns with the "standard" functional
APIs (e.g. `Map`) rather than the "standard" imperative APIs which
take the data container as their first argument (e.g. `Hashtbl`).
Moreover, this change also aligns the interface with `Map`, `Set`,
and `Value.Env` in our codebase.

### DIFF
--- a/bin/driver.ml
+++ b/bin/driver.ml
@@ -130,7 +130,7 @@ module Phases = struct
       (fun name var ->
         let (datatype : string) =
           Types.string_of_datatype
-            (Env.String.lookup Lib.typing_env.Types.var_env name)
+            (Env.String.find name Lib.typing_env.Types.var_env)
         in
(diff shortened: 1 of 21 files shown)


dhil:
Further down the road, we can remove `Value.Env` as it ought to be an instance of `Env.S`. However, a little more plumbing is required to make it work.

PR #782, opened 2019-11-07
Structured variable usage tracking.

dhil:
This patch introduces a thin abstraction for variable usage tracking
in the frontend type checker. Usages are now manipulated via a
structured interface `Usage`, as a result it is no longer necessary
(nor possible) to manipulate them on an ad-hoc basis via bare
`StringMap`s, `StringSet`s, and lists.

This code is extracted and adapted from the name hygiene patch which I
have been working on lately. I think this code is useful on its own as
it makes the type checker more robust. Ancedotally, I did find and fix
a couple of linearity bugs in the implementation of effect handlers
and session exceptions, where usage weren't tracked properly.

This patch adds a small compatibility module `Ident` which is
essentially an artefact of my hygienic names patch. I have added it to
make it easier for me to eventually merge my name patch.

### DIFF
--- a/core/typeSugar.ml
+++ b/core/typeSugar.ml
@@ -1623,20 +1623,38 @@ module Usage: sig
   [@ocaml.warning "-32"]
   type t
 
+  (* Empty usage container. *)
   val empty : t
+  (* Adds an entry to a given container. Any preexisting entry with
+     the same identifier will be overwritten. *)
   val add : Ident.t -> int -> t -> t
+  (* Removes an entry from a given container. *)
   val remove : Ident.t -> t -> t
+  (* Tests whether an entry exists. *)
   val mem : Ident.t -> t -> bool
+  (* Returns the uses of an identifier. Returns 0 if `mem` returns
+     false. *)
   val uses_of : Ident.t -> t -> int
+  (* Increments the usage of an identifier by a positive amount. *)
   val incr : ?by:int -> Ident.t -> t -> t
-  val singleton : Ident.t -> t
+  (* Constructs a new container with a single usage mapping. *)
+  val singleton : ?count:int -> Ident.t -> t
+  (* Combines two containers. *)
   val combine : t -> t -> t
+  (* Combines many containers. *)
   val combine_many : t list -> t
+  (* Returns the contents of the container as an association list. *)
   val bindings : t -> (Ident.t * int) list
+  (* Merges two containers. *)
   val merge : (Ident.t -> int option -> int option -> int option) -> t -> t -> t
+  (* Filters entries in a given container. *)
   val filter : (Ident.t -> int -> bool) -> t -> t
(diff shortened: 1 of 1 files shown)


core/typeSugar.ml
```diff
@@ -2314,56 +2430,10 @@ let make_ft_poly_curry declared_linearity ps effects return_type =
...
-           matter whether the variable in question is used anywhere else or not, it must be
-           unlimited. *)
-          StringMap.add v 2 resulting_usages in
-      List.fold_right f vs StringMap.empty in
-    List.fold_right same us u
+let make_mono pats = List.iter (List.iter (fun (_, _, t) -> Types.Mono.make_type t)) pats
 
 let usages_cases bs =
-  usage_compat (List.map (fun (_, (_, _, m)) -> m) bs)
+  Usage.align (List.map (fun (_, (_, _, m)) -> m) bs)
```
SimonJF: Ah -- I see now -- `align` is `usage_compat`

dhil: I decided to use the name `align` after I read Garrett's explanation of `usage_compat`.

core/typeSugar.ml
```diff
@@ -1605,28 +1605,144 @@ let unbind_var context v = {context with var_env = Env.unbind v context.var_env}
...
+  [@ocaml.warning "-32"]
+  type t
+
+  val empty : t
+  val add : Ident.t -> int -> t -> t
+  val remove : Ident.t -> t -> t
+  val mem : Ident.t -> t -> bool
+  val uses_of : Ident.t -> t -> int
+  val incr : ?by:int -> Ident.t -> t -> t
+  val singleton : Ident.t -> t
```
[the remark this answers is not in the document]
dhil: A singleton operation on a map is well-defined, however, the signature probably ought to be `signature : Ident.t -> int -> t` such that it takes the number of uses too. I think the rationale behind this decision is that we have multiple occurrences of `StringMap.add var 1 StringMap.empty`.

dhil:
> This is lovely; thanks @dhil. The patch abstracts the various usage functions to a separate module, and makes the code nicer to read as a result. I don't feel the actual program logic has changed anywhere, so this is safe to go in.
>
> I have only minor gripes about commenting, but this PR has my wholehearted approval.

Indeed, there should not be any logic changes. The code should be equivalent, but arguably more robust.

SimonJF:
OK. I'll leave it up to you whether you want to tweak any comments but I'm happy for you to merge when ready.

dhil:
@SimonJF I've added some comments now.

SimonJF:
Thanks, @dhil. Merging once CI passes.

dhil:
> Thanks, @dhil. Merging once CI passes.

Thanks. Please use "rebase and merge", because I am working on another patch on top of this one.

PR #783, opened 2019-11-13
Foreign Function Interface clean up.

dhil:
This patch refactors the frontend implementation of FFI. The changes
are only cosmetic.

This is backported from my work on name hygiene/compilation
units. Ultimately, I would like to use the FFI to bootstrap (in the
frontend) the Links builtins library (aka `lib.ml`). By doing this
`lib.ml` need not be treated specially by the frontend, which makes it
significantly easier to implement a uniform infrastructure for
compilation units.

### DIFF
--- a/core/commonTypes.ml
+++ b/core/commonTypes.ml
@@ -199,6 +199,19 @@ module Name = struct
     [@@deriving show]
 end
 
+module ForeignLanguage = struct
+  type t =
+    | JavaScript
+    [@@deriving show]
+
+  let of_string = function
+    | "javascript" -> JavaScript
+    | _ -> raise (Invalid_argument "of_string")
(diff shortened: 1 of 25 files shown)


core/commonTypes.ml
```diff
@@ -199,6 +199,19 @@ module Name = struct
     [@@deriving show]
 end
 
+module ForeignLanguage = struct
+  type t =
+    | JavaScript
+    [@@deriving show]
+
+  let of_string = function
+    | "javascript" -> JavaScript
```
dhil: Maybe we should normalise the "language strings" such that a user can type `alien JavaScript ...` or `alien jAvAsCrIpT ...` or whatever.

PR #784, opened 2019-11-14
Remove `native` location.

dhil:
This patch removes the `native` location which, judging from the
source code, never seems to have been implemented. I seem to recall
that values marked native were intended to be FFI-like, i.e. their
implementations were to be given outside of Links. We already support FFI calls
to JavaScript via `alien` declarations. I think further development of
FFI should happen on `alien`.

This patch garbage collects the keyword `native`, that is, it can now
be used as an identifier.

### DIFF
--- a/bin/repl.ml
+++ b/bin/repl.ml
@@ -301,7 +301,7 @@ let handle previous_context current_context = function
                    `FunctionPtr (var, None)
                 | Location.Client ->
                    `ClientFunction (Js.var_name_binder (var, finfo))
-                | Location.Native -> assert false in
+              in
               let t = Var.info_type finfo in v, t
            | _ -> assert false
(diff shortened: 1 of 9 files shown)


dhil:
It'd be good if someone could sign-off on this as it breaks any Links program which uses `native` location annotations. That said, I would be surprised to find any **working** Links program which uses those annotations as they weren't supported by `evalir` or `irtojs`.

frank-emrich:
The patch looks good to me, but I wonder if @slindley knows what the original intended semnatics of `native` was supposed to be?

slindley:
I don't think `native` was ever implemented, and I don't know if it ever had a well-defined semantics. It should be fine to remove it.

PR #785, opened 2019-11-14
Implement character operations in Links.

dhil:
This patch removes the following character operations from `lib.ml`
and implements them in Links itself (they are now defined in
`prelude.links`):

* `toUpper : (Char) -> Char`
* `toLower : (Char) -> Char`
* `isUpper : (Char) -> Bool`
* `isLower : (Char) -> Bool`
* `isDigit : (Char) -> Bool`
* `isAlnum : (Char) -> Bool`
* `isXDigit : (Char) -> Bool`
* `isBlank : (Char) -> Bool`

### DIFF
--- a/core/lib.ml
+++ b/core/lib.ml
@@ -72,22 +72,6 @@ let string_to_xml : Value.t -> Value.t = function
   | _ -> raise (runtime_type_error "non-string value passed to xml conversion routine")
 
 (* The following functions expect 1 argument. Assert false otherwise. *)
-let char_test_op fn pure =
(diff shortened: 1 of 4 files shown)


jstolarek:
I wonder what are the performance differences between the two implementations. I suppose we we don't have a way of testing that? And even if we did it's network that's probably going to be a bottleneck anyway.

dhil:
I should add that defining them in the prelude is a temporary solution. Ultimately, I want to define these functions in a compilation unit called `Char`, which we can implicitly load for backwards compatibility.

PR #791, opened 2019-12-02
Handle free type variables in function types during closure conversion

frank-emrich:
This fixes a bug @jstolarek discovered while working on #774:

Consider the following IR pesudo-code where square brackets denote the `tyvars`:

```
fun[e] (f : \forall e. (Int) -e-> Int) (x : Int) {
  fun[] (g : (Int) -e-> Int) (y : Int) {
    y
  }
  g x
}
```

Closure conversion should detect that the inner function uses `e` from the oute function and must be generalized over it when hoisting it. However, the current implementation would only look at the body of `g` as well as its parameter `y` when looking for free type variables that must be generalized. Instead, it should also check the type of g itself, where it would discover `e`.

### DIFF
--- a/core/closures.ml
+++ b/core/closures.ml
@@ -215,6 +215,13 @@ struct
           let o = o#reset in
 
 
+          (* We must process the binder f to check its type for free type variables.
+             This must happen before adding the tyvars to o, as they are not bound in
+             the annotation on f.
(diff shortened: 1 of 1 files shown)


frank-emrich:
merging now, the failed tests are just due to #792

PR #796, opened 2019-12-03
Simplify typeSugar

jstolarek:
Remove some duplicated logic. Instead of inlining pattern environment computation inside `type_pattern` just call already existing `pattern_env` after typing the pattern.

### DIFF
--- a/core/typeSugar.ml
+++ b/core/typeSugar.ml
@@ -2179,203 +2179,6 @@ let check_unsafe_signature context unify inner = function
      unsafe
   | Some (_, None) -> raise (internal_error "Sugartypes.datatype' without a Types.typ instance")
 
-let type_pattern closed : Pattern.with_pos -> Pattern.with_pos * Types.environment * Types.datatype =
-  let make_singleton_row =
-    match closed with
-      | `Closed -> Types.make_singleton_closed_row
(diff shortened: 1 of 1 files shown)


slindley:
This is an unsound refactoring. The `pattern_env` function is only valid for let-patterns (not case-patterns) because it does not distinguish the inner and outer types. See the `Negative` and `As` cases of `type_pattern`. I'm a bit dismayed that our tests don't seem to detect the regression introduced here.

jstolarek:
`Negative` cases actually seem identical - both return an empty environment. I agree about the difference in the `As` case though. I missed that one.

slindley:
The `Negative` case is a source of different inner and outer types

    let outer_type = `Variant (positive, row_var, false) in
    let inner_type = `Variant (negative, row_var, false) in
    Negative names, Env.empty, (outer_type, inner_type)

as is the `HasType` case.

The `As` case puts the inner type into the environment.

slindley:
Hmmm... looking more closely at this, I think this refactoring is correct after all! Environment bindings come from variables and `As` patterns. For the `Variable` case there is no difference between the inner and outer type. For `As` patterns `pattern_env` extracts the type from the binder, which is already annotated with the inner type as required.

jstolarek:
Closed via b46003ebc7c44151b1a50120b7b2e0022c593260

PR #802, opened 2020-02-01
Add support for OCaml 4.10 (disable warnings as errors for releases)

kit-ty-kate:
Warnings should never be turned into errors in release mode. Currently `links` does not compile with OCaml 4.10 because of this.

*Side note: You won't be able to compile `links` with this PR with `4.10.0+beta1` or `4.10.0+beta2` (due to this bug: https://github.com/ocaml/ocaml/pull/9269) but it does work with `4.10.0+trunk` and will with `4.10.0+rc1` coming up very soon.*


dhil:
Hi @kit-ty-kate 

Thank you for your contribution. 

I am not sure why this change is suddenly necessary. I suppose OCaml 4.10 produces a new warning for our code base. Our policy is that we do not release code that produces warnings. Thus in the long time I would much rather fix whatever warning OCaml 4.10 emits in a backwards compatible way with the compiler versions that we support.

kit-ty-kate:
why not both? Fixing the warnings AND fixing this turn of warnings into errors in releases.

> Our policy is that we do not release code that produces warnings

Precisely, the dev profile is used before the release and is fine as is, and the release profile is used after (in the release)

dhil:
I somehow managed to push my changes to another branch rather than to your branch. PR #809 includes the changeset of this PR along with my own changes.

PR #804, opened 2020-02-05
LWT 5 support

dhil:
The continuous integration seems broken since Lwt 5.*. It produces the
following error

```
#=== ERROR while compiling lwt.5.1.1 ==========================================#

# context     2.0.4 | linux/x86_64 | ocaml-base-compiler.4.06.0 | https://opam.ocaml.org#4ddb5d06

# path        ~/.opam/ocaml-base-compiler.4.06.0/.opam-switch/build/lwt.5.1.1

# command     ~/.opam/ocaml-base-compiler.4.06.0/bin/dune build -p lwt -j 1

# exit-code   1

# env-file    ~/.opam/log/lwt-28397-aee82e.env

# output-file ~/.opam/log/lwt-28397-aee82e.out

### output ###

# [...]

# lwt_unix.h:101:9: error: unknown type name ‘CRITICAL_SECTION’

#  typedef CRITICAL_SECTION lwt_unix_mutex;

#          ^

#       ocamlc src/unix/lwt_libev_stubs.o (exit 2)

# (cd _build/default/src/unix && /home/travis/.opam/ocaml-base-compiler.4.06.0/bin/ocamlc.opt -g -I /home/travis/.opam/ocaml-base-compiler.4.06.0/lib/bytes -I /home/travis/.opam/ocaml-base-compiler.4.06.0/lib/mmap -I /home/travis/.opam/ocaml-base-compiler.4.06.0/lib/ocaml/threads -I /home/travis/.opam/ocaml-base-compiler.4.06.0/lib/ocplib-endian -I /home/travis/.opam/ocaml-base-compiler.4.06.0/[...]

# In file included from lwt_libev_stubs.c:9:0:

# lwt_unix.h:100:9: error: unknown type name ‘DWORD’

#  typedef DWORD lwt_unix_thread;

#          ^

# lwt_unix.h:101:9: error: unknown type name ‘CRITICAL_SECTION’

#          ^
```

However, I cannot seem to reproduce this error locally.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,11 +6,12 @@ services:
 before_install:
   - sudo add-apt-repository -y ppa:avsm/ppa
   - sudo apt-get -qq update
-  - sudo apt-get install -y ocaml-nox ocaml-native-compilers python3-setuptools python3-pip
+  - sudo apt-get install -y ocaml-nox ocaml-native-compilers python3-setuptools python3-pip libev-dev
   - pip3 install -U Sphinx
   - ./install_local_opam2.sh
   - ./opam2_local switch create ocaml-base-compiler.4.06.0
   - eval `./opam2_local config env`
+  - ./opam2_local install conf-libev
   - ./opam2_local install -y postgresql ounit2 ocamlformat
   - make rule-check
   - psql -c 'create database links;' -U postgres
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -35,7 +35,7 @@ steps:
 
   - bash: |
       case $AGENT_OS in
-        "Linux") sudo apt-get install ocaml python3-setuptools python3-pip libev-dev;;
+        "Linux") sudo apt-get install ocaml python3-setuptools python3-pip;;
         "Darwin") brew install ocaml;;
         *) exit 1 ;;
       esac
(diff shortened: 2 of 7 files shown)


SimonJF:
Lwt 5 has exposed an issue with the scheduler (in particular, `Lwt_main.run` being called multiple times). This has existed, I believe, for as long as we've used it, but Lwt has only just added an assertion failure.

I'm working to fix this now.

dhil:
I managed to get Lwt 5.* to compile after instailling `conf-libev` as suggested by https://github.com/ocsigen/lwt/issues/760.

SimonJF:
Alright -- that should hopefully do the trick to bring us up to Lwt 5 support (I've taken the liberty of renaming the PR, too). I've tested the web side of things and this is also fine.

Let's see how the CI goes. @dhil, mind trying it out? It might be worth having someone else take a look before merging.

SimonJF:
Hurrah!

dhil:
This seems to do the trick modulo the recently discovered issues #807 and #808 which are unrelated to this patch as they are present on master.

dhil:
As discussed offline, the morally right thing to do is probably to not implement `ApplyPure` in terms of `Apply`/`apply`, but rather as its own separate thing. There is a semantic difference between `ApplyPure` and `Apply`, but we never seem to really have cared about this in the evaluator. If we at some point in the future start caring about this then we may want to revert some of the changes this patch introduces.

aantron:
FYI the error in the top comment was diagnosed in ocsigen/lwt#761, fixed, and the fix released as part of Lwt 5.1.2.

PR #806, opened 2020-02-05
Operator associativity and precedence

dhil:
Prior to this patch the associativity and precedence of user-definable
operators would be stored in the internal state of the lexer. This
state was not preserved between invocations of the lexer on per
compilation unit and per REPL entry. As a consequence, during
subsequent parses every previously defined operator would default to
be left associative and have precedence 9.

This patch removes the operator associativity and precedence
information from the lexer. Instead of handling associativity and
precedence in the lexer and parser, they are now handled by the
transformation `shunting` which applies a variation of the Shunting
Yard algorithm on expressions. The transform uses our transformation
infrastructure to preserve and pass an operator table between
subsequent invocation of this transform. As a result both the lexer
and parser are greatly simplified. I believe this patch opens the door
for more rule simplifications in the parser, but I will leave that for
future work.

Currently, a default operator table is hard-coded in the `shunting`
module. My front-end bootstrapping patch hoists this table out of the
compiler and into a Links source file.

### DIFF
--- a/core/parser.mly
+++ b/core/parser.mly
@@ -289,21 +289,11 @@ let parse_foreign_language pos lang =
 %token <string> QUOTEDMETA
 %token <string> SLASHFLAGS
 %token UNDERSCORE AS
-%token <Operators.Associativity.t -> int -> string -> unit> INFIX INFIXL INFIXR PREFIX POSTFIX
+%token <Operators.Associativity.t> FIXITY
 %token TYPENAME
 %token TYPE ROW PRESENCE
 %token TRY OTHERWISE RAISE
-%token <string> PREFIXOP POSTFIXOP
-%token <string> INFIX0 INFIXL0 INFIXR0
-%token <string> INFIX1 INFIXL1 INFIXR1
-%token <string> INFIX2 INFIXL2 INFIXR2
-%token <string> INFIX3 INFIXL3 INFIXR3
-%token <string> INFIX4 INFIXL4 INFIXR4
-%token <string> INFIX5 INFIXL5 INFIXR5
-%token <string> INFIX6 INFIXL6 INFIXR6
-%token <string> INFIX7 INFIXL7 INFIXR7
-%token <string> INFIX8 INFIXL8 INFIXR8
-%token <string> INFIX9 INFIXL9 INFIXR9
+%token <string> OPERATOR
 
 %start just_datatype
 %start interactive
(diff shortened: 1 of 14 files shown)


core/parser.mly
```diff
@@ -436,8 +423,8 @@ fun_kind:
 tlfunbinding:
 | fun_kind VARIABLE arg_lists perhaps_location block           { ($1, $2, $3, $4, $5)                }
 | OP pattern sigop pattern perhaps_location block              { ((dl_unl, false), WithPos.node $3, [[$2; $4]], $5, $6) }
-| OP PREFIXOP pattern perhaps_location block                   { ((dl_unl, false), $2, [[$3]], $4, $5)          }
-| OP pattern POSTFIXOP perhaps_location block                  { ((dl_unl, false), $3, [[$2]], $4, $5)          }
+| OP OPERATOR pattern perhaps_location block                   { ((dl_unl, false), $2, [[$3]], $4, $5)          }
+| OP pattern OPERATOR perhaps_location block                  { ((dl_unl, false), $3, [[$2]], $4, $5)          }
```
SimonJF:
```suggestion
| OP pattern OPERATOR perhaps_location block                   { ((dl_unl, false), $3, [[$2]], $4, $5)          }
```

(nit: alignment)

core/parser.mly
```diff
@@ -625,113 +595,25 @@ perhaps_exps:
...
-infixl_5:
-| infixr_5                                                     { $1 }
-| infixl_5 INFIXL5 infixr_5                                    { infix_appl ~ppos:$loc $1 $2 $3 }
-
-infixr_4:
-| infixl_5                                                     { $1 }
-| infixl_5 INFIX4    infixl_5
-| infixl_5 INFIXR4   infixr_4                                  { infix_appl ~ppos:$loc $1 $2 $3 }
-| infixr_5 EQUALSTILDE regex                                   { let r, flags = $3 in
+| unary_expression OPERATOR                                    { unary_appl ~ppos:$loc (UnaryOp.Name $2) $1 } /* TODO(dhil): This is more general than the previous rule as in it will permit more expressions to be chained together. */
```
SimonJF: I wonder whether it's worth getting rid of the comment here and opening an issue, if it's problematic?

core/shunting.ml
```diff
@@ -0,0 +1,296 @@
+(* An implementation of Dijkstra's Shunting Yard algorithm. *)
+open Sugartypes
+open SourceCode.WithPos
+open Operators
+
+let name_of_binop = function
```
SimonJF: Does this really belong here? I would have thought that this would have belonged more in either `Operators` or `Sugartypes` rather than a transformation pass.

core/shunting.ml
```diff
@@ -0,0 +1,296 @@
...
+  let name = "shunting"
+
+  let get_operator_table context =
+    match Context.operator_table context with
+    | None -> Optable.defaults
+    | Some table -> table
+
+  open Transform.Untyped
+  let program state program =
+    (* Printf.fprintf stderr "Before shunting:\n%s\n%!" (Sugartypes.show_program program); *)
```
SimonJF: Might be worth either sticking these as `Debug.print` or removing.

core/parser.mly
dhil: I have thought some more about it. It shouldn't be a problem. I think I eventually convinced myself that it allows no longer sentences than before to be parsed. What I would like to do is to collapse the whole expression grammar into a single `expression:` rule. It would require a bit of experimentation which I will defer for later.

core/parser.mly
dhil: There is a note to be made here, that I didn't remark in the source code. With this change the source language becomes more "permissive" in some sense. Previously it would be a syntax error to define a prefix operator using the postfix operator definition syntax. For example, the following would be a syntax error
```
prefix ++;
op v ++ { v + 1 }
```
with this patch it is permitted, because we cannot tell a priori whether `++` is a prefix, postfix, or infix operator.

Personally, I think we shouldn't spend any effort on rejecting the above example, it is just bad code practice.

core/shunting.ml
dhil: I'd like to keep it local for now. It will go away with the introduction of hygienic names.

SimonJF: OK, if this is a conscious choice then that's fine.

dhil:
I've added the request tests now.

PR #811, opened 2020-02-19
Fix compilation of negative patterns.

### LINKED ISSUE #810: Implementation of negative patterns is broken
The implementation of negative patterns appears to be broken. Consider the following:
```
links> fun g(-Foo as r) { r };
g = fun : ([|Foo:_::Any|b::Any|]) -> [|Foo-|b::Any|]
links> g(Foo);
Foo : [|Foo-|_::Any|]
```
The result is nonsensical: `Foo` inhabits a type which explicitly prohibits `Foo`.

The expectation is a runtime error:
```
links> g(Foo);
***: Error: Links_core.Evalir.Exceptions.Wrong 
```

dhil:
This patch resolves #810 by implementing a compilation rule for
negative patterns in let expressions. The implementation simply
expands a negative pattern match into a switch-case expression as
follows.

```
  [| var -(l1,...,lN) = value; body |]
= switch (value) {
    case l1 -> Wrong
        ...
    case lN -> Wrong
    case _  -> body
  }
```

### DIFF
--- a/core/compilePatterns.ml
+++ b/core/compilePatterns.ml
@@ -268,9 +268,37 @@ let let_pattern : raw_env -> Pattern.t -> value * Types.datatype -> computation 
             let body = lp case_type patt (Variable case_variable) body in
             let cases = StringMap.singleton name (case_binder, body) in
               [], Case (value, cases, None)
-        | Pattern.Negative _ ->
-            (* TODO: compile this properly! *)
-            body
+        | Pattern.Negative names ->
+           (* The following expands the negative pattern into
+              a switch-case expression:
+
+              [| var -(l1,...,lN) = value; body |]
+            = switch (value) {
+                case l1 -> Wrong
+                     ...
+                case lN -> Wrong
+                case _  -> body
+              }
+            *)
+            let negative_cases, t' =
+              StringSet.fold
+                (fun label (cases, t) ->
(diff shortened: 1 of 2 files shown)


dhil:
Elaborating a bit, I think the pass is doing some extraneous work as it builds the type for the success case binder
```
  [| var -(l1,...,lN) = value; body |]
= switch (value : [|l1:a1|...|lN:aN|b|]) {
    case l1 (x : a1) -> Wrong
        ...
    case lN (x : aN) -> Wrong
    case (y : [|l1-|...|lN-|b|])  -> body
  }
```
However, this binder `y` is synthetic and can never be used in `body`, although less precise I suppose we could ascribe `y` the input type `[|l1:a1|...|lN:aN|b|]`.

jamescheney:
I wondered whether/how `switch`/`case` would work, given that it fails if the negated patterns are present in the scrutinee instead of, say, falling through to a continuation for the next case.  Looking at the code, the change @dhil made is just to the part that deals with let-patterns (i.e. forced pattern matching), there is separate code for general pattern-matching compilation.  One wonders whether we could DRY this redundancy out a bit.  In any case, +1 for having test cases for negated patterns in switch also.

dhil:
I have added two test cases for negative patterns in switch expressions.

PR #813, opened 2020-03-04
Refactoring DesugarDatatypes into different passes

frank-emrich:
This PR splits DesugarDatatypes into 3 individual, self-contained passes. I found this to be a necessary step to continue my work on explicit type abstractions, as `DesugarDatatypes` had become an unmaintainable mess doing too many things at once. The new passes only perform a single, dedicated task each. This should make them easier to understand and maintain.

The new passes are the following:

1. `DesugarTypeVariables` "resolves" type variables, meaning that it determines their scope and replaces  their syntactic representation (i.e., their string name) with a `Unionfind` point.
Even though `DesugarTypeVariables` works very differently from how type variables were resolved before, the goal of this PR is not to change the scoping behavior of type variables. I'll make those changes in a separate PR, for which this is a precondition.

2. `DesugarEffects` performs some mandatory massaging of the function arrows in effect rows. Further, it does all of the magic for the `effect_sugar` mode. This includes rewriting typename definitions and type applications to account for implicit effect variables.

3. `DesugarDatatypes` does everything left over: Translating "syntactic" types (from `SugarTypes.Datatype`) to "semantic" ones (from `Types`). In addition, it contains the remaining logic for handling typename defs.

## Other notes

* The new type `Sugartypes.SugarTypeVar.t` has constructors to represent type variables coming out of the parser (`TUnresolved`) and ones whose scope has been determined (`TResolved[Type|Row|Presence] `).
   The latter constructors directly carry the `Unionfind` points for the type variable in question.
   Similary, `Sugarypes.SugarQuantifier` can represent quantifiers before and after they have been resolved.
  The bulk of the changes outside of `desugar*` come from adapting the AST to use these new types.

*  `DesugarTypeVariables` cannot resolve or detect errors regarding anonymous type variables. This is because `DesugarEffects` has to do its magic on them first, potentially removing some, unifying others.
   Hence, anonymous effect variables are resolved as part of `DesugarEffects`. I've made sure that this separation of work is documented properly.

*  There is some duplication of work, which seems to be the inevitable result of splitting up the different transformations over different passes:
   Both `DesugarTypeVariables` and `DesugarEffects` need to know where implicitly bound
   type variables are allowed to appear.
   Both `DesugarEffects` and `DesugarDatatypes` need to carry some kind of
   type constructor environment along.

### DIFF
--- a/core/renamer.ml
+++ b/core/renamer.ml
@@ -131,6 +131,7 @@ let renaming_type_visitor instantiation_map =
          end
       | _ -> super#meta_presence_var point
 
+
   end
 
 
@@ -250,7 +251,11 @@ let renamer qs_from qs_to =
           `ForAll (qs_to, t')
        | t -> snd (o#typ t)
 
-    end
+
+     (* We do not rename inside the syntactic representation of types *)
+     method! type_variable x = o, x
+
+  end
 
 let rename_function_definition : function_definition -> function_definition =
   fun { fun_binder
--- a/core/renamer.ml
+++ b/core/renamer.ml
@@ -131,7 +131,6 @@ let renaming_type_visitor instantiation_map =
          end
       | _ -> super#meta_presence_var point
 
-
   end
 
 
--- a/core/sugarTraversals.ml
+++ b/core/sugarTraversals.ml
@@ -112,7 +112,7 @@ class map =
       | Open _x ->
           let _x = o#type_variable _x in Open _x
       | Recursive ((_x, _x_i1)) ->
-          let _x = o#name _x in
+          let _x = o#type_variable _x in
           let _x_i1 = o#row _x_i1 in Recursive ((_x, _x_i1))
 
     method row : Datatype.row -> Datatype.row =
@@ -613,7 +613,7 @@ class map =
           let _x_i1 = o#row _x_i1 in
           let _x_i2 = o#datatype _x_i2 in Lolli (_x, _x_i1, _x_i2)
       | Mu (_x, _x_i1) ->
-          let _x = o#name _x in
+          let _x = o#type_variable _x in
           let _x_i1 = o#datatype _x_i1 in Mu (_x, _x_i1)
       | Forall (_x, _x_i1) ->
           let _x_i1 = o#datatype _x_i1 in Forall (_x, _x_i1)
@@ -897,7 +897,7 @@ class fold =
       | Open _x ->
           let o = o#type_variable _x in o
       | Recursive ((_x, _x_i1)) ->
-          let o = o#name _x in let o = o#row _x_i1 in o
+          let o = o#type_variable _x in let o = o#row _x_i1 in o
 
     method row : Datatype.row -> 'self_type =
       fun (_x, _x_i1) ->
@@ -1332,7 +1332,8 @@ class fold =
           let o = o#list (fun o -> o#datatype) _x in
           let o = o#row _x_i1 in let o = o#datatype _x_i2 in o
       | Mu (_x, _x_i1) ->
-          let o = o#name _x in let o = o#datatype _x_i1 in o
+          let o = o#type_variable _x in
+          let o = o#datatype _x_i1 in o
       | Forall (_x, _x_i1) ->
           let o = o#datatype _x_i1 in o
       | Unit -> o
@@ -1609,7 +1610,7 @@ class fold_map =
       | Open _x ->
           let (o, _x) = o#type_variable _x in (o, (Open _x))
       | Recursive ((_x, _x_i1)) ->
-          let (o, _x) = o#name _x in
+          let (o, _x) = o#type_variable _x in
           let (o, _x_i1) = o#row _x_i1 in (o, Recursive ((_x, _x_i1)))
 
     method row : Datatype.row -> ('self_type * Datatype.row) =
@@ -2172,7 +2173,7 @@ class fold_map =
           let (o, _x_i2) = o#datatype _x_i2
           in (o, (Lolli (_x, _x_i1, _x_i2)))
       | Mu (_x, _x_i1) ->
-          let (o, _x) = o#name _x in
+          let (o, _x) = o#type_variable _x in
           let (o, _x_i1) = o#datatype _x_i1 in (o, (Mu (_x, _x_i1)))
       | Forall (_x, _x_i1) ->
           (*let (o, _x) = o#list (fun o -> o#quantifier) _x in*)
--- a/core/sugarTraversals.ml
+++ b/core/sugarTraversals.ml
@@ -14,7 +14,6 @@ open SourceCode
 open Sugartypes
 
 
-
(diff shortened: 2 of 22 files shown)


core/sugarTraversals.ml
```diff
@@ -795,7 +807,7 @@ class map =
       | `Type t -> `Type (o#typ t)
       | `Row r -> `Row (o#type_row r)
       | `Presence p -> `Presence (o#type_field_spec p)
-    method tyvar : tyvar -> tyvar = o#unknown
+    (* method tyvar : tyvar -> tyvar = o#unknown *)
```
jamescheney: Is there a reason for this to remain but be commented out?

core/sugarTraversals.ml
```diff
@@ -1291,13 +1311,13 @@ class fold =
       fun (_x, _x_i1) ->
         let o = o#string _x in let o = o#list (fun o -> o#string) _x_i1 in o
 
-    method tyvar : tyvar -> 'self_type = fun _ -> o
+    (* method tyvar : tyvar -> 'self_type = fun _ -> o *)
```
jamescheney: Is there a reason to keep this commented out?

core/sugarTraversals.ml
```diff
@@ -76,26 +78,37 @@ class map =
...
+      function
+        | TUnresolved (name, subkind_opt, freedom) ->
+           let name' = o#name name in
+           let subkind_opt' = o#option (fun o -> o#subkind) subkind_opt in
+           let freedom' = o#freedom freedom in
+           TUnresolved (name', subkind_opt', freedom')
+        | v -> o#unknown v
+
+
+  method quantifier : SugarQuantifier.t -> SugarQuantifier.t =
```
jamescheney: Inconsistent indentation

core/renamer.ml
```diff
@@ -131,6 +131,7 @@ let renaming_type_visitor instantiation_map =
          end
       | _ -> super#meta_presence_var point
 
+
```
jamescheney: Stray whitespace

core/sugarTraversals.ml
```diff
@@ -13,6 +13,8 @@ open CommonTypes
 open SourceCode
 open Sugartypes
 
+
```
jamescheney: Stray whitespace

core/sugarTraversals.ml
```diff
@@ -2369,8 +2400,8 @@ class fold_map =
       | `Row r -> let o, r = o#type_row r in o, `Row r
       | `Presence p -> let o, p =o#type_field_spec p in o, `Presence p
 
-    method tyvar : Quantifier.t -> ('self_type * Quantifier.t) =
-      o#unknown
+    (* method tyvar : Quantifier.t -> ('self_type * Quantifier.t) =
```
jamescheney: Still need this commented out?

core/sugarTraversals.mli
```diff
@@ -138,7 +138,7 @@ class fold :
     method fieldspec       : Datatype.fieldspec -> 'self
     method fieldconstraint : fieldconstraint -> 'self
     method directive       : directive -> 'self
-    method tyvar           : tyvar -> 'self
+    (* method tyvar           : tyvar -> 'self *)
```
jamescheney: Still need this commented out?

core/sugarTraversals.mli
```diff
@@ -219,12 +218,12 @@ object ('self)
...
   method freedom         : Freedom.t -> 'self * Freedom.t
-  method type_variable   : type_variable -> 'self * type_variable
-  method known_type_variable : known_type_variable -> 'self * known_type_variable
+  method quantifier      : SugarQuantifier.t -> 'self * SugarQuantifier.t
+  method type_variable   : SugarTypeVar.t -> 'self * SugarTypeVar.t
   method typ             : Types.datatype -> ('self * Types.datatype)
   method type_row        : Types.row -> ('self* Types.row)
   method type_arg        : Datatype.type_arg -> 'self * Datatype.type_arg
-  method tyvar           : Quantifier.t -> ('self * Quantifier.t)
+  (* method tyvar           : Quantifier.t -> ('self * Quantifier.t) *)
```
jamescheney: Still need this commented out?

core/sugartypes.ml
```diff
@@ -55,7 +58,7 @@ end = struct
 end
 
 (* type variables *)
-type tyvar = Quantifier.t
+(* type tyvar = Quantifier.t *)
```
jamescheney: This seems strange, what does the [@@deriving] on the next line refer to?

core/sugartypes.ml
```diff
@@ -73,36 +76,140 @@ let default_effect_subkind : Subkind.t = (lin_unl, res_any)
...
 let rigidify (name, kind, _) = (name, kind, `Rigid)
 
-let string_of_type_variable ((var, (kind, subkind), _) : type_variable) =
-  match kind with
-  | None -> var
-  | Some kind ->
-     let subkind = OptionUtils.opt_app Subkind.to_string "" subkind in
-     var ^ "::" ^ PrimaryKind.to_string kind ^ subkind
+let string_of_type_variable (_v : SugarTypeVar.t) =
+  failwith "fixme"
```
jamescheney: What would it mean to fix this?  Is this function still used anywhere?  It doesn't appear in the diff.

core/sugarTraversals.ml
frank-emrich: removed

core/sugarTraversals.ml
frank-emrich: fixed

core/renamer.ml
frank-emrich: fixed

core/sugarTraversals.ml
frank-emrich: fixed

core/sugarTraversals.ml
frank-emrich: removed

core/sugarTraversals.mli
frank-emrich: removed

core/sugarTraversals.mli
frank-emrich: removed

core/sugartypes.ml
frank-emrich: good point. This was supposed to go away, together with   `Sugartypes.tyvar`, `Sugartypes.type_variable`, and  `Sugartypes.quantifier`. Done now.

core/sugartypes.ml
frank-emrich: Ah, good point. I disabled this function and used an auto-generated printer in the one place where this function was originally used. I've now restored this (nicer) version and given it a proper implementation.

frank-emrich:
If nobody objects and there are no other reviews I would like to merge this at the end of the week.

PR #814, opened 2020-03-05
Fix compilation warning 67 emitted by OCaml 4.10.0

dhil:
Our codebase triggers numerous instances of warning 67 with OCaml
4.10.0. This patch fixes those instances such that our codebase no
longer emits any warnings under any of our compilation profiles.

### DIFF
--- a/dune
+++ b/dune
@@ -1,19 +1,19 @@
 (env
  (dev
   (flags (:standard
-          -strict-formats  ;; Disallows legacy formats.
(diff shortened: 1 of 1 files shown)


dhil:
I'll merge this as soon as the CI passes as we are otherwise stalled by using OCaml 4.10.0.

dhil:
Interesting, syntax for anonymous functor arguments is apparently new in OCaml 4.10.0, and thus not backwards compatible. In the light of this discovery, I suppose we silence warning 67 by adding it to our whitelist.

Issue #816, opened 2020-03-06
Test script executor leaks memory


dhil:
The test script executor (`test-harness`) can leak memory when a test script times out. The reason is that the executor does not perform proper clean up of timed out scripts, allowing their processes to keep running forever. 

Here is a small reproduction `hungry.links`.
```
Hungry
fun f() { 1 :: f() } f()
stdout : () : ()
```
Be careful if you run the above snippet as it will quickly consume all the memory on your machine.
```
$ ./test-harness hungry.tests
!FAILURE: Hungry [TIMED OUT]
1 failures (+0 ignored)
0 successes
$ echo $?
1
$ ps -ef | grep links.exe
joe     15972 15971 99 21:52 pts/1    00:00:18 _build/default/bin/links.exe -e fun f() { 1 :: f() } f()
```
The process is kept alive even after `test-harness` has completed.

jamescheney:
Is this now closed by #821?

frank-emrich:
Yes, apparently the auto-closing isn't triggered from comments, just from the main description of a PR. Learned something new today!

jamescheney:
It seemed to get this right using other examples (e.g. #810/#811).  The NLP is obviously unreliable (like all ML/AI) but it can be explicitly set using the "Linked pull requests" sidebar.

PR #821, opened 2020-04-10
test suite: kill offending Links process in case of timeout

frank-emrich:
Currently, the test harness detects if a test case times out, but fails to actually kill the Links process executing the timed-out test case.

The only minor caveat is that the test harness doesn't spawn the Links processes directly. Instead, it creates a shell which in turn starts the Links process. Hence, killing just the shell after a timeout leaves the Links process running. Therefore, I had to add some process group magic such that we can easily kill the shell and all of its sub-processes (i.e., the Links process).

### DIFF
--- a/test-harness
+++ b/test-harness
@@ -2,7 +2,7 @@
 
 links = './links'
 
-import sys, re, time
+import sys, re, time, os, signal
 import concurrent.futures, threading, multiprocessing
 from os import path
 from subprocess import Popen, PIPE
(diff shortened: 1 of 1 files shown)


frank-emrich:
This fixes #816

dhil:
Sounds good to me, although I haven't tested it. I assume you have tried example from #816.

jstolarek:
How do I trigger a timeout in the testsuite so that I can test this patch?

frank-emrich:
@dhil Yes I have

@jstolarek 
You can create the a file `nonterm.tests` with the following content and run `./test-harness nonterm.tests`
```
Hungry
fun f() {  f() } f()
stdout : () : ()
```

This should timeout after 15 seconds and more importantly, kill the links process

PR #823, opened 2020-04-13
Make InstantiateTypes visitor correctly handle rows and field_specs 

### LINKED ISSUE #574: "Type mismatch" IR typing error produced by `curry` Prelude function
(results below were obtained from commit dccf4aa, currently in #572 PR)

`curry` function from Prelude:

```
fun curry(f) {
  fun (x)(y) { f(x,y) }
}
```

causes a "Type mismatch" IR typing error:

```
Type mismatch:
 Expected:
  (229:(a::Any, b::Any) -c-> d::Any,230:a::Any)
 Actual:
  (229:(a::Any, b::Any) -e-> d::Any,230:a::Any)
```

Config settings:

```
typecheck_ir=true
prelude=empty.links
optimise=false
debug=true
hide_fresh_type_vars=false
show_flavours=true
```

Error goes away with `optimise=true`.

frank-emrich:
The `InstantiateTypes` visitor is used to instantiate all types found in an IR program. It had a bug where it may not have worked properly on rows and field specs. This was the root cause for  #574. Thus, this fixes  #574


tests/typed_ir/T574.links
```diff
@@ -0,0 +1,4 @@
+#Unchanged from prelude
```
jstolarek, line 1: You need to register this test in ` tests/typed_ir.tests`. At the moment it's not being run by the testsuite.

frank-emrich: Good catch, thanks!

tests/typed_ir.tests
```diff
@@ -41,3 +41,8 @@ Bound quantifiers in a query (#694)
 tests/typed_ir/T694.links
 filemode : true
 stdout : fun : forall a::Row,b::Row(Any,Base).() -a-> [(|b::(Any,Base))]
+
+Curry (#574)
+tests/typed_ir/T574.links
+filemode : true
+stdout : () : ()
```
jstolarek, line 48: Notice how github complains about lack of newline at the EOF. Running `git diff` before committing should show you a warning about this:

frank-emrich: That's how it's supposed to be. The test suite will complain if the `.tests` file ends with a newline.

jstolarek: It will complain if you add **two** newlines at the end of file, but not when you add **one** newline. Note that `git diff` will also mark empty lines at the end of file, again making it easy to spot:

frank-emrich: You're right, sorry. I fixed this now.

Issue #698, opened 2019-08-01
Type mismatch IR error in Prelude.asList


jstolarek:
code (from prelude):
```
fun asList(t) server {
  for (x <-- t) [x]
}
```
error:
```
Type mismatch:
 Expected:
  TableHandle((|a::Base),(|b::Base),(|c::Base))
 Actual:
  (|a::Base)
occurring in IR binding:
(Ir.Let (
   (794,
    (TableHandle((|a::Base),(|b::Base),(|c::Base)), "_for__g1",
     Var.Scope.Local)),
   ([], (Ir.Return (Ir.Variable 796)))))
```
(there's also an "incompatible effect" error - see #697). Config:
```
typecheck_ir=true
prelude=empty.links
optimise=false
debug=true
hide_fresh_type_vars=false
show_flavours=true
show_quantifiers=true
```

jstolarek:
Here's some investigation into what's going on.  During desugaring `desugarFors` generates a fragment that looks like this:

```
Block ([(Fun
  { fun_binder =
    ("_fun__g2", ((|a::Base)) {}-> [(|a::Base)]);
    fun_linearity = CommonTypes.DeclaredLinearity.Unl;
    fun_definition =
    ([], ([[(Pattern.As (
           ("_for__g1", TableHandle((|a::Base),(|b::Base),(|c::Base))),
           (Pattern.Variable ("x", (|a::Base))))) ] ],
      (ListLit ([(Var "x")], (Some (|a::Base))))));
    fun_location = CommonTypes.Location.Unknown;
    fun_signature = None;
    fun_unsafe_signature = false;
    fun_frozen = true })
  ], (FreezeVar "_fun__g2"))
```

after translating to IR this becomes:

```
(Fun ((245,
     (forall a::Row(Unl,Base),b::Row(Unl,Base),c::Row(Unl,Base).((|a::Base)) {}-> [(|a::Base)],
      "_fun__g2", Var.Scope.Global)),
    ([a::Row(Unl,Base); a::Row(Unl,Base); a::Row(Unl,Base)],
     [(244, ((|a::Base), "", Var.Scope.Local))],
     ([(Let ((242,
           (TableHandle((|a::Base),(|b::Base),(|c::Base)), "_for__g1",
            Var.Scope.Local)),
          ([], (Return (Variable 244)))));
        (Let ((243, ((|a::Base), "x", Var.Scope.Local)),
           ([], (Return (Variable 244)))))
        ],
      (Return
[...]

PR #824, opened 2020-04-14
Properly register test for #698 in typed_ir.tests

frank-emrich:
In another PR, @jstolarek observed that I had created a test file for an IR test, but forgot to register it in `typed_ir.tests`. The same problem happened in #820, where I added a test for  #698.

### DIFF
--- a/tests/typed_ir.tests
+++ b/tests/typed_ir.tests
@@ -46,3 +46,8 @@ Curry (#574)
 tests/typed_ir/T574.links
 filemode : true
 stdout : () : ()
+
+asList (#698)
+tests/typed_ir/T698.links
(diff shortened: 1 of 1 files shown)


jamescheney:
Maybe there should be a check that all of the tests are registered? Who wants to write some Python?

frank-emrich:
One heuristic for making sure that all tests are "registered" may be the following:
 For every `.links` file somewhere in `tests`, the path of the `.links` file should appear in a `.tests` file in `tests`. I'll see if this works.

frank-emrich:
Merging this, I've made progress on the automated checking for non-registered test files, but it belongs in a separate PR.

Issue #826, opened 2020-04-26
ocamlformat_check rule isn't executed by rule_check


frank-emrich:
As with the tab checking script problem (#825), `tools/rules/ocamlformat_check` wasn't marked as executable and was therefore ignored by `toolsl/rule_check`.

Running it manually makes it complain about the following files:
```
core/lens_errors.ml
core/lens_ir_conv.ml
core/lens_value_conv.ml
core/lens_type_conv.ml
core/lens_database_conv.ml
core/lens_sugar_conv.ml
core/lens_sugar_conv.mli
lens/sorted_records.mli
lens/alias.ml
lens/phrase.ml
lens/lens_list.ml
lens/value.ml
lens/lens_list.mli
lens/fun_dep.mli
lens/eval.mli
lens/type.ml
lens/lens_map.ml
lens/phrase_value.mli
lens/sort.ml
lens/type.mli
lens/eval_classic.ml
lens/phrase_value.ml
lens/fun_dep.ml
lens/eval_incremental.ml
lens/column.ml
lens/phrase_typesugar.ml
lens/eval.ml
lens/database.mli
lens/lens_seq.ml
lens/phrase_typesugar.mli
lens/statistics.ml
lens/database.ml
lens/sorted_records.ml
lens/sort.mli
lens/value.mli
lens/column.mli
lens/lens_option.ml
lens/lens_set.ml
tests/unit/unitTestsLensSetOperations.ml
tests/unit/unitTestsLensPerformance.ml
tests/unit/unitTestsLensPrimitives.ml
tests/unit/unitTestsLensLang.ml
tests/unit/unitTests.ml
tests/unit/unitTestsLensCommon.ml
tests/unit/unitTestsLensAlias.ml
tests/unit/unitTestsLensFDHelpers.ml
tests/unit/unitTestsLensSort.ml
tests/unit/unitTestsLensDatabase.ml
```

@rudihorn could you investigate if those complains are legitimate?

frank-emrich:
This was fixed in #828

PR #830, opened 2020-04-29
Check that there are no orphaned .links files in tests directory

frank-emrich:
This is a follow-up to @jamescheney's suggestion in #824 that we should have some way of automatically checking the *.links files in tests/ are somehow executed.

To this end, my script checks that for any .links file in tests/ if it is either
- mentioned in a .tests file
- mentioned in any of the testsuite.config files for database tests
- starts with `#IGNORED_BY_TEST_REGISTRATION_RULE_CHECKING_SCRIPT` to indicate that the file wasn't overlooked, but doesn't have to satisfy the first to conditions (for example, because it is included by another file)

I've also made some tweaks to tools/rule-check, which can now execute any kind of rule checking script (rather than just bash) and also complains if it finds a non-executable file in tools/rules/, which caused #825 and #826.


tools/rules/test_files_registered.py
```diff
@@ -0,0 +1,67 @@
...
+links_files = set()
+
+#Collect .links files that should appear in a ,tests file
+for links_file in glob.glob(os.path.join(TEST_DIR, "**", "*.links"), recursive=True):
+    with open(links_file) as f:
+        first_line = f.readline().strip()
+        if(first_line != BLACKLIST_STRING):
+            normed = os.path.normpath(links_file)
+            links_files.add(normed)
+            #print("normed : %s" % normed)
```
SimonJF, line 18: Consider deleting

tools/rules/test_files_registered.py
```diff
@@ -0,0 +1,67 @@
...
+
+            if line == "":
+                test_offset = 0
+            else:
+                test_offset += 1
+
+            #If this is the second line in a test case def, it may be the path of a  .links file
+            if(test_offset == 2 and re.match(".*\.links", line)):
+                normed = os.path.normpath(line)
+                #print("normed: %s" % normed)
```
SimonJF, line 39: Consider deleting

tools/rules/test_files_registered.py
```diff
@@ -0,0 +1,67 @@
...
+                if normed in links_files:
+                    #print("found %s" %line)
```
SimonJF, line 41: Consider deleting

tools/rules/test_files_registered.py
```diff
@@ -0,0 +1,67 @@
...
+    path_dir = os.path.dirname(testsuite_file)
+
+    with open(testsuite_file, "r") as f:
+        for line in f:
+            path_file = line.strip() + ".links"
+
+            combined_path = os.path.join(path_dir,path_file)
+            normed = normed = os.path.normpath(combined_path)
+            if normed in links_files:
+                    #print("found %s" %line)
```
SimonJF, line 57: Consider deleting

SimonJF:
(The "Consider deleting"s were referring to the commented debug prints, not the entire blocks of code GitHub has chosen to show...)

frank-emrich:
> Have you considered, rather than having a per-file blacklist string, having a JSON config file containing the lists of blacklisted files? 

I've now included the blacklist as a simple list within the script. If we need the list in multiple places, we can still promote it to its own file, but that seemed unnecessary for now.

I've removed the dead print statements

SimonJF:
Looks great to me. Happy for this to go in when you're ready for it to.

PR #831, opened 2020-04-29
Remove Ocamlformat from blacklist

rudihorn:
This branch updates our `.ocamlformat` file to be compatible with Ocamlformat 0.14.1. It also fixes ocamlformat to version 0.14.1 so that any future compatability updates don't affect the codebase. I've moved the installation of ocamlformat to `./tools/install_ocamlformat` which is then used by both `.travis.yml` as well as `azure-pipelines.yml`.

### DIFF
--- a/.ocamlformat
+++ b/.ocamlformat
@@ -1,5 +1,6 @@
 break-cases=all
 doc-comments=before
+doc-comments-val=before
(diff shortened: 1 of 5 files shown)


frank-emrich:
LGTM

PR #832, opened 2020-05-05
Abstract syntax for SQL modifications

SimonJF:
# Motivation
Previously, SQL update and deletion operations were constructed as strings in `query.ml`. Since insertions are implementation-dependent, the picture there is even worse: we needed to do some gymnastics to get a bunch of strings, which were then passed to the `Value.database` implementation.

When we're wanting to do query rewrites (say, if we wanted to do temporal database operations, for example), the string-based approach is very inflexible. Furthermore, insertions are treated non-uniformly, which is a pity.

# Changes
This patch adds abstract syntax for SQL insertions, updates, and deletions in the `Sql` module, and adapts the rest of the codebase to use them. More specifically:

  - `Sql` contains the new abstract syntax nodes `Insert`, `Update`, and `Delete`.
  - `Sql.string_of_query` takes a field quotation function of type `string -> string`, rather than taking a `Value.database`. In turn, this breaks the dependency from `Sql` to `Value`.
  - We add a `string_of_query` method to `Value.database`, which is an overridable implementation-specific query serialisation function which calls into `Sql.string_of_query` by default.
  - The `make_insert_query` method is deleted from `Value.database`. Instead, implementation-dependent insertion functionality is achieved by overriding `string_of_query` for `Insert` AST nodes.
  - The `make_insert_returning_query` method is simplified to take an insertion query and a `returning` field, rather than serialised values.

### DIFF
--- a/core/query/sql.ml
+++ b/core/query/sql.ml
@@ -2,27 +2,24 @@ open Utility
 open CommonTypes
 
 type index = (Var.var * string) list
-type range = (int * int) option
+type range = int * int
 
 type query =
   | UnionAll  of query list * int
   | Select    of select_clause
-  | Insert    of insert_query
-  | Update    of update_query
-  | Delete    of delete_query
+  | Insert    of {
+      ins_table: string;
+      ins_fields: string list;
+      ins_records: base list list
+    }
+  | Update    of {
+      upd_table: string;
+      upd_fields: (string * base) list;
+      upd_where: base option
+    }
+  | Delete    of { del_table: string; del_where: base option }
   | With      of Var.var * query * Var.var * query
 and select_clause = (base * string) list * (string * Var.var) list * base * base list
-and insert_query = {
-  ins_table: string;
-  ins_fields: string list;
-  ins_records: base list list
-}
-and update_query  = {
-  upd_table: string;
-  upd_fields: (string * base) list;
-  upd_where: base option
-}
-and delete_query  = { del_table: string; del_where: base option }
 and base =
   | Case      of base * base * base
   | Constant  of Constant.t
--- a/core/query/sql.ml
+++ b/core/query/sql.ml
@@ -2,12 +2,27 @@ open Utility
 open CommonTypes
 
 type index = (Var.var * string) list
+type range = (int * int) option
 
 type query =
   | UnionAll  of query list * int
   | Select    of select_clause
+  | Insert    of insert_query
+  | Update    of update_query
+  | Delete    of delete_query
   | With      of Var.var * query * Var.var * query
 and select_clause = (base * string) list * (string * Var.var) list * base * base list
+and insert_query = {
+  ins_table: string;
+  ins_fields: string list;
+  ins_records: base list list
+}
+and update_query  = {
+  upd_table: string;
+  upd_fields: (string * base) list;
+  upd_where: base option
+}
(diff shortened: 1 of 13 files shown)


core/value.ml
```diff
@@ -74,16 +74,13 @@ class virtual database = object(self)
...
-        "("^String.concat "," field_names ^") values "^
-        String.concat "," (List.map (fun vs -> "(" ^ String.concat "," vs ^")") vss)
-  method make_insert_returning_query : (string * string list * string list list * string) -> string list =
-    fun _ ->
+  method make_insert_returning_query : string -> Sql.query -> string list =
+    fun _ _ ->
     raise (raise (internal_error ("insert ... returning is not yet implemented for the database driver: "^self#driver_name())))
   method virtual supports_shredding : unit -> bool
 
+  method string_of_query range q =
```
dhil, line 82: Consider making `range` an optional parameter.

core/query/sql.ml
```diff
@@ -208,21 +257,21 @@ and string_of_base db one_table b =
...
+and string_of_projection quote one_table (var, label) =
   if one_table then
-    db#quote_field label
+    quote label
   else
-    string_of_table_var var ^ "." ^ (db#quote_field label)
+    string_of_table_var var ^ "." ^ (quote label)
 
-let string_of_query db range q =
+let string_of_query quote range q =
```
dhil, line 271: Possible alternative interface:
```ocaml
val string_of_query : ?quote:(string -> string) -> ?range:(int * int) -> query -> string
```
with defaults
```ocaml
let string_of_query ?(quote=id) ?range q = ...
```

core/query/sql.ml
```diff
@@ -125,9 +140,10 @@ let order_by_clause n =
    returned. This allows these operators to take lists that have any
    element type at all. *)
 
-let rec string_of_query db ignore_fields q =
-  let sq = string_of_query db ignore_fields in
-  let sb = string_of_base db false in
+let rec string_of_query quote ignore_fields q =
```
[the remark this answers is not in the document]
SimonJF: Absolutely. I'll make an issue.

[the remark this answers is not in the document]
jamescheney: Introducing an intermediate pretty-printing stage would be nice also, since for debugging the printed queries are entirely unreadable...

core/value.ml
SimonJF: Good idea!

core/query/sql.ml
SimonJF: Hopefully you'll forgive me for not taking your suggestion here:

  - I'm reluctant to give `quote` a default argument, since the field quotation is something which should almost certainly be set to `db#quote_field` and I fear that having this filled implicitly would open the door to subtle bugs for others later on.

  - Changing `range` to take an `int * int` pair would complicate matters since there's a difference as to whether `range` is specified in a call, or whether `range` is indeed set to `None` at a program level. Conceptually, this would end up with us having an `(int * int) option option` type, which is something I'd like to avoid.

dhil: You can provide `None` as default for `range` to flatten it. Or alternatively change the `type range = ...` definition.

SimonJF: Huh, that works nicely. Thanks.

SimonJF:
Thanks for the review, @dhil. Is this good to go after the CI passes?

PR #838, opened 2020-06-02
fix Issue 835

### LINKED ISSUE #835: Query flatness checking is too strict
If we run this query (with shredding/relax_query_type_constraint disabled):

```
 # should succeed but fails because of nested intermediate type/type error

 query { for (y <- for (x <-- factorials) [(b=[])])
         [(a=0)]};
```
 we get a type error:
```
Unification error: Cannot unify the Base row variable 6645 with the non-Base row b:[_::Any].
<stdin>:1: Type error: A database comprehension must return a list of records of base type, but the expression
    `[(b=[])]'
has type
    `[(b:[a::Any])]'
In expression: for (x <-- factorials) [(b=[])].

```
This is wrong because the above query only introduced nesting inside the query and the result type is not nested.  Therefore normalization will get rid of it.

If instead we write this equivalent query:
```
# should be equivalent but succeeds
 query { for (y <- for (x <- asList( factorials)) [(b=[])])
         [(a=0)]};
```
it typechecks and we do get the right answer.

Moreover, if we enable shredding and relax_query_type_constraint and run these queries with the flat query evaluator we get a run-time error both times:

```
***: Runtime error: Incompatible query evaluation annotations. Expected flat, got nested. 
```

jamescheney:
This attempts to fix #835 by simply disabling the syntactic heuristic for deciding whether an iteration is a query in the typechecker.  

This has the possibly unwelcome side effect that for any `Iteration` comprehension that would have previously been assumed to be a query (because of the presence of `<--`), if a tame function is called inside that `Iteration` block, we now need to explicitly enclose it in a `query` block when we didn't before.  I'm personally fine with this because it makes it more explicit what is going on, and I would also be fine with requiring `query` keywords to enclose any `Iteration` that is expected to evaluate as an external query, or rejecting attempts to use `<--` outside of a `query` block, but these would be more invasive and separate changes.

In our test suite there was exactly one place where this mattered and meant code needed to change: in `wine.links`.

It also means that we decease the number of places we use the `relax_query_type_constraint` setting to 1 : the only remaining use is in `irTyping`.  I think that the logic for determining whether a query has to have a flat return type is now captured in the `policy` parameter and so we should inspect that instead of using this flag.  I also think shredding should just be enabled by default now since we have a way within Links to choose whether to use it or not on a given query, so having the flag provide default behavior unnecessarily complicates matters.

To do:
- [ ] Argue about whether the proposed breaking change is reasonable.
- [x] Adjust irTyping to use the query policy flag to decide whether to enforce flat query results instead of `relax_query_type_constraint`
- [x] Cleanup and remove dead code

### DIFF
--- a/core/database.ml
+++ b/core/database.ml
@@ -9,11 +9,12 @@ let connection_info
               |> convert Utility.some
               |> sync)
 
+    (* XXX
 let relax_query_type_constraint =
   Settings.(flag "relax_query_type_constraint"
             |> convert parse_bool
             |> sync)
-
+*)
 let shredding =
   Settings.(flag "shredding"
             |> synopsis "Enables database query shredding"
--- a/core/database.ml
+++ b/core/database.ml
@@ -9,12 +9,6 @@ let connection_info
               |> convert Utility.some
               |> sync)
 
-    (* XXX
-let relax_query_type_constraint =
-  Settings.(flag "relax_query_type_constraint"
-            |> convert parse_bool
-            |> sync)
-*)
 let shredding =
   Settings.(flag "shredding"
             |> synopsis "Enables database query shredding"
--- a/core/database.mli
+++ b/core/database.mli
@@ -1,7 +1,7 @@
 (** A generic interface for SQL-style databases. Vendor-specific implementations are elsewhere *)
 
 val connection_info : string option Settings.setting
-val relax_query_type_constraint : bool Settings.setting
+(* XXX val relax_query_type_constraint : bool Settings.setting *)
 val shredding : bool Settings.setting
 
 class virtual db_args : string -> object
@@ -27,4 +27,3 @@ val execute_select : (string * Types.datatype) list -> string -> Value.database 
 val execute_untyped_select : string -> Value.database -> Value.t
 
 val execute_insert_returning : string -> Sql.query ->  Value.database -> Value.t
-
--- a/core/database.mli
+++ b/core/database.mli
@@ -1,7 +1,6 @@
 (** A generic interface for SQL-style databases. Vendor-specific implementations are elsewhere *)
 
 val connection_info : string option Settings.setting
-(* XXX val relax_query_type_constraint : bool Settings.setting *)
 val shredding : bool Settings.setting
 
 class virtual db_args : string -> object
--- a/core/database.mli
+++ b/core/database.mli
@@ -26,3 +26,4 @@ val execute_select : (string * Types.datatype) list -> string -> Value.database 
 val execute_untyped_select : string -> Value.database -> Value.t
 
 val execute_insert_returning : string -> Sql.query ->  Value.database -> Value.t
+
--- a/core/evalir.ml
+++ b/core/evalir.ml
@@ -699,13 +699,6 @@ struct
               value env offset >>= fun offset ->
               Lwt.return (Some (Value.unbox_int limit, Value.unbox_int offset))
        end >>= fun range ->
-       let evaluator =
-         let open QueryPolicy in
-         match policy with
-           | Flat -> `Flat
-           | Nested -> `Nested
-           | Default ->
-               if Settings.get Database.shredding then `Nested else `Flat in
 
        let evaluate_standard () =
          match EvalQuery.compile env (range, e) with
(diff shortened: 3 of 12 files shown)


jamescheney:
I believe this is finished now pending review.  I did further experimentation to see what the impact of requiring `asList` and `<--` to be inside query blocks would be in #841.

SimonJF:
I'm very much in favour of this change. It's always seemed counterintuitive that unenclosed `<--` comprehensions were allowed. I'll check it on my GtoPdb code, but I doubt it'll be problematic modulo maybe some superficial changes.

slindley:
The design of `<-- ` has always been controversial. IIRC the thinking was that it would be harmless, and sometimes convenient, to bake an implicit `query` into table comprehensions, given that they are syntactic sugar anyway and not part of the core language proper.

Now (over a decade later!) that we have much more experience using language integrated query in Links, it seems like a good time to revisit this design.

@SimonJF, I'm sure that it's easy to update GtoPdb to work with the alternative syntax, but it would be interesting to know whether there are any cases in which it is mildly irritating to have to write the extra code warranted by the alternative design.

jamescheney:
@SimonJF: This PR does not make it an error to use table comprehensions outside a query block.  I left that (more-breaking) change for #841/#842 since I thought it might be more controversial.

@slindley: This PR keeps (for now) the implicit insertion of query{} around table comprehensions, but does makes a slight breaking change, which is:
- if we call a tame function from a query whose context is wild, then we must explicitly write `query` around the block to avoid an effect mismatch.  There is one example of this in the wine.links example.

[The query{} blocks are still implicitly added around table comprehensions, but this does not happen until after the initial typechecking pass.]

I think it is possible to keep the old behavior, but if we agree on the stricter typechecking proposed in #841 then this is moot.

Discussion of whether it is desirable to explicitly require query{} to be written around all table comprehensions / asList calls belongs in #841/#842.  Likewise to assess the impact of that change the issue841 branch should be checked out.

Are there any objections to merging this?

SimonJF:
@jamescheney: Ah, of course. That makes sense -- thanks for the clarification.

I'm absolutely fine with this being merged.

slindley:
Before incorporating this kind of change I would like to see a couple of concrete examples (e.g. taken from GtoPdb) to illustrate the kind of changes required to code. Ideally these would illustrate the maximum amount of pain we would likely encounter, from which we can hopefully conclude that the change is not painful at all.

SimonJF:
This present PR doesn't affect GtoPdb at all (I've checked). In terms of actual Links code that's modified, apart from that single example in `wine.links` (which morally should be in a `query` block anyway) it seems almost completely uninvasive. We may have to think harder regarding #842 though.

jamescheney:
Yes, I was initially trying to avoid any breaking change and the winestore example is the only one in the test suite.  Moreover, this change seems necessary to restore consistency between how `asList` and table comprehensions are treated, since the following (supposedly) equivalent code already doesn't typecheck:
```
sig firstField : (TableHandle ((|r::Base), (|w::Base), (|n::Base)), ((|r)) {}-> Bool, ((|r)) {}-> a) ~e~> (a::Base) 
fun firstField(t, p, body) { 
  var matches = 
    for (r <- asList(t)) 
    where (p(r)) 
    [(v=body(r))]; 
  hd(matches).v
}
```
That is, we were already requiring `query` blocks around queries written using `asList` in order to call the tame query code from a wild context.  So something has to give if we want `for(x<--t)...` and `for(x<--asList(t))...` to do the same thing.

If there are strong objections to this breaking change I can try to revert it (thus preserving the existing broken, but more permissive behavior for now).

jamescheney:
is this an error outside a query block?
```
for (x <-- factorials) [(b=[])]
```

jamescheney:
Yes, this does lead to a run-time failure - so maybe we should go straight for #841 
```
links> for (x <-- factorials) [(b=[])];
Not a base expression: (Query.Lang.Concat [])
***: Error: "Assert_failure core/query/query.ml:1118:6" 
```

jamescheney:
Closing since #841/842 seem necessary to preserve safe behavior so we should just discuss that.

Issue #841, opened 2020-06-07
Stricter enforcement of query effects?


jamescheney:
Following discussion in #835 I propose the following (mildly) breaking change:

- The use of `asList` and `for(<--)` is limited to tame contexts, i.e. the wild effect is required to be definitely absent.
- As a consequence, it becomes a type error to write a query that accesses database tables without explicitly writing `query` around it.
- Therefore, the implicit insertion of `query` keywords around `Iteration`s that happen to contain table references is no longer necessary; moreover, it would be possible to desugar such comprehensions to calls to `asList` (or vice versa, as currently `asList` is defined in terms of a table comprehension), since the two supposedly equivalent constructs are actually equivalent.

I think this would be cleaner and matches what we say about `query` and `asList` in various papers.

It does potentially break existing code, in the following sense:
- any code that explicitly relied on the implicit insertion of `query` keywords around table comprehensions, now fails to typecheck.
- relatedly, any type annotations that previously used effect / presence polymorphism for code that runs on the database, now need to be changed to specify that the code must be tame (i.e. the wild effect is absent).

I am creating a PR building on #838 that shows all of the places that need to be fixed in the test suite.  There aren't that many: the most affected example is the winestore which uses queries with implicit table comprehension syntax freely, and there are a couple of other places where query keywords need to be added or type annotations strengthened.

My sense is the number of changes is small and it's worth it to get rid of the implicit `wrapTableIterators` pass, which was always ad hoc (i.e. applied to queries with table comprehensions but not those containing asList), and makes it crystal clear when/where queries are actually issued (with the caveat that query expressions that don't actually refer to the database will still execute in-memory.)

However I'd like there to be a discussion of this (and ideally for the impact on larger applications like GtoPdb to be assessed) first.

SimonJF:
I'm very much in favour of this change. It's always seemed counterintuitive that unenclosed <-- comprehensions were allowed. I'll check it on my GtoPdb code, but I doubt it'll be problematic modulo maybe some superficial changes.

jamescheney:
Per #838, something like this needs to be done to resolve #835 without introducing new bugs due to failure to check that query results are flat.

PR #842, opened 2020-06-07
Initial stab at #841 for discussion

jamescheney:
This PR shows what happens if we make the use of `asList` and table comprehensions (`for (..<--..)`) stricter as proposed in #841.  This bakes good style into the type system, so that queries now always happen only when the `query` keyword appears.

The type system is still somewhat relaxed about allowing `query` to appear nested inside another `query`, and `query` keywords can also not result in queries being issued if the contents don't actually refer to a database table.  On the other hand, `query nested`s can generate several SQL queries.  So the correspondence is not 1-1.  But I still think this is an improvement.

To do:
- [x] Discussion whether this is reasonable / assess impact on other code
- [x] Remove commented code
- [x] Remove `wrapTableIterators` module


jamescheney:
@SimonJF could you have a look to see how painful this is to accommodate in GtoPdb?

@slindley I've come to the conculsion that #838 was an unsustainable half measure.  So I'd rather either do this or leave as is (i.e. keep the inconsistent treatment of `<--` and allow use of `asList` if we really want to work with intermediate nested structure.)

slindley:
@jamescheney I agree.

I'm happy to either stick with the current elaboration or to change the behaviour of `<--` to not insert a query providing we're stricter about the effect typing on `<--` and `asList` in order to prevent accidentally loading in whole tables.

The intermediate half-measure is not viable.

On balance, I think I am persuaded that thinking about where queries happen is important enough that it is a good idea to insist that the programmer always makes this explicit - so we should enact #841 and strive to avoid implicitly including `query` in any syntactic sugar.

jamescheney:
OK, sounds good.  As we (hopefully) further extend the query capabilities I think being explicit about where queries may / cannot happen will help.  Can you / @SimonJF have a look to see if there are any unintended consequences on the tests or GtoPdb?

SimonJF:
Sorry, missed this. Will have a look wrt. GtoPdb now.

SimonJF:
This seems too restrictive at present. Take the following for example:

```
var furtherReading =
          for (fr <-- Tables.grac_further_reading)
            where (fr.family_id == familyId)
            for (ref <-- Tables.reference)
              where (ref.reference_id == fr.reference_id)
              [
                (makeReference(ref.reference_id,
                   ref.type, ref.title, ref.article_title, ref.year,
                   ref.volume, ref.issue, ref.pages, ref.publisher,
                   ref.publisher_address, ref.editors, ref.pubmed_id,
                   ref.isbn, ref.authors, fr.key_ref))
              ];
        var furtherReading =
          sortBy(fun(ref) { ref.authors }, furtherReading);
```

This is an "in the wild" instance of a "bare" `<--`. Adding the `query` block gets me:

```
        var furtherReading =
          query {
            for (fr <-- Tables.grac_further_reading)
              where (fr.family_id == familyId)
              for (ref <-- Tables.reference)
                where (ref.reference_id == fr.reference_id)
                [
                  (makeReference(ref.reference_id,
                     ref.type, ref.title, ref.article_title, ref.year,
                     ref.volume, ref.issue, ref.pages, ref.publisher,
                     ref.publisher_address, ref.editors, ref.pubmed_id,
                     ref.isbn, ref.authors, fr.key_ref))
                ]
          };
        var furtherReading =
          sortBy(fun(ref) { ref.authors }, furtherReading);
```

However, I still get an error:

```
[simon@dazzle skye-gtopdb]$ ./run.sh 
Unification error: Rows
 
and
 wild|_
 could not be unified because the former is rigid and the latter contains fields not present in the former, namely `wild'.
src/web/familyDisplay.links:430: Type error: The function
    `sortBy'
has type
    `(((...)) ~a~> String, [(...)]) ~a~> [(...)]'
    `((...)) ~a~> String'
and
    `[(...)]'
    `'
In expression: sortBy(fun(ref) { ref.authors }, furtherReading).
```

jamescheney:
That's strange.  Are you sure the reason for the error is not the earlier unguarded `for` expression:
```
var basicInfo =
      for (f <-- Tables.family)
        where (f.family_id == familyId)
        [(name = f.name, type = f.type,
            grac_texts =
              for (g <-- Tables.grac_family_text)
                where (g.family_id == familyId)
                [g])];
```
which will force the whole function body to be tame?

jamescheney:
(or the other later unguarded queries `subcommitteeMembers` and `additionalContributors`)?

SimonJF:
Ah, of course! Thanks for pointing that out -- I'd missed it.

The actual edits I needed to make to GtoPdb were minimal. I was taking advantage of the automatic insertion in places, but nothing really changed modulo a few query blocks being added (you can see the diffs in the repository).

So I'd be happy with this.

jamescheney:
OK. Sam seems happy also so will merge after #852.

jamescheney:
Obviously the error message could be more helpful when this happens also...

PR #845, opened 2020-06-08
Fix Issue 833

Emanon42:
This attempts using `Module Buffer` instead of native `^` string concatenation to implement `string_of_query` for more efficient SQL query building.


Emanon42:
Thanks for review! I’m checking PP.ml.

jamescheney:
Having looked around a little more, I'd suggest using OCaml's Format library if possible, which is standard and better documented, to avoid increasing the number of uses of PP.ml.

https://caml.inria.fr/pub/docs/manual-ocaml/libref/Format.html

There is already code doing some query serialization using Format in `lens/database.ml`.

jamescheney:
@Emanon42, @SimonJF, is there any progress on this?

SimonJF:
Still ongoing -- @Emanon42 and I are meeting later today to make some more progress on it.

Emanon42:
Still working on it - it needs more work than my expectation.
Also, the `Format` API is quite terrible.

rudihorn:
@Emanon42 https://github.com/links-lang/links/blob/master/lens/database.ml has an example of how to use the `Format` libraries `fprintf` function for doing something quite close to what you are doing.

core/query/sql.ml
```diff
@@ -150,157 +150,196 @@ let order_by_clause n =
...
+        pr_delete ppf del_table del_where
+    | Update { upd_table; upd_fields; upd_where } ->
+        pr_update ppf upd_table upd_fields upd_where
+    | Insert { ins_table; ins_fields; ins_records } ->
+        pr_insert ppf ins_table ins_fields ins_records
+    | With (_, q, z, q') ->
+        match q' with
+        | Select (fields, tables, condition, os) ->
+            (* Inline the query *)
+            let tables = List.map (fun (t, x) -> buffer_concat [quote t; " as "; (string_of_table_var x)]) tables in
```
rudihorn, line 258: You can rewrite this buffer concat as something like:
`Format.asprintf "%a" pp v`.

core/query/sql.ml
```diff
@@ -150,157 +150,196 @@ let order_by_clause n =
...
-      | Apply ("RLIKE", [v; w]) -> buf_add "("; sb v; buf_add ")"; buf_add " RLIKE "; buf_add "("; sb w; buf_add ")"
-      | Apply ("LIKE", [v; w]) -> buf_add "("; sb v; buf_add ")"; buf_add " LIKE "; buf_add "("; sb w; buf_add ")"
-      | Apply (f, args) when SqlFuns.is f -> buf_add (SqlFuns.name f); buf_add "("; buf_mapstrcat buf args (fun x -> sb x) ","; buf_add ")"
-      | Apply (f, args) -> buf_add f; buf_add "("; buf_mapstrcat buf args (fun x -> sb x) ","; buf_add ")"
-      | Empty q -> buf_add "not exists ("; string_of_query buf quote true q; buf_add ")"
-      | Length q -> buf_add "select count(*) from ("; string_of_query buf quote true q; buf_add ") as "; buf_add (fresh_dummy_var ())
-      | RowNumber [] -> buf_add "1"
-      | RowNumber ps ->
-        buf_add "row_number() over (order by "; buf_add (String.concat "," (List.map (string_of_projection quote one_table) ps)); buf_add ")"
+      | Apply ("not", [Empty q]) -> Format.fprintf ppf "exists (%a)" pr_q q
```
rudihorn, line 300: It might be nice to introduce a constant like:
```ocaml
let binary_ops = StringSet.of_list ["&&", "||", "==", ...]
let binary_map op =
  match op with 
  | "&&" -> "and"
  | "||" -> "or"
  | "==" -> "="
  | _ -> ""
```

And then in your code you can just have:
```ocaml
 | Apply (op, [v; w]) when StringSet.mem op binary_ops -> Format.fprintf ppf "(%a) %s (%a)" pr_b v (binary_map op) pr_b w
```

It might also be nice to do the same with unary operators.

core/query/sql.ml
```diff
@@ -150,157 +150,196 @@ let order_by_clause n =
...
+      | RowNumber ps -> Format.fprintf ppf "row_number() over (order by %a)" pp_constant (String.concat "," (List.map (string_of_projection quote one_table) ps))
 
 let string_of_base quote one_table b =
   let buf = Buffer.create 0 in
-  string_of_base buf quote one_table b; Buffer.contents buf
+  let ppf = Format.formatter_of_buffer buf in
+  pr_base ppf buf quote one_table b; Format.pp_print_flush ppf (); Buffer.contents buf
 
 let string_of_query ?(range=None) quote q =
+  let pp_constant ppf s = Format.fprintf ppf "%s" s in
```
rudihorn, line 332: Instead of creating a buffer etc., I would suggest just using `Format.asprintf "%a" pp v`.

core/query/sql.ml
```diff
@@ -150,157 +150,196 @@ let order_by_clause n =
...
+            let tables = List.map (fun (t, x) -> buffer_concat [quote t; " as "; (string_of_table_var x)]) tables in
+            let buf2 = Buffer.create 0 in
+            let ppf2 = Format.formatter_of_buffer buf2 in
+            let q2 = pr_query ppf2 buf2 quote ignore_fields q; Format.pp_print_flush ppf2 (); Buffer.contents buf2 in
+            let q = buffer_concat ["("; q2; ") as "; string_of_table_var z] in
+            pr_select ppf fields (q::tables) condition os
+        | _ -> assert false
 
-and string_of_base buf quote one_table b =
+and pr_base ppf buf quote one_table b =
```
rudihorn, line 266: This function still takes a buffer argument. 
I would also try to have functions of the type signature `val myformatter : <args> -> formatter -> <value> -> unit`. This way your formatter can always be used within other format statements (i.e. `Format.fprintf f "%a" (myformatter <args>) val`).

Emanon42: Changed. Also, is explicitly declaring the type sign for different formatter functions in `mli` a good idea? Currently these formatter functions are only used in `sql.ml`. Not sure if somebody would require them outside.

rudihorn: Its sufficient to only document the public ones used outside of the `sql` module.

core/query/sql.ml
```diff
@@ -137,138 +152,163 @@ let order_by_clause n =
...
+    match option with
+      | None -> ()
+      | Some x -> Format.fprintf ppf fmt_str f x
+  in
+
+  let pr_fields ppf fields =
+    let pp_field ppf (b, l) = gen_pp_pair ppf "(%a) as %a" pr_b pp_constant (b, l) in
     if ignore_fields then
-      "0 as \"@unit@\"" (* SQL doesn't support empty records! *)
+      Format.fprintf ppf "%a" pp_constant "0 as \"@unit@\"" (* SQL doesn't support empty records! *)
```
jamescheney, line 158: This is equivalent to `pp_constant ppf "0 as \"@unit@\""` which is in turn equivalent to `Format.pp_print_string ppf "0 as \"@unit@\""`.

core/query/sql.ml
```diff
@@ -137,138 +152,163 @@ let order_by_clause n =
...
     else
       match fields with
-        | [] -> "0 as \"@unit@\"" (* SQL doesn't support empty records! *)
-        | fields ->
-          mapstrcat ","
-            (fun (b, l) ->
-              "(" ^ sb b ^ ") as "^ quote l) (* string_of_label l) *)
-            fields
+        | [] -> Format.fprintf ppf "%a" pp_constant "0 as \"@unit@\"" (* SQL doesn't support empty records! *)
```
jamescheney, line 161: In general, `Format.fprintf ppf "%a" pp_x x` is equivalent to `pp_x ppf x`, except that the former does more work.

core/query/sql.ml
```diff
@@ -137,138 +152,163 @@ let order_by_clause n =
...
+
+  let pr_select ppf fields tables condition os =
+    let pp_os_condition ppf a = Format.fprintf ppf "%a" pr_b a in
+    let pp_orderby ppf os =
       match os with
-        | [] -> ""
-        | _ -> " order by " ^ mapstrcat "," sb os in
-    let where =
+        | [] -> ()
+        | _ -> Format.fprintf ppf " order by %a" (Format.pp_print_list ~pp_sep:pp_comma pp_os_condition) os in
```
jamescheney, line 170: You could also make an abbreviation for `Format.pp_print_list ~pp_sep:pp_comma` since this is used in several places.  Perhaps called `pp_comma_separated`.

core/query/sql.ml
```diff
@@ -137,138 +152,162 @@ let order_by_clause n =
...
-      List.map (fun (k, v) -> quote k ^ " = " ^ sbt v) fields
-      |> String.concat ", " in
 
-    let where =
-      OptionUtils.opt_app
-        (fun x -> "where (" ^ sbt x ^ ")") "" where in
-    Printf.sprintf "update %s set %s %s" table fields where
+  let pr_delete ppf table where =
+    let pp_where ppf where = gen_pp_option ppf "where (%a)" pr_b_ignore_fields where in
+    Format.fprintf ppf "delete from %a%a" pp_constant table pp_where where
```
jamescheney, line 185: If the table name is foo, this will generate `"delete from foowhere ...`.

core/query/sql.ml
```diff
@@ -137,138 +152,162 @@ let order_by_clause n =
...
-      values
-      (* Concatenate and bracket the values in each row *)
-      |> List.map ((List.map sbt) ->- String.concat ", " ->- Printf.sprintf "(%s)")
-      (* Join all rows *)
-      |> String.concat ", " in
-    Printf.sprintf "insert into %s (%s) values %s"
-      table fields values
+  let pr_update ppf table fields where =
+    let pp_field ppf (k, v) = gen_pp_pair ppf "%a = %a" pp_quote pr_b_ignore_fields (k, v) in
+    let pp_where ppf where = gen_pp_option ppf "where (%a)" pr_b_ignore_fields where in
```
jamescheney, line 189: This code is identical to the definition above for `delete` and could be reused.

core/query/sql.ml
```diff
@@ -137,138 +152,162 @@ let order_by_clause n =
...
-              (* Inline the query *)
-              let tables = List.map (fun (t, x) -> quote t ^ " as " ^ (string_of_table_var x)) tables in
-              let q = "(" ^ sq q ^ ") as " ^ string_of_table_var z in
-              string_of_select fields (q::tables) condition os
-          | _ -> assert false
+  let pr_insert ppf table fields values =
+    let pp_value ppf x = Format.fprintf ppf "(%a)" (Format.pp_print_list ~pp_sep:pp_comma pr_b_ignore_fields) x in
+    Format.fprintf ppf "insert into %a (%a) values %a"
+      pp_constant table
+      (Format.pp_print_list ~pp_sep:pp_comma Format.pp_print_string) fields
```
jamescheney, line 196: use pp_constant / pp_string here instead of Format.pp_print_string for consistency?

core/query/sql.ml
```diff
@@ -69,6 +71,19 @@ let string_of_label label =
   else
     label
 
+(* concatenation implement with Buffer module*)
```
SimonJF, line 74: This can be removed

core/query/sql.ml
```diff
@@ -137,138 +152,162 @@ let order_by_clause n =
...
+let rec pr_query ppf quote ignore_fields q =
+  let pp_quote ppf q = Format.fprintf ppf "%s" (quote q) in
+
+  let pr_q ppf q = pr_query ppf quote ignore_fields q in
+  let pr_b ppf q = pr_base ppf quote false q in
+  let pr_b_ignore_fields ppf q = pr_base ppf quote true q in
+
+  let gen_pp_pair ppf fmt_str fl fr (l, r) = Format.fprintf ppf fmt_str fl l fr r in
+  let gen_pp_option ppf fmt_str f option =
+    match option with
```
SimonJF, line 167: You could probably use `OptionUtils.opt_iter` here (defined in `Utililty`) rather than doing the split explicitly

core/query/sql.ml
```diff
@@ -137,138 +152,162 @@ let order_by_clause n =
...
+    | Select (fields, [], Constant (Constant.Bool true), _os) ->
+      Format.fprintf ppf "select %a" pr_fields fields
+    | Select (fields, [], condition, _os) ->
+      Format.fprintf ppf "select * from (select %a) as %a where %a"
+        pr_fields fields
+        pp_constant (fresh_dummy_var ())
+        pr_b condition
+    | Select (fields, tables, condition, os) ->
+        (* using quote_field assumes tables contains table names (not nested queries) *)
+        let tables = List.map (fun (t, x) -> Format.asprintf "%a as %s" pp_quote t (string_of_table_var x)) tables in
```
jamescheney, line 217: This is OK for now, but the use of asprintf to serialize an intermediate pretty-printed value to a string and then re-pretty-print it is a symptom of a deeper problem here: the type of `Select` forces tables to be a list of pairs of string table names and variables, but in general selection queries can have subqueries instead of plain table names.  This is not something you have to fix, but I'd appreciate if you could put a FIXME comment here as a reminder to revisit this.

Also, the definition of `tables` is repeated verbatim in the `With` case, so could be hoisted out as a function and reused.

core/query/sql.ml
```diff
@@ -137,138 +152,162 @@ let order_by_clause n =
...
+        pr_delete ppf del_table del_where
+    | Update { upd_table; upd_fields; upd_where } ->
+        pr_update ppf upd_table upd_fields upd_where
+    | Insert { ins_table; ins_fields; ins_records } ->
+        pr_insert ppf ins_table ins_fields ins_records
+    | With (_, q, z, q') ->
+        match q' with
+        | Select (fields, tables, condition, os) ->
+            (* Inline the query *)
+            let tables = List.map (fun (t, x) -> Format.asprintf "%a as %s" pp_quote t (string_of_table_var x)) tables in
```
jamescheney, line 229: I'm surprised by this!  I hadn't noticed before that we were pretty-printing `With` queries by inlining the let-bound query into the select clause.  We should really be emitting standard SQL WITH expressions from this AST, and if we want to inline such queries we can do that as a separate preprocessing step.  But the type of select_clause is not general enough to permit that.  Not something you need to fix, but please put a `FIXME` comment here.

core/query/sql.ml
```diff
@@ -137,138 +152,162 @@ let order_by_clause n =
...
-and string_of_base quote one_table b =
-  let sb = string_of_base quote one_table in
+and pr_base ppf quote one_table b =
+  let string_of_projection quote one_table (var, label) =
+  if one_table then
+    quote label
+  else
+    buffer_concat [string_of_table_var var; "."; (quote label)]
+  in
+  let pr_b ppf b = pr_base ppf quote one_table b in
```
jamescheney, line 242: The definitions of pr_b and pr_q here shadow earlier ones, which might be confusing.

core/query/sql.ml
```diff
@@ -137,138 +152,162 @@ let order_by_clause n =
...
+      | "||" -> "or"
+      | "==" -> "="
+      | _ -> op
+  in
+  let unary_map op =
+    match op with
+      | "floatToInt" -> "floor"
+      | "not" -> "not "
+      | "negate" | "negatef" -> "-"
+      | _ -> ""
```
jamescheney, line 258: This is because the other operations are type conversions, but looks like a potential bug.  (and is fragile: it depends on the fact that all the operations aside from floatToint, not and negate correspond to silent type conversions in SQL, and could become a bug if we add other unary operations for which this isn't the case).  Could you add a comment or make the pattern match explicit so that it's more obvious that "" is an appropriate translation for the remaining operations?

core/query/sql.ml
```diff
@@ -137,138 +152,162 @@ let order_by_clause n =
...
-          "case when " ^ sb c ^ " then " ^sb t ^ " else "^ sb e ^ " end"
-      | Constant c -> Constant.to_string c
-      | Project (var, label) -> string_of_projection quote one_table (var, label)
-      | Apply (op, [l; r]) when Arithmetic.is op
-          -> Arithmetic.gen (sb l, op, sb r)
-      | Apply (("intToString" | "stringToInt" | "intToFloat" | "floatToString"
-                | "stringToFloat"), [v]) -> sb v
-      | Apply ("floatToInt", [v]) -> "floor("^sb v^")"
+      | Case (c, t, e) -> Format.fprintf ppf "case when %a then %a else %a end" pr_b c pr_b t pr_b e
+      | Constant c -> Format.fprintf ppf "%a" pp_constant (Constant.to_string c)
```
jamescheney, line 262: Redundant `Format.fprintf ppf "%a"` here and next line

core/query/sql.ml
```diff
@@ -137,138 +152,162 @@ let order_by_clause n =
...
+      | Project (var, label) -> Format.fprintf ppf "%a" pp_constant (string_of_projection quote one_table (var, label))
+      | Apply (op, [l; r]) when Arithmetic.is op ->
+        let pr_b ppf b = pr_base ppf quote one_table b in
+        let l = Format.asprintf "%a" pr_b l in
```
jamescheney, line 266: Fine for now but asprintf should eventually be avoided

core/query/sql.ml
```diff
@@ -137,138 +152,162 @@ let order_by_clause n =
...
+        let r = Format.asprintf "%a" pr_b r in
+        Format.fprintf ppf "%a" pp_constant (Arithmetic.gen (l, op, r))
+      | Apply ("not", [Empty q]) -> Format.fprintf ppf "exists (%a)" pr_q q
+      | Apply (uop, [v]) when StringSet.mem uop unary_ops -> Format.fprintf ppf "%s(%a)" (unary_map uop) pr_b v
 
       (* optimisation *)
-      | Apply ("not", [Empty q]) -> "exists (" ^ string_of_query quote true q ^ ")"
+      | Apply (op, [v; w]) when StringSet.mem op binary_ops -> Format.fprintf ppf "(%a) %s (%a)" pr_b v (binary_map op) pr_b w
+      | Apply (f, args) when SqlFuns.is f ->
+        let pp_value ppf x = Format.fprintf ppf "%a" pr_b x in
```
jamescheney, line 275: Redundant `Format.fprintf ppf "%a"` here and on line 295

core/query/sql.ml
```diff
@@ -137,138 +152,162 @@ let order_by_clause n =
...
-      | Apply (op, [l; r]) when Arithmetic.is op
-          -> Arithmetic.gen (sb l, op, sb r)
-      | Apply (("intToString" | "stringToInt" | "intToFloat" | "floatToString"
-                | "stringToFloat"), [v]) -> sb v
-      | Apply ("floatToInt", [v]) -> "floor("^sb v^")"
+      | Case (c, t, e) -> Format.fprintf ppf "case when %a then %a else %a end" pr_b c pr_b t pr_b e
+      | Constant c -> Format.fprintf ppf "%a" pp_constant (Constant.to_string c)
+      | Project (var, label) -> Format.fprintf ppf "%a" pp_constant (string_of_projection quote one_table (var, label))
+      | Apply (op, [l; r]) when Arithmetic.is op ->
+        let pr_b ppf b = pr_base ppf quote one_table b in
```
jamescheney, line 265: Another (redundant?) shadow of pr_b.

core/query/sql.ml
```diff
@@ -137,138 +152,162 @@ let order_by_clause n =
...
-      | RowNumber [] -> "1"
-      | RowNumber ps ->
-        "row_number() over (order by " ^ String.concat "," (List.map (string_of_projection quote one_table) ps) ^ ")"
-and string_of_projection quote one_table (var, label) =
-  if one_table then
-    quote label
-  else
-    string_of_table_var var ^ "." ^ (quote label)
+let string_of_base quote one_table b =
+  let pr_b ppf b = pr_base ppf quote one_table b in
```
jamescheney, line 286: This definition of pr_b is used exactly once.

core/query/sql.ml
```diff
@@ -137,138 +152,162 @@ let order_by_clause n =
...
+  Format.asprintf "%a" pr_b b
 
 let string_of_query ?(range=None) quote q =
-  let range =
+  let pr_q ppf q = pr_query ppf quote false q in
```
jamescheney, line 290: This definition of pr_q is used exactly once.

It might be an idea to rearrange the arguments to pr_query to be pr_query quote b ppf q, so that instead of using this custom definition of pp_q (or having to write something like (fun ppf -> pr_query ppf quote false q) you can write

```
Format.asprintf "%a%a" (pr_query quote false) q
```
that is, so that the last two arguments are ppf and q and the behavior-modifying  parameters come before ppf.

core/query/sql.ml
Emanon42: `Arithmetic.gen` requires `string` as arguments. I think `asprintf` is necessary to generate intermediate results here?

core/query/sql.ml
```diff
@@ -30,6 +30,8 @@ and base =
   | RowNumber of (Var.var * string) list
     [@@deriving show]
 
+type 'a fmt_fn = Format.formatter -> 'a -> unit
```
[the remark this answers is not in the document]
Emanon42: I am planning to add type constraint for all my `pp_` functions.

core/query/sql.ml
SimonJF: It's a little worrying that this wasn't picked up by the CI...

core/query/sql.ml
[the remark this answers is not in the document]
jamescheney: OK, that's fine.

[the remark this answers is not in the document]
jamescheney: OK

SimonJF:
Thanks, @Emanon42, this is looking in good shape now. I've tested with my GtoPdb application and everything works as expected. If you could fix @jamescheney's comments, that would be excellent.

Since the discussion also highlighted a hole in our CI, I'd be really grateful if you could add a deletion test with a `where` clause to `tests/database/factorials.links` too, just so that code path's exercised, but then I think we're good to go.

jamescheney:
> if you could add a deletion test with a where clause

Good point.  Equally, we could be doing unit testing of pretty-printing of SQL examples using ounit, much like @rudihorn's code for unit testing of the lens subsystem (though not asking for this here).

jamescheney:
Looks good!  Is this ready to merge now?

Emanon42:
> Looks good! Is this ready to merge now?

yes

Issue #792, opened 2019-12-02
Azure Mac CI breaks during database setup


frank-emrich:
This [build](https://dev.azure.com/links-lang/links/_build/results?buildId=621) failed repeatedly with the following error. Those changes are unrelated to DB-stuff; I guess master would equally fail.

```
cmnd=psql -h localhost -p 5432 -v ON_ERROR_STOP=1 -q -d links -f tests/shredding/travel_portal.sql
psql:tests/shredding/travel_portal.sql:1: NOTICE:  table "agencies" does not exist, skipping
psql:tests/shredding/travel_portal.sql:2: NOTICE:  table "externaltours" does not exist, skipping
psql:tests/shredding/travel_portal.sql:9: ERROR:  syntax error at or near "OIDS"
LINE 6: ) WITH OIDS;
               ^
```

Maybe the mac setup uses a new version of postgres that's not compatible with our script anymore?

slindley:
We should either fix this ASAP or remove the Azure Mac CI.

Is anyone attempting to maintain the Mac CI?

If a Mac user isn't actively maintaining this CI, then I suspect keeping it may cause us more trouble than it's worth.

jamescheney:
As "the"  user of Links on mac, I have had nothing to do with this so far.  I don't think fixing it requires Mac know-how.  I suspect the reason things broke is that the brew postgres package just went to postgres 12.  Pinning a specific stable version that Links is believed to be compatible with (such as 9.5) ought to work.  I'll see if my minimal knowledge/competence to edit the Azure configuration suffices, but if not I don't object to retiring the Mac CI.

jamescheney:
The reason appears to be that prior to Postgres 12, if you wanted OIDs to be added to a table then you have to explicitly say WITH OIDS.  As of version 12, you don't have to say this and it happens automatically, and they retired the now-redundant syntax non-backwards-compatibly.  So, for this exampel to work on both <12 and >=12 versions will require conditional code depending on the Postgres version.

This will presumably also affect the Azure Linux config at whatever point the default version of Postgres you get increases high enough.  At that point, we should probably just stop supporting versions < 12.

jamescheney:
[...]

PR #846, opened 2020-06-10
Fix #575

### LINKED ISSUE #575: "Attempt to split row" IR typing error produced by `isInt` Prelude function
(results below were obtained from commit dccf4aa, currently in #572 PR)

`isInt` function from Prelude:

```
sig isInt : (String) -> Bool
fun isInt (x) { x =~ /-?[0-9]+$/ }
```

causes a ridiculously long "Attempt to split row" IR typing error that I'm not going to quote here.  Minimal example I was able to come up with is:

```
fun f (x) { x =~ /0+$/ }
```

Config settings:

```
typecheck_ir=true
prelude=empty.links
print_types_pretty=false
optimise=false
debug=true
hide_fresh_type_vars=false
show_flavours=true
fail_on_ir_type_error=true
```

Error goes away with `optimise=true`.

jamescheney:
There were three problems:
- Aliases were treated opaquely in some places in irCheck --> call TypeUtils.concrete_type in more places.
- Question, plus, star flags were ascribed Regex type --> introduced Repeat type alias and use that instead.
- There was a stray extra list constructor --> removed

This seems not to break anything, but I'm not sure if I have run all of the ir typing regression tests.  @jstolarek ?

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -87,7 +87,7 @@ steps:
       case $AGENT_OS in
         "Linux") sudo apt-get install postgresql libpq-dev sqlite3 libsqlite3-dev &&
                  sudo LINKS_CI=1 tools/configure-postgresql &&
-                 sudo systemctl reload postgresql &&
+                 sudo systemctl start postgresql &&
                  sudo -u postgres createuser -s $(whoami) &&
                  sudo -u postgres createdb -O $(whoami) links;;
         "Darwin") brew install postgresql@9.5 libpq sqlite3 &&
--- a/core/desugarRegexes.ml
+++ b/core/desugarRegexes.ml
@@ -21,15 +21,15 @@ let star_str         = "Star"
 let plus_str         = "Plus"
 let question_str     = "Question"
 
-let desugar_regex phrase regex_type regex : phrase =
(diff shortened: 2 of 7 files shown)


jamescheney:
I'd like to merge this now, since it (a) seems to fix #575 without introducing any other problems and (b) contains the fix to the azure setup needed for portability across postgres versions.  (This might also have fixed #792 had we thought of it at the time... I think I had assumed that the WITH OIDS behavior was integral to the travel portal example, but it is only being used for provenance which isn't being tested.)

Any objections?  Janek maybe you could take a quick look and let me know if there are any IR typing ramifications?  As far as I can tell all of the tests in typed_ir.tests are still passing.  I'll add a test for `isInt`.

jstolarek:
I'm looking at this at the moment, just making sure everything works as expected. Can we merge this after the Links meeting?

jamescheney:
No hurry as far as I am concerned.  (but note the meeting had to be pushed back to tomorrow).

jstolarek:
Yes, I just caught up with email backlog. I'll finish reviewing this by noon. Tests pass without problems but I want to have a moment to look at the code.

PR #847, opened 2020-06-11
Test for a clean clone as part of CI

### LINKED ISSUE #507: Test for a clean clone as part of CI
Following discussion in #503, I think that CI should test whether the working tree in a cloned Links repo is clean. This will catch situations where we get the `.gitattributes` wrong and files are unexpectedly converted on clone. @SimonJF, you are familiar with CI - would it be a difficult thing to do? `git diff-index --quiet HEAD --` is the command to call. It returns 0 if the working tree is clean and 1 if it is not.

jstolarek:
What the title says. If the working tree isn't clean after cloning (e.g. because newlines are automatically converted in a way they shouldn't -see #5030) CI will catch it. Closes #507.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ language: c
 services:
   - postgresql
 before_install:
+  - git diff-index --quiet HEAD --
   - sudo add-apt-repository -y ppa:avsm/ppa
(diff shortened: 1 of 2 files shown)


jstolarek:
Ready for review.

Issue #522, opened 2019-03-11
Preserve aliases when dualising types


SimonJF:
In `dual_type` in `types.ml`, we have:

```
      (* it sometimes seems tempting to preserve aliases here, but it
         won't always work - e.g. when we use dual_type to expose a
         concrete type *)
      (* | `Alias _ as t         -> `Dual t *)
      (* Still, we might hope to find a way of preserving 'dual
         aliases' in order to simplify the pretty-printing of types... *)
      | `Alias (_, t) -> dt t
```

This has the disadvantage of making the type errors more difficult to read. It would be good to refactor things such that aliases can be preserved when dualised.

jamescheney:
So just leaving the `Dual` constructor in the code (as in the commented-out version) looks like it would mostly just break things in unification, where we currently pattern match against the result to determine whether something is a dualized type variable.  

What about adding a "dual" flag to Alias which could be used to short-circuit type printing  in error messages, and defining `dual_type` to preserve the Alias constructor (with the flag flipped) and recursively dualize the body?

This would necessitate doing an additional `Types.concrete_type` call in a few places where `dual_type` is called, but that seems like an acceptable cost.  Note that `Types.concrete_type` calls `dual_type` in the `Dual` case though.

PR #848, opened 2020-06-12
Attempt at retaining dual aliases (fixing #522)

jamescheney:
This appears to pass all tests and keeps aliases in the type syntax with an extra flag indicating whether the alias is dualized.  This should have no impact on typechecking, but just means that aliases don't evaporate when a type is dualized and when we print out a type containing residual dualized aliases we print the alias with `~` before it rather than expanding out the definition.

It would be helpful to have a test case illustrating the desired error message matches what @SimonJF had in mind, since my familiarity with session types/error messages is not sufficient to come up with a simple example.

It is also possible that this change breaks something non-obvious (that is not exercised by the test suite).  In particular I'm not especially confident that this change does not break unification somehow.  I would hope that this is dealt with correctly already by the code that just discards `Alias` nodes during unification.

WIP for discussion after #637.

### DIFF
--- a/core/desugarPages.ml
+++ b/core/desugarPages.ml
@@ -51,7 +51,7 @@ let rec desugar_page (o, page_type) =
             let formlet_type = Types.concrete_type formlet_type in
             let a = Types.fresh_type_variable (lin_any, res_any) in
             let b = Types.fresh_type_variable (lin_any, res_any) in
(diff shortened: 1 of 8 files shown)


slindley:
This will need to be redone once the types refactoring is merged, so it may make most sense to hold off pulling it into master and just use the changes as a guide for manually updating the refactored code.

jamescheney:
Closing since I screwed up merging.

PR #850, opened 2020-06-19
Fix154

### LINKED ISSUE #154: using "client" keyword breaks todo list example
If you try to run the todo list example using the appserver ([todo.links.txt](https://github.com/links-lang/links/files/876461/todo.links.txt)) then the presence/absence of the "client" keyword on the "todo" function seems to determine whether the program works correctly.  If the "client" keyword is used then the page renders as a line or so of gobbledegook, if it is omitted then everything is fine. 

I don't know whether this is a bug in the interpreter (i.e. client calls should be allowed in server page handlers and aren't handled correctly) or in the typechecker (i.e. client calls shouldn't be allowed in server page handlers but this isn't being checked).  It is also not clear to me whether this is related to the appserver at all, as opposed to realpages or something else.

This is observed with Links 0.6.1.  I'm guessing that someone (Sam) ran into and avoided this problem when updating todo.links, so perhaps it duplicates an already-known bug, but here's the issue if not.

Emanon42:
Fix #154 by raising runtime error when detecting the client calling when the server is still generating page

### DIFF
--- a/core/evalir.ml
+++ b/core/evalir.ml
@@ -126,8 +126,9 @@ struct
      Value.continuation ->
      Value.t list ->
      result =
-
        fun req_data name cont args ->
+         if not(Utility.is_ajax_call(RequestData.get_cgi_parameters req_data)) then
+           raise (Errors.RuntimeError "Tried to make a client call on the server.");
          if not(Settings.get Basicsettings.web_mode) then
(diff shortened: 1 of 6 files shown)


Emanon42:
sorry, PR from the wrong branch.

PR #851, opened 2020-06-19
Fix #154

### LINKED ISSUE #154: using "client" keyword breaks todo list example

Emanon42:
Fix #154 by raising runtime error when detecting the client-side RPC calling during page generating in the server-side.


core/utility.ml
```diff
@@ -1459,4 +1459,13 @@ struct
...
 
+(** remote client->server call *)
+let is_remote_call params =
+  mem_assoc "__name" params && mem_assoc "__args" params
 
+(** return __result from server->client call with server continuation __continuation *)
+let is_client_return params =
+  mem_assoc "__continuation" params && mem_assoc "__result" params
+
+let is_ajax_call cgi_args =
```
SimonJF, line 1470: If you're moving this from `webif.ml`, could you make sure you delete it from there, and fix up the references to it from other files?

(Maybe this would be better in `RequestData`, actually).

Emanon42: Sure

core/webif.ml
```diff
@@ -88,14 +80,11 @@ struct
       body ^
       "\n  </body></html>\n"
 
-  let is_ajax_call cgi_args =
-    (is_remote_call cgi_args) || (is_client_return cgi_args)
-
   (* jcheney: lifted from serve_request, to de-clutter *)
```
jamescheney, line 83: I think this ancient comment of mine can probably go now

core/utility.ml
```diff
@@ -1458,5 +1458,3 @@ struct
         (sequence xs) >>= fun xs ->
         Lwt.return (x :: xs)
 end
-
```
jamescheney, line 1461: Please avoid changing trailing whitespace for no reason

core/evalir.ml
```diff
@@ -128,6 +128,8 @@ struct
      result =
 
        fun req_data name cont args ->
+         if not(RequestData.is_ajax_call (RequestData.get_cgi_parameters req_data)) then
+           raise (Errors.runtime_error "Tried to make a client call on the server.");
```
jamescheney, line 132: This should be a call to a new `Error`, similar to `client_call_outside_webmode` that is raised below if a client function is called before the web server is started.

Emanon42: Changed to new `Error` type `ClientCallBeforeServerStart`

core/utility.ml
jamescheney: Actually, maybe it is better to remove both trailing whitespace lines, these probably should have been removed already.  In any case, there is no point in just removing one.

Emanon42: Is there a code style I can check?

core/errors.ml
```diff
@@ -186,6 +187,8 @@ let format_exception =
      pos_prefix (Printf.sprintf "Error: Cannot load plugin '%s' (link error: %s)\n" file (Dynlink.error_message err))
   | ClientCallOutsideWebMode fn ->
      pos_prefix (Printf.sprintf "Error: Cannot call client side function '%s' outside of web mode\n" fn)
+  | ClientCallBeforeServerStart fn ->
+     pos_prefix (Printf.sprintf "Error: Cannot call client side function '%s' before server page is ready\n" fn)
```
jamescheney, line 191: Not to nitpick, but perhaps this and the previous error could be merged into one `ForbiddenClientCall` that takes two parameters, `fn` the function name and `reason` the reason the client call is not possible.

Emanon42: Sure, it's better for possible future code reuse.

core/utility.ml
jamescheney: Supposedly the script `src/links/tools/rules/trailing_whitespace.sh` checks for trailing whitespace.  But it doesn't seem to object to newlines at the end of a file - I keep tripping over this because my editor is configured to remove them from ocaml files.

Here I'm just reacting to the fact that utility.ml seems to have been changed for no reason possibly because your editor is doing the same thing.  Probably we should standardize on removing spurious newlines from the ends of files (or at least not complain reflexively about them being removed like I have).

jamescheney:
I've resolved a conflict with the master branch due to #849, and reordered the checks so that we check for being in web mode before checking the request data (the other order was OK, but relies on the fact that the empty request data is not an ajax request; I think it's better to check for being in web mode before checking the request data.)  Will merge once CI passes.

core/errors.ml
```diff
@@ -184,8 +184,8 @@ let format_exception =
      pos_prefix (Printf.sprintf "Error: Cannot load plugin dependency '%s' (link error: %s)\n" file (Dynlink.error_message err))
   | LoadFailure (file, err) ->
      pos_prefix (Printf.sprintf "Error: Cannot load plugin '%s' (link error: %s)\n" file (Dynlink.error_message err))
-  | ClientCallOutsideWebMode fn ->
-     pos_prefix (Printf.sprintf "Error: Cannot call client side function '%s' outside of web mode\n" fn)
+  | ForbiddenClientCall (fn, reason) ->
```
jamescheney, line 187: One more tiny thing: could you remove the space at the end of this line?  This is why the CI is failing at the moment.

PR #852, opened 2020-06-22
fix #754 

### LINKED ISSUE #754: Regexes in nested queries do not work
Looking at the dictionary suggestion example, (`examples/dictionary/dictSuggest.links`), we can use regex syntax to translate to SQL `LIKE` predicates. However, this is unsupported when using nested queries.

MFE:

```
var db = database "test";

var testTable = table "testTable" with (field: String) from db;

query {
  for (x <-- testTable)
    where (x.field =~ /^{"foo"}.*/)
    [x]
}
```

yields

```
   (Query.Lang.Concat
      [(Query.Lang.Singleton
          (Query.Lang.Variant ("StartAnchor", (Query.Lang.Record {}))));
        (Query.Lang.Singleton
           (Query.Lang.Variant ("Quote",
              (Query.Lang.Variant ("Simply",
                 (Query.Lang.Constant (CommonTypes.Constant.String "foo"))))
              )));
        (Query.Lang.Singleton
           (Query.Lang.Variant ("Repeat",
              (Query.Lang.Record
                 {1 => (Query.Lang.Variant ("Star", (Query.Lang.Record {}))); 
                 2 => (Query.Lang.Variant ("Any", (Query.Lang.Record {}))); })
              )))
        ])
   ))
```

jamescheney:
This PR attempts to fix #754 by adding cases that recognize regular expression matching tests in the let-insertion and flattening stages of shredding.  Previously these were handled generically and failing because these stages didn't provide any cases to handle variants (which is how regular expressions are represented at runtime in the IR).  

This makes simple tests using regexes inside nested queries, like the one in #754 work, but does not make regular expression handling on the database any more complete/safe than it already was.

I've added debug messages to the two new cases in case we should be doing something more with the regex expressions, so that if we run into problems there will be a debugging message that gives more of a clue what the problem is.

### DIFF
--- a/core/query/evalNestedQuery.ml
+++ b/core/query/evalNestedQuery.ml
@@ -386,7 +386,6 @@ struct
       | Apply (Primitive "tilde", [s; r]) as e ->
           Debug.print ("Applying lins_inner to tilde expression: " ^ QL.show e);
           Apply (Primitive "tilde", [lins_inner (z, z_fields) ys s; r])
-
       | Apply (Primitive f, es) ->
         Apply (Primitive f, List.map (lins_inner (z, z_fields) ys) es)
       | Record fields ->
--- a/core/query/evalNestedQuery.ml
+++ b/core/query/evalNestedQuery.ml
@@ -804,3 +804,4 @@ let compile_shredded : Value.env -> Ir.computation
           let t = Q.type_of_expression v in
           let p = unordered_query_package t v in
             Some (db, p)
+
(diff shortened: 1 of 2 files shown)


jamescheney:
I added a test for this in `jrules.links`, is that sufficient?

SimonJF:
I was more thinking of something like 

```
sig marketers : () -> [(name:String, clients:[Int])]
fun marketers () {
  for (m <-- marketers_names)
    [(name=m.name, clients = for (x <-- mc) where (x.m == m.name && x.name =~ /a/) [x.c])]
}
```

just to make sure it works within nested queries, if possible. I'm sure it'll be the same and still work, but it's worth checking.

jamescheney:
OK,done.  Will merge when CI passes.

PR #854, opened 2020-06-22
Fix #808

### LINKED ISSUE #808: Winestore example can produce a runtime error
The Winestore example can produce the following runtime error
```
Links Error
***: Runtime error: range: argument less than zero
```
Steps to reproduce:
1. In the "filter wines by" section select:
   - Region: Barossa Valley
   - Wine type: Sweet
2. Click "Show wines"

jamescheney:
Replaces errorful `range` function with builtin range function  `[0..n]`.  

This should be more correct (i.e. not leave off the final partial page) but I haven't tested it.  It may still have an off-by-one error if the number of wines is exactly divided by the page size.

But it should not result in an error no matter what, because `[0...n]` is defined (as the empty list) even if `n` is negative.


jamescheney:
Will update and merge after #858.

Issue #855, opened 2020-06-22
Implement INSERT RETURNING correctly on MySQL and SQLite3


jamescheney:
INSERT RETURNING isn't directly supported in MySQL or SQLite3.  The driver for the latter just fails if you try; for MySQL we attempt to simulate INSERT RETURNING by returning the last inserted rowid, which is correct iff the field we want to return is declared as autoincrement (or, happens to always equal the implicit rowid field because we were lucky), but we don't check this so it would be easy to write something that seems like it should work but doesn't.  

SQLite3 also supports a similar function with the same limitation, which block§ #322 because some of the tests use `returning`.

Moreover in general, INSERT RETURNING on Postgres supports returning any expression (or expression sequence) definable in terms of the inserted row's field names.  I think we can support this, either the special or general case, on other databases as follows:

- generate the insert directive
- generate a second query of the form `SELECT e1...en FROM inserted_into_table WHERE rowid = most_recently_inserted_rowid()`
- send both the insert and query in one transaction
- parse the result into a Links value.

We could keep the classic `returning i` syntax and just allow returning one field at a time (which is all we really need anyway), or generalize it to `returning (x.e)` where `e` is an arbitrary SQL-izable expression that is allowed to refer to the inserted row, `x`.  

- [x] Implement correct support for INSERT RETURNING of a single field for MySQL and SQLite3 
- [X] ~~Generalize to support returning arbitrary expressions?~~
- [X] ~~Arguably, `insert returning` makes no sense if we insert multiple (or zero) rows so maybe we should also rationalize the syntax and typing to forbid this at the same time.~~

jamescheney:
First part done in #858.

jamescheney:
Regarding limiting INSERT RETURNING to singletons, this is done already - the syntax only allows RETURNING if a single tuple is being inserted.

jamescheney:
And allowing returning an expression rather than variable would create lots of interesting parser conflicts.  I'm inclined to drop this feature request, hence, this will be closed by #858.

Issue #655, opened 2019-06-22
Support for queries with nulls


jamescheney:
Links queries assume a "pure" (2 valued logic, no nulls) semantics, whereas real SQL and databases use 3 valued logic and nulls.  Other LINQ systems, such as C# or F#, work around this either by allowing null references or using option types.  Neither approach is particularly appealing in Links because we'd rather not

This issue is to collect thoughts / design plans for how to handle nulls in Links in a principled way.  One particular issue is that Links's query normalization rules may not be sound in the presence of nulls.  Another is that within a query, we might want to interpret types and primitive operators in SQL's null-propagating way, but then there is an(other) impedance mismatch when we return the data to Links server space.  A third problem is that it seems  commonplace that `NOT NULL` is omitted from fields in SQL table definitions, even when actual nulls in these fields are rare.

I can think of several strategies:

* require all database tables we use to be null-free.  A nonstarter if we want to use Links on already-existing databases (such as GtoPDB).
* generalize #643 to allow defaults at any base type.  Not very elegant, and doesn't account for difference in semantics inside queries operating on nulls.
* Map any (nullable) base type values in input tables to options, then we can use `switch` inside queries to pattern match and eventually return queries with flat results.  Should work, provided we can translate any remaining `switch` statements in normal forms to SQL `IS NULL` and `CASE`, but seems likely to make queries much more verbose, even in common cases, if the underlying tables aren't declared with `NOT NULL` types.
* Alternatively, run queries as normal (with whatever adjustments are necessary for soundness in the presence of nulls), and convert any nullable result values to options when returning data to Links server space.  Con: still need awkward case analysis on query results.
* I have some other ideas that are long shots/research ideas.

If nothing else, we should certainly first be checking the database query results for nulls in a robust way, using the `ocaml-postgresql` or other driver APIs, rather than just assuming that a string value that doesn't match the expected datatype is a null.  (this certainly won't work for strings)

jamescheney:
Planned first step in this direction: change interface for returning values from the database to return string options (in OCaml) instead of punning NULL as the empty string.

jamescheney:
Concrete proposal:

1.  Add `null`, an expression that can be of any base type (in a databaseable effect), and null-sensitive functions such as `isNull`, `isTrue`, `isFalse`. maybe exact equality (which translates to SQL "NOT DISTINCT FROM".  These are constrained to only occur in queries/databaseable functions.
[...]

PR #858, opened 2020-06-22
Restore MySQL driver and fix ordering problem

### LINKED ISSUE #856: Ordering behavior is incorrect/inconsistent on some database drivers
`orderby` appears to produce inconsistent results (as already documented in `factorials.links`).  A minimal example:
```
var db = database "links";
var factorials = table "factorials" with (i : Int, f : Int) from db;

fun insertSome() {
  delete (x <-- factorials);
  insert factorials
  values (f, i)
     [(f=1, i=1),(f=2, i=2),(f=6, i=3)];
}
var asdf = insertSome();
query {
    for (row <-- factorials)
    where (row.i <= 5)
    orderby (row.i)
      [(i=row.i, f=row.f)]
  }
```
We get the result list in reverse order in Postgres and MySQL, while for SQLite3 the order is correct.

The SQL query that is generated for each database is the same, so presumably somewhere in the Postgres and MySQL database drivers the order of the returned results is being reversed.

This blocks #322 since it prevents us running the tests on all three databases and getting the same results.

jamescheney:
This PR restores the MySQL driver, a precondition for #322, and fixes #856 (albeit this could be cleaned up a bit).

Marked as WIP because the way query results are constructed by Sqlite3 is still a little inefficient and could be improved, and I'd like to make progress towards #855 also before merging, unless it turns out to be harder than I think.

- [x] Avoid unnecessary `List.rev`s in sqlite3 driver and value.ml
- [X] Basic support for INSERT RETURNING for all supported dbs.
- [x] Delete remaining dead code


jamescheney:
This fixes parts of #856 except for the extension to support RETURNING multiple columns which I leave as a todo.

Something horrible has happened so the PR / branch probably needs to be redone from a clean checkout of master.

SimonJF:
Regarding something horrible happening, have you tried:

> Here's a good workaround. Use the Edit button when viewing the PR in GitHub to change the base branch to something other than master. Then switch it back to master and it will now correctly show only the changes from the most recent commits.

(from the following link; there are also some other solutions)
 
https://stackoverflow.com/questions/16306012/github-pull-request-showing-commits-that-are-already-in-target-branch

jamescheney:
I seem to have found / remembered the incantation needed to get the branches synced up again, which was `git merge --no-ff master` - if that doesn't cause problems I'm inclined to leave as is.

SimonJF:
Ah, good to know!

jamescheney:
So the CI is failing because of an apparent regression on #835...

jamescheney:
I take that back, I only thought that because I had a stale build.  Looks like I just needed to adjust the factorials example again.

jamescheney:
I've decided to weaken #855 sufficiently that this fixes it.

jamescheney:
Now fixed sqlite3 inefficiency and realized that mysql driver suffered from a different problem (whose solution is similar).  Previously, we would randomly access the mysql library's result which was apparently slow.  Now, instead I load it all into an append-only buffer.  Possibly doing this for postgres results would be better and more uniform but this PR is already getting too big.

jamescheney:
The code for handling null integers in mySQL was broken, this is now fixed and I've added a database test that checks that null handling works correctly, as a first step towards #655.  This should be ready for review now.

core/database.ml
```diff
@@ -91,13 +90,10 @@ let execute_insert_returning returning q db =
...
               match result#status with
                | `QueryOk ->
-                   let rows = result#get_all_lst in
-                     begin
-                       match rows with
-                         | [[id]] -> Value.box_int (int_of_string id)
-                         | _ ->
-                             raise (runtime_error ("Returned the wrong number of results executing " ^ q))
-                     end
+                  if result#nfields == 1 && result#ntuples == 1
```
dhil, line 93: Structural equality `=`.

core/utility.ml
```diff
@@ -1458,3 +1458,66 @@ struct
...
+    then Array.get (Array.get buf.pages (i/buf.pagesize)) (i mod buf.pagesize)
+    else raise Not_found
+
+  let append buf x =
+    (* first, check if there is enough space or allocation is needed *)
+    if (buf.nextitem == buf.pagesize)
+    then begin
+      buf.nextitem <- 0;
+      buf.currpage <- buf.currpage+1;
+      if (buf.currpage == buf.numpages)
```
dhil, line 1509: Structural equality `=`.

core/utility.ml
```diff
@@ -1458,3 +1458,66 @@ struct
...
+                    currpage = 0;
+                    nextitem = 0;
+                    pages = Array.init n (fun _ -> Array.init m (fun _ -> x)) }
+
+  let length buf = buf.currpage*buf.pagesize + buf.nextitem
+
+  let set buf i x =
+  if 0 <= i && i < buf.currpage*buf.pagesize + buf.nextitem
+    then Array.set (Array.get buf.pages (i/buf.pagesize)) (i mod buf.pagesize) x
+    else raise Not_found
```
dhil, line 1496: Maybe use `raise (Invalid_argument "out of bounds")`.

core/utility.ml
```diff
@@ -1458,3 +1458,66 @@ struct
...
+    then Array.get (Array.get buf.pages (i/buf.pagesize)) (i mod buf.pagesize)
+    else raise Not_found
+
+  let append buf x = 
+    (* first, check if there is enough space or allocation is needed *)
+    if (buf.nextitem == buf.pagesize)
+    then begin 
+      buf.nextitem <- 0;
+      buf.currpage <- buf.currpage+1;
+      if (buf.currpage == buf.numpages)
```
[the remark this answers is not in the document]
jamescheney: Done.  (But structural equality on integers behaves the same as `==`.)

core/utility.ml
jamescheney: OK, matching what OCaml does with ordinary arrays.

jamescheney:
This should await #859.

jamescheney:
Miraculously, this seems to have commited with 859 and #861, so now that they are merged, will merge this once CI passes if no one stops me.

Issue #691, opened 2019-07-29
IR typing is unaware of unsafe signatures


jstolarek:
code:
```
unsafe sig map : ((a) -b-> c, [a]) -b-> [c]
fun map(f, l) {
  switch (l) {
    case []     -> []
    case hd::tl -> f(hd) :: map(f,tl)
  }
}
```
error:
```
***: IR Type Error: Type mismatch:
 Expected:
  `Function ((`Record (({1 => `Present (`MetaTypeVar (Point @ 0x7f707d60bd90 `Var (
                                                    (350,
                                                     (CommonTypes.Linearity.Unl,
                                                      CommonTypes.Restriction.Any),
                                                     `Rigid))));
                       },
                      Point @ 0x7f707dca2cc8 `Closed, false)),
            ({wild => `Present (`Record (({}, Point @ 0x7f707dca2cc8 `Closed,
                                          false)));
              },
             Point @ 0x7f707d60bf68 `Var ((351,
                                           (CommonTypes.Linearity.Unl,
                                            CommonTypes.Restriction.Any),
                                           `Rigid)),
             false),
            `MetaTypeVar (Point @ 0x7f707d60bff8 `Var ((352,
                                                        (CommonTypes.Linearity.Unl,
                                                         CommonTypes.Restriction.Any),
                                                        `Rigid)))))
 Actual:
  `Function ((`Record (({1 => `Present (`MetaTypeVar (Point @ 0x7f707d5b2400 `Var (
                                                    (350,
                                                     (CommonTypes.Linearity.Unl,
                                                      CommonTypes.Restriction.Any),
                                                     `Rigid))));
                       },
                      Point @ 0x7f707dca2cc8 `Closed, false)),
            ({},
             Point @ 0x7f707d5b2278 `Var ((351,
                                           (CommonTypes.Linearity.Unl,
                                            CommonTypes.Restriction.Any),
                                           `Rigid)),
             false),
            `MetaTypeVar (Point @ 0x7f707d5b22e8 `Var ((352,
                                                        (CommonTypes.Linearity.Unl,
                                                         CommonTypes.Restriction.Any),
                                                        `Rigid)))))
occurring in IR binding:
(Ir.Let (
   (230,
    (`Function ((`Record (({1 => `Present (`MetaTypeVar (Point @ 0x7f707d60bd90 `Var (
                                                         (350,
                                                          (CommonTypes.Linearity.Unl,
                                                           CommonTypes.Restriction.Any),
[...]

Issue #47, opened 2016-07-28
Make List a user-defined type constructor


slindley:
We could define List as a variant type. A possible disadvantage is that lists may be harder to optimise if they are not built in.

Issue #37, opened 2016-07-28
Pattern matching on lists fails in queries


slindley:
Pattern-matching is always typed as being compileable to the database, but pattern matching on lists compiles to functions `hd` and `tl` that are not yet compileable to the database.

jamescheney:
Is there a strong reason why list pattern matching has to be handled by desugaring to code with `hd`/`tl`?  Given that we may want to handle it differently depending on whether we are running on the server, client or normalizing queries before sending to the database, baking this implementation decision in at an early stage seems undesirable.  

What if instead we define a standard list datatype in the prelude / `defaultAliases.ml`, and desugar list pattern matching to that?  This would mean a little more work in irtojs if we want to detect places where we can translate list pattern matching to `hd`/`tl`, but it seems to me that that is the place where that implementation choice belongs.   It seems like this would help with #24 also.

jamescheney:
I think the problem is with desugaring (list) pattern matching to conditionals with any kind of partial accessor function, not `hd` and `tl` specifically.  `hd` and `tl` are not currently handled by the normalization rules (and make no sense in queries actually because within a query, `[a]` is usually an unordered collection).  But defining lists as a user-defined datatype (#47) and translating `hd`, `tl`, them to `Project` would not help, I think.

Pattern matching on datatypes (and datatype constructors) are allowed, but expected to completely normalize away at query-generation time.  This only works because there is no other way of creating something of type "datatype" inside a query aside from using a constructor, or referencing a variable of type "datatype" that will become a value with a head-constructor by the time normalization starts.  However, since we can construct things of list-type inside queries in other ways besides `Nil` and `Cons`,  pattern-matching on lists is a special case that is not handled correctly by the normalization algorithm.  So for example the following query expression 
```
query { tl(t) }
```
where t is a table, has no straightforward /deterministic  SQL equivalent.

I'm inclined to say that within queries, pattern-matching on lists specifically should be disallowed, i.e. `wild`, like `hd` and `tl` already are.  Or, maybe we should revisit the punning of `[a]` as "list on server/client but bag on the database".

PR #864, opened 2020-06-26
Proposal for simpler handling of unsafe sig and fix for #691

jamescheney:
This PR disables the previous code for handling unsafe signatures and implements a new approach:

- `unsafe sig foo` means that the declared identifier is not treated as recursive, for the purpose of effect typing; that is, recursive calls to `foo` are tame.
- type inference is otherwise as normal, and we do not (and do not need to) adjust the inferred type

It also adds treatment of unsafe-signature behavior to the IR.  This unfortunately touches a lot of code because the function definition data is a tuple rather than record.  In IR typechecking, we treat identifiers in recursive blocks as non-recursive if they are declared with `unsafe sig`.

Finally, a remaining problem is that pattern matching against List cons is translated (by `compilePatterns.ml`) to calls to `hd` and `tl`, which are wild.  This PR introduces a small hack to work around this: adding tame versions of `hd` and `tl` which have names that can never appear in Links source code, and are only used for desugaring of pattern matching.  This is slightly bad though because it would mean that these versions of `hd` and `tl` could be present in query code and the IR typechecker would allow this; but this is a problem we already have because of #37.  I have some longer-term ideas how to remove this related to #47 and #37.

### DIFF
--- a/core/typeSugar.ml
+++ b/core/typeSugar.ml
@@ -362,8 +362,6 @@ sig
   val bind_rec_annotation : griper
   val bind_rec_rec : griper
 
-(*XXX  val bind_unsafe_fun_annotation : griper *)
-
   val bind_exp : griper
 
   val list_pattern : griper
@@ -2444,6 +2442,28 @@ let resolve_type_annotation : Binder.with_pos -> Sugartypes.datatype' option -> 
       | Some (_, t) -> t
     end
   | t -> Some t
+
+(* NOTE: Unsafe signatures hack
+
+   The syntax "unsafe sig f : ty" asserts that f has type t, *provided* that recursive
+   calls to f are not treated as wild.  This is used to allow recursive definitions for
+   specific operations such as map, concatMap, filter, sortByBase which previously required
+   an even more awkward workaround.  "unsafe" should not be used outside of the prelude since it would
+   enable constructing queries that typecheck but cannot be translated to SQL and
+   perhaps we should check for this.
+
+   The initial implementation of "unsafe" ran type inference as normal, and then performed
+   surgery on the type.  Unfortunately this meant that other occurrences of the wild effect
+   arising because of the recursive occurrence were left as is, and doing more surgery to fix
+   these up seemed nontrivial.  See #691 and #864.
+
+   The unsafe flag is currently propagated to the IR and has the same meaning there -
+   calls involving function names are given the wild effect only if they are recursive
+   and not declared with the unsafe flag.  It is desirable to move to a design where "unsafe"
+   is not needed in the IR (because the staging distinction between queries and normal code
+   has already been made explicit) or at all (because query syntax is desugared to an interface
+   with sufficiently effect-polymorphic operations).
+ *)
 
 let rec type_check : context -> phrase -> phrase * Types.datatype * Usage.t =
   fun context {node=expr; pos} ->
--- a/core/typeSugar.ml
+++ b/core/typeSugar.ml
@@ -362,7 +362,7 @@ sig
   val bind_rec_annotation : griper
   val bind_rec_rec : griper
 
-  val bind_unsafe_fun_annotation : griper
+(*XXX  val bind_unsafe_fun_annotation : griper *)
 
   val bind_exp : griper
 
@@ -2166,7 +2166,7 @@ let check_for_duplicate_names : Position.t -> Pattern.with_pos list -> string li
     else
       List.map fst (StringMap.bindings binderss)
 
-let map_effects fn =
+(* XXX let map_effects fn =
   let transform = object(o)
     inherit Types.Transform.visitor as super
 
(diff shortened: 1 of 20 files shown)


jamescheney:
Mental note: before merging update typed IR tests that currently have dummy code for functions with unsafe sig.

frank-emrich:
This sounds like a reasonable approach, given that it makes it much less invaisve to handle unsafe signatures in the IR type-checker.
I've had a look at the changes to `IrCheck`, and they look good to me.

jamescheney:
The following six ways of defining `hd` typecheck (and are tame) in Links:
```
fun ([x]) {x};
fun (x::_) {x};
fun (l) {var [x] = l; x};
fun (l) {var x::_ = l; x};
fun (l) {switch (l) {case [x] -> x}};
fun (l) {switch (l) {case x::_ -> x}};
```
However, in this branch, the first two don't pass the IR typechecker, while the remaining ones do.  This means that an occurence of the first one in `prelude.links` `formP` also fails to IR typecheck.  

The problem was that for functions (and only functions) the ambient environment of the function expression is not the one you want to use for the desugared pattern code.  Thus, when translating to IR, the wrong effect (e.g. wild) was being given as a type argument to the effect-polymorphic occurrences of `hd` and `tl` generated by the pattern compiler.

The fix is to use the inner environment (body_env) when translating the desugared patterns to IR, not the outer environment (which might have a different ambient effect, incompatible with the inner one).

jamescheney:
I'm really tempted to make `Ir.fun_def` into a record though, since this would pretty much touch all the same code.

jamescheney:
Just realized that I would have saved myself some trouble if I'd reread the comments and seen that @jstolarek already did some of the work that I just redid in a branch 
`js-T691-ir-unsafe-sigs`
Of course, I made different, incompatible  arbitrary decisions.

core/typeSugar.ml
```diff
@@ -362,7 +362,7 @@ sig
   val bind_rec_annotation : griper
   val bind_rec_rec : griper
 
-  val bind_unsafe_fun_annotation : griper
+(*XXX  val bind_unsafe_fun_annotation : griper *)
```
jstolarek, line 365: Why leave this?

jamescheney: I meant to remove it and missed it.  Thanks.

jamescheney:
> What exactly happens? Do these definitions result in one of the known IR errors (there shouldn't be any by now except for #865?) or some new IR typing error? Either way I think this should be recorded, either as a new bug report or a comment on existing bug report. 


It's basically an effect-mismatch type error.  It wasn't a bug in the IR typechecker, instead, the IR typechecker was correctly catching the fact that earlier stages were putting incorrect `wild` annotations on the code.

To illustrate what was happening in a hopefully somewhat readable way...

We compile desugared Links code with patterns in function parameters by, first, doing a little extra desugaring.  (Which could perhaps be hoisted out as a separate desugaring step, but currently happens in tandem with sugartoir/compilePatterns).  A function with patterns in its parameters like
```
fun ([x]) {x}
```
gets desugared to
```
fun (z) {var [x] = z; x}
```
Then this in turn gets pattern-compiled to some IR code along these lines:
```
fun (z) { 
  if (empty(x) )
   {Wrong} 
  else { 
    var hx = ##hd(z); 
    var tz = ##tl(z); 
    if (empty(tz)) { var x = hx; x }
    else {Wrong}
  }
```
But wait, you may ask, that Links code you just wrote isn't what we're really working with, we're working with desugared, typechecked ASTs and we need to maintain the type information.  Indeed we do.  Also, `##hd` and `##tl` are tame (i.e. fully effect-polymorphic) versions of `hd` and `tl`, so we need to plug in an effect for them.  What effect should we use?  

The right answer, of course, is the effect that we have already inferred for the body of the (in this case, anonymous) function, because this should match whatever the rest of the function does.  However, what was happening, possibly due to a typo, was that the ambient effect that has been inferred for code *outside* of the function was being used instead - in this case, since the function was checked at the top level, this would be `wild`.  The solution, done in 581ca95b4dbcc59df867b4237abf180a09a3aa31, is to modify `sugartoir` so that the correct environment is indeed used.  Because of the way the translation to IR was defined, however, for recursive functions this requires further refactoring because *the effects inferred for each function in a mutual block might be different* (though they all might be wild, some might have other effects, or be polymorphic over different effect variables bound in different places).

I thought it would be a little silly to make an issue about this and immediately close it, but I can if you want - that would help document this.  Or maybe this is enough.

> Also, given that there are alternative ways of defining hd do we need the ##hd hack? Couldn't we redefine hd in prelude instead?

We could, but I wouldn't - it's a longstanding bug that `hd`/`tl` can be emulated in queries using pattern matching (#37) and one way of fixing this is to make pattern-matching against list `cons` attract the wild effect so we would wind up back where we started.

> 
> I found the explanations in the first comment of the PR helpful - could they make it into the code as a Note somewhere?

OK
- [ ] TODO: add explanatory note about unsafe sig

jstolarek:
> I thought it would be a little silly to make an issue about this and immediately close it, but I can if you want - that would help document this.

If this is fixed as part of this PR then it doesn't make sense to make a separate bug report - disregard my earlier comment.

PR #867, opened 2020-06-30
Implement Lens Serialization

### LINKED ISSUE #436: Serialization for lenses is not implemented
It would be good to have this (maybe just using storing on server and using handles for them on client)  since otherwise declaring a lens at the toplevel and then using it (even in a server-only function) fails with a "lens not serializable" error.

rudihorn:
My attempt to fix #436. As part of it the database handle has become a parameter of the `get` / `put` functions rather than a part of the lens value as this is a bit cleaner. Links still treats a lens handle as both a lens and a database handle though.

Still need to test this properly, but uploading it to check CI.

### DIFF
--- a/core/evalir.ml
+++ b/core/evalir.ml
@@ -579,7 +579,7 @@ struct
              | `Bool false    -> e
              | _              -> eval_error "Conditional was not a boolean")
   and special env (cont : continuation) : Ir.special -> result =
-    let get_lens l = match l with | `Lens l -> l | _ -> raise (internal_error "Expected a lens.") in
+    let unpack_lens l = match l with | `Lens l -> l | _ -> raise (internal_error "Expected a lens.") in
     let invoke_session_exception () =
       special env cont (DoOperation (Value.session_exception_operation,
         [], Types.Not_typed)) in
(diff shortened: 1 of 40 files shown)


core/evalir.ml
```diff
@@ -638,11 +639,11 @@ struct
...
           |> Lens_errors.unpack_sort_select_result ~die:(eval_error "%s")
         in
-        apply_cont cont env (`Lens (Value.LensSelect {lens; predicate; sort}))
+        apply_cont cont env (`Lens (db, Value.LensSelect {lens; predicate; sort}))
     | LensJoin { left; right; on; del_left; del_right; _ } ->
         let open Lens in
-        value env left >|= get_lens >>= fun lens1 ->
-        value env right >|= get_lens >>= fun lens2 ->
+        value env left >|= unpack_lens >>= fun (db1, lens1) ->
+        value env right >|= unpack_lens >>= fun (_db2, lens2) ->
```
jamescheney, line 646: Should it be allowed to build a join lens where the underlying databases are different?

core/evalir.ml
```diff
@@ -656,24 +657,24 @@ struct
             (Lens.Value.sort lens1)
             (Lens.Value.sort lens2) ~on
           |> Lens_errors.unpack_sort_join_result ~die:(eval_error "%s") in
-        apply_cont cont env (`Lens (Value.LensJoin {left; right; on; del_left; del_right; sort}))
+        apply_cont cont env (`Lens (db1, Value.LensJoin {left; right; on; del_left; del_right; sort}))
     | LensCheck (lens, _typ) ->
```
jamescheney, line 661: Here it looks like lens checking is a no-op.  Is that correct or is the plan to put more code here?

rudihorn: This is currently a no-op which is just a slight divergence from the theoretical behaviour. In theory the dynamic checks happen during evaluation of the `LensCheck` expression, in practice the dynamic checks happen during lens construction (but can still require a `LensCheck` expression to statically acknowledge runtime behaviour).

core/evalir.ml
rudihorn: I've just committed a check for this. It used to happen in the old code.

rudihorn:
@jamescheney  yes the added library `ppx_sexp_conv` provides the s expression derivers which essentially just produce the `sexp_of_t : t -> Sexp.t` and `t_of_sexp: Sexp.t -> t` functions. `Sexp.t` can then be converted into a string using `Sexp.to_string` and back with `Sexp.of_string`. For the purposes of Links, the lens library provides the `Lens.Value.serialize` and `Lens.Value.deserialize` blackboxes.

rudihorn:
I tried the following example which seems to work:
```
var db = database "links";

var albumsTable =
    table "albums"
    with (album: String, quantity: Int)
    tablekeys [["album"]]
    from db;

var tracksTable =
    table "tracks"
    with (track: String, date: Int, rating: Int, album: String)
    tablekeys [["track", "album"]]
    from db;

var albumsLens = lens albumsTable with { album -> quantity };
var tracksLens = lens tracksTable with { track -> date rating };

var joinedLens = lensjoin albumsLens with tracksLens on album delete_left;
var droppedLens = lensdrop date determined by track default 2018 from joinedLens;
var filteredLens = lensselect from droppedLens by fun (x) { x.quantity > 2 };

var myFormlet =
  formlet
    <#>
      { submitButton("Refresh") -> checkout }
    </#>
  yields ();

mutual {
  fun submit(fd) {
      mainPage("")
  }

  fun mainPage (_) {
    page
      <html>
      <body>
      <h1>Hello world!</h1>
      <ol>
      {
        for (track <- lensget tracksLens)
            <#>
              <li>{stringToXml(track.track)}</li>
            </#>
      }
      </ol>
      { myFormlet => submit }
      </body>
      </html>
  }
}


fun main () {
  addRoute("",mainPage);
  servePages()
}

main ()
```

core/evalir.ml
rudihorn: You don't want to eliminate `lenscheck`, because it is used to differentiate between statically checkable lenses and dynamically checkable lenses. So the difference between the first and third example you had given is that the third example might be fine, if <expression> can be typechecked statically, then it would compile and run. If it could not be typechecked statically, then the first example would run and the third would refuse to compile if used by a `lensput` expression.

jamescheney: Ah, OK, I remember now.  So maybe what I want is for lenscheck to be translated away after typechecking but that would be a separate issue.

rudihorn: I think it would be better for it to eventually be the deferred location for catching errors. I'm abusing this PR to also add a comment in the code and will merge it after the CI passes.

PR #868, opened 2020-06-30
Rename type variables in function signatures

### LINKED ISSUE #865: Desugaring revalidation errors in prelude
If validation after every pass is enabled, typechecking the `reproduce` function from `prelude.links` yields the following mysterious error:

```
error: transformation pass 'cp' produced an ill-typed program.

Type error: Inconsistent quantifiers, expected: 
    forall a::Session,b::Row.(AP (a::Session), (a::Session) {SessionFail:() {}-> [||]}~> ()) ~b~> ()
actual: 
    forall a::Session,b::Row.(AP (a::Session), (a::Session) {SessionFail:() {}-> [||]}~> ()) ~b~> ()

In expression: fun reproduce(ap, f) {
  var x = accept(ap);
  var _ = spawn {f(x)};
  reproduce(ap, f)
}.
```
The quantifiers don't look inconsistent to me, however, maybe this is hidden by pretty-printing.  Looking at the raw types wasn't that much more enlightening.

Also, after commenting that out, typechecking `makeEndBang` yields:
```
error: transformation pass 'funs' produced an ill-typed program.

Type error: Because of the value restriction there can be no
free rigid type variables at an ungeneralisable binding site,
but the type `Process ({ |a })' has free rigid type variables.
In expression: var _ = spawn {
     var c = accept(ap);
     wait(c)
  };.
```

jstolarek:
Fixes #865. See https://github.com/links-lang/links/issues/865#issuecomment-651879149 for description.

### DIFF
--- a/core/renamer.ml
+++ b/core/renamer.ml
@@ -54,10 +54,20 @@ let renaming_type_visitor instantiation_map =
 
     method! typ : datatype -> (datatype * 'self_type) = function
       | ForAll (qs, t) ->
-         let inst_map = shadow_vars inst_map qs in
-         let t', _ = {< inst_map >}#typ t in
-         ForAll (qs, t'), o
+         let qs', _ = List.split (List.map {< inst_map >}#quantifier qs) in
+         let t' , _ = {< inst_map >}#typ t in
+         ForAll (qs', t'), o
       | t -> super#typ t
+
+    method! quantifier : Quantifier.t -> (Quantifier.t * 'self_type) =
+      fun (var, kind) ->
+      begin match IntMap.lookup var inst_map with
+      | Some t -> begin match Unionfind.find (unwrap_meta t) with
+                  | Var (var', _, _) -> (var', kind), o
+                  | _ -> assert false
+                  end
+      | None -> (var, kind), o
+      end
 
     method! meta_type_var : meta_type_var -> (meta_type_var * 'self_type) =
       fun point ->
--- a/tests/functions.tests
+++ b/tests/functions.tests
@@ -146,3 +146,10 @@ Linearity (9) - linear recursive functions should be disallowed
 linfun f(x) {f(x)} f
 stderr : @.*cannot be linear.*
 exit : 1
+
+Type variable freshening in type signatures (#865)
+./tests/functions/T865.links
+filemode : true
+exit : 0
+args : --config=tests/functions/T865.config
+stdout : () : ()
--- a/tests/functions/T865.config
+++ b/tests/functions/T865.config
@@ -1,0 +1,3 @@
+prelude=empty.links
+optimise=false
+recheck_frontend_transformations=true
--- a/tests/functions/T865.config
+++ b/tests/functions/T865.config
@@ -1,3 +1,3 @@
-prelude=empty.links
+prelude=tests/empty_prelude.links
 optimise=false
 recheck_frontend_transformations=true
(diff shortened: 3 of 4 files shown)


core/renamer.ml
```diff
@@ -54,11 +54,21 @@ let renaming_type_visitor instantiation_map =
 
     method! typ : datatype -> (datatype * 'self_type) = function
       | ForAll (qs, t) ->
-         let inst_map = shadow_vars inst_map qs in
-         let t', _ = {< inst_map >}#typ t in
-         ForAll (qs, t'), o
+         let qs', _ = List.split (List.map {< inst_map >}#quantifier qs) in
```
jamescheney, line 57: This also seems wrong though: the problem isn't that we want to allow renaming *all* forall-quantified variables, it is that we want to rename *just* the ones in signatures that need to stay synchronized with those in the type annotation on the binder.

If you disable all shadowing then if the function signature is something like `forall a b. (forall c. a -> c) -> b` and the renamer applies a renaming such as `[a -> c, b -> d]` to the outer two quantifiers.  If shadowing is completely disabled then you get `forall c d. (forall c. c -> c) -> d` which is alpha-inequivalent.

This seems unlikely to actually happen provided the RHS of the renaming is all fresh variables, and maybe that is enough to ensure that this is correct, but it still feels a little wrong.  In that case, it would be a good idea to add a sanity check that whenever we encounter a variable to rename, that variable does *not* occur in the RHS of the renaming (that is, the freshly chosen variables were indeed fresh).  If we're generating fresh names this should always be true, but it would prevent inadvertently using the renamer unsoundly.

jamescheney:
This does not look like it solves the second part of #865 (`makeEndBang`) which appears to be an unrelated issue.  We should not close #865 until that is also fixed.

jstolarek:
> This does not look like it solves the second part of #865 (makeEndBang) which appears to be an unrelated issue. We should not close #865 until that is also fixed.

I think we should have separate bug reports for separate bugs. I say we close #865 and create a new ticket for `makeEndBang`, since it seems like a completely unrelated problem.

core/renamer.ml
jstolarek: > we want to rename just the ones in signatures that need to stay synchronized with those in the type annotation on the binder.

But we rename all variables in a signature, and all need to stay synchronized.

> If you disable all shadowing then (...)

Like you say, capture shouldn't be a problem if all variables in the codomain are fresh.

> it would be a good idea to add a sanity check that whenever we encounter a variable to rename, that variable does not occur in the RHS of the renaming

I'm hesitant to implement such a check for performance reasons and the fact that this seems rather unlikely to happen.

jamescheney:
> I say we close #865 and create a new ticket for makeEndBang, since it seems like a completely unrelated problem.

OK with me.

core/renamer.ml
```diff
@@ -54,11 +54,21 @@ let renaming_type_visitor instantiation_map =
...
+         let t' , _ = {< inst_map >}#typ t in
+         ForAll (qs', t'), o
       | t -> super#typ t
 
+    method! quantifier : Quantifier.t -> (Quantifier.t * 'self_type) =
+      fun (var, kind) ->
+      begin match IntMap.lookup var inst_map with
+      | Some t -> begin match Unionfind.find (unwrap_meta t) with
+                  | Var (var', _, _) -> (var', kind), o
+                  | _ -> assert false
```
jstolarek, line 67: Did I got this line right?

core/renamer.ml
jamescheney: OK.  Given that the external interface doesn't allow calling the renamer with arbitrary renamings this seems fine.

jstolarek:
Closed via  2f71a06

PR #883, opened 2020-07-03
Expect Tests Example

rudihorn:
For the fun of it I was experimenting with what replacing the links test harness with expect tests would look like.

Essentially the process is taking tests such as:

```
Alien declaration below toplevel
if (true) { alien javascript "foo.js" foo : () ~> (); foo() } else { () }
stderr : @.*
exit : 1
```

and replacing them with something like:
```ocaml
let%expect_test "Alien declaration below toplevel" =
  run_expr {|if (true) { alien javascript "foo.js" foo : () ~> (); foo() } else { () }|}
```

Then the test can be run using `dune runtest`, and it will produce a diff which can be accepted using `dune promote`. This updates the above code to:
```ocaml
let%expect_test "Alien declaration below toplevel" =
  run_expr {|if (true) { alien javascript "foo.js" foo : () ~> (); foo() } else { () }|}
  [%expect {|
      exit: 1
      ***: Parse error: <string>:1

      if (true) { alien javascript "foo.js" foo : () ~> (); foo() } else { () }
                       ^ |}]
```

It is also possible to write an expect regex pattern to simulate existing behaviour. 

This approach with less test harness code can also do a little bit of parallelism as supported by dune, where each folder runs in one job. I'm not sure how performance compares, as my laptop isn't a real benchmark for this but will check CI.

Such an approach might make it easier to support more complicated database tests, and I would be happy to migrate lens tests if we decided to adopt such an approach.

I wrote a small powershell script `transcribe.ps1` for translating existing expect tests into ocaml code. The script isn't perfect, but its only worth fixing if we actually decided to adopt such an approach.


dhil:
I both like and dislike an approach like this. What I like is that our testing infrastructure would be written in the same language as the majority of our code base. What I do not like is the tight coupling with `dune`. The more dune kool aid we drink, the more we depend on the dune developers to do "the right thing" for us. If the dune developers suddenly decide to make a breaking change then we might find ourselves in trouble that is not really worth our time. Changing the build system, although, it is tedious, it is something that can be done rather swiftly. Migrating the whole of our testing infrastructure is another matter, and I think the current state of your patch illustrates this point very well: new directory structure, loads of config files, new test framework clutter, etc.

My general opinion is that we should minimise the number of external dependencies we require, and we should also aim to have a loose coupling with those dependencies that are really necessary. In fact we are already slightly in trouble as Links is currently not compatible with the latest versions of cohttp, because another dependency we use is incompatible with it. This means that any OPAM user of Links will be forced to use old, potentially stale, versions of an array of libraries which would make Links OPAM-uninstallable along side tools or libraries that are built on newer versions of the dependencies we use.

rudihorn:
I think the probability of us wanting to suddenly change build systems is probably low, but if this situation ever arrises it should be possible to run the tests without dune. The tests produce a library, and the library can be run using a test runner which is mentioned here https://github.com/janestreet/ppx_inline_test#building-and-running-the-tests-outside-of-jane-street. The libraries biggest dependency is the `base` package.

jamescheney:
As discussed on Monday, closing but @rudihorn may create an enhancement issue describing this for future discussion.

PR #884, opened 2020-07-03
Don't rename tyvars in mutually recursive definitions

### LINKED ISSUE #882: Type variable renamer does not work for mutually recursive functions
**EDIT**: see comments below for an updated explanation of what's going on.

Compiling the following code (reduced from `examples/draggable.links`) with `recheck_frontend_transformations=true`:
```
mutual {
  fun foo(x) client {
   receive { case Foo -> bar(x) }
  }

  fun bar(x) client {
   receive { case Foo -> foo(x) }
  }
}
```
produces a unification error:
```
Unification error: Rigid type variables 429 and 465 do not match
error: transformation pass 'Conditional(session_exceptions (typeable))' produced an ill-typed program.

../sandbox/links/T446.links:3: Type error: The function
    `bar'
has type
    `(429::Any) {:[|Foo|]|430}~> 431::Any'
    `465::Any'
    `|hear:%478,wild|%486'
In expression: bar(x).
```
This further leads to IR typechecking error:
```
***: IR Type Error: Incompatible effects:
 Allowed:
  hear:[|Foo|],wild|380
 Actual:
  hear:[|Foo|],wild|%382
occurring in IR tail computation:
(Ir.Apply ((Ir.TApp ((Ir.Variable 231), [%381::Any; { |%382 }; %383::Any])),
   [(Ir.Variable 227)]))
```
Reproduced on 5d312665 with following config:
```
optimise=false
debug=true
recheck_frontend_transformations=true
hide_fresh_type_vars=false
show_flavours=true
show_quantifiers=true
print_types_pretty=true
show_raw_type_vars=true
```

jstolarek:
Fixes #882

### DIFF
--- a/core/renamer.ml
+++ b/core/renamer.ml
@@ -1,5 +1,8 @@
 (* Rename for type variables in terms *)
 
+(* At the moment no renaming for mutually recursive functions definitions - see
+   #882 *)
+
 open Utility
 open CommonTypes
 open Sugartypes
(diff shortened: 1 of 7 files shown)


jamescheney:
So if disabling renaming for mutually recursive things doesn't break anything, and avoids #882, what is the benefit of renaming for non-mutually-recursive things?

jstolarek:
> what is the benefit of renaming for non-mutually-recursive things?

We can do the escapes check correctly - see https://github.com/links-lang/links/issues/734#issuecomment-526570323

jamescheney:
Sorry, my question wasn't clearly expressed.   I remember the reasons for wanting to rename, but was unsure why disabling renaming for just recursive functions makes sense.  I'll add further discussion to a review.

slindley:
Something that would be (more generally) good to strive for is a situation in which as little type inference as possible has to happen after the first pass. Morally, subsequent passes should be entirely about type checking. Desugaring of typed terms should not introduce new terms whose types must be inferred.

Another approach that may have a similar effect to what @jamescheney is suggesting (delete the signature) would be to manually update the signature to be the type on the binder after doing any renaming (assuming the renaming itself doesn't already do this). I'm not keen on the idea of deleting a signature, as it is the one representation of that type that should be irrefutable as it derives directly from an annotation provided by the programmer.

Ultimately, though, it may be that type-checking can be simplified in the case that the algorithm detects that the type of a function has already been inferred so that only checking is now needed. OTOH, that may be an ideal that we could only meet by writing new code, as I expect we're relying on unification in many places where it wouldn't be necessary if the only goal was checking. An extreme approach (that has some merits) would be to write a completely separate (unification-free) type-checker for the frontend representation along the lines of the checker we already have for the IR.

jstolarek:
> In particular @slindley mentioned such an example in a comment to #678.

For the record, @slindley's original example requires a slight modificaton in order to exhibit the problem (behaviour changed in a0d79aed968f4e937e801a9ae6c312bac198e2e6):

```
var f = id(id);

sig g : forall a. (a) {}~> a
fun g(x) { ignore(g); f(x) }
```

I could add this to the testsuite but it is a bit disapointing that there is no way of saying that the fact this code compiles is a bug and the expected behaviour is to fail the compilation.

For now I'll send this PR to `/dev/null` and see whether I can fix recursive function renaming so that it behaves correctly (even if it doesn't fix the escapes check, which is a separate issue).

jstolarek:
> naively renaming the type parameters of functions is not the right thing to do because frontend AST expects the inferred types to match the type signatures

Also, I'm not sure whether type signatures are relevant here, since a minimal example I was able to find does not have user supplied signatures. (Obviously, the renaming must work with and without explicit signatures present.)

PR #885, opened 2020-07-03
Abstract IR binders

dhil:
This patch makes the representation of IR binders abstract, specifically the type definitions `Var.var_info` and `Var.binder`  are abstract, whilst `Var.var` is left concrete for now as making it abstract amounts to a more extensive change. Thus I will leave `Var.var` for a later patch.

The motivation is behind this change is to make it easier to make the notion of binders (and eventually "vars") richer to support #337. This patch includes only the first step. The next step is to make `Var.var` abstract and reorganise the interface of `Var` into two interfaces: one for binders and another for vars.

### DIFF
--- a/core/irtojs.ml
+++ b/core/irtojs.ml
@@ -729,7 +729,7 @@ end = functor (K : CONTINUATION) -> struct
   struct
     let rec fun_def : Ir.fun_def -> code -> code =
       fun ((fb, (_, xsb, _), zb, location, _unsafe) : Ir.fun_def) code ->
-        let f_var, _ = fb in
+        let f_var = Var.var_of_binder fb in
         let bs = List.map name_binder xsb in
         let _, xs_names = List.split bs in
 
--- a/core/irtojs.ml
+++ b/core/irtojs.ml
@@ -343,14 +343,14 @@ end
 let cps_prims = ["recv"; "sleep"; "spawnWait"; "receive"; "request"; "accept"]
 
 (** Generate a JavaScript name from a binder, wordifying symbolic names *)
-let name_binder (x, info) =
-  let name = Js.name_binder (x, info) in
-  if (name = "") then
-    prerr_endline (Ir.show_binder (x, info))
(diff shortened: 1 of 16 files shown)


core/irtojs.ml
```diff
@@ -343,14 +343,14 @@ end
...
 (** Generate a JavaScript name from a binder, wordifying symbolic names *)
-let name_binder (x, info) =
-  let name = Js.name_binder (x, info) in
-  if (name = "") then
-    prerr_endline (Ir.show_binder (x, info))
-  else
-    ();
+let name_binder b =
+  let name = Js.name_binder b in
+  (* TODO(dhil): The below check is unnecessary as `Js.name_binder`
```
SimonJF, line 348: Why do you need it, in that case?

core/query/query.ml
```diff
@@ -521,8 +521,10 @@ struct
     match Tables.lookup Tables.fun_defs f with
     | Some (finfo, (xs, body), z, location) ->
       Some
-      begin
-        match Var.name_of_binder (f, finfo) with
+        begin
+          (* TODO(dhil): This is a bit of an around-about way to
```
SimonJF, line 525: Nittiest of nit-picks -- round-about, rather than around-about ;P

Is there any way you could package this into a nicer API?

core/irtojs.ml
dhil: It was already there. I did not want to change "too much".

core/query/query.ml
dhil: Typo, thanks! I think we can do something much nicer. But I would rather keep this particular patch minimal (the whole point of these "small" patches I have been upstreaming over the past year is to implement a big change gradually).

SimonJF:
Cool, thanks -- feel free to merge when you're ready.

dhil:
I've tested the web examples as well now. I'll merge once the CI finishes (with success).

PR #886, opened 2020-07-03
Fix web bug introduced by #864

SimonJF:
"#" cannot be used in JS identifiers, so this led to problems executing web things.

This patch replaces `##hd` and `##tl` with `$$hd` and `$$tl`, since `$`
is valid in JS identifiers. It also adds the necessary `_$$hd` and `_$$tl`
stubs in jslib.


dhil:
Be cautious! The current implementation of name prettier printer uses `$` as a special delimiter. I am not sure what the consequence would be, if any since the users cannot type `_$$hd` themselves.

SimonJF:
@dhil :

> Be cautious! The current implementation of name prettier printer uses `$` as a special delimiter. I am not sure what the consequences would be, if any since the users cannot type `_$$hd` themselves.

Thanks for pointing this out; it's an important one to consider. Having checked the HTML output, these seem to be printed fine in the JS. The one place I see the dollar signs being important is `Name.prettify`, which seems only to work on types (aliases, specifically), which doesn't apply here. Since `$$hd` and `$$tl` are introduced in the pattern compilation pass during IR conversion, I don't think they'll be needed to show any errors, so I'm fairly confident there will be no effect (and even if there were -- it'd just be `..hd` and `..tl` in the error messages).

PR #890, opened 2020-07-07
Preserve ordering in Draggable lists (database version) example.

dhil:
This commit fixes a bug in the database version of the draggable lists
example that would cause the list of items to be displayed in inverse
order after a save.

### DIFF
--- a/examples/draggableDb.links
+++ b/examples/draggableDb.links
@@ -58,7 +58,7 @@ fun load() server {
   var t = itemsTable();
   var items =
     query { for (item <-- t)
-             orderby (-item.i)
+             orderby (item.i)
               [(name=item.name)]};
   map ((.name), items)
 }


dhil:
Fixing this I experienced a sort of déjà vu. I believe I have previously fixed the exact same bug... I will try to dig through the logs later.

jamescheney:
> Fixing this I experienced a sort of déjà vu. I believe I have previously fixed the exact same bug... I will try to dig through the logs later.

That wouldn't surprise me, since the previous version code exhibited the now-fixed bug #856.

dhil:
I should add, I have only tested with `postgresql` database driver. I have yet to setup SQLite3 and MySQL databases on my machine. I would appreciate if someone has those databases set up already could test the draggable lists example with this patch.

Issue #871, opened 2020-07-01
Generalisation breaks the value restriction


jstolarek:
The program:

```
var foo = spawn { () }; ()
```
compiled with following configuration on commit 2f71a065:
```
prelude=tests/empty_prelude.links
optimise=false
recheck_frontend_transformations=true
print_types_pretty=false
hide_fresh_type_vars=false
show_flavours=true
```
produces a type error:
```
error: transformation pass 'funs' produced an ill-typed program.

Type error: Because of the value restriction there can be no
free rigid type variables at an ungeneralisable binding site,
but the type `(Types.Meta
   Point @ 0x7f1fd9f22978 (Types.Application
                             ({ Types.Abstype.id = "Process";
                                name = "Process";
                                arity =
                                [(CommonTypes.PrimaryKind.Row,
                                  (CommonTypes.Linearity.Any,
                                   CommonTypes.Restriction.Effect))
                                  ]
                                },
                              [(CommonTypes.PrimaryKind.Row,
                                (Types.Row
                                   ({},
                                    Point @ 0x7f1fd95c0d60 (Types.Var
                                                              (394,
                                                               (CommonTypes.PrimaryKind.Row,
                                                                (CommonTypes.Linearity.Unl,
                                                                 CommonTypes.Restriction.Any)),
                                                               `Rigid)),
                                    false)))
                                ])))' has free rigid type variables.
In expression: var foo = spawn { () }.
```

Below I'm copying @slindley explanation from [here](https://github.com/links-lang/links/issues/865#issuecomment-650811461):

The error in function desugaring arises because of the generalisation that occurs inside function literals. This seems harmless enough, and indeed it would be if the implementation of generalisation was a bit more careful. Here's roughly what's going on. The following
```
var (z : { |%e}) = spawnAt { |%e} (fun () {...});
```
is intended to be desugared into code that explicitly names, and generalises, the anonymous function:
```
var (z : { |%e}) = spawnAt [{ |%e}] (fun f[e'::Row]() {...}; f[{ |%e}])
```
but instead it comes out more like this
```
var (z : { |e}) = spawnAt [{ |e}] (fun f[e::Row]() {...}; f[{ |e}])
```
(using suitable notation for type abstraction and type application).

The row variable `%e` doesn't appear in the type environment so it gets generalised in the type of the anonymous function. But this operation incorrectly mutates the `%e`s appearing in the type of `z` and applied as a type parameter to `spawnAt` turning them into rigid variables.

[...]

PR #891, opened 2020-07-08
Don't discard typechecked AST with recheck_frontend_transformations

jstolarek:
I see what's going on in #882 and I believe I have a solution - that is, if everyone agrees it is a good solution.

When `recheck_frontend_transformations` is enabled we typecheck the AST after every desugaring pass and discard the resulting typechecked AST. Except that typechecking is side-effecting and we don't really fully discard the result - unified flexible variables remain unified and that is what causes the problem here.

After the inners pass we have something like this in the AST (simplified and prettified):

```
f : forall a. a -> ...
f (x : a) = g @%b x
```

where `@` is a type application of a flexible type variable `%b`. Type variable renamer is run right at the beginning of `Funs` case in the typechecker and turns the above code into:

```
f : forall c. c -> ...
f (x : c) = g @%b x
```

Note how rigid type variable `a` is renamed into `c`, but flexible variable `%b` remains unchanged. The typechecking continues and right before finishing the `Funs` case the code is:

```
f : forall c. c -> ...
f (x : c) = g @c x
```

Note that flexible variable `%b` was unified with type variable `c`. We now discard the typechecked AST, except that the flexible variable remains unified. So at the beginning of next pass (`session_exceptions`) the AST looks like this:

```
f : forall a. a -> ...
f (x : a) = g @c x
```

The renamer turns that into:

```
f : forall d. d -> ...
f (x : d) = g @c x
```

and at this point typechecker concludes that the type of `g` is `c -> ...` but the argument is of type `d`.

So it looks like the renamer is doing the right thing and the problem arises from side-effecting nature of typechecking. My proposed solution is to keep the typechecked AST when `recheck_frontend_transformations=true` is enabled.

One thing I am not certain of is whether the `state` should be updated in any way. For now I am leaving it unchanged, but I'm wondering whether the second component returned by typechecking (`Types.datatype`) should be put into `state.datatype`? The tests pass without updating that field, but that of course doesn't mean anything.

### DIFF
--- a/tests/typed_ir.tests.config
+++ b/tests/typed_ir.tests.config
@@ -4,4 +4,3 @@ prelude=tests/empty_prelude.links
 optimise=false
 generalise_toplevel=true
 show_quantifiers=true
-recheck_frontend_transformations=true
--- a/tests/typed_ir.tests.config
+++ b/tests/typed_ir.tests.config
@@ -4,3 +4,4 @@ prelude=tests/empty_prelude.links
 optimise=false
 generalise_toplevel=true
 show_quantifiers=true
+recheck_frontend_transformations=true
--- a/core/frontend.ml
+++ b/core/frontend.ml
@@ -223,7 +223,7 @@ module Typeability_preserving = struct
         Debug.if_set Basicsettings.show_stages (fun () -> T.Typeable.name ^"...");
         T.Typeable.program state program
       in
-      (if verify_transformation T.Typeable.name then
+      if verify_transformation T.Typeable.name then
          let tyenv =
            Context.typing_environment payload.state.Transform.Typeable.context
          in
@@ -233,14 +233,16 @@ module Typeability_preserving = struct
             check *against* the datatype in transformation state
             [payload]. *)
          try
-           ignore (TypeSugar.Check.program
+           let (program, _, _) =
+             TypeSugar.Check.program
                      { tyenv with Types.desugared = true }
-                     payload.program)
+                     payload.program
                   (* TODO(dhil): Verify post-transformation invariants. *)
+            in (Transform.Typeable.Result { state; program })
          with exn ->
            let stacktrace = Printexc.get_raw_backtrace () in
-           trace_type_error T.Typeable.name Sugartypes.pp_program program payload.program stacktrace exn);
-      result
+           trace_type_error T.Typeable.name Sugartypes.pp_program program payload.program stacktrace exn
+      else result
     in
     let (Transform.Typeable.Result { state; program }) =
       let initial_state = Transform.Typeable.{ datatype; context = context' } in
@@ -261,7 +263,7 @@ module Typeability_preserving = struct
       let (Transform.Typeable.Result payload) as result =
         T.Typeable.sentence state program
       in
-      (if verify_transformation T.Typeable.name then
+      if verify_transformation T.Typeable.name then
          let tyenv =
            Context.typing_environment payload.state.Transform.Typeable.context
          in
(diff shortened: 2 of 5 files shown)


dhil:
If you want to keep the program then you should also update the `datatype` and `context` components. The `datatype` component is supposed to be the type of the `program` component, and the `context` component provide the necessary context to interpret `datatype.

jstolarek:
Ok, I'm not surprised. Updating the datatype stored inside the `state` seems to work without problems - I pushed an update to the branch.  However, attempting to put the typing environment returned from typechecking into the state context results in a lot of internal errors in `instantiate.ml`. So basically this doesn't work:

```
let (program, datatype, tyenv) =
  TypeSugar.Check.sentence
          { tyenv with Types.desugared = true }
          payload.program in
let context = { payload.state.Transform.Typeable.context with
                  Context.typing_environment = tyenv } in
let state   = { Typeable.datatype = datatype
              ; Typeable.context  = context }
```

dhil:
I think you need to extend the previous context with the "new" context for that kind of thing to work.

jstolarek:
Yes, that's precisely what I did in an earlier commit today. However, this has an unexpected effect: enabling `recheck_frontend_transformations` hides #871 (me earlier marking this as fixed was wrong). More precisely, if `recheck_frontend_transformations` is disabled then #871 occurs but is invisible as nothing catches that error. Presence of #871 can only be witnessed by printing the program AST and seeing that printing produces the following debug message:

```
Row variable where row expected:(Types.Meta
   Point @ 0x7fdc53e69820 (Types.Var
                             (350,
                              (CommonTypes.PrimaryKind.Row,
                               (CommonTypes.Linearity.Unl,
                                CommonTypes.Restriction.Any)),
                              `Rigid)))
```
(generated [here](https://github.com/links-lang/links/blob/e879056874a8b099773c7ea62791f840c976d4e9/core/types.ml#L2353) and marked as something that shouldn't happen). Enabling `recheck_frontend_transformations` makes #871 go away. For a test program:

```
var foo = spawn { () }; ()
```

this is a fragment of the AST without rechecking:

```
(Sugartypes.Pattern.Variable ("foo", Process ({ |350 })))
```

and with rechecking:

```
(Sugartypes.Pattern.Variable ("foo", Process ({ |%394::(Unl,Mono) })))
```

I am guessing that with rechecking enabled and AST/environment being retained the flexible type variable makes it into the environment and isn't generalised, c.f. Sam's comment on #871:

> The row variable `%e` doesn't appear in the type environment so it gets generalised in the type of the anonymous function. But this operation incorrectly mutates the `%e`s appearing in the type of `z` and applied as a type parameter to `spawnAt` turning them into rigid variables.

@slindley, what are your thoughts here? For now I marked the test for #871 as ignored since I don't have a better idea what to do.

jstolarek:
@jamescheney suggested to write a test for #871 by running the IR typechecker, which means we now have a working test to catch #871.

tests/typed_ir.tests.config
```diff
@@ -4,4 +4,3 @@ prelude=tests/empty_prelude.links
 optimise=false
 generalise_toplevel=true
 show_quantifiers=true
-recheck_frontend_transformations=true
```
jamescheney, line 7: Could the recheck flag be disabled just for the new test #871?  I think in general it is good to do the revalidation for the typed IR tests, so that IR typing failures that arise because of a bug in an earlier desugaring stage are correctly attributed.  (On the other hand, this situation also illustrates that revalidation can mask problems that would otherwise be caught by IR typechecking, because revalidation is currently side-effecting.  So maybe we should be checking the typed IR tests both with and without revalidation.)

jstolarek: Done now.

jamescheney:
Meeting discussion: Double check whether this closes #882 and @slindley to have a look.

jstolarek:
> Double check whether this closes #882

I checked - it does.

jamescheney:
So I think this means we can update and merge this and close #882.

jstolarek:
Closed in  527c0bd

Issue #862, opened 2020-06-25
null.links: Database test failure


dhil:
The test `tests/database/null.links` fails on my machine. 
```
$ ./run-tests db-only
cmnd=psql -h localhost -U links -p 5432 -v ON_ERROR_STOP=1 -q -d links -f tests/database/factorials.sql
cmnd=psql -h localhost -U links -p 5432 -v ON_ERROR_STOP=1 -q -d links -f tests/database/null.sql
cmnd=psql -h localhost -U links -p 5432 -v ON_ERROR_STOP=1 -q -d links -f tests/database/xpath-reduced.sql
cmnd=psql -h localhost -U links -p 5432 -v ON_ERROR_STOP=1 -q -d links -f tests/database/xpath.sql
cmnd=./links --config=tests/database/config tests/database/factorials.links
() : ()
cmnd=./links --config=tests/database/config tests/database/factorials-large.links
() : ()
cmnd=./links --config=tests/database/config tests/database/empty.links
[] : [(|_::Base)]
cmnd=./links --config=tests/database/config tests/database/emptyfun.links
[] : [(f:Int,i:Int)]
cmnd=./links --config=tests/database/config tests/database/null.links
Raised at file "core/database.ml", line 63, characters 12-92
Called from file "core/database.ml", line 144, characters 23-51
Called from file "core/database.ml", line 163, characters 29-50
Called from file "list.ml", line 69, characters 12-15
Called from file "core/database.ml", line 162, characters 9-19
Called from file "core/evalir.ml", line 723, characters 34-71
Called from file "src/core/lwt.ml", line 818, characters 19-23
Re-raised at file "src/core/lwt.ml", line 823, characters 6-15
Called from file "core/proc.ml", line 288, characters 5-64
Called from file "core/proc.ml" (inlined), line 297, characters 4-13
Called from file "core/proc.ml", line 300, characters 17-27
Called from file "core/evalir.ml", line 932, characters 10-67
Called from file "camlinternalLazy.ml", line 31, characters 17-27
Re-raised at file "camlinternalLazy.ml", line 36, characters 4-11
Called from file "bin/driver.ml", line 84, characters 8-75
Called from file "camlinternalLazy.ml", line 31, characters 17-27
Re-raised at file "camlinternalLazy.ml", line 36, characters 4-11
***: Runtime error: Attempted to read null integer from the database 
FAILED TEST null
cmnd=./links --config=tests/database/config tests/database/unit.links
[(), ()] : [()]
[...]

Issue #844, opened 2020-06-08
Rationalize/remove shredding settings?


jamescheney:
Since #757 we now support both flat and shredded queries in the same program.  We anticipate supporting another query policy (at least experimentally) for queries with set/bag semantics (DISTINCT and friends).

The only effect the `shredding` setting actually has anymore is to determine what happens if the `query` keyword is used on its own without a policy: if `shredding` is off then the flat query evaluator is used, if it is on then the nested query evaluator is used.  (But even if `shredding` is off, both behaviors are available by using `query flat` or `query nested`.)

So, I suggest that we change the `shredding` setting to something like `default_query_evaluator`, or just drop it entirely and let `query {...}` stand for `query flat {...}`.   

A related question is whether to get rid of, or internalize, the `use_keys_in_shredding` flag.  Using this means that just the declared key fields of a table (if present) are used for indexing instead of all fields.  I would suggest to remove this option and instead support this behavior (if at all) on a table-by-table basis by using the key fields if present, otherwise defaulting to using all fields, which is already the default behavior if no keys are declared anyway.

jamescheney:
Another observation is that the OCaml MySQL library does not yet support MySQL 8, and we cannot support shredding on MySQL versions < 8.  This is another reason why I think the query policy should default to flat.

jamescheney:
Having discussed this change a while ago with @SimonJF,  I think it is reasonable to remove the shredding setting, and thus require explicitly saying "query nested" if the query is intended to run using the shredding code, since the impact on GtoPdb is minimal.  The shredding examples do need to be updated to explicitly say `query nested`, though.

PR #893, opened 2020-07-10
WIP: fix #322

### LINKED ISSUE #322: Add MySql and Sqlite drivers/packages to CI testing

jamescheney:
This is work in progress towards a new, hopefully cleaner infrastructure for running database tests, which is intended to fix #322.

The script `run-database-tests` replaces and generalizes the parts of `run-tests` that set up and run the database tests.  It takes an argument which is a directory containing database tests.  The tests are described in a json file (called testsuite.json by default).  The databases, setup files (SQL files to be run), and Links test files are listed in testsuite.json.  Test suite files have been written for the database, shredding, and relational-lenses directories and all work (locally).  

Host, port, username and password arguments are parsed from config files as in the original script.  There is a separate config file for each database, called config.<databasename>.

The listed sql files are each first run using the appropriate database command line client.

Custom SQL setup files are possible, using the database name as suffix to deal with situations where the SQL syntax needed varies.  If a custom file (e.g. factorials.mysql) is present when setting up tests for a database (e.g. mysql).

Once each database is set up, each test listed in the testsuite.json file is run.  The number of failures (per database) is tracked and printed.  If there are any failures the test runner returns 1, otherwise 0.

More work needs to be done:

- [x] Confirm that `run-tests` does the right thing for Postgres and Sqlite3
- [x] Enable mysql services in Travis and Azure VMs and test that compiling Links with mysql enabled works
- [x] Configure MySQL service and initialize to a state that the test runner can handle; adjust `config.mysql` and enable mysql tests.
- [x] Confirm that things break if a test is broken (e.g. add garbage to factorials.links).
- [x] Decide whether to continue to call `run-database-tests` from `run-tests` or directly from the Travis and Azure scripts.
- [x] Once everything is working, clean up previous database test running infrastructure: remove code from run-tests, remove old config.sample and testsuite.config files.
- [x] Replace tabs with spaces because I can't figure out how to get Emacs to do the right thing when indenting Python. 
- [x] Update wiki page: https://github.com/links-lang/links/wiki/Running-the-test-suite

I may need to pester people who have more experience with Travis / Azure ship-in-a-bottle building.


jamescheney:
This now mostly works.  I'm configuring MySQL directly creating a Links user and dummy password.  I think Postgres could be configured the same way instead of setting it up to allow local connections from the links user but won't fiddle with that for now.

One thing I'm not sure about is whether we want the test runner (run-tests) to run tests for all three databases by default, or only if explicitly requested during CI.  I think most people not doing development of Links database (i.e. sane people) will only have (at most) one database installed.  (Indeed, using sqlite3 is so much easier than postgres that I'm inclined to say that it should be the default for casual/tutorial use, and it would be good to see whether it's possible to install recent versions of sqlite3 that support shredding in the CI.)

I'm not super-enthused about making the test script detect which database(s) are installed before deciding which tests to run, and instead suggest making the database a command-line parameter.  The CI will run all of the tests, but whoever is running tests locally can choose to run them for only their chosen database.

Another useful piece of functionality that isn't yet preserved is the ability to have a local config file that overrides the default one.  While this can cause problems (e.g. #862) it still seems handy.  Perhaps this suggests adjusting the database test script so that optionally, it uses the database specified by the "config" file on the theory that that's likely to be the one that contains the preferred setup for the locally-installed database.

jamescheney:
I settled on adjusting the database test script to take a flag `--database` to specify whether to run all tests or just a single database (which currently defaults to pgsql), and another flag --local that says "if a local `config` file is present, use that and ignore any --database option".  This approximates the current behavior, and is a bit more flexible, for example it parses the config file to work out what database is locally installed.

I also decided to keep the calls to database tests from `run-tests` but disable them by default since we don't want to run the same tests on the CI as people will normally want to run locally.  Thus, to run the database tests using `run-tests` you now have to say explicitly one of:

```
./run-tests database
./run-tests shredding
./run-tests relational-lenses
```

In each case this will use the local `config` if present, otherwise the `pgsql` configuration.

Incidentally, it would be cleaner to use the same suffix for the configurations as for the Links database driver names - currently we write `database_driver=postgresql` but use `config.pgsql`, so there's an annoying conditional that checks for this and translates.  Further confusing matters is that the Links driver file is called `pg_database.ml`.  I'm not sure this is really worth the effort though.

jamescheney:
Huh.  Turns out there were some dangling references to tests/shredding/config.sample, which are no longer needed anyway because of #844.  After looking around, I determined that there are only a few places where we still need a config file, since it is now possible to override settings locally using the `--set` command line option.  So I'm opportunistically cleaning up the configurations used in the tests a bit, and renaming the remaining config files to consistently be `foo.config` where used with test file `tests.config`.

jamescheney:
This is ready for review, and I will add candidate new text to the wiki page shortly.

jamescheney:
Wiki page now updated, but planning to merge #896 in to this first before merging, assuming CI is happy.

jamescheney:
I just realized that I inadvertently disabled the unit tests, which also run against a database configuration.  More work is needed before this is ready to merge.

jamescheney:
Adapted/restored the original unit tests setup; it should probably be generalized to MySQL and SQLite3 but let's add that to the todo list for #894.

run-database-tests
```diff
@@ -0,0 +1,152 @@
...
+    verbose = args.verbose
+
+    linksdir = os.getcwd()
+    linksexec = os.path.join(linksdir,"links")
+
+    def debug(s):
+        if verbose:
+            print(s)
+
+    def linkscmd(config,file):
```
SimonJF, line 40: Nit: `links_cmd` is more widely-accepted Python style

run-database-tests
```diff
@@ -0,0 +1,152 @@
...
+        debug("%s --config=%s %s %s" % (linksexec,config,setdebug,file))
+        if verbose:
+            completed_process = subprocess.run([linksexec, "--config=%s" % config, setdebug, file])
+        else:
+            completed_process = subprocess.run([linksexec, "--config=%s" % config, setdebug, file],
+                                               stdout=subprocess.PIPE,stderr=subprocess.PIPE)
+        return completed_process.returncode
+
+    matcher = re.compile("database_args=([^:]*):([^:]*):([^:]*):([^\n:]*)")
+    def parsedbconfig(config):
```
SimonJF, line 51: `parse_db_config`, `parse_db_driver` etc. are more readable

run-database-tests
```diff
@@ -0,0 +1,152 @@
...
+
+    os.chdir(working_dir)
+
+    with open(config_file) as fp:
+        config = json.load(fp)
+
+    debug(config)
+
+    if args.database == "all":
+        dbconfigs = [(db, parsedbconfig("config.%s" % db)) for db in config['databases']]
```
SimonJF, line 129: How does this work in the case of `(None, None, None, None)` being returned?

database/sqlite3-driver/lite3_database.ml
```diff
@@ -94,6 +94,13 @@ class lite3_database file = object(self)
   method exec query : Value.dbvalue =
     Debug.print query;
     let stmt = prepare connection query in
+    let rec last_res r =
```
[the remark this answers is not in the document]
rudihorn: Yes this can be considered Sqlite3 boilerplate. The manual essentially states that calling `prepare` may not execute all of the command, and to determine if this is the case `prepare_tail` can be called on the resulting statement. It returns `None` if the computation is finished, otherwise `Some x`, where `x` is a new statement. This is used to ensure that multiple statements are executed.

lens/database.ml
```diff
@@ -63,7 +63,10 @@ let fmt_cols ~db f cols =
 let fmt_phrase_value ~db f v =
   Format.fprintf f "%s"
     ( match v with
-    | LPV.Bool b -> string_of_bool b
+    | LPV.Bool b -> (
+        match b with
+        | true -> "(1=1)"
```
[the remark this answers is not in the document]
rudihorn: The fairly old `sqlite3` version used by our CI does not support these constants which is indeed depressing. This method is luckily portable, but it might be nice to change it back to the old code once `true` and `false` become more widespread.

database/sqlite3-driver/lite3_database.ml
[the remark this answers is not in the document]
jamescheney: Yes. this is discussed further in #894.  I'll add a comment documenting this.  It would probably be good practice to qualify the Sqlite3 library imports too.

lens/database.ml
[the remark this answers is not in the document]
jamescheney: Yes, we have the same issue with shredding (`row_number` not supported by Sqlite3 < 3.25).  On one hand it is good to support old versions, on the other hand the default ubuntu 16 version of Sqlite3 is pretty old.  I think we should manually install a newer version in the CI and recommend people use the most recent version.

run-database-tests
jamescheney: Done

run-database-tests
jamescheney: Done

run-database-tests
jamescheney: In the case of sqlite3, this is fine because no configuration is needed, the sqlite3 just looks for a sqlite3 database file in the filename described by the database name.

In that case, if you're trying to run mysql or postgres tests, the configuration is wrong and you get an error like "Expected string, got None".  An error that says the configuration is wrong would be better, btu configurations can also be wrong for other reasons that are more work to check, so  I'm OK with leaving this particular yak a little unshaven.

(It would probably be better to generalize this so that we can point to the location of the sqlite3 database file in the config file, but this isn't currently supported.)

jamescheney:
I've pushed but for some reason nothing is happening. Will merge when my changes appear and CI is happy.

PR #896, opened 2020-07-13
Database ci: get SQLite3 working

jamescheney:


jamescheney:
@rudihorn, this is not passing the SQLite3 relational lens tests.  Can you confirm they running locally for you with 
```
./run-database-tests tests/relational-lenses -d sqlite3 -v
```
to show what the successful run looks like?

Then if they work for you but not in the CI we can try adding `-v` to see what is going wrong.

rudihorn:
Locally I had the issue that the database is called `links`, and so is the executable. Let me try this again.

rudihorn:
I just ran this and got:

```
./run-database-tests tests/relational-lenses -d sqlite3
Local configuration file not found; testing databases specified by --database parameter
tests/relational-lenses: Setting up sqlite3
tests/relational-lenses: Testing sqlite3
 SUCCESS: Test 01_cds on database sqlite3
 SUCCESS: Test 02_cds on database sqlite3
 SUCCESS: Test 03_cds on database sqlite3
tests/relational-lenses: 0 failures for sqlite3
```

jamescheney:
OK, I'll try it locally (or in CI) when I get a chance. 

I just realized that I accidentally disabled the unit testing part of the CI, and that this relies on the database configuration too.

rudihorn:
Seems the bug is that the CI installs SQLite 3.22 and `TRUE`/`FALSE` constants were added in SQLite version 3.23 https://sqlite.org/changes.html 2 years and 3 months ago. Ubuntu adds this from release 19.10 onwards.

jamescheney:
Ah yes, I had to fix that already in the main Links DB code (use 0=1 instead of FALSE).

rudihorn:
That should be it solved, if the CI passes it can be merged in.

jamescheney:
Thanks.  I think the true/false thing is probably also broken / nonportable elsewhere in the query code...

PR #897, opened 2020-07-16
Relational Lens MySQL support + Sqlite3 and MySQL serial column support

rudihorn:
This pull request adds support for databases that don't support multi command execution such as MySQL. This is done by introducing a lens `Crud` module, that has an `exec_multi` command. The command either formats all the commands into a single string, or executes them seperately using `exec_multi_slow` depending on the db driver. 

Multiple insert returning is done by either generating an insert returning statement or performing each insert value seperately and then querying for the last inserted row id.

- [x] Create test and check it works


rudihorn:
Okay this now works and is ready for review and merging.

lens/database.ml
```diff
@@ -302,5 +336,78 @@ module Insert = struct
...
+      Debug.print getid;
+      Statistics.time_query (fun () ->
+          db.execute_select getid ~field_types:[ ("id", Phrase_type.Serial) ])
+    in
+    no_returning data |> split |> List.map ~f:exec_insert |> List.concat
+
+  let exec_insert_returning ~db ~field_types data =
+    match db.driver_name () with
+    | "sqlite3"
+     |"mysql" ->
```
jamescheney, line 373: indentation/spacing

lens/database.ml
```diff
@@ -302,5 +336,78 @@ module Insert = struct
...
+    let { table; columns; values; returning } = v in
+    let f v =
+      let values = [ v ] in
+      { table; columns; values; returning }
+    in
+    List.map ~f values
+
+  let no_returning v = { v with returning = [] }
+
+  let exec_insert_returning_hack ~db data =
```
jamescheney, line 349: There is also code accomplishing this in the main core/database.ml and driver modules.  It is unfortunate that our current architecture for splitting Links into core and lens libraries leads to repetition of code for essentially the same purpose.  It would be good to find a way to avoid this someday...

lens/database.ml
```diff
@@ -302,5 +336,78 @@ module Insert = struct
...
+        exec_insert_returning_hack ~db data
+    | _ ->
+        let cmd = Format.asprintf "%a" (fmt ~db) data in
+        Debug.print cmd;
+        Statistics.time_query (fun () -> db.execute_select ~field_types cmd)
+
   (* optionally add a returning statement if required *)
 end
+
+module Crud = struct
```
jamescheney, line 383: I'd prefer a more descriptive name such as "Update" or "DatabaseUpdate"

tests/relational-lenses/serial.sql
```diff
@@ -0,0 +1,15 @@
+DROP TABLE IF EXISTS t1;
+
+CREATE TABLE t1 (
+    id1 SERIAL NOT NULL,
```
jamescheney, line 4: I think SERIAL is potsgres-specific while the sqlite3 version below is generic SQL.  So I'd be inclined to name this file "serial.pgsql" and rename the sqlite3-specific one to use the .sql suffix.

It is too bad that they didn't standardize this, since in my experience every nontrivial database application uses serial/auto-increment etc.

lens/database.ml
rudihorn: This is part of the current ocamlformat settings, where the alternative syntax is used to highlight that fact that the second case is a fallthrough. I think this setting should be kept, or at the very least not changed in this PR. The docs for this formatting behaviour:

```
       --indicate-nested-or-patterns={unsafe-no|space}
           Control whether or not to indicate nested or-pattern using
           indentation. unsafe-no does not indicate nested or-patterns.
           Warning: this can produce confusing code where a short body of a
           match case is visually hidden by surrounding long patterns, leading
           to misassociation between patterns and body expressions. space
           starts lines of nested or-patterns with " |" rather than "| ". The
           default value is unsafe-no.
```

jamescheney: Oh, OK.

lens/database.ml
rudihorn: There is code which does something similar, but it is a bit more primitive in the sense that it requires a links style insert query and does not always handle the multivalue case (and therefore not efficiently).

lens/database.ml
rudihorn: I will rename this to `Change`, as anything `Update` related is even more confusing. Admittedly `Crud` requires querying as well, so wasn't correct anyways.

tests/relational-lenses/serial.sql
rudihorn: I have renamed it to `serial.pgsql`, but I don't think the sqlite3 one deservers to receive the generic `.sql` either. I find Sqlite3's behaviour very implicit, as setting a column type to `INTEGER PRIMARY KEY` converts it into a column alias for rowid. Having something slightly different like `INT PRIMARY KEY` already seemingly breaks this behaviour. I personally find postgres' behaviour of introducing a new datatype with a different behaviour better, but agree it should probably not be the default file.

Issue #878, opened 2020-07-02
Make it possible to write Links IR programs


frank-emrich:
In the Links meeting we discussed that it may be helpful for different purposes to write Links IR programs directly. One use case would be testing the IR type-checker.

We discussed different ways of implementing this, like the following:
1. Use the existing derived parsers for JSON or s-expression representation. This is easy to do, but yields syntax that is probably very painful to use.
2. We could define a proper syntax for the IR and write/generate a parser for it. This may or may not require implementing a similar parser for a simplified language of Links types. I specifically mention the types here because I would like to be able to write types in a way so that we have more direct control about their internal representation (e.g., being able to directly specify the integer identifiers for type variables).
3. We could use a mixture of the two.

jstolarek:
While I like the idea of being to supply an externally-generated IR to the IR compiler (this creates a possibility for writing a new frontend that works with Links' IR backend), I am *very* sceptical about writing IR programs directly (in a sense of having a language that is parsed to the IR). What is the exact motivation/use case for having such a feature? If `links` library interface exposed the IR AST and the entry point to the IR compiler/typechecker, we could write IR programs by building their ASTs directly, with no need for any parsing. This would allow to test the IR typechecker fairly easily.

There is a similar feature in GHC but the idea of writing Core programs by hand never took off. See [An External Representation for the GHC Core Language (For GHC 6.10)](http://www.haskell.org/ghc/docs/6.10.4/html/ext-core/core.pdf) paper and [GHC wiki](https://gitlab.haskell.org/ghc/ghc/-/wikis/external-core).

rudihorn:
When used directly ocaml can also be pretty good at writing small embedded DSLs for defining these types of things. I've used the `let` syntax for defining mini lambda calculi which has worked quite well.

e.g. the following example I used to compile a lambda abstraction `\x -> (let y = x in if true then 20 else y)`. It's not perfect, but for internal purposes it can work well.
```ocaml
  let open Lang.O in
  abs1 "x" (fun x ->
      let@ y = x in
      ifelse ctrue (i 20) y)
```

Regarding james' PPX question: Because ocaml just derives the `pp` function it is easy to override behaviour to a certain extent, but becomes problematic in the recursive case. The following example might make what I mean more obvious:

```ocaml
(* derived function *)
let rec pp x = ... pp subterm ...

(* our function *)
let pp_ours x = special case of x | pp x
```

frank-emrich:
@jamescheney 
[...]

PR #898, opened 2020-07-18
A DSL for the Links IR

frank-emrich:
Following the discussion in #878, I decided to implement a DSL for the Links IR, called Schinks (Scheme-y Looking Links). The purpose is to be able to write test cases for the IR type-checker directly.

I've integrated this with the existing unit testing framework used for relational lenses.

An example IR program (more specifically, a binding) would look like the following, as found in ` tests/unit/ir/test_cases/binding.ml`:

```
fun_ "f" ([int] |--> int)
    [("x", int)]
    (tc_to_comp (return (var "x")))
```

Things get a bit more verbose when using quantifiers:
```
fun_ "f"  (forall [q "T1"; q "T2"] ([tvar "T1"; tvar "T2"] |--> int))
      ~tparams:[q "Q1"; q "Q2"]
      [("x", tvar "Q1"); ( "y", tvar "Q2")]
      (tc_to_comp (return (i 3)));
```

I haven't implemented all constructs of the IR in the DSL yet, but will do so once I write more tests.


jamescheney:
> Schinks (Scheme-y Looking Links)

Perhaps following tradition of naming Links versions after Edinburgh neighborhoods we should call the IR DSL "Leith".  (There is a Leith Links, though it no longer is an active golf course, and it is in some sense "scheme-y" though this is a bit unkind... Also happens to be the place where golf's roles were first codified before moving to St Andrews.)

frank-emrich:
The original motivation for naming it Schinks was to be able to add ASCII art of the following very happy looking skink:

```
              ____...---...___
___.....---"""        .       ""--..____
     .                  .            .
 .             _.--._       /|
        .    .'()..()`.    / /
            ( `-.__.-' )  ( (    .
   .         \        /    \ \
       .      \      /      ) )        .
            .' -.__.- `.-.-'_.'
 .        .'  /-____-\  `.-'       .
          \  /-.____.-\  /-.
           \ \`-.__.-'/ /\|\|           .
          .'  `.    .'  `.
          |/\/\|    |/\/\|
jro
```
... but the copyright situation wasn't clear and we may get sued until all eternity.

frank-emrich:
This is currently blocked by #899.

jamescheney:
One aspect of #878 that I'm not sure was discussed is that it would also be nice to provide a way to *serialize* IR code as concrete syntax or DSL code.  Maybe not necessary for this PR if not straightforward, but it would be good to have.  One use case would be to serialize the Links IR for a program to an OCaml file and compile/link it with the rest of the backend, producing a standalone executable that typechecks and runs the Links IR code.  (Of course this would still essentially be interpreting the Links IR code, but it would enable deploying executables without necessarily having all of hte OCaml ecosystem present on a given remote VM, and would avoid the overhead of Links frontend typechecking etc, though this is a fixed cost anyway.)

frank-emrich:
> it would also be nice to provide a way to serialize IR code as concrete syntax or DSL code.

I'm not quite sure if I understand why the serialization should use concrete syntax or DSL code. If we had a parser for concrete IR syntax then sure, we could just use that. But in the absence of that, why not serialize to something less human-readable for the use cases you mention? 
Given that we already serialize things to JSON for the web interaction, it should be possible to serialize IR programs to JSON such that they can be de-serialized and processed later on. The bigger challenge would be the serialization of the value environment, but I think some or most of that already exists in order to support the web-related serialization.

jamescheney:
Right, I wasn't necessarily saying we should come up with another separate, human writable concrete syntax (i.e. something requiring its own parser in addition to a pretty printer.)  Serializing to JSON or to Schinks, or just directly to IR expressions that OCaml correctly parses (which I'm not sure is the case for the existing pretty printer)  would all make sense.  I was thinking that serializing IR datatypes to the DSL (compared to JSON or raw IR) would yield something both executable and slightly more readable.  Currently, the pretty-printed IR code is neither executable (as far as I know) nor readable.

I'll think about this, no reason to worry about it in this PR.

tests/unit/ir/schinks.ml
```diff
@@ -0,0 +1,406 @@
...
+    in
+    let varname_to_id =
+      add_binding maker.reserved_ids 0 maker.used_names StringMap.empty
+    in
+    let tname_to_id =
+      add_binding maker.reserved_tids 0 maker.used_tnames StringMap.empty
+    in
+    maker.mk_ir_form { varname_to_id; tname_to_id }
+
+  (* Note that this isn' quite monadic bind due to the param type of the passed function *)
```
jamescheney, line 46: This seems rather strange.  I strongly suspect that the `([-] maker, string) Result.t` is really a well-behaved monad in disguise, probably a combination of Reader/Writer and exception monads.  (Possibly also an applicative/monoidal functor as well.)

Concretely, an `'a t` is something which:
- can fail, returning a string
- if not failing, returns:
  - a function from name_maps to 'a
  - other stuff which was collected up during construction of 'a (`reserved_*ids`, `used_*names`)

I think it does no harm to hoist the name_maps, to give type
```
type 'a maker = name_maps -> {ir_form: 'a, reserved_tids: ...} Result.t
```
and then return would be the obvious thing (ignore the name_maps, generate empty maps/lists), and bind would just be the composition of the exception, writer and reader monad binds.

I'm not sure how much the rest of the code depends on the non-standard bind though - my hope would be that using a more standard type and setup would simplify matters but YMMV.

frank-emrich: Thanks, that's a good point. I suspected that there's a proper monad in there, but didn't quite manage to find it. @rudihorn expressed some interest in working on this, otherwise I will refactor it myself at the end of the week.

rudihorn:
As @frank-emrich  mentioned I'm happy to help out with this one a little. I've been trying to experiment with a few different things in https://github.com/rudihorn/links/tree/ir_dsl. The first is using a state library https://github.com/rudihorn/links/blob/ir_dsl/tests/unit/ir/state.mli and then modeling @frank-emrich's version using a double state monad, where the first stage determines reserved ids and names, then binds the names to ids and recreates the IR node looking up the variable in the state in the second stage. An example of this is the `binder` definition which now looks like the following:

```ocaml
let binder ?(scope = Var.Scope.Local) name ty =
  let* ty = ty in
  let+ () = Repr.add_name ~name in
  let stage2 =
    let* ty = ty in
    let var_info = Var.make_info ty name scope in
    let+ id = Repr.lookup_name ~name in
    Var.make_binder id var_info
  in
  stage2
```

I've had some success defining `let**` and `and**` so that e.g. `if` can nicely be defined as:
```ocaml
let if_ cond then_ else_ =
  let** c = cond and** t = then_ and** e = else_ in
  Ir.If (c, t, e) |> State.return
```

The second thing I have been trying to implement was working with binders (https://ocaml.org/releases/4.08/htmlman/manual046.html) to perform a shallow embeddeding of the Ir language into ML. This allows us to produce expressions like the following, preventing us from having to add stray variable nodes (e.g. `var "x"`) in the code. 

```ocaml
let _foo =
  let g = unbound "g" in
  let f = unbound "f" in
  def "f"
    ([ int ] |~~> int)
    [ ("x", int) ]
    Comp2.(
      (let* x = param "x" string (apply1 g (s "bla")) in
       let+ _y = param "y" string (apply1 f (i 3)) in
       return x)
      |> make)
```

I've also tried a small version which doesn't perform name binding at all (https://github.com/rudihorn/links/blob/ir_dsl/tests/unit/ir/schinks_nost.ml), which should prevent the requirement for two processing stages. With this example, a function definition could look something like the following:

```ocaml
let _comp =
  let* f = unbound in
  let+ g = unbound in
  fun_ T.int []
    Computation.(
      let$* x = bind T.string (apply1 g (s "foo")) in
      let$+ y = bind T.int (apply1 f x) in
      return y)
```

It would need some further modules for handling top level definitions though, as `bind` creates a binder in the background which is used to generate a computation, while binding `var <id_of_x>` to `x` to be used in the tail computation. A similar thing needs to happen with function definitions, which also generate bindings where variable references might be required in further definitions.

For now I think it would be good to discuss how we would like to proceed before adapting any further.

frank-emrich:
> As @frank-emrich mentioned I'm happy to help out with this one a little. I've been trying to experiment with a few different things in https://github.com/rudihorn/links/tree/ir_dsl. The first is using a state library https://github.com/rudihorn/links/blob/ir_dsl/tests/unit/ir/state.mli and then modeling @frank-emrich's version using a double state monad, where the first stage determines reserved ids and names, then binds the names to ids and recreates the IR node looking up the variable in the state in the second stage.

Thanks, this looks quite nice. I haven't looked at the implementation in detail, but I'm happy to hear that you managed to express the creation of IR nodes in terms of state monads, rather than the non-standard thing I used in my version. I would be happy to use this for the implementation instead.


> The second thing I have been trying to implement was working with binders (https://ocaml.org/releases/4.08/htmlman/manual046.html) to perform a shallow embeddeding of the Ir language into ML. This allows us to produce expressions like the following, preventing us from having to add stray variable nodes (e.g. var "x") in the code

To be honest, I'm strongly opposed to the idea of mixing OCaml's custom binders and object-level (i.e., Links IR/Schinks) binders. We cannot express all of the possibilities of binding variables in the Links IR anyway (e.g., closure variable bindings), so there will always be the need for `(var "x")` nodes. Using both looks horrible to me.

Further, I think that using the OCaml custom binders greatly obfuscates things. looking at just
```
let* x = param "x" string (apply1 g (s "bla")) in
let+ _y = param "y" string (apply1 f (i 3) in
...
```
I have very little clue what's going on, which only gets worse in the second example you've shown (after the "I've also tried a small version which doesn't perform name binding at all"). Having `let*` and `let+` looks very unintuitive.
My design goal for the DSL was that by just opening the mli files of the Links IR and the DSL, you should be able to use the DSL and know exactly what the resulting Links IR will look like. I feel like this is not the case any longer when bringing the custom OCaml bindings into the mix.

I'm curious what other people think about this, but my suggestion would be to stick with creating binders the way I've suggested in my original design.

rudihorn:
My second example shows how closure bindings / unbound variables can be declared, which in my opinion makes things more explicit when defining variables and respecting scoping. Take the `_comp` example from above. Unlike in the `var "f"` declaration you know that `f` and `g` are fresh variables, even when the expression is composed. If you don't want them to be fresh, then you have to explicitly pass them as variables and it becomes the calling sites responsibility to instantiate them.

I also think that there is nothing wrong with mixing the bindings. It would be just as legal and probably often a good idea to write an expression `let x = var "x" in ... x ...` in the non-binding version, except here you can save yourself the trouble. 

And while it looks fancy, there is nothing complicated going with the computation declaration which essentially just behaves like a monad. Assuming `type 'a t = ('a, St.t) State.t` and the `_comp` example, `x` and `y` type to regular `Ir.value t` values, `return y` 
types ot a `Ir.tail_computation t`, the `let$+` expressions behaves like a `map`, where the monadic `return` wraps a tail_computation into a computation with no bindings, and `map` adds an extra binding to the list. The `let$*` is then just the equivalent `bind`. Once you have worked with ppx_let / ocaml binding syntax / Haskell do-notation for a little bit this becomes quite intuitive / natural.

frank-emrich:
> My second example shows how closure bindings / unbound variables can be declared, which in my opinion makes things more explicit when defining variables 

So we would need `let*`, `let+`, let?` for binding parameters, function, etc? 

> and respecting scoping.
But that's literally the opposite of what I want. I don't want to have a DSL that a) pretends that the scoping of the Links IR can neatly be expressed using the scoping of OCaml let bindings and b) makes it impossible or more difficult to write certain deliberately ill-scoped IR code.
One of the original example tests I uploaded checks that non-recursive bindings are indeed non-recursive, testing that the IR type-checker throws an error when trying to use a non-recursive function recursively. You cannot write this program if the DSL already enforces scoping rules like this.

In general it seems to me that we simply have different design goals. I want a DSL that makes it easier to write Links IR ASTs. You seem to want something more high-level that makes it easier to write actual programs. Therefore, I don't see any of the advantages you see in using the custom let bindings as advantages for me when writing test cases for the IR type-checker. Instead, I see a great increase in complexity without any benefits for my particular use case.

Therefore, I would suggest that
a) you find a way of making the original syntax a subset of the final syntax of the DSL. Concretely, it means that it's possible to use the DSL without needing to use the custom let bindings, ever.
b) If that's not possible, I would suggest splitting these DSL projects and having two separate DSLs (one "low-level", one "high-level"), possibly sharing the same state monad(s) underneath for the translation to the actual Links IR.

slindley:
Whilst it can be fun to implement various kinds of embedded DSL, this is really not called for here. A better path to follow is to define a really simple grammar for the IR - e.g. based on S-expressions. It doesn't have to be that pretty or even particularly easy to write. The crucial functionality that would be useful is to have a way of serialising (in a human-readable format), editing by hand, and parsing the IR.

Getting something like this working shouldn't take much effort.

We aren't going to be writing anything complicated in this syntax - we already have a nice syntax for more complex examples, and that's the source language!

frank-emrich:
I've reworked this now, which mostly just meant taking Rudis's implementation using two nested state monads (thanks again!) and fixing a few bugs in it.
Happy for people to have a look at it again now.

frank-emrich:
would anyone mind having a look at this? I would like to merge this before doing the Links release

dhil:
I don't see this PR as being critical for the release; from my quick glance over it I don't see anything that would be exposed to users.

Also, I suppose you are likely to be the primary, perhaps only, (internal) user of this framework, so I think you should just go ahead and merge it in (those who want to object have had ~2months to do so).

PR #899, opened 2020-07-18
Raise minimum required OCaml version to 4.08

frank-emrich:
OCaml 4.08 added new customizable let bindings. Defining
```
let ( let* ) = bind
```

means that `let* x = M in N` is desugared to `bind M (fun x -> N)`. The same can be done for some other characters (e.g., `let+`). I'm using this extensively in the (semi-) monadic implementiation of the Links IR DSL (#898), which does not compile with OCaml 4.06.


jamescheney:
As mentioned in discussion of #858, the `-linkall` flag seems necessary in OCaml 4.06 to avoid a problem with dynamic linking of ocaml-mysql, which uses `StdLabels`.  This issue seems to be fixed in OCaml 4.10.  If it is also already fixed in OCaml 4.08, then we can remove these flags (which makes the binary somewhat smaller).  This can be tested by removing the three occurrences of `-linkall` in the main `dune` file.

frank-emrich:
I removed the flag. I assume this would show up in the test suite if this broke something? So it looks like it worked.

jamescheney:
Yes as of #893 the CI exercises the MySQL driver.  The logs show that things are working so looks like `-linkall` is no longer needed in 4.08.  Hurray!

dhil:
If we are going to bump required the version of OCaml, then I suggest we bump it all the way to 4.10.

jamescheney:
Going to 4.10 also seems fine to me.

frank-emrich:
I'm fine with 4.10 if nobody objects, given that it's only been released in February.

jamescheney:
Having had a quick look at the 4.08, 4.09 and 4.10 change logs, there don't seem to be many changes that affect Links after 4.08's addition of definable binding forms.  (Also, reading the changelog, I noted that they made many fixes/improvements to Dynlink in v4.08 which probably explains why the driver dynamic loading works better now.)  On balance, I'd be in favor of supporting the earliest version we can, and increasing this only when forced to, since OCaml now seems to be a bit more of a work in progress.

We could potentially add a CI run that tests with 4.10 (or whatever is the most recent version) as well as the earliest compatible version, to get an early warning if breaking changes to the bleedingest-edge OCaml version affect us.

frank-emrich:
> We could potentially add a CI run that tests with 4.10 (or whatever is the most recent version) as well as the earliest compatible version, to get an early warning if breaking changes to the bleedingest-edge OCaml version affect us.

I think we would have to check what degree of parallelization the existing CI pipelines give us for this. Otherwise, always testing both 4.08 and 4.10 could double our (already annoyingly long) running time of the CI.

jamescheney:
That's a good point, I was thinking it could be two separate / parallel jobs but don't know what is possible/easy to setup.  Not sure it is critical for now, so merging this as is.

Issue #441, opened 2018-12-13
Cleanup block_contents and binding productions in the parser


jstolarek:
In the parser there's a group of productions that look fishy (slightly reformatted and reordered here):

```
block_contents:
| bindings exp SEMICOLON               { ($1 @ [with_pos $loc($2) (`Exp $2)],
                                          with_pos $loc (`RecordLit ([], None))) }
| bindings exp                         { ($1, $2) }
| exp SEMICOLON                        { ([with_pos $loc($1) (`Exp $1)],
                                          with_pos $loc (`RecordLit ([], None))) }
| exp                                  { ([], $1) }
| SEMICOLON | /* empty */              { ([], with_pos $loc (`TupleLit [])) }

bindings:
| binding                              { [$1]      }
| bindings binding                     { $1 @ [$2] }

binding:
| VAR pattern EQ exp SEMICOLON         { with_pos $loc (`Val ($2, ([], $4), `Unknown, None)) }
| exp SEMICOLON                        { with_pos $loc (`Exp $1) }
| signature FUN var arg_lists block    { annotate $loc($1) $1 $loc
                                          (`Fun ($3, `Unl, ($4, with_pos $loc($5) (`Block $5)), `Unknown)) }
| signature LINFUN var arg_lists block { annotate $loc($1) $1 $loc
                                          (`Fun ($3, `Lin, ($4, with_pos $loc($5) (`Block $5)), `Unknown)) }
| FUN var arg_lists block              { with_pos $loc (`Fun (make_untyped_binder $2,
                                                          `Unl, ([], ($3, with_pos $loc($4) (`Block $4))), `Unknown,
                                                          None)) }
| LINFUN var arg_lists block           { with_pos $loc (`Fun (make_untyped_binder $2,
                                                          `Lin, ([], ($3, with_pos $loc($4) (`Block $4))), `Unknown,
                                                          None)) }
| typed_handler_binding                { with_pos $loc (`Handler $1) }
| typedecl SEMICOLON | links_module
| alien_block | links_open { $1 }
```

We could in theory define `bindings` production using Menhir combinators:

```
bindings:
| binding+  { $1 }
```

but that gets us into a shift/reduce conflict as explained here https://github.com/links-lang/links/pull/434#issuecomment-445784429.  TL;DR: a `binding` production has this rule:

```
| exp SEMICOLON  { with_pos $loc (`Exp $1) }
```
  
[...]

Issue #103, opened 2016-12-17
Empty case defaults to unit


dhil:
An empty case expression defaults to unit. For example:
```javascript
links> switch (Bar) { 
                 case Foo -> 
                 case Bar -> print("Hello") 
           };
Hello
() : ()
```
Is this intended? This "feature" seems rather useless to me.

dhil:
I will fix this when I come around to do #57.

dhil:
After studying the grammer in [parser.mly](https://github.com/links-lang/links/tree/master/core/parser.mly) I understand why Links currently exhibits this behaviour. This is because the production rules for `case` and `block_contents`
```
case:
| CASE pattern RARROW block_contents { $2, block ~ppos:$loc($4) $4 }

block_contents:
| ...
| /* empty */ { ([], with_pos $loc (TupleLit [])) }
```
The expansion of `block_contents` may be the empty string, in which case it is interpreted as a nullary tuple literal.

So it has little to nothing to do with #57. I do not think this is bug per see, but rather I think it is a questionable design decision. When we refactor the parser and/or change the grammar or syntax of Links we may want to revisit whether we want to support this special case.

SimonJF:
Ran into this today and was mightily confused for a second. I think this should be a parse error, and will make the necessary changes unless others object.

dhil:
> Ran into this today and was mightily confused for a second. I think this should be a parse error, and will make the necessary changes unless others object.

I think you need to be careful. The change would be rather pervasive. Also, you may have to make the empty block `{ }` a parse error, which means `fun() { }` would be a parse error. The whole parser logic regarding implicit units is rather intricate.

SimonJF:
Perhaps this is one for the Links developer meeting next week.

dhil:
> Perhaps this is one for the Links developer meeting next week.

Just for the record, I am in favour of getting rid of the implicit units.

jamescheney:
This issue seems related to the proposal to get rid of optional semicolons at the end of blocks (#900).  However, this behavior arises because of the line https://github.com/links-lang/links/blob/faa49cd09cc20429122c9def648b25cc3cab991b/core/parser.mly#L696

which allows `block_contents` to appear on the RHS of a `case -> ` branch *without* surrounding braces.  This is, I suppose, to avoid even more curly braces.  But since block_contents also matches an empty token sequence we get the implicit unit behavior, which is undesirable.

[...]

PR #900, opened 2020-07-23
remove option of having a semicolon at the end of a block and forbid empty cases

jamescheney:
This PR resolves part of #441 by disabling the productions in the parser that permissively allow a semicolon at the end of a block (which are interpreted as adding a `()` expression to the end of the block).  Since we require expressions appearing in a block to have type unit if they are evaluated solely for side-effects, I don't believe this change affects the behavior of any well-typed Links programs, and I have (tediously) fixed all of the tests that relied on the permissive behavior so they parse again.

In practice this mostly means that where we previously allowed both this:
```
fun foo(x) {
  bar(y);
  baz(x);
}
```
the parser now forbids the first form and instead you have to write
```
fun foo(x) {
  bar(y);
  baz(x)
}
```
I think this is better because the existing permissive approach allows bad style.  It isn't immediately clear whether semicolons are optional/inferred (like in Scala) or required (like in C), nor is it clear whether the meaning changes depending on whether the semicolon is present or absent.  The new approach exhibits a uniform rule: a (nonempty) block is an expression consisting of a sequence of bindings followed by an expression; its value is returned as the value of the block; if you want to evaluate an expression that returns a non-unit value and ignore its result, you have to explicitly ignore it or bind to `var _` (but this was already the case).

Empty blocks `{}` are still allowed to save us having to write `{()}` though you still can if you want.  The syntax `{;}` which used to be allowed and equivalent to `{()}` or `{}` is now disallowed.

#441 also suggests to replace the productions for `bindings` with:
```
bindings:
| binding+  { $1 }
```
Since then, we have added mutual recursive blocks so we now need to do something like this instead:
```
binding_or_mutual:
| binding                                                      { [$1]      }
| mutual_binding_block                                         { $1        }

bindings:
| binding_or_mutual+                                            { List.concat $1 }
```
However, when I do this (or other equivalent-seeming things such as using `nonempty_list`), menhir complains with various errors, and I'm not sure what I'm doing wrong, since the `+` notation and `nonempty_list` supposedly expand to something like:
```
bindings:
| binding_or_mutual                                            { $1 }
| bindings binding_or_mutual                                   { $1 @ $2 }
```
which does work; I can also see other places where we are using `+` and it seems to work.

Marking as WIP because of the above problem and because this is a breaking change to the syntax (though I would argue, just codifies what is already good style).

TODO:
- [x] See if people like this change
- [x] Figure out what menhir's problem is
- [x] Remove commented out code and cleanup


jamescheney:
As discussed in #103, modified to stop empty cases defaulting to unit.

jstolarek:
> when I do this (or other equivalent-seeming things such as using `nonempty_list`), menhir complains with various errors, and I'm not sure what I'm doing wrong

This is the exact same problem I've run into the previous time. Here's a peek at parser debug files - maybe someone will be able to make more sense of this. Assuming the `bindings` production is defined in a way that does not lead to conflicts:

```
bindings:
| binding_or_mutual             { $1 }
| bindings binding_or_mutual    { $1 @ $2 }
```

We have the following states in the automaton:

```
State 835:
bindings -> bindings . binding_or_mutual [ VARIABLE ... ]
-- On VARIABLE shift to state 1

State 837:
bindings -> bindings binding_or_mutual . [ VARIABLE ... ]
-- On VARIABLE reduce production bindings -> bindings binding_or_mutual

State 838:
binding_or_mutual -> binding . [ VARIABLE ... ]
-- On VARIABLE reduce production binding_or_mutual -> binding

State 839:
bindings -> binding_or_mutual . [ VARIABLE ... ]
-- On VARIABLE reduce production bindings -> binding_or_mutual
```

If we now define `bindings` as:

```
bindings:
| binding_or_mutual+            { List.concat $1 }
```

we have:

```
State 838:
nonempty_list(binding_or_mutual) -> binding_or_mutual . [ VARIABLE ... ]
nonempty_list(binding_or_mutual) -> binding_or_mutual . nonempty_list(binding_or_mutual) [ VARIABLE ... ]
-- On VARIABLE shift to state 1
-- On VARIABLE reduce production nonempty_list(binding_or_mutual) -> binding_or_mutual

State 839:
nonempty_list(binding_or_mutual) -> binding_or_mutual nonempty_list(binding_or_mutual) . [ VARIABLE ... ]
-- On VARIABLE reduce production nonempty_list(binding_or_mutual) -> binding_or_mutual nonempty_list(binding_or_mutual)

State 840:
binding_or_mutual -> binding . [ VARIABLE ... ]
-- On VARIABLE reduce production binding_or_mutual -> binding
```

where State 838 is the one that causes a shift/reduce conflict, and I believe that state is a product of merging states 835 and 837 from the conflict-free grammar.  I looked at other definitions that use `+` operator and it seems the reason why these other productions don't cause conflicts is because they are always followed by a terminal symbol.  So for example if we write:

```
bindings:
| binding_or_mutual+ SEMICOLON      { List.concat $1 }
```

then the shift/reduce conflict goes away. (I am not proposing we actually do this, just demonstrating what happens.)  I speculate that the conflict does not happen in the original grammar because when we define `bindings` as two separate productions we force there to be two separate parser states which, when merged (as in second definition) are actually conflicting.

At this point I think that making the parser one line shorter is not worth investing further time into this.  I added a comment pointing to this and previous discussion so that anyone who sees `bindings` production and thinks it can be simplified has a starting point.

jamescheney:
That's pretty strange.  I wonder if it is worth shrinking this down to a minimal example and reporting as a possible bug in menhir; one certainly gets the impression from the documentation that `foo+ {e}` is just syntactic sugar for two rules
```
nonempty_foo_list {e}

nonempty_foo_list: 
| foo {[$1]}
| nonempty_foo_list foo { $1 @ [$2] }
```
but clearly that's not the case...

I agree not really worthwhile to try to work around here.

jamescheney:
Meeting discussion: I think we agreed this is reasonable but perhaps someone could take a closer look before merging.  Meanwhile I'll remove the commented code and sync with the current branch.

jstolarek:
I can take another look at parser changes once the code is cleaned up.

jamescheney:
Thanks, this is now done.

core/parser.mly
```diff
@@ -839,26 +839,27 @@ mutual_bindings:
...
-                                                                  record ~ppos:$loc []) }
+case_contents:
 | bindings exp                                                 { ($1, $2) }
-| exp SEMICOLON                                                { ([with_pos $loc($1) (Exp $1)],
-                                                                  record ~ppos:$loc []) }
 | exp                                                          { ([], $1) }
-| SEMICOLON | /* empty */                                      { ([], with_pos $loc (TupleLit [])) }
+
+block_contents:
+case_contents                                                  { $1 }
```
jstolarek, line 861: We always start the first production with a leading `| `.

jamescheney: Done

dhil:
> That's pretty strange.  I wonder if it is worth shrinking this down to a minimal example and reporting as a possible bug in menhir; one certainly gets the impression from the documentation that `foo+ {e}` is just syntactic sugar for two rules
> ```
> nonempty_foo_list {e}
>
> nonempty_foo_list: 
> | foo {[$1]}
> | nonempty_foo_list foo { $1 @ [$2] }
> ```
> but clearly that's not the case...

I originally intended to comment on this, but I placed it on my stack. Now my stack has been popped.

According to the documentation of Menhir `foo+` is syntactic sugar for `nonempty_list(foo)` which is implemented as
```
%public nonempty_list(X):
  x = X
    { [ x ] }
| x = X; xs = nonempty_list(X)
    { x :: xs }
```
(see line 219 in [standard.mly](https://gitlab.inria.fr/fpottier/menhir/blob/master/src/standard.mly#L219))

Thus I think the behaviour you experience is the "intended" behaviour -- at least from the point of view of the tool.

Issue #902, opened 2020-07-26
Radio button groups in raw forms do not work


jamescheney:
Here is a minimal example:
```
fun handleForm(i) {
  page
    <html>
      <body>
      <p>param: {stringToXml(i)} </p>
      </body>
    </html>
}

fun mainPage (_) {
  page
    <html>
    <body>
      <h1>Example form</h1>
      <form l:action="{handleForm(i)}">
		<input type="radio" l:name="i" value="A"> A</input>
		<input type="radio" l:name="i" value="B"> B</input>
        <button type="submit">Submit!</button>
      </form>
      </body>
    </html>
}

fun main () {
  addRoute("",mainPage);
  servePages()
}

main()
```

`desugarLAttributes` fails with the following message:
```
issue208b.links:15: Error compiling attributes: Duplicate l:name binding: i
In expression: <form l:action="{handleForm(i)}">
		<input type="radio" l:name="i" value="A"> A</input>
		<input type="radio" l:name="i" value="B"> B</input>
        <button type="submit">Submit!</button>
      </form>.
```
This is because `desugarLAttributes` assumes that `l:name` fields are unique in a form, whereas in HTML, radio button groups are defined by using the same `name` attribute across several instances of `radio` input fields.  This behavior should be relaxed to allow radio buttons with the same name.

PR #903, opened 2020-07-26
handle checkbox and radio group form elements

### LINKED ISSUE #208: checkboxes in raw HTML forms
Example code:
```

fun handleForm(p) {
  page
    <html>
      <body>
      <p>param: {stringToXml(p)} </p>
      </body>
    </html>
}

fun mainPage (_) {
  page
    <html>
    <body>
      <h1>Example form</h1>
      <form l:action="{handleForm(i)}">
		<input type="checkbox" l:name="i" /> Test
      </form>
      </body>
    </html>
}

fun main () {
  addRoute("",mainPage);
  servePages()
}

main()
```

Desired output:
The resulting page should show "param: on" if the checkbox is checked on submission, "param: off" or something like that if not.

Actual output:
"param: on" does correctly appear if the checkbox is checked, but if not we get an error 

```
Links Fatal Error
Not found lname__g126 
```
because in checkboxes in HTML forms, nothing is sent if the checkbox is not checked. (see [HTML5 spec](https://www.w3.org/TR/2014/REC-html5-20141028/forms.html#constructing-form-data-set))

Proposed fix: 
Instead of raising an error, the missing form fields should be initialized with a default empty-string value (because the "value" of the checkbox can be set to any string by the user).  Using "option" as the type of submitted checkbox fields would be slightly cleaner.

Anyway, I know that the right answer is "use formlets", but given that we still support raw HTML-style forms, checkboxes should work.

jamescheney:
This PR adjusts the code inserted when desugaring `l:` attributes.  

On the server side, if an expected `l:` attribute name/id is not present in the field parameters, it is initialized to the empty string.

On the client side, if the element is a checkbox or radio button, then the `value` field is returned only if the `checked` field is true.  Otherwise, the empty string is returned instead.

This fixes #208, but radio buttons/groups of more than one radio button still do not work. See #902, which I'll try to fix as well.

### DIFF
--- a/lib/js/jslib.js
+++ b/lib/js/jslib.js
@@ -1766,6 +1766,21 @@ function _getInputValue(id) {
 
 var getInputValue = LINKS.kify(_getInputValue);
 
+// getRadioGroupValue : [String] -> String
+function _getRadioGroupValue(ids) {
+  let ptr = ids;
+  while ( LINKEDLIST.isNil(ptr) == false ) {
+    var element = document.getElementById(ptr._head);
+    if (element.checked) {
+      return element.value;
+    }
+    ptr = ptr._tail;
+  }
+  return "";
+}
+
+var getRadioGroupValue = LINKS.kify(_getRadioGroupValue);
+
 //getValue : domRef -> xml
 //    NOTE: this is recursive.
 function _getValue(nodeRef) {
--- a/lib/js/jslib.js
+++ b/lib/js/jslib.js
@@ -251,7 +251,7 @@ const LINKEDLIST =  (function(){
     some: function(f, xs) {
       let ptr = xs;
       let found = false;
-      while ( (found || LINKEDLIST.isNil(ptr)) == false ) {
+      while ( (found || LINKEDLIST.isNil(ptr)) === false ) {
           found = f(ptr._head);
           ptr = ptr._tail;
       }
@@ -1754,10 +1754,10 @@ function _nodeTextContent(node) {
 
 // getInputValue : String -> String
 function _getInputValue(id) {
-  var element = document.getElementById(id);
+  const element = document.getElementById(id);
   DEBUG.assert(element != null, "invalid input node (id " + id + ")");
   DEBUG.assert(element.value != undefined, "invalid input value in id " + id)
-  if ((element.type != "radio" && element.type != "checkbox") || element.checked) {
+  if ((element.type !== "radio" && element.type !== "checkbox") || element.checked) {
     return element.value;
   } else {
     return "";
@@ -1769,8 +1769,8 @@ var getInputValue = LINKS.kify(_getInputValue);
 // getRadioGroupValue : [String] -> String
 function _getRadioGroupValue(ids) {
   let ptr = ids;
-  while ( LINKEDLIST.isNil(ptr) == false ) {
-    var element = document.getElementById(ptr._head);
+  while ( LINKEDLIST.isNil(ptr) === false ) {
+    const element = document.getElementById(ptr._head);
     if (element.checked) {
       return element.value;
     }
--- a/lib/js/jslib.js
+++ b/lib/js/jslib.js
@@ -1757,7 +1757,11 @@ function _getInputValue(id) {
   var element = document.getElementById(id);
   DEBUG.assert(element != null, "invalid input node (id " + id + ")");
   DEBUG.assert(element.value != undefined, "invalid input value in id " + id)
-  return element.value;
+  if ((element.type != "radio" && element.type != "checkbox") || element.checked) {
+    return element.value;
+  } else {
+    return "";
+  }
 }
 
 var getInputValue = LINKS.kify(_getInputValue);
--- a/core/desugarLAttributes.ml
+++ b/core/desugarLAttributes.ml
@@ -30,6 +30,9 @@ let server_use name =
 let client_use id =
   apply "getInputValue" [constant_str id]
 
+let client_radiogroup_use ids =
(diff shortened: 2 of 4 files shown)


jamescheney:
Now updated to also handle radio groups.

To test, it is necessary to replace the jslib.js file with the one in this branch since changes / additions were made to the javascript to account for checkboxes and radio buttons.

lib/js/jslib.js
```diff
@@ -1757,11 +1757,30 @@ function _getInputValue(id) {
...
+    return "";
+  }
 }
 
 var getInputValue = LINKS.kify(_getInputValue);
 
+// getRadioGroupValue : [String] -> String
+function _getRadioGroupValue(ids) {
+  let ptr = ids;
+  while ( LINKEDLIST.isNil(ptr) == false ) {
```
dhil, line 1772: Dare I say that we ought to use strict equality here `===`? Without having looked at the implementation of `isNil`, my faint hope is that it returns either `true` or `false` and not any other truthy or falsy value.

lib/js/jslib.js
```diff
@@ -1757,11 +1757,30 @@ function _getInputValue(id) {
...
+    var element = document.getElementById(ptr._head);
```
dhil, line 1773: Use `const` rather than `var` here. The former is block-scoped, whilst the latter is function-global scoped.

lib/js/jslib.js
```diff
@@ -1757,11 +1757,30 @@ function _getInputValue(id) {
...
+    if (element.checked) {
+      return element.value;
+    }
+    ptr = ptr._tail;
+  }
+  return "";
```
dhil, line 1779: Comment: we ought to define `String.empty` for the empty string somewhere in jslib.

lib/js/jslib.js
```diff
@@ -1757,11 +1757,30 @@ function _getInputValue(id) {
   var element = document.getElementById(id);
   DEBUG.assert(element != null, "invalid input node (id " + id + ")");
   DEBUG.assert(element.value != undefined, "invalid input value in id " + id)
-  return element.value;
+  if ((element.type != "radio" && element.type != "checkbox") || element.checked) {
```
dhil, line 1760: If `element.type` is always (supposed to be) a string, then I think you want to use strict inequality `!==` in both instances.

lib/js/jslib.js
jamescheney: Happy to fix this, but I based this on https://github.com/links-lang/links/blob/faa49cd09cc20429122c9def648b25cc3cab991b/lib/js/jslib.js#L254 

so maybe that should be changed to use `===` too.

lib/js/jslib.js
jamescheney: Hmm.  it might be undefined (if the element whose id we looked up lacked the `type` attribute).  If the element type is undefined then I think we should just return the value (this is most likely due to a bug in the form, and we are not trying very hard to check for this class of bugs).  We could be more careful here by checking that each input field in a form does have a type in `desugarLAttributes`.

lib/js/jslib.js
jamescheney: Again this was based on 
https://github.com/links-lang/links/blob/faa49cd09cc20429122c9def648b25cc3cab991b/lib/js/jslib.js#L1757

so should that be changed too?

lib/js/jslib.js
jamescheney: There are around 20 occurrences of `""` in `jslib.js`.  Should they all be replaced by `String.empty` or just some of them?

lib/js/jslib.js
dhil: We should possibly make an issue to clean up `jslib.js`.

lib/js/jslib.js
dhil: Using `const` over `var` for L1757 instance should not make much of a difference as the variable is function-global scoped regardlessly. However, `const` (or `let`) might help the compiler apply some optimisations, although in this specific case I don't think there is much room for optimisations.

lib/js/jslib.js
dhil: Quite possibly! I am not familiar with every instance of `""` in `jslib.js`. My thinking is that we ought to have a module for each primitive type to ensure some consistency. But I think this would be implemented in a separate patch.

lib/js/jslib.js
dhil: > We could be more careful here by checking that each input field in a form does have a type in `desugarLAttributes`.

That sounds entirely sensible to me!

lib/js/jslib.js
jamescheney: OK

lib/js/jslib.js
jamescheney: While that would be a good idea, it would be stricter than (de facto) HTML requires... in HTML5 an `<input>` tag without a `type` is assumed to hhave `type=text`.  So, as long as the behavior is the same for `element.type === undefined` as for `text`, we should be OK.  

There are other things we are not really checking such as that `l:name` fields are only used in sensible places inside forms, for example on `input` or `textarea` but not ordinary content tags.  (Of course, they should have no effect there and this patch will result in their variables being bound to `""`.)  

On balance, having dipped a toe in the chilly waters of `l:` attribute desugaring and raw forms, I lean towards thinking that modernizing and extending formlets would probably be a better use of time, to cover the use cases currently only supported by raw forms while retaining composability and nicer high-level typechecking.

lib/js/jslib.js
jamescheney: Be my guest - I have only looked at enough of `jslib` to figure out how to solve this specific problem.

PR #904, opened 2020-07-27
Fix #776 and reorder irTraversals return values to match

### LINKED ISSUE #776: Inconsistent order of return arguments between TYPE_VISITOR and other traversals
`SugarTraversals.fold_map` and `TransformSugar.transform` return `self` object as the first component of the returned tuple, but `Types.Transform` (of type `TYPE_VISITOR`) returns `self` as the last component. We should make this consistent by swapping order of return components in `TYPE_VISITOR`.

jamescheney:
This attempts to fix #776, a tedious refactoring cleanup which I found strangely relaxing.  Keeping as draft since I also wondered whether we might want to further refactor the IR traversals to have the same argument order, see #776.


jamescheney:
Transient CI error?

PR #906, opened 2020-07-31
attempt to fix #219 by forcing pickleCont to be performed on the server

### LINKED ISSUE #219: replaceNode fails silently if the xml argument contains a l:action form
# Input
The following code snippet uses `replaceNode` to add two XML form snippets to a document, one using `l:action` and the other using `l:onsubmit`.

```
fun testPage(msg){
  page
  <html>
    <body>
      <div id="t">
        <h4>HAHAH</h4>
      </div>
      <div id="p">
        <p>LALA</p>
      </div>
      <a href="" l:onclick="{changeContent()}">Click me</a>
    </body>
  </html>
}

fun changeContent(){
  replaceNode(
    <div id="t">
      <form l:onsubmit="{}" method="post">
        <input type="text" l:name="i" value="" />
      </form>
    </div>,
    getNodeById("t")
  );
  replaceNode(
    <div id="p">
      <form l:action="{testPage(i)}" method="post">
        <input type="text" l:name="i" value="" />
      </form>
    </div>,
    getNodeById("p")
  )
}


fun main () {
  addRoute("",testPage);
  servePages()
}

main()
```

# Expected output

When the button is clicked, both HAHAH and LALA should change to input fields; the first should have no results and the second should call testPage again with a new message (this can never actually happen because there is no way to submit the forms, the main thing is that the content should change.)

# Actual output

HAHAH changes to an input field but LALA does not.  Inspecting the (Firefox) JS console suggests that what is happening is a failure in jslib when base64decode is called during a call to _remoteCallHandler (at the point where the server response is sent to _debug).  This in turn suggests that something is going wrong somewhere else, but I don't know what.

I'm uncertain whether this is something that just should not work (in which case we should catch the error somehow) or it should work but doesn't.

Swapping the order of the two replaceNode calls also confirms that the `l:action` call fails, since if we try to do that replacement first then nothing happens.

jamescheney:
Since #219 happens because calling `pickleCont` from the client is nonsensical, one way to fix this is to force this to happen on the server.  This entails additional round-trips from the client to the server, but otherwise seems unproblematic.  Let's see if the CI agrees.


jamescheney:
I also wondered whether we should force `unsafePickleCont` to be evaluated on the server, but actually it already is; so weirdly, things work if `pickleCont` is evaluated on the server (and all it does is call a server function wiht a server closure as its argument) but not if the exact same code is evaluated on the client.

So perhaps the other place `unsafePickleCont` is called, in `freshResource`, should also be explicitly marked as a server function.

jamescheney:
I thought this might be because `pickleCont` calls `unsafePickleCont` with an eta-expanded argument, but this does not matter (eta-reducing and removing the `server` annotation yields the same problem).  There appears to be no way to force a local function definition to be a client or server function since location annotations are only allowed on toplevel function definitions.  So I guess that the location of an inner function is always inherited from that of the toplevel function in which it is defined (or wherever the closure is created) which means that if we want to ensure the `unsafePickleCont` call and its argument are both available to the server then the call to `pickleCont` has to be marked server.

I am curious what happens if we call `pickleCont` from the client with a closure that refers to variables defined on the client.   Does the right thing happen?

Short answer: no, in fact the continuations that should be called by submitting the form aren't valid, and result in the error:
```
***: Error: Links_core.Evalir.Exceptions.EvaluationError("Attempting to apply (_closureTable = 1) (need function instead)") 
```

On the other hand, the snippet-ized versions do work.  
```
mutual {
fun testPage(msg) {
  page
  <html>
    <body>
	<div>Message: {stringToXml(msg)}</div>
      <div id="t">
        <h4>HAHAH</h4>
      </div>
      <div id="p">
        <p>LALA</p>
      </div>
      <div id="x">
        <p>GAGA</p>
      </div>
      <a href="" l:onclick="{changeContent()}">Click me</a>
    </body>
  </html>
}

fun snippet1 (time) server {
      <form l:action="{testPage(intToString(time))}" method="post">
        action handled OK at <b>{intToXml(time)}</b>
      </form>
}

fun snippet2(time) server {
        <a l:href="{testPage(intToString(time))}">href handled OK at <b>{intToXml(time)}</b></a>
}

fun changeContent() client {
  var time = clientTime();
  replaceNode(
      <form l:onsubmit="{error("asdf")}" method="post">
        onSubmit handled OK
      </form>
    ,
    getNodeById("t")
  );
  replaceNode(snippet1(time),getNodeById("p"));
  replaceNode(snippet2(time),getNodeById("x"))
}
}

fun main () {
  addRoute("",testPage);
  servePages()
}

main()

```
So perhaps a way to fix this would be to desugar forms to server-side functions that do the form construction on the server so that the right thing happens.  Even though there is no source-level syntax for labeling local functions as client or server, we can (and `desugarLAttributes` already does) do this in desugaring.  So hopefully locally desigaring into a server call would do the right thing and be handled correctly by closure conversion (much) later.

jamescheney:
As an experiment, I tried enabling client and server annotations for local functions.  This results in assertion failures in evalir so I suspect we are assuming in various places that only toplevel functions carry location annotations.  I think that should be fixed first and then #219 can be fixed by desugaring forms that contain server-side continuations to code that RPCs back to the server.  I still think documenting that pickleCont only makes sense on the server is a good idea so keeping this open for now for discussion.

jamescheney:
Closing pending #907 as discussed in #219

jamescheney:
The error message above that made me give up earlier seems to be a duplicate of #1050.  Reopening to see  if #1118 fixes this or at least leads to some progress.

jamescheney:
The error message above that made me give up earlier seems to be a duplicate of #1050.  Reopening to see  if #1118 fixes this or at least leads to some progress.

jamescheney:
It doesn't immediately help (some problems reported in #219 persist even with the server annotation).

Issue #869, opened 2020-06-30
Proposal: SML-style function definition groups


dhil:
I'd like to propose augmenting Links with SML-style function definition groups. This proposal is entirely about improving the quality of life for Links programmers.

## What: Function definition group
In functional languages such as SML and Haskell the definition of function can be given be a group of partial definitions, technically known as a function definition group. For example, in SML the following three definitions form a function definition group which defines the Ackermann function:

```sml
fun ack (0, n) = n + 1
  | ack (m, 0) = ack (m - 1, 1)
  | ack (m, n) = ack (m - 1, ack (m, n - 1))
```

Each line provides a partial definition of the function. The first definition of the group is prefixed by `fun` and subsequently definitions are prefixed by `|`. When calling `ack` at runtime the temporal order of definitions is used to determine which definition to execute.

## Motivation

Function definition groups provide a form of declarative programming that can aid readability and writability of programs by liberating the programmer from performing control flow analysis. For example consider the `ack` function above in implemented in Links

```links
fun ack(m, n) {
  if (m == 0) n + 1
  else if (n == 0) ack(m - 1, 1)
  else ack(m - 1, ack(m, n - 1))
}
```

The concrete shape of the control flow is an implementation detail, here is an equivalent implementation with a slightly more involved control flow

```links
fun ack'(m, n) {
  if (m <> 0 && n == 0) ack'(m - 1, 1)
  else if (m <> 0 && n <> 0) ack'(m - 1, ack'(m, n - 1))
  else n + 1
}
```

The point is the concrete control flow is not illuminating. By using function definition groups a clever compiler can potentially pick the optimal implementation for the programmer.

In Links, one could make use of a `switch`-expression to implement the `ack` function

```links
fun ack''(m, n) {
  switch ((m, n)) {
    case (0, n) -> n + 1
    case (m, 0) -> ack''(m - 1, 1)
    case (m, n) -> ack''(m - 1, ack''(m, n - 1))
  }
}
```

Arguably this implementation is more readable than the previous implementation `ack'`, and it has a declarative flavour closer to that of function definition groups. Though, there is one caveat. This particular implementation is potentially less efficient as it requires the "arguments" to be boxed. A clever compiler may be able to figure out that this boxing is unnecessary.

## Concrete syntax

I propose to adopt a syntax similar to the syntax used by SML, i.e. the first definition of a definition group is prefixed by `fun` and subsequent definitions are prefixed by `|`. Concretely, the `ack` function would be implemented as

```links
[...]

PR #908, opened 2020-08-04
Proposal 869: pattern-matching syntax

Emanon42:
This pull request implements #869 

To enable this syntax sugar, set `pattern_matching_sugar=true` in config  

### Syntax
It adds syntax of SML-style pattern-matching like that:
```
fun ack(_,_) switch {
  case (0, n) ->  n + 1
  case (m, 0) -> ack(m - 1, 1)
  case (m,n) ->  ack(m - 1, ack(m, n - 1)) 
}
```
### Semantics
Desugaring the cases defined in pattern-matching function to a `Switch` phrase and adding a wild card case to raise an error to handle the non-exhaustive case (with position of wrong code).


SimonJF:
CI failure appears transient.

dhil:
Thanks for opening this PR! I'll take a thorough look at this later.

core/desugarModules.ml
```diff
@@ -307,16 +307,28 @@ and desugar ?(toplevel=false) (renamer' : Epithet.t) (scope' : Scope.t) =
...
+          let visitor = self#clone in
+          let paramss' =
+            List.map
+              (fun params ->
+                List.map (fun param -> visitor#pattern param) params)
+              paramss
+          in
+          let body' = visitor#phrase body in
+          NormalFunlit (paramss', body')
+        | MatchFunlit (paramss, body) ->
```
jamescheney, line 322: If I understand correctly, `MatchFunLit` will have been desugared away before module desugaring starts.  So this case can just be `assert false`.

core/frontend.ml
```diff
@@ -123,7 +123,8 @@ module Untyped = struct
 
   (* Collection of transformers. *)
   let transformers : transformer array
-    = [| (module ResolvePositions)
+    = [| (module DesugarMatching)
```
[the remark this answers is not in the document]
jamescheney: (but before `desugarModules` if my earlier comment is adopted.)

core/lexer.mll
```diff
@@ -134,6 +134,7 @@ let keywords = [
  "var"      , VAR;
  "where"    , WHERE;
  "with"     , WITH;
+ "match"    , MATCH;
```
jamescheney, line 137: I had suggested `fun (...) match {case ...}` but I think @dhil preferred to use `switch` and not add a new keyword, which I'd also be happy with.

core/parser.mly
```diff
@@ -443,6 +446,15 @@ tlfunbinding:
...
 | OP pattern OPERATOR perhaps_location block                   { ((dl_unl, false), $3, [[$2]], $4, $5)          }
 
+match_tlfunbinding:
+| fun_kind VARIABLE arg_lists perhaps_location match_body      { ($1, $2, $3, $4, $5)                }
+
+match_body:
+| MATCH LBRACE match_cases* RBRACE                             { $3 }
+
+match_cases:
+| VBAR pattern RARROW block_contents                           { ($2, block ~ppos:$loc $4) }
```
jamescheney, line 456: This production matches a single case, not cases.  (so call the nonterminal `match_case`)

core/parser.mly
```diff
@@ -426,6 +427,8 @@ fun_declarations:
 fun_declaration:
 | tlfunbinding                                                 { fun_binding ~ppos:$loc($1) None $1 }
 | signatures tlfunbinding                                      { fun_binding ~ppos:$loc($2) (fst $1) ~unsafe_sig:(snd $1) $2 }
+| match_tlfunbinding                                           { match_fun_binding ~ppos:$loc($1) None $1 }
+| signatures match_tlfunbinding                                { match_fun_binding ~ppos:$loc($2) (fst $1) ~unsafe_sig:(snd $1) $2 }
```
jamescheney, line 432: It seems that `match` can be used in toplevel bindings or anonymous functions but not ordinary function bindings (or toplevel operator definitions).  Why not?  I think the desugaring should work for them too without problems.

On the other hand, I'm not a huge fan of the duplication of rules here and below.

core/parser.mly
```diff
@@ -547,6 +559,7 @@ primary_expression:
 | LBRACKET exp DOTDOT exp RBRACKET                             { with_pos $loc (RangeLit($2, $4))   }
 | xml                                                          { $1 }
 | linearity arg_lists block                                    { fun_lit ~ppos:$loc $1 $2 $3 }
+| linearity arg_lists match_body                               { match_fun_lit ~ppos:$loc $1 $2 $3 }
```
jamescheney, line 562: Are there tests exercising this syntax (anonymous match functions)?

core/sugartypes.ml
```diff
@@ -369,7 +369,10 @@ and regex =
   | Splice    of phrase
   | Replace   of regex * replace_rhs
 and clause = Pattern.with_pos * phrase
-and funlit = Pattern.with_pos list list * phrase
+and funlit = NormalFunlit of normal_funlit | MatchFunlit of match_funlit
+and match_funlit = Pattern.with_pos list list * match_body
+and match_body = (Pattern.with_pos * phrase) list
```
jamescheney, line 374: I think this should be `(Pattern.with_pos list * phrase) list`, where if the (curried) function takes several lists of patterns the first argument is the list of curried patterns rather than a single tuple pattern. Indeed, the data of a single case should be the same as `funlit`, so you could just define `match_body` as `funlit list`.

Alternatively, forbidding `match` on functions that take multiple parameter lists and specializing to 
```
and funlit = NormalFunlit of normal_funlit | MatchFunlit of match_funlit
and match_funlit = Pattern.with_pos list * match_body
and match_body = (Pattern.with_pos * phrase) list
```
would be fine.

core/lexer.mll
jamescheney: On the other hand, since `switch` is paired with `case` and doesn't handle curried arguments, maybe having a different keyword is justifiable - it just seems strange to have both, with different case syntax.

tests/functions.tests
```diff
@@ -146,3 +146,19 @@ Linearity (9) - linear recursive functions should be disallowed
 linfun f(x) {f(x)} f
 stderr : @.*cannot be linear.*
 exit : 1
+
+Pattern Matching (1):
+fun foo(_) match {| x -> x} foo
```
jamescheney, line 151: As a matter of visual style I suggest having a space between `{` and `|` since otherwise it looks like `{|` is a single token - I might then expect to see a closing `|}` at the end.  (Links uses some other similar-looking tokens as parentheses such as `[| ... |]` for variants.)

Alternatively perhaps the initial "|" in a match block could be optional, or perhaps we could explicitly require "match braces" around match blocks, i.e. 

```
 fun foo(_) match {| x -> x |}
 fun ack(_,_) match {| (0, n) -> n + 1 | (m, 0) -> ack(m - 1, 1) | (m, n) -> ack(m - 1, ack(m, n - 1)) |} 
```

(In fact if we did this (by adding new tokens `{|`, `|}` then the `match` keyword would be redundant...)

jamescheney:
Discussed at the meeting: `match` -> `switch`, '|' -> `case`, make it an error to use with curried functions.

jamescheney:
Plus I forgot to record that someone suggested guarding this behavior behind a flag.  Perhaps this could be implemented simply by having the `desugarMatch` desugaring pass be controlled by an option.  Then later passes will fail somewhere if the new syntax is used, so perhaps it would also be necessary to put a flag-dependent error message somewhere.

core/sugartypes.ml
Emanon42: I think the second one looks more sensible. What is the example of a curried function with pattern matching?

core/parser.mly
Emanon42: Added support to ordinary function bindings. Consider they are calling different sugar constructors, I don't know how to merge them.

Emanon42:
> Plus I forgot to record that someone suggested guarding this behavior behind a flag. Perhaps this could be implemented simply by having the `desugarMatch` desugaring pass be controlled by an option. Then later passes will fail somewhere if the new syntax is used, so perhaps it would also be necessary to put a flag-dependent error message somewhere.

Is there a similar feature controlled by a flag that I can reference?

jamescheney:
The `relational_lenses` setting works this way: if you search for where it is defined and used, and do something similar (e.g. create a guard/error that checks the flag before doing any desugaring, and raises an error if the `switch` form is used when it's not enabled) that should work.

Incidentally, the tests are failing because they need to be updated to use `switch` now.  I'm not sure where this was left, hopefully we can finalize this soon.

Emanon42:
Sure. The thing left is not much. We can make it done soon.

core/renamer.ml
```diff
@@ -213,11 +215,12 @@ let renamer qs_from qs_to =
 let rename_function_definition : function_definition -> function_definition =
   fun { fun_binder
       ; fun_linearity
-      ; fun_definition = (tyvars_from, (pats, body))
+      ; fun_definition = (tyvars_from, f)
       ; fun_location
       ; fun_signature
       ; fun_frozen
       ; fun_unsafe_signature } ->
+  let (pats, body) = match f with | NormalFunlit (ps, bd) -> (ps, bd) | MatchFunlit (_,_) -> assert false in
```
dhil, line 223: Please split this over multiple lines...
```suggestion
  let (pats, body) = 
     match f with 
     | NormalFunlit (ps, bd) -> (ps, bd) 
     | MatchFunlit (_,_) -> assert false 
  in
```
... or define and use a suitable extraction function, e.g.
```ocaml
let normal_funlit = function
   | NormalFunlit def -> def
   | _ -> assert false
```

core/sugartoir.ml
```diff
@@ -1180,9 +1180,10 @@ struct
                     let defs =
                       List.map
                         (fun { rec_binder           = bndr;
-                               rec_definition       = ((tyvars, _), (pss, body));
+                               rec_definition       = ((tyvars, _), fnlit);
                                rec_location         = location;
                                rec_unsafe_signature = unsafe; _ } ->
+                          let (pss, body) = match fnlit with | NormalFunlit (pss, body) -> (pss, body) | MatchFunlit (_,_) -> assert false in
```
dhil, line 1186: Please split this over multiple lines.

core/transformSugar.ml
```diff
@@ -130,6 +130,9 @@ let on_effects o (eff : Types.row) fn x =
   let (o, x, y) = fn o x in
   (o#with_effects effect_row, x, y)
 
+let get_normal_funlit fnlit =
+  match fnlit with | NormalFunlit x -> x | MatchFunlit _ -> assert false
```
dhil, line 134: Please put each case on its own line.

core/typeSugar.ml
```diff
@@ -4187,13 +4188,13 @@ and type_binding : context -> binding -> binding * context * Usage.t =
...
                fun_linearity = lin;
-               fun_definition = (_, (pats, body));
+               fun_definition = (_, fnlit);
                fun_location;
                fun_signature = t_ann';
                fun_frozen;
                fun_unsafe_signature = unsafe } =
            Renamer.rename_function_definition def in
-
+          let (pats, body) = match fnlit with | NormalFunlit (pats, body) -> (pats, body) | MatchFunlit (_,_) -> assert false in
```
dhil, line 4197: Layout.

dhil (changes requested):

core/desugarMatching.ml
```diff
@@ -54,6 +62,7 @@ object ((self : 'self_type))
           (* bind the arguments with unique var name *)
           let name_list = List.map (fun pats -> List.map (fun pat -> (pat, Utility.gensym())) pats) patterns in
           let switch_tuple = List.map (fun (_, name) -> with_pos (Var name)) (List.flatten name_list) in
+          nullary_guard switch_tuple pos;
```
dhil, line 65: This is somewhat nitpick, but we might as well move this check up such that it is executed right after the flag check (`pattern_matching_sugar_guard`). It'd require the implementation of `nullary_guard` to match on `[[]]` rather than `[]`.

```suggestion
          nullary_guard patterns pos;
```

core/desugarMatching.ml
```diff
@@ -43,6 +43,14 @@ let pattern_matching_sugar_guard pos =
...
   then raise (pattern_matching_sugar_disabled pos)
 
+let nullary_guard tuple pos =
+  let nullary_error pos =
+    Errors.desugaring_error ~pos:pos ~stage:Errors.DesugarMatching ~message:"Can't match over nullary function"
+  in
+  match tuple with
+    | [] -> raise (nullary_error pos)
+    | _ -> ()
+
```
dhil, line 53:
```suggestion
let nullary_guard pss pos =
  let nullary_error pos =
    Errors.desugaring_error ~pos:pos ~stage:Errors.DesugarMatching ~message:"Can't match over nullary function"
  in
  match pss with
    | [[]] -> raise (nullary_error pos)
    | _ -> ()

```

dhil:
@Emanon42 while I remember it, for consistency we should probably change all occurrences of `MatchFunlit` to `SwitchFunlit` and the `match_` prefixed type aliases to have `switch_` as a prefix instead.

core/desugarMatching.ml
```diff
@@ -43,12 +43,12 @@ let pattern_matching_sugar_guard pos =
...
-let nullary_guard tuple pos =
+let nullary_guard pss pos =
   let nullary_error pos =
     Errors.desugaring_error ~pos:pos ~stage:Errors.DesugarMatching ~message:"Can't match over nullary function"
   in
-  match tuple with
-    | [] -> raise (nullary_error pos)
+  match pss with
+    | [[]] -> raise (nullary_error pos)
     | _ -> ()
```
dhil, line 52: We also need to check whether the function is curried, as we decided not to support this just yet. The following should be rejected
```
links> fun f(x)(y) switch { case x -> x };
f = fun : (a) -> (c) ~> (a, c)
links> f(2)(4);
(2, 4) : (Int, Int)
```

dhil (changes requested):

core/parser.mly
```diff
@@ -169,7 +169,10 @@ let parseRegexFlags f =
               | 'g' -> RegexGlobal
               | _ -> assert false) (asList f 0 [])
 
-
+let switch_fun_currying_guard p args =
+  match args with
+  | [arg] -> arg
+  | _ -> raise (Errors.Type_error (pos p, "Not support pattern matching function with multiple parameter lists yet"))
```
dhil, line 175: Small typo: `Not` => `No`. I suggest changing the text slightly.
```suggestion
  | _ -> raise (Errors.Type_error (pos p, "Curried switch functions are not yet supported."))
```

core/parser.mly
```diff
@@ -169,7 +169,10 @@ let parseRegexFlags f =
               | 'g' -> RegexGlobal
               | _ -> assert false) (asList f 0 [])
 
-
+let switch_fun_currying_guard p args =
```
dhil, line 172: It is better to define this in the desugar pass module such that you only need to call it once.

core/parser.mly
```diff
@@ -443,6 +448,12 @@ tlfunbinding:
 | OP OPERATOR pattern perhaps_location block                   { ((dl_unl, false), $2, [[$3]], $4, $5)          }
 | OP pattern OPERATOR perhaps_location block                   { ((dl_unl, false), $3, [[$2]], $4, $5)          }
 
+switch_tlfunbinding:
+| fun_kind VARIABLE arg_lists perhaps_location switch_body     { ($1, $2, (switch_fun_currying_guard $loc($3) $3), $4, $5)   }
```
dhil, line 452: Note an alternative to the dynamic check `switch_fun_currying_guard`, would be to restrict the grammar such that only uncurried function definitions are admissible -- amounts to a static enforcement of `switch_fun_currying_guard`.

I am not advocating you should restrict the grammar as we ultimately want to support curried function definitions, I just wanted to make sure you were aware of this alternative design decision.

core/desugarMatching.ml
```diff
@@ -0,0 +1,95 @@
+open Sugartypes
+open Utility
+open SourceCode
+
+(* This module desugars pattern-matching functions
```
dhil, line 5: I suggest renaming this file to `desugarSwitchFuns.ml` for consistency.

core/frontend.ml
```diff
@@ -125,6 +125,7 @@ module Untyped = struct
   let transformers : transformer array
     = [| (module ResolvePositions)
        ; (module CheckXmlQuasiquotes)
+       ; (module DesugarMatching)
```
dhil, line 128: ... when the file name is changed this should read `DesugarSwitchFuns`.

core/parser.mly
Emanon42: I have tried changing grammar before - menhir gave me errors of `one state has shift/reduce conflicts.` I think its because the new grammar is not LL(1). So I just use dynamic checking.

dhil: Menhir is a LR(1) parser generator. Note that `LL != LR`.

There are typically ways in which one can "massage" the grammar to eliminate shift/reduce conflicts. It ultimately depends on how you factor the production rules.

core/parser.mly
Emanon42: Since I have changed the type of `switch_funlit` to `Pattern.with_pos list * switch_funlit_body`, I think it is a syntax error and should be checked in parsing phase?

dhil: Conceptually it is a syntax error, and thus I understand why you would wire this check into the parser. Syntax errors do not exclusively arise during parsing as sometimes you can choose to relax the implemented language grammar a bit in favour of, say, a simpler parser. To catch the false-positives produced by the parser you can subsequently traverse syntax tree to check that it has the intended form.

The duplication of function definition production rules in the parser means you have multiple call sites for `switch_fun_currying_guard`. You can get away with having only one call site if invoke along with the other guards in the desugaring pass. This design is leaner,  and will make the code base more robust for future changes, say, if we need to duplicate rules in the parser again.

Emanon42: I see. Consider we may support currying for it in future. I think its better to change the type of `switch_funlit` back to `Pattern.with_pos list list * switch_funlit_body` and put the guard in `desugarSwitchFuns` pass to raise an error if the user gives it multiple parameter list. (If that's a bad design please correct me)

dhil: I think it is fine to slightly generalise the shape `switch_funlit`.

jamescheney (approved):
This looks good now.  My main suggestion besides whitespace nitpicking is to make `get_normal_funlit` into a library function and use it in more places to avoid repetition (and save effort if we ever later refactor the funlit cases).  Otherwise looks ready to merge.

Emanon42:
> This looks good now. My main suggestion besides whitespace nitpicking is to make `get_normal_funlit` into a library function and use it in more places to avoid repetition (and save effort if we ever later refactor the funlit cases). Otherwise looks ready to merge.

Looks like put `get_normal_funlit` in `utility.ml` could have the problem of the dependency cycle, so I just put it in `sugartypes`.

core/desugarInners.ml
```diff
@@ -172,7 +172,12 @@ object (o : 'self_type)
                let o = o#with_visiting (StringSet.add (Binder.to_name rec_binder) visiting_funs) in
                let (o, tyvars) = o#quantifiers tyvars in
                let (o, inner) = o#datatype inner in
-               let inner_effects = TransformSugar.fun_effects inner (fst lam) in
+               let lam_in =
```
SimonJF, line 175: Take this as a comment rather than an explicit request for changes, but I notice this pattern a lot: knowing that something must be a `NormalFunLit`, and then matching, and asserting false in the impossible case. You may want to abstract over this by moving this logic to `Sugartypes` by having a function there:

```
let get_normal_funlit = function
  | NormalFunLit x -> x
  | _ -> internal_error "SwitchFunLit where NormalFunLit expected"
```

and then you could just call it as

`let lam_in = get_normal_funlit lam in ...`

and save yourself the unwrapping effort at the callsites.

core/desugarSwitchFuns.ml
```diff
@@ -0,0 +1,101 @@
...
+
+  This transformation convert function like that:
+
+  fun foo(a1, ..., an) switch {
+    case (p1_1, ..., p1_n) -> b_1
+    ...
+    case (pm_1, pm_n) -> b_m
+  }
+
+  to function with switch body like that:
```
SimonJF, line 15: Good comment; minor grammar tweaks:

```suggestion
(* This module desugars pattern-matching functions.

  This transformation converts `switch` functions of the form:

  fun foo(a1, ..., an) switch {
    case (p1_1, ..., p1_n) -> b_1
    ...
    case (pm_1, pm_n) -> b_m
  }

  to standard functions of the form:
```

core/parser.mly
```diff
@@ -443,6 +445,12 @@ tlfunbinding:
 | OP OPERATOR pattern perhaps_location block                   { ((dl_unl, false), $2, [[$3]], $4, $5)          }
 | OP pattern OPERATOR perhaps_location block                   { ((dl_unl, false), $3, [[$2]], $4, $5)          }
 
+switch_tlfunbinding:
+| fun_kind VARIABLE arg_lists perhaps_location switch_body     { ($1, $2, $3, $4, $5)   }
+
+switch_body:
```
SimonJF, line 451: Could you rename this to `switch_funlit_body`? Otherwise, it may be confused with the body of an actual switch statement by someone who is unfamiliar with the codebase.

core/sugarTraversals.ml
```diff
@@ -559,9 +559,14 @@ class map =
...
-        let _x = o#list (fun o -> o#list (fun o -> o#pattern)) _x in
-        let _x_i1 = o#phrase _x_i1 in (_x, _x_i1)
+      fun f ->
+        match f with
+          | NormalFunlit (_x, _x_i1) ->
+            let _x = o#list (fun o -> o#list (fun o -> o#pattern)) _x in
+            let _x_i1 = o#phrase _x_i1 in NormalFunlit (_x, _x_i1)
+          | SwitchFunlit (pat, body) ->
+            let pat = o#list (fun o -> o#list (fun o -> o#pattern)) pat in
+            let body = o#list (fun o (p, c) -> let p = o#pattern p in let c = o#phrase c in (p, c)) body in SwitchFunlit (pat, body)
```
SimonJF, line 569: This is currently a bit hard to read without the proper line breaks:

```suggestion
          | SwitchFunlit (pat, body) ->
            let pat = o#list (fun o -> o#list (fun o -> o#pattern)) pat in
            let body =
              o#list (fun o (p, c) ->
                let p = o#pattern p in
                let c = o#phrase c in (p, c)) body in
            SwitchFunlit (pat, body)
```

core/sugarTraversals.ml
```diff
@@ -1287,9 +1292,15 @@ class fold =
...
-        let o = o#list (fun o -> o#list (fun o -> o#pattern)) _x in
-        let o = o#phrase _x_i1 in o
+      fun f ->
+        match f with
+          | NormalFunlit (_x, _x_i1) ->
+            let o = o#list (fun o -> o#list (fun o -> o#pattern)) _x in
+            let o = o#phrase _x_i1 in o
+          | SwitchFunlit (pat, body) ->
+            let o = o#list (fun o -> o#list (fun o -> o#pattern)) pat in
+            let o = o#list (fun o (p, c) -> let o = o#pattern p in let o = o#phrase c in o) body in o
```
SimonJF, line 1302: Same here:

```suggestion
          | SwitchFunlit (pat, body) ->
            let o = o#list (fun o -> o#list (fun o -> o#pattern)) pat in
            let o = o#list (fun o (p, c) ->
              let o = o#pattern p in
              let o = o#phrase c in o) body in
            o
```

core/sugarTraversals.ml
```diff
@@ -2115,9 +2126,14 @@ class fold_map =
...
-        let (o, _x) = o#list (fun o -> o#list (fun o -> o#pattern)) _x in
-        let (o, _x_i1) = o#phrase _x_i1 in (o, (_x, _x_i1))
+      fun f ->
+        match f with
+          | NormalFunlit (_x, _x_i1) ->
+            let (o, _x) = o#list (fun o -> o#list (fun o -> o#pattern)) _x in
+            let (o, _x_i1) = o#phrase _x_i1 in (o, NormalFunlit (_x, _x_i1))
+          | SwitchFunlit (pat, body) ->
+            let (o, pat) = o#list (fun o -> o#list (fun o -> o#pattern)) pat in
+            let (o, body) = o#list (fun o (p, c) -> let (o, p) = o#pattern p in let (o, c) = o#phrase c in (o, (p, c))) body in (o, SwitchFunlit (pat, body))
```
SimonJF, line 2136: As above:

```suggestion
          | SwitchFunlit (pat, body) ->
            let (o, pat) = o#list (fun o -> o#list (fun o -> o#pattern)) pat in
            let (o, body) =
              o#list (fun o (p, c) ->
                let (o, p) = o#pattern p in
                let (o, c) = o#phrase c in
                (o, (p, c))) body in
            (o, SwitchFunlit (pat, body))
```

core/sugartypes.ml
```diff
@@ -566,6 +569,11 @@ let tappl' : phrase * tyarg list -> phrasenode = fun (e, tys) ->
        in
        TAppl (e, List.map make_arg tys)
 
+let get_normal_funlit fnlit =
```
SimonJF, line 572: Ah -- I see you have done this already! Why not replace all the explicit unwrapping elsewhere with calls to this?

core/transformSugar.ml
```diff
@@ -724,13 +724,22 @@ class transform (env : Types.typing_environment) =
...
+            let (o, pss) = listu o (fun o -> listu o (fun o -> o#pattern)) pss in
+            let o = o#with_effects inner_eff in
+            let (o, e, t) = o#phrase e in
+            let o = o#restore_envs envs in
+            (o, NormalFunlit (pss, e), t)
+          | SwitchFunlit (pss, body) ->
+            let envs = o#backup_envs in
+            let (o, pss) = listu o (fun o -> listu o (fun o -> o#pattern)) pss in
+            let o = o#with_effects inner_eff in
+            let (o, body) = listu o (fun o (p, c) -> let (o, p) = o#pattern p in let (o, c, _) = o#phrase c in (o, (p, c))) body in
```
SimonJF, line 740:
```suggestion
            let (o, body) =
              listu o (fun o (p, c) ->
                let (o, p) = o#pattern p in
                let (o, c, _) = o#phrase c in
                (o, (p, c))) body in
```

core/typeSugar.ml
```diff
@@ -4356,12 +4357,17 @@ and type_binding : context -> binding -> binding * context * Usage.t =
...
               (fun (inner_rec_vars, inner_env, patss)
                    {node= { rec_binder = bndr; rec_linearity = lin;
-                            rec_definition = ((_, def), (pats, _));
+                            rec_definition = ((_, def), fnlit);
                             rec_signature = t_ann';
                             rec_unsafe_signature = unsafe;
                             rec_frozen = frozen;
                             _ }; _ } ->
                  let name = Binder.to_name bndr in
+                 let pats =
```
SimonJF, line 4366: Replace with call to `get_normal_funlit`

core/typeSugar.ml
```diff
@@ -4428,8 +4434,13 @@ and type_binding : context -> binding -> binding * context * Usage.t =
                 (List.fold_left2
                    (fun defs_and_uses
                         {node={ rec_binder = bndr; rec_linearity = lin;
-                                rec_definition = (_, (_, body)); _ } as fn; pos }
+                                rec_definition = (_, fnlit); _ } as fn; pos }
                         pats ->
+                      let body =
```
SimonJF, line 4439: Same here -- to get the body and patterns, you can just do something like:

```
let (_, body) = get_normal_funlit fnlit
```

SimonJF (approved):
Thanks for doing all these revisions, @Emanon42! It was good working with you during your internship.

I've got a few final comments, but all of the major things have been addressed now. I'd be happy for this to go in.

core/sugarTraversals.ml
Emanon42: Done. However, there are comments at the beginning of `sugarTraversals` saying that it should be automatically generated by Camlp4 (but I had to handly update it). Is readability a problem here?

core/sugartypes.ml
Emanon42: Done. This one is newly added, there were some historical boilerplate code I forgot to check :p

core/sugarTraversals.ml
jamescheney: I think those comments are (a) very old and (b) no longer accurate - in particular camlp4 is no longer used at all by Links.  We discussed trying to reinstate the auto-generation of visitors/traversals a while ago (#255 ) but decided not to; we might revisit that now that Links has mostly been rewritten to use standard datatypes instead of polymorphic variants.

It would probably be reasonable to update the comment to state that this file is no longer autogenerated.

jamescheney:
Hmm, the CI is failing due to the documentation not building for some reason - maybe a transient problem?  (this only happens on Travis so it might be a Travis problem).

@dhil, any further remarks before this gets merged?  It seems that your most recent review comments have been addressed.

dhil:
It seems that anonymous switch functions are not being desugared properly:
```links
links> fun(s) switch { case x -> x };
***: Error: "Assert_failure core/desugarModules.ml:322:15"
```

core/desugarSwitchFuns.ml
```diff
@@ -0,0 +1,115 @@
...
+  let exhaustive_position = Format.sprintf "non-exhaustive pattern matching at %s" (SourceCode.Position.show pos) in
+  let exhaustive_case = FnAppl (with_pos (Var "error"), [with_pos (Constant (CommonTypes.Constant.String exhaustive_position))]) in
+  let normal_args =
+    List.map
+      (fun (pat, name) -> with_pos (Pattern.As (with_pos (Binder.make ~name ()), pat)))
+      name_list
+  in
+  let cases = cases@[(exhaustive_patterns, with_pos exhaustive_case)] in
+  let switch_body = Switch (with_pos (TupleLit switch_tuple), cases, None) in
+  let normal_fnlit = NormalFunlit ([normal_args], with_pos switch_body) in
```
dhil, line 77: I think you should reuse the source position of `fun f(...) switch` for `switch_body` to improve error messages. Right now you get error messages like
```
:0: Type error: The type of an input to a switch should match the type of its patterns, but the expression
    `<dummy>'
```
where `<dummy>`, in this instance, is the `switch_body`.

Emanon42: Sorry, forget to check the error message.

dhil: No worries. They have sort of secondary status anyways -- the average quality of the error messages in Links is pretty poor.

Generally, it is a good idea to map source position from sugared nodes to desugared nodes whenever possible. It may mean that several desugared nodes may share the same position, because they were derived from the same sugared node -- which is fine. At least this provides the user with a minimal degree of provenance.

As an aside: One thing we lack internally is the ability to tell whether a given node is synthetic, i.e. desugared.

Emanon42: Fixed. Would you like to give me an example for testing it?

dhil: Try some that are obvious type errors:
```
fun f(xs) switch { case [] -> xs case x :: _ -> x } # body types do not match
fun g(Foo) switch { case 1 -> 0 } # incompatible scrutinee and pattern types.
```

core/desugarSwitchFuns.ml
```diff
@@ -0,0 +1,116 @@
...
+let construct_normal_funlit funlit_pos patterns cases =
+  pattern_matching_sugar_guard funlit_pos;
+  let patterns = switch_fun_currying_guard funlit_pos patterns in
+  nullary_guard patterns funlit_pos;
+  (* bind the arguments with unique var name *)
+  let pat_first_pos = WithPos.pos (List.nth patterns 0) in
+  let name_list = List.map (fun pat -> (pat, Utility.gensym())) patterns in
+  let switch_tuple = List.map (fun (_, name) -> with_pos (Var name)) name_list in
+  (* assemble exhaustive handler *)
+  let exhaustive_patterns = with_pos (Pattern.Any) in
```
dhil, line 68: Note: This pattern/case may be redundant, though, currently we do not check for redundancy or exhaustiveness of patterns.

dhil (approved):
Looks good to me! I will merge this in shortly.

SimonJF:
Congratulations, @Emanon42!

Emanon42:
> Congratulations, @Emanon42!
Thanks again for mentoring me! It is really happy to work with you all. 
and new avatar is cool

PR #913, opened 2020-09-15
Various IR type-checking cleanups

frank-emrich:
This PR 
- removes some comments about the earlier handling scoping of type variables being broken (this was already fixed by @slindley)
- adds test ensuring that the scoping of such type variables is now indeed correct
- adds new features to the IR DSL (closures, records, ...)
- slightly stratifies the types used within the IR DSL
- changes the generation and type-checking of closures such that it coincides with the description in the IR formalization. This only changes that the type of the binder of the closure variable will now carry the quantifiers specific to the closure.
  Overall, this doesn't fix any existing bugs, the only difference in behavior is that we can now detect the following, which wasn't possible earlier: When supplying a closure environment to a function as well as type arguments specific to the closure, we can now detect better if the number of these type arguments is wrong. This is because we can now distinguish better between quantifiers of a function that are specific to the closure (i.e., added during closure conversion), and those that were originally part of the function (i.e., before closure conversion). 
- adds a series of tests for the type-checking of closures

### DIFF
--- a/core/closures.ml
+++ b/core/closures.ml
@@ -710,7 +710,8 @@ struct
             let tyvars = inner_quantifiers @ fn_tyvars in
             let (o, z) = match z with
               | Some zbinder ->
-                let zbinder = Var.update_type (Instantiate.datatype inner_maps (Var.type_of_binder zbinder)) zbinder in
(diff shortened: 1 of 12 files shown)


frank-emrich:
This should be pretty un-controversial as it touches mostly code in the IR type-checker and the IR DSL. If nobody objects I would like to merge this tomorrow.

PR #915, opened 2020-09-16
Change Ir.fun_def to be a record instead of a tuple 

frank-emrich:
This changes the type `Ir.fun_def` from being a 7-tuple to a record. A tedious, but completely mechanical change. As a pleasant side-effect, this makes the output of `show_compiled_ir` much more readable!


core/closures.ml
```diff
@@ -363,8 +376,8 @@ struct
                treat them as globals *)
           let o, _ =
             List.fold_right
-              (fun (f, _, _, _, _) (o, fs) ->
-                 let o, f = o#binder f in
+              (fun fundef (o, fs) ->
```
dhil, line 379: Maybe a bit late now, but for future for these sort of projections I think it is nicer to perform them by pattern matching; in this instance `{ binder; _ }`.

core/ir.mli
```diff
@@ -59,7 +59,16 @@ and tail_computation =
   | Special    of special
   | Case       of value * (binder * computation) name_map * (binder * computation) option
   | If         of value * computation * computation
-and fun_def = binder * (tyvar list * binder list * computation) * binder option * location * bool
+and fun_def =
+  {
+    binder   : binder;
```
dhil, line 64: Ideally each of these field names would be prefixed, e.g. `fn_binder : binder`.

core/closures.ml
frank-emrich: I've done that in most other cases

core/ir.mli
frank-emrich: Unfortunately, these prefixes make record field punning less convenient: For example, we always use `tyvars` as the corresponding variable name, which would require `{fn_tyvar = tyvar; ... }` every single time.

frank-emrich: Do you (or others) feel strongly enough about this so I should change this or not?

jamescheney: I think the reason for such prefixing is that in OCaml there's no way to disambiguate between field names defined in different modules that happen to both be in the same scope: whoever was declared last wins.  Hence, prefixing is a way of defensive programming around this limitation.  (I had a vague recollection that something may have been done about this in modern OCamls but can't remember how/what.)  In any case, I think it's probably better to stick with this convention uniformly even though it makes code a little more painful.

frank-emrich: I've update the PR to prefix all the fields with `fn_`

frank-emrich:
I assume that people are okay with this and will merge it tomorrow if nobody objects.

Issue #807, opened 2020-02-06
Todo list (server) is broken on master


dhil:
The `Todo list (server)` example from the example page is broken on the master branch. It produces the following error upon attempting to complete an item

```
Links Error

***: Error: "Assert_failure core/query/query.ml:1102:6" 
```

Steps to reproduce:
1. Enter an item, e.g. "test"
2. Press "Add item"
3. Press "Done"

dhil:
The assertion failure is triggered by the default case in the `base` function in `query/query.ml`
```
and base : Sql.index -> Q.t -> Sql.base = fun index ->
  let open Q in
  function
    ...
    | e ->
      Debug.print ("Not a base expression: " ^ Q.show e);
      assert false
```

wricciot:
No big deal: the delete query wasn't normalized. I'm surprised this went overlooked for such a long time.

jamescheney:
I just ran across this bug again since it is present in v0.9.1 and breaks the todo_db tutorial example.  It looks like this was introduced inadvertently by #624 which changed `Eval.computation` so that it just translates IR to query code, but did not normalize.  It would be good to get an update out that includes this fix.  

I guess this is a further argument for checking the web examples (and tutorials) before creating a new version (i.e. headless web testing #851).  Though in this case the problem could have been caught by a database test; right now there aren't any database tests that exercise this code path.

PR #918, opened 2020-09-20
Changelog 0.9.2

frank-emrich:
This is a draft for the changelog for the 0.9.2 release.

I vaguely remember that this release fixes some problem in 0.9.1 where many (all?) web or database examples were broken, but I may be wrong. Does this ring a bell? I went through all commits sine 0.9.1 to create the changelog, but I couldn't find something sounding like that.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,7 @@
 # 0.9.2
 
-This minor release contains various bug fixes, improvements, and a **breaking** change.
+This minor release contains various bug fixes, improvements, and a **breaking**
+change.
 
 ## Breaking change: Trailing semicolons are no longer permitted
 The surface syntax of Links has been changed.  Up until now it was possible to
@@ -41,13 +42,13 @@ fun foo(x) {
 A third option is to simply drop the trailing semicolon, though, this only works
 as intended if the type of the last expression is `()`.
 
-## TODO: Bug breaking web examples
 
 ## SML-style function definitions
 
-Links now supports "switch functions", a new syntax for defining functions in terms of match clauses
-directly, similar to SML. This allows writing the following function
-```
+Links now supports "switch functions", a new syntax for defining functions in
+terms of match clauses directly, similar to SML. This allows writing the
+following function
+```links
 fun ack(_,_) switch {
   case (0, n) -> n + 1
   case (m, 0) -> ack(m - 1, 1)
@@ -56,7 +57,7 @@ fun ack(_,_) switch {
 ```
 instead of the following, more verbose version:
 
-```
+```links
 fun ack(a, b) {
   switch(a, b) {
     case (0, n) -> n + 1
@@ -67,7 +68,7 @@ fun ack(a, b) {
 ```
 
 Switch functions can also be anonymous, allowing function like the following:
-```
+```links
 fun(_, _) switch {
   case (0, n) -> 0
   case (m, n) -> m + n
@@ -80,17 +81,19 @@ fun(_, _) switch {
 The minimum required OCaml version has been raised to 4.08.
 
 
-## Miscellaneous:
+## Miscellaneous
 
+- Fixed a bug breaking the TODO list example (#812)
 - Checkboxes and radio groups in form elements are now handled correctly (#903)
 - Links supports MySQL databases again! (#858)
 - Fixed a bug where the effect of `orderby` was inconsistent between database
   drivers w.r.t. reversing the order of results (#858)
-- Relational lenses can now be used with MySQL and Sqlite3  databases, too (#897)
+- Relational lenses can now be used with MySQL and Sqlite3 databases, too (#897)
 - Remove setting `use_keys_in_shredding`, behaving as if it was always true (#892)
 - Remove setting `query`, behaving as if it was off
   (i.e., `query` behaves like `query flat`) (#892)
-- Fixed a bug where regular expressions in nested queries did not work correctly (#852)
+- Fixed a bug where regular expressions in nested queries did not work correctly
+  (#852)
 - Implemented support for negative patterns in let bindings (#811)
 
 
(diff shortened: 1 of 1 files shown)


CHANGES.md
```diff
@@ -1,3 +1,61 @@
...
+
+
+## Require OCaml 4.08
+
+The minimum required OCaml version has been raised to 4.08.
+
+
+## Miscellaneous:
+
+- The last expression in a block may no longer be followed by a semicolon (#900)
```
jamescheney, line 46: I would be inclined to say more about this, since it is potentially a breaking change - perhaps making its own subsection and showing an example of how code with the (now) forbidden semicolons needs to be adapted.  Of course this is also explained in more detail in #900 .

CHANGES.md
```diff
@@ -1,3 +1,61 @@
...
+- Checkboxes and radio groups in form elements are now handled correctly (#903)
+- Links supports MySQL databases again! (#858)
```
jamescheney, line 48: Perhaps also worth mentioning that several bugs affecting the order of query results in different drivers were fixed along the way, so that things now (hopefully) are correct and consistent across drivers.

jamescheney: But maybe I only think this because it took an embarrassingly long time to notice and fix this.

jamescheney (changes requested):
I believe the bug that made it into 0.9.1 was #807 (fixed by #812).  This doesn't break all examples but is a showstopper for any database example involving delete.

It also illustrates that we should be doing basic sanity checking on a version planned for release before pressing the final version, including not just running all the automated test cases but looking at all of the web examples and testing basic functionality.  (of course, this is not systematic or repeatable.)

CHANGES.md
```diff
@@ -1,3 +1,61 @@
+# 0.9.2
+
+This minor release contains various bug fixes and improvements.
+
+## TODO: Bug breaking web examples
```
dhil, line 5:
```suggestion
## Breaking change: Trailing semicolons are no longer permitted

The surface syntax of Links has been changed. Up until now it was possible to end a block with a semicolon. A trailing semicolon was interpreted as implicitly ending the block with a `()` expression.  The rationale for this change is to make the Links syntax more consistent, i.e. now all blocks must end with an explicit expression. 

To sum up, previously both of the following were allowed

```links
fun foo(x) {
  bar(y);
  baz(x);
}

fun foo'(x) {
  bar(y);
  baz(x);
}
```

Now the first form is no longer accepted. Instead you have to drop the semicolon and either end the block with an explicit `()` or wrap the last expression in an `ignore` application.

```links
fun foo(x) {
  bar(y);
  baz(x); 
  ()
}

fun foo(x) {
  bar(y);
  ignore(baz(x))
}
```

A third option is to simply drop the trailing semicolon, though, this only works as intended if the type of the last expression is `()`.

## TODO: Bug breaking web examples
```

File: CHANGES.md [hunk 4]
```diff
@@ -1,3 +1,61 @@
+# 0.9.2
+
+This minor release contains various bug fixes and improvements.
```
dhil, line 3:
```suggestion
This minor release contains various bug fixes, improvements, and a **breaking** change.
```

CHANGES.md
```diff
@@ -1,3 +1,61 @@
...
+}
+```
+
+
+## Require OCaml 4.08
+
+The minimum required OCaml version has been raised to 4.08.
+
+
+## Miscellaneous:
```
dhil, line 44:
```suggestion
## Miscellaneous
```

dhil (commented):
I have attached some suggestions.

CHANGES.md
dhil: It seems like the Github UI have cropped my suggestion, you may have to "edit" my comment and copy-paste the source code if you want to adopt it.

frank-emrich: I assume that in `foo'` (after "To sum up, previously both of the following were allowed"), you meant to omit the second `;`?

dhil: Absolutely!

frank-emrich:
I've incorporated all the suggestions

CHANGES.md
```diff
@@ -1,3 +1,104 @@
...
+```
+
+Switch functions can also be anonymous, allowing function like the following:
+```links
+fun(_, _) switch {
+  case (0, n) -> 0
+  case (m, n) -> m + n
+}
+```
+
```
dhil, line 77:
```suggestion
Note: currently switch function syntax is only supported for uncurried functions.
```

CHANGES.md
```diff
@@ -1,3 +1,104 @@
...
+
```
dhil, line 78:
```suggestion
As switch functions have experimental status they are disabled by default. To enable them you must set the option `pattern_matching_sugar=true` in a configuration file. 
```

dhil: I wonder whether we can make a swift change. Rather than naming the option "pattern_mathcing_sugar" it ought to be named "switch_funs" or something along those lines. 

A further thought would be to adopt an extension scheme similar to that of GHC, where an extension is toggled by passing `-Xextension_name` on the commandline. In fact, I deliberately engineered the settings framework to support this kind of interface. But let's leave this for a later discussion.

jamescheney (approved):
Looks reasonable to me now, pending @dhil's suggestions

CHANGES.md
frank-emrich: Sure. Do you want to make the change yourself?

dhil: I've opened a PR to rename the option "pattern_matching_sugar" to "switch_functions", c.f. #919

frank-emrich:
@jamescheney 
> It also illustrates that we should be doing basic sanity checking on a version planned for release before pressing the final version, including not just running all the automated test cases but looking at all of the web examples and testing basic functionality. (of course, this is not systematic or repeatable.)

The  [pre-release checklist ](https://github.com/links-lang/links/wiki/Pre-release-sanity-checking) already suggests doing this. However, it is indeed a tedious process that involves setting up the database as well as creating and populating the tables needed for the examples. I'm not surprised that we frequently break the examples in master.

dhil:
I think this is ready to be merged now. I don't think you need to waste compute resources on Travis, so I suggest you use your executive admin powers to force merge this thing.

PR #919, opened 2020-09-21
Rename option to toggle switch functions.

dhil:
This commit renames the option `pattern_matching_sugar` to
`switch_functions` as the former is somewhat generic and the latter is
the name of the feature actually being enabled/disabled by toggling
the option.

In addition, this commit also cleans up the test suite for
functions. The non-experimental function features should not be kept
in the same test file as the experimental features -- in particular,
the non-experimental features should not **always** be tested under
the assumption that an experimental flag is enabled.

### DIFF
--- a/core/desugarSwitchFuns.ml
+++ b/core/desugarSwitchFuns.ml
@@ -28,19 +28,18 @@ open SourceCode
 
 let with_pos = SourceCode.WithPos.make
 
-let pattern_matching_sugar =
+let enabled =
   Settings.(
-    flag "pattern_matching_sugar"
-    |> synopsis
-         "Toggles whether to enable the switch pattern matching syntax sugar"
+    flag ~default:false "switch_functions"
+    |> synopsis "Toggles whether to enable the switch function syntax"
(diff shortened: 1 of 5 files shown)


core/desugarSwitchFuns.ml
```diff
@@ -28,19 +28,18 @@ open SourceCode
 
 let with_pos = SourceCode.WithPos.make
 
-let pattern_matching_sugar =
+let enabled =
```
frank-emrich, line 31: We usually have the convention that the OCaml variable is named the same as the setting itself.

dhil: I don't have strong feelings about this. I think this convention is accidental, and I may be sole responsible for it, when I centralised most of the settings years ago as a quick fix to make all the side effects observable. With centralised settings, the names are global, and thus it makes sense to name them according to their configuration key name. With decentralised settings, some names are local, and thus we can choose names that better convey the nature of settings objects in their local context.

frank-emrich (approved):
LGTM

PR #923, opened 2020-10-04
Testing the IR type-checker's handling of effects

frank-emrich:
This extends the IR DSL with constructs for dealing with effects. It also adds some test cases for handling the effects of functions, their invocations, and bindings. Luckily, this didn't uncover any bugs.

Note that this doesn't test effect handlers (i.e., `do` and `handle`) themselves.


frank-emrich:
Merging as this is un-controversial and only affects IR type-checking

PR #925, opened 2020-10-08
Testing the IR type-checker's handling of variants and case constructs

frank-emrich:
This mostly contains tests for the IR type-checker for variants and case constructs.

In the process, I discovered the following bugs:
- In `TypeUtils.split_row`, removing a field from an open row would not make it `Absent`, but just remove it altogether. Hence, the resulting row may become ill-kinded (if we had a mechanism for checking that, which we don't)
- The IR type-checker didn't allow matching over a variant type with presence polymorphism in any of its fields. However, the frontend does allow this, if the presence polymorphism only affects labels handled by the default case. I've added test cases for both type-checkers.
- The IR type-checker didn't check that the binder for the default case (if it exists) has the right type, which can be deduced from the type of the expression matched over and the variants present in the match clauses.

I also realized that all the front-end tests for patterns pass the IR type-checker now, which is why I've enabled IR type-checking for these tests.


jamescheney:
As discussed in Monday's meeting, this is still planned to be merged once updating work is done.

frank-emrich:
This is currently blocked by #955, will merge after that is fixed.

tests/unit/unitTests.ml
```diff
@@ -27,6 +27,7 @@ let suites =
            Effects.suite;
            Quantifiers.suite;
            Closures.suite;
+           Variants.suite;
```
dhil, line 30: Can we split this unit test definition into two, such that the database and non-database suites are in separate files?

dhil (changes requested):

tests/unit/unitTests.ml
frank-emrich: Not sure if I understand what you mean. None of these tests should interact with a database?

dhil: So the lens tests do not use the database?

.github/workflows/default.yml
```diff
@@ -106,6 +106,7 @@ jobs:
         run: |
          eval $(opam env)
          make tests
+	 ./run-tests nodb-unit-tests
```
dhil: I think YAML insists that you line up L107-109 on the same column.

dhil: You also need to install `ounit2` in the job, e.g. on line https://github.com/links-lang/links/blob/3ba9c18bd77dc6ec7aa1d655d5ef404308c07948/.github/workflows/default.yml#L91

tests/unit/unitTests.ml
frank-emrich: as discussed elsewhere: you're right, and I split the unit tests into those using the database and those that don't

dhil (approved):

PR #934, opened 2020-11-03
Distribute MVU JavaScript source files.

### LINKED ISSUE #930: JavaScript files missing in opam installation
The files `vdom.js` and `virtual-dom.js` are not part of the opam installation. There is also a more general problem about paths to these files that will submitted as a separate issue (#931), as it may be independent of the opam install.

### LINKED ISSUE #931: Hardcoded paths to JavaScript files
The paths to `vdom.js` and `virtual-dom.js` appear to be hardcoded in a way which is incompatible with the location of the JavaScript files in the opam install. The opam install uses the following config options
```
jsliburl=/lib/js
jslibdir=/Users/vashti/.opam/default/lib/links/js
```
but if the two js files above (which are missing in the opam install --see issue #930) are copied into the jslibdir then they are not found when running links.

A solution was to create a local lib directory containing those two files plus jslib.js and changing the config to
```
jsliburl=/lib
jslibdir=lib 
```
but moves the js files out of the opam lib structure.

(BTW, this problem was hard to debug because links tries to do something reasonable with some other files and one gets a half functional webpage.)

dhil:
This commit modifies the install script such that the MVU dependencies
`vdom.js` and `virtual-dom.js` are installed along with the MVU
examples.

To make the MVU examples work out of the box I have chosen to piggyback
on hardcoded resolution for `jslib.js`. This is a horrible hack. I
also had to change the default configuration file to set the stdlib
path in order for the system to find the MVU Links source files.

I was able to run the MVU todo example application on a fresh
installation of Links with the following command.

```
$ linx $OPAM_SWITCH_PREFIX/share/links/examples/mvu/todomvc/todoMVC.links
```

Resolves #930. Resolves #931.

### DIFF
--- a/lib/js/dune
+++ b/lib/js/dune
@@ -1,4 +1,6 @@
 (install
   (section lib)
-  (files (jslib.js as js/jslib.js))
+  (files (jslib.js as js/jslib.js)
+         (vdom.js as js/vdom.js)
+         (virtual-dom.js as js/virtual-dom.js))
   (package links))
--- a/lib/stdlib/mvu.links
+++ b/lib/stdlib/mvu.links
@@ -7,14 +7,14 @@ open import MvuCommands;
 
 # Needed to ensure that virtual-dom is open
 module VirtualDom {
-  alien javascript "/lib/virtual-dom.js" {
+  alien javascript "/lib/js/virtual-dom.js" {
     dummy : ();
   }
 }
 
 # Loads the "vdom.js" file, exposes runDom and updateDom.
 module VDom {
-  alien javascript "/lib/vdom.js" {
+  alien javascript "/lib/js/vdom.js" {
     runDom : forall a :: Type(Any, Any), e :: Row .
       (String, HTML(a), AP(?a.End), Sub(a)) ~e~> ();
     updateDom : forall a :: Type(Any, Any), e :: Row . (HTML(a), Sub(a)) ~e~> ();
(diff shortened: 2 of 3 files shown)


jamescheney:
This looks reasonable to me, I'm curious how this compares to how @SimonJF normally runs things though.

vcgalpin (commented):
Looks OK to me, but I'm also interested in how this compares with how @SimonJF achieves this.

jamescheney (approved):
Seems reasonable to me.  Can we document (e.g. in an issue) that a better solution is desired here, perhaps including an explanation of the current design and desiderata for a new one?

jamescheney:
Added an issue to signpost that further cleanup is desirable.  Hoping @SimonJF can take a quick look before merging.

SimonJF:
Sorry, the "review" nudge just alerted me -- I missed this! All looks very sensible to me, happy for it to go in.

@vcgalpin: I tend to use a git checked-out version (which contains the `js` files) and an explicit `jslibdir`, so I think that's probably why I've not run into this yet. Good spot; I guess the issue is that this isn't picked up by the examples CI since it's a dynamic error.

SimonJF (approved):

dhil:
>  Can we document (e.g. in an issue) that a better solution is desired here, perhaps including an explanation of the current design and desiderata for a new one?

I can give it a stab later this week when I need a break from dissertation writing.

jamescheney:
> I can give it a stab later this week when I need a break from dissertation writing.

Thanks, please add to #938.

PR #935, opened 2020-11-10
Allow table projection from within likeifyable SQL-compiled regular expressions

SimonJF:
# Issue

(Related in part to #754).

Previously, compiled regular expressions failed at runtime. As an example from GtoPdb:

```
for (strI <-- Tables.structural_info)
  where (strI.object_id == objectId && strI.species_id == 2)
  for (al <-- Tables.allele)
    where (strI.official_gene_id <> "" && al.accessions =~ /.*{strI.official_gene_id}.*/)
```

would fail due to the clause

```
al.accessions =~ /.*{strI.official_gene_id}.*/
```

within the `where` clause, since we're including a regular expression which projects from the `strI` table.

# Scope

This patch rectifies the above problem for `LIKE`-translatable queries (but not yet for arbitrary `RLIKE` queries).

# Proposal

To fix this, we need to do two things: generalise the return type of `likeify` to include SQL which allows strings, concatenation, and projection operators; and ensure the nested query translations occur inside regular expressions.

### DIFF
--- a/core/query/evalNestedQuery.ml
+++ b/core/query/evalNestedQuery.ml
@@ -397,7 +397,7 @@ struct
       | Apply (Primitive f, es) ->
         Apply (Primitive f, List.map li es)
       | Record fields ->
-        Record (StringMap.map (li) fields)
+        Record (StringMap.map li fields)
       | Primitive "out" ->
         (* z.2 *)
         Project (Var (z, z_fields), "2")
--- a/core/query/evalNestedQuery.ml
+++ b/core/query/evalNestedQuery.ml
@@ -375,7 +375,10 @@ struct
 
   let rec lins_inner (z, z_fields) ys : QL.t -> QL.t =
     let open QL in
+    let li () = lins_inner (z, z_fields) ys in
+    let liq = lins_inner_query (z, z_fields) ys in
     function
+      (* Need to make sure this happens on the regex, too. *)
       | Project (Var (x, fields), l) ->
         begin
           match position_of x ys with
@@ -386,20 +389,34 @@ struct
                 (Project
                     (Project (Var (z, z_fields), "1"), string_of_int i), l)
         end
-      | Apply (Primitive "Empty", [e]) -> Apply (Primitive "Empty", [lins_inner_query (z, z_fields) ys e])
-      | Apply (Primitive "length", [e]) -> Apply (Primitive "length", [lins_inner_query (z, z_fields) ys e])
+      | Apply (Primitive "Empty", [e]) -> Apply (Primitive "Empty", [liq e])
+      | Apply (Primitive "length", [e]) -> Apply (Primitive "length", [liq e])
       | Apply (Primitive "tilde", [s; r]) as e ->
           Debug.print ("Applying lins_inner to tilde expression: " ^ QL.show e);
-          Apply (Primitive "tilde", [lins_inner (z, z_fields) ys s; r])
+          Apply (Primitive "tilde", [li () s; li () r])
       | Apply (Primitive f, es) ->
-        Apply (Primitive f, List.map (lins_inner (z, z_fields) ys) es)
+        Apply (Primitive f, List.map (li ()) es)
       | Record fields ->
-        Record (StringMap.map (lins_inner (z, z_fields) ys) fields)
+        Record (StringMap.map (li ()) fields)
       | Primitive "out" ->
         (* z.2 *)
         Project (Var (z, z_fields), "2")
       | Primitive "in"  -> Primitive "index"
       | Constant c      -> Constant c
+      (* Regex variants. *)
+      | Variant ("Simply", x) ->
+          Variant ("Simply", li () x)
+      | Variant ("Seq", Singleton r) ->
+          Variant ("Seq", Singleton (li () r))
+      | Variant ("Seq", Concat rs) ->
+          Variant ("Seq",
+            Concat (List.map (
+              function | Singleton x -> Singleton (li () x) | _ -> assert false) rs))
+      | Variant ("Quote", Variant ("Simply", v)) ->
+          Variant ("Quote", Variant ("Simply", li () v))
+      (* Other regex variants which don't need to be traversed *)
+      | Variant (s, x) when s = "Repeat" || s = "StartAnchor" || s = "EndAnchor" ->
+          Variant (s, x)
       | e ->
         Debug.print ("Can't apply lins_inner to: " ^ QL.show e);
         assert false
@@ -499,7 +516,7 @@ struct
       | Apply (Primitive "length", [e]) -> Apply (Primitive "length", [flatten_inner_query e])
       | Apply (Primitive "tilde", [s; r]) as e ->
           Debug.print ("Applying flatten_inner to tilde expression: " ^ QL.show e);
-          Apply (Primitive "tilde", [flatten_inner s; r])
+          Apply (Primitive "tilde", [flatten_inner s; flatten_inner r])
       | Apply (Primitive f, es) -> Apply (Primitive f, List.map flatten_inner es)
       | If (c, t, e)  ->
         If (flatten_inner c, flatten_inner t, flatten_inner e)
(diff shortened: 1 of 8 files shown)


jamescheney:
Huh.  While it's nice to be able to support this, I admit to being a little surprised it is even possible - I guess because SQL LIKE does its regular expression compiling as a run-time "eval"-like operation which is late enough that we can do other SQL string operations first.

I am wondering whether this same limitation also applies to flat queries - did we already support regexp antiquotation referring to query-generator-bound variables in flat queries, or would it be good to do the same thing to the flat query engine so that the same support is available tehre?

SimonJF:
> Huh. While it's nice to be able to support this, I admit to being a little surprised it is even possible - I guess because SQL LIKE does its regular expression compiling as a run-time "eval"-like operation which is late enough that we can do other SQL string operations first.

Yes, it's pretty neat -- the nested query test case gets compiled as:

```
select (2) as "1@1",(t2345."2") as "1@2",(t2341."c") as "2"
from (select (t2340."name") as "1@1@name",(row_number() over (order by t2340."name")) as "2"
from "marketers" as t2340) as t2345,"mc" as t2341
where ((t2341."m") LIKE ('%' || t2345."1@1@name" || '%')) and ((t2341."m") = ('a'))
```

noting `LIKE ('%' || t2345."1@1@name" || '%')`.

> I am wondering whether this same limitation also applies to flat queries - did we already support regexp antiquotation referring to query-generator-bound variables in flat queries, or would it be good to do the same thing to the flat query engine so that the same support is available tehre?

My hunch is that the only thing stopping the flat evaluator from working was the `likeify` return type limitation: the edits to the nested query evaluator were only necessary in order to ensure that we were projecting from the correct table after shredding. That said, I'll add in a test.

SimonJF:
Sorry, CI and PRs are being funky. The first run passed with the nested query. I've just updated with the flat query test (which passes locally) so hopefully that should go through, too.

dhil:
> Sorry, CI and PRs are being funky. The first run passed with the nested query. I've just updated with the flat query test (which passes locally) so hopefully that should go through, too.

I think Travis CI missed you!

SimonJF:
:upside_down_face: 

Seems like the build is queued at least, hopefully if it goes through, it'll trigger the necessary hooks...

https://travis-ci.org/github/links-lang/links/builds/742710312

SimonJF:
Of course MySQL does things differently 🙄

SimonJF:
Putting this on hold until #937 is done (thanks, MySQL)

SimonJF:
Woo! OK, after that adventure, I think this is finally ready for review. 

As a pleasant side-effect, I also fixed string concatenation in MySQL more generally.

core/query/evalNestedQuery.ml
```diff
@@ -375,7 +375,10 @@ struct
 
   let rec lins_inner (z, z_fields) ys : QL.t -> QL.t =
     let open QL in
+    let li () = lins_inner (z, z_fields) ys in
```
jamescheney, line 378: Is there a reason for making li be a function of type `unit -> QL.t -> QL.t` rather than just leaving out the unit?  I guess this avoids nontermination but eta-expanding the RHS by the argument should also do it, and then you'd not have to write `li () xyzzy` everywhere.

```suggestion
    let li e = lins_inner (z, z_fields) ys e in
```

core/query/evalNestedQuery.ml
```diff
@@ -386,20 +389,34 @@ struct
...
                     (Project (Var (z, z_fields), "1"), string_of_int i), l)
         end
-      | Apply (Primitive "Empty", [e]) -> Apply (Primitive "Empty", [lins_inner_query (z, z_fields) ys e])
-      | Apply (Primitive "length", [e]) -> Apply (Primitive "length", [lins_inner_query (z, z_fields) ys e])
+      | Apply (Primitive "Empty", [e]) -> Apply (Primitive "Empty", [liq e])
+      | Apply (Primitive "length", [e]) -> Apply (Primitive "length", [liq e])
       | Apply (Primitive "tilde", [s; r]) as e ->
           Debug.print ("Applying lins_inner to tilde expression: " ^ QL.show e);
-          Apply (Primitive "tilde", [lins_inner (z, z_fields) ys s; r])
+          Apply (Primitive "tilde", [li () s; li () r])
```
jamescheney, line 396:
```suggestion
          Apply (Primitive "tilde", [li s; li r])
```

core/query/evalNestedQuery.ml
```diff
@@ -386,20 +389,34 @@ struct
...
       | Apply (Primitive f, es) ->
-        Apply (Primitive f, List.map (lins_inner (z, z_fields) ys) es)
+        Apply (Primitive f, List.map (li ()) es)
```
jamescheney, line 398:
```suggestion
        Apply (Primitive f, List.map li es)
```

core/query/evalNestedQuery.ml
```diff
@@ -386,20 +389,34 @@ struct
...
       | Record fields ->
-        Record (StringMap.map (lins_inner (z, z_fields) ys) fields)
+        Record (StringMap.map (li ()) fields)
```
jamescheney, line 400:
```suggestion
        Record (StringMap.map li fields)
```

core/query/evalNestedQuery.ml
```diff
@@ -386,20 +389,34 @@ struct
...
       | Primitive "out" ->
         (* z.2 *)
         Project (Var (z, z_fields), "2")
       | Primitive "in"  -> Primitive "index"
       | Constant c      -> Constant c
+      (* Regex variants. *)
+      | Variant ("Simply", x) ->
+          Variant ("Simply", li () x)
```
jamescheney, line 408:
```suggestion
          Variant ("Simply", li x)
```

core/query/evalNestedQuery.ml
```diff
@@ -386,20 +389,34 @@ struct
...
+      | Variant ("Seq", Singleton r) ->
+          Variant ("Seq", Singleton (li () r))
```
jamescheney, line 410:
```suggestion
          Variant ("Seq", Singleton (li r))
```

core/query/evalNestedQuery.ml
```diff
@@ -386,20 +389,34 @@ struct
...
+      | Variant ("Seq", Concat rs) ->
+          Variant ("Seq",
+            Concat (List.map (
+              function | Singleton x -> Singleton (li () x) | _ -> assert false) rs))
```
jamescheney, line 414:
```suggestion
              function | Singleton x -> Singleton (li x) | _ -> assert false) rs))
```

core/query/evalNestedQuery.ml
```diff
@@ -386,20 +389,34 @@ struct
...
+      | Variant ("Quote", Variant ("Simply", v)) ->
+          Variant ("Quote", Variant ("Simply", li () v))
```
jamescheney, line 416:
```suggestion
          Variant ("Quote", Variant ("Simply", li v))
```

core/query/sql.ml
```diff
@@ -26,9 +26,16 @@ and select_clause = (base * string) list * from_clause list * base * base list
 and from_clause =
   | TableRef of table_name * Var.var
   | Subquery of query * Var.var
+and sql_like =
```
jamescheney, line 29: I'm curious what is the benefit of having a separate case for LIKE strings as opposed to just using the string constant, ordinary projection, and string concatenation operators.  Clearly doing it this way makes it easier to recognize where match expressions have been translated to (LIKE) strings since regexp compilation is the only place where we do that.  However, what would break (or be unpleasantly harder) if we just compiled regexps to the existing Constant(String -), Project and Append operations?

jamescheney (changes requested):
Looks reasonable.  I have a few nitpicky questions - I'm not insisting on changes to the code but if not it would be good to document the reasoning with comments in a few places.

SimonJF:
Thanks for the excellent suggestions, @jamescheney! I've removed thunking in favour of eta-expansion (good trick), and removed the `SqlLike` construct in favour of compiling directly to the query language.

jamescheney (commented):
A few tiny nitpicks but nothing that should require discussion or block merging.

SimonJF:
Fab, thanks! This good to go when CI passes?

jamescheney (approved):

jamescheney:
huh, I thought I chose "approve" - should be fixed now.

SimonJF:
Wonderful, thanks!

PR #936, opened 2020-11-10
Heterogeneous queries and delateralization

wricciot:
This PR adds to Links set-based database queries, which can be freely composed with (standard) bag queries. No new syntax is provided (presently) for such queries; however, the new `dedup` function in the prelude, which removes duplicates from a list, is recognized by the database query implementation and used to produce `select distinct/union` statements rather than the usual `select/union all`.

The internal representation of queries is extended with `Prom` and `Dedup` constructors which embed a set query in a bag query and deduplicate a bag query to obtain a set query. The normalization code has been changed to handle these extra constructors,
and some of the changes (particularly those involving `For` comprehensions are rather complicated (perhaps too much). The most radical change is that `norm` now accepts queries containing `For` terms (whereas previously it expected all the input comprehensions to be expressed using `ConcatMap`). Getting this extension to work correctly with an incremental changed proved rather difficult and it's probably a good idea to rewrite the code that performs the normalization of `For` from scratch.

In addition to the `flat` and `nested` query policies for homogeneous queries, two new experimental policies are provided: `mixing` and `delat`. Both new policies know how to handle sets and bags, but they differ in the treatment of "lateral" variables. While normalized homogeneous comprehensions always have base tables as their generators, normalized heterogeneous comprehensions allow general subqueries in that position, which makes it possible for a generator to reference a variable introduced by a sister generator as in (pseudocode)
```
for (x <-- table1, y <- Q(x))
 ...

```
To express such a query in SQL directly, we have to use the SQL:1999 keyword `lateral`, which is what the `mixing` policy does. If a DBMS does not support `lateral` (or is particularly inefficient when handling lateral subqueries),  the `delat` policy employs a multi-step "delateralization" technique to produce an equivalent query without lateral references.

The two new policies employ a simplified SQL code generator (`query/SimpleSqlGen.ml`) due to my belief that the original code could be rewritten in a cleaner way knowing the exact shape of a normalized query. However, it is not as complete as the corresponding code in `query/sql.ml` (a quick look suggests that it doesn't support indices and ranges). At some point, the two SQL generators will have to be merged.


wricciot:
Based on @slindley's suggestion, I have now reinstated the old normaliser into this PR. The mixing normaliser is now implemented in a separate module `MixingQuery`, which is called when evaluating `query mixing { ... }` or `query delat { ... }`. It is also possible to switch to the new normaliser for _all_ flat queries (but not for shredded queries), by adding a `mixing_norm=on` flag to the configuration file.

There is, of course, a lot of duplicated code in `Query` and `MixingQuery`, but it's not very easy to factorize due changes to the `(Mixing)Query.t` data type. However, eventually only one of the two should remain, so that shouldn't be a problem.

By comparing `EvalQuery` and `EvalMixingQuery` you'll see that the latter is rather naive (especially concerning ordering), so that's something I will have to address in the future.

Finally, the mixing normaliser presents a regression wrt issue #619, which I'm investigating (something to do with the light-weight typechecking performed within query normalisation, which changed considerably).

jamescheney:
I'll have a look in Monday.

core/query/delateralize.ml
```diff
@@ -0,0 +1,247 @@
...
+        m
+        None
+
+let list_tryPick f l =
+    List.fold_left
+        (fun acc x -> f x ||= acc)
+        None
+        l
+
+let rec subst t x u =
```
jamescheney, line 58: Should this function be in mixingQuery?

core/query/delateralize.ml
```diff
@@ -0,0 +1,247 @@
+(*****************************************************************************
+ ** Delateralize.ml - Implements the delateralization algorithm             **
+ **                                                                         **
+ ** author: Wilmer Ricciotti                                                **
+ *****************************************************************************)
+open Utility
```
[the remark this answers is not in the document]
jamescheney: for example, explaining a bit of terminology ("graph" etc.)

core/query/delateralize.ml
```diff
@@ -0,0 +1,247 @@
...
+    in
+    let q1_rp = subst q1 y rp
+    in
+    Q.For (None,
+        gs @ [(p, Q.Prom graph)],
+        [],
+        Q.If (eq_query, q1_rp, Q.nil))
+
+(* Returns (Some ty) if v occurs free with type ty, None otherwise *)
+let occurs_free (v : Var.var) =
```
jamescheney, line 133: Also seems like it could go in mixingQuery

core/query/evalMixingQuery.ml
```diff
@@ -0,0 +1,121 @@
+(*****************************************************************************
+ ** simpleSqlGen.ml - Simple Sql syntax generator                           **
```
jamescheney, line 2: This comment seems outdated

core/query/evalMixingQuery.ml
```diff
@@ -0,0 +1,121 @@
...
+        <| List.map (fun (f,x) -> (base_exp x, f)) (StringMap.to_alist fields)
+        <| Sql.Constant (Constant.Bool true)
+    | Q.If (c, Q.Singleton (Q.Record fields), Q.Concat []) ->
+        selquery
+        <| List.map (fun (f,x) -> (base_exp x, f)) (StringMap.to_alist fields)
+        <| base_exp c
+    | _ -> Debug.print ("error in SimpleSqlGen.body: unexpected j = " ^ Q.show j); failwith "body"
+
+and base_exp = function
+(* XXX: Project expects a (numbered) var, but we have a table name
```
jamescheney, line 69: Perhaps document this expectation executably with an assertion or error message

core/query/mixingQuery.ml
```diff
@@ -0,0 +1,1551 @@
...
+    | Apply (Primitive "Empty", [v]) ->
+        Sql.Empty (unit_query v)
+    | Apply (Primitive "length", [v]) ->
+        Sql.Length (unit_query v)
+    | Apply (Primitive f, vs) ->
+        Sql.Apply (f, List.map (base index) vs)
+    | Project (Var (x, _tyx), name) ->
+        Sql.Project (x, name)
+    | Constant c -> Sql.Constant c
+    | Primitive "index" ->
```
jamescheney, line 1423: Perhaps leave this out since we are not currently using "index" in mixing queries?

core/query/evalMixingQuery.ml
```diff
@@ -0,0 +1,121 @@
...
+                S.Constant (Constant.String (Regex.string_of_regex (Linksregex.Regex.ofLinks (Q.value_of_expression r))))
+            in
+                Sql.Apply ("RLIKE", [base_exp s; r])
+    end
+| Q.Apply (Q.Primitive "Empty", [v]) -> S.Empty (sql_of_query false v)
+| Q.Apply (Q.Primitive "length", [v]) -> S.Length (sql_of_query false v)
+| Q.Apply (Q.Primitive f, vs) -> S.Apply (f, List.map base_exp vs)
+| Q.Constant c -> S.Constant c
+(* WR: we don't support indices in this simple Sql generator *)
+(* | Q.Primitive "index" -> ??? *)
```
jamescheney, line 94: Perhaps leave out / remove comment?

core/query/evalMixingQuery.ml
```diff
@@ -0,0 +1,121 @@
...
+| e ->
+    Debug.print ("Not a base expression: " ^ (Q.show e) ^ "\n");
+    failwith "base_exp"
+
+let compile_mixing : delateralize:QueryPolicy.t -> Value.env -> (int * int) option * Ir.computation -> (Value.database * Sql.query * Types.datatype) option =
+  fun ~delateralize env (range, e) ->
+    (* Debug.print ("env: "^Value.show_env env);
+    Debug.print ("e: "^Ir.show_computation e); *)
+    (* XXX: I don't see how the evaluation here is different depending on the policy *)
```
jamescheney, line 103: Is this comment still relevant?

core/query/evalMixingQuery.ml
```diff
@@ -0,0 +1,121 @@
...
+    if range != None then eval_error "Range is not (yet) supported by the new mixing normaliser";
+    let evaluator =
+        (* FIXME *)
```
jamescheney, line 106: What needs to be fixed here?

core/query/mixingQuery.ml
```diff
@@ -0,0 +1,1551 @@
...
+
+  let rec flattened_pair x y =
+    match x, y with
+    | Var (_nx, Types.Record row), _ ->
+        let x' = Record (StringMap.fold (fun f _ acc -> StringMap.add f (Project (x,f)) acc) (field_types_of_row row) StringMap.empty)
+        in flattened_pair x' y
+    | _, Var (_ny, Types.Record row) ->
+        let y' = Record (StringMap.fold (fun f _ acc -> StringMap.add f (Project (y,f)) acc) (field_types_of_row row) StringMap.empty)
+        in flattened_pair x y'
+    (* XXX: using a field with an empty name to deal with variables of non-record type ... will it work? *)
```
jamescheney, line 301: Is this comment still relevant?

core/query/mixingQuery.ml
```diff
@@ -0,0 +1,1551 @@
...
+      | (Concat _ | Singleton _) as v ->
+          implode
+            (List.map
+               (function
+                  | Constant (Constant.Char c) -> c
+                  | _ -> raise (runtime_type_error "failed to unbox string"))
+               (unbox_list v))
+      | _ -> raise (runtime_type_error "failed to unbox string")
+
+  (* let rec field_types_of_list =
```
jamescheney, line 364: Is this still needed?

core/query/mixingQuery.ml
```diff
@@ -0,0 +1,1551 @@
...
+        reduce_artifacts (Q.Apply (xlate env f, List.map (xlate env) ps))
+    | Closure (f, _, v) ->
+      let open Lang in
+      let (_finfo, (xs, body), z_opt, _location) = Tables.find Tables.fun_defs f in
+      let z = OptionUtils.val_of z_opt in
+      (* Debug.print ("Converting evalir closure: " ^ Var.show_binder (f, _finfo) ^ " to query closure"); *)
+      (* yuck! *)
+      let env' = bind (empty_env env.policy) (z, xlate env v) in
+      Q.Closure ((xs, body), env')
+      (* (\* Debug.print("looking up query closure: "^string_of_int f); *\) *)
```
jamescheney, line 887: Are these commented out lines still relevant?

core/query/mixingQuery.ml
```diff
@@ -0,0 +1,1551 @@
...
+        Q.Case (v', cases', default')
+    | If (c, t, e) ->
+      let c = xlate env c in
+      let t = computation env t in
+      let e = computation env e in
+        Q.If (c, t, e)
+
+  (** returns true iff the input query contains any free variable in the list;
+   * the query is assumed to be in normal form! *)
+  let contains_free fvs =
```
jamescheney, line 970: Could this function live in mixingQuery?

core/query/mixingQuery.ml
```diff
@@ -0,0 +1,1551 @@
...
+    in
+    (* let newx = Var.fresh_raw_var () in *)
+    let vx = Q.Var (x, ty_elem) in
+    let cenv = Q.bind env (x, vx) in
+    (* Debug.print ("mk_for_term: " ^ string_of_int newx ^ " for " ^ string_of_int x); *)
+    Q.For (None, [x,xs], [], body_f cenv)
+
+  let rec norm in_dedup env : Q.t -> Q.t =
+    function
+    (* XXX: this is a quick and dirty fix to implement substitution into a term via normalization:
```
jamescheney, line 1005: Is this comment still relevant?

core/query/mixingQuery.ml
```diff
@@ -0,0 +1,1551 @@
...
+                  Q.type_of_expression u'
+                  |> TypeUtils.element_type
+                in
+                let vz = Q.Var (z, tyz) in
+                (* Debug.print ("norm.For fresh var: " ^ Q.show vz); *)
+                Q.reduce_for_source env (z, u', tyz) (fun env' os_f' ->
+                  Q.For (None, [], List.map (norm false env') (os_f' (os_f [])),
+                    norm in_dedup env' (Q.Singleton vz)))
+            | u' ->
+              (* Q.For (None, [], List.map (norm false env) (os_f []), u') *)
```
jamescheney, line 1113: Is this debugging code still needed?  It looks like the initial commented-out version is equivalent.

core/query/mixingQuery.mli
```diff
@@ -0,0 +1,93 @@
+open Utility
```
jamescheney, line 1: Looking at this vs. query.* it seems like there is a lot of overlap/code duplication.  This in turn seems to offer lots of opportunities for regression/having to fix the same bug in two places.

Would it be possible to unify the common parts of query.* with mixingQuery.*, for example by providing two Eval modules, one that handles queries containing Dedup and one without?

core/query/sql.ml
```diff
@@ -235,7 +249,19 @@ class virtual printer =
             pr_b_one_table l
             (Arithmetic.sql_name op)
             pr_b_one_table r
-
+(*
```
jamescheney, line 252: Can this comment go now?

jamescheney (changes requested):
This mostly looks ok.  My main high-level comment is it would be nice to avoid code duplication between query.* and mixingQuery.*, or if not then perhaps the latter could be simplified somewhat to leave out things that are only needed by the Nested or Flat policies (e.g. index, or ordering stuff respectively).  But I'm happy with cleaning up and merging and leaving how best to do this for later.

core/query/mixingQuery.ml
jamescheney: Yes (per @wricciot )

core/query/mixingQuery.ml
```diff
@@ -0,0 +1,1551 @@
...
+      | Project   of t * string
+      | Erase     of t * StringSet.t
+      | Variant   of string * t
+      | XML       of Value.xmlitem
+      | Apply     of t * t list
+      | Closure   of (Ir.var list * Ir.computation) * env
+      | Case      of t * (binder * t) StringMap.t * (binder * t) option
+      | Primitive of string
+      (* | Var       of Var.var * Types.datatype StringMap.t *)
+      | Var       of Var.var * Types.datatype
```
jamescheney, line 48: Could this be unified with Query.t by abstracting over the variable annotation type?  (Make a note of this as a comment in the code for future reference)

prelude.links
```diff
@@ -97,7 +97,10 @@ fun the(l) {
 }
 
 fun elem(x, l) {
-  not(l==[]) && (x == hd(l) || x `elem` tl(l))
```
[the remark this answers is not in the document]
wricciot: Looks like old code from an intermediate version. Will revert.

core/query/sql.ml
```diff
@@ -173,16 +177,25 @@ class virtual printer =
...
+        | _ -> Format.fprintf ppf "\nwhere %a" pp_os_condition condition
+      in
+      if fDistinct then
+        Format.fprintf ppf "select distinct %a\nfrom %a%a%a"
+        self#pp_fields fields
+        (self#pp_comma_separated pp_from_clause) tables
+        pp_where condition
+        pp_orderby os
+      else
+        Format.fprintf ppf "select %a\nfrom %a%a%a"
```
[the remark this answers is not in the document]
wricciot: Ok. I'd still prefer to generate "select" and "union" rather than "select all" and "union distinct", so the code is specialized in such a way that the unnecessary specification is omitted.

prelude.links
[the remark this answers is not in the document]
wricciot: Ok (probably the remainder of an intermediate version that didn't use a local `elem` inside `dedup`).

core/query/delateralize.ml
```diff
@@ -0,0 +1,247 @@
...
+let rec (>>==) (l : 'a list) (f : 'a -> 'a option) : 'a list option =
+    match l with
+    | [] -> None
+    | a::al ->
+        match f a, al >>== f with
+        | None, None -> None
+        | fa, fal ->
+            Some (unopt_default fa a::unopt_default fal al)
+
+(* not the most efficient way of doing these two functions: why is this not in the standard library? *)
```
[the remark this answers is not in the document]
wricciot: Yes. My efficiency complaint is based on the fact that when we find an element `x` of the collection for which `f x` is defined, we don't need to evaluate `f` on the remaining elements, nor do we need to scan the rest of the collection.

For lists, I can achieve that by ad-hoc recursion rather than using a fold. The same won't work on StringMap.t because its implementation is hidden: the best "standard" way I have to avoid evaluating `f` unnecessarily is by means of `lazy`/`force`, but `StringMap.fold` will still scan the whole map no matter what. Exceptions can be used to efficiently exit a loop in such a case, but since I need to extract a value of generic type `'a option`, I can't use exceptions without resorting to `Obj.(black)magic`, which I prefer to avoid.

core/query/delateralize.ml
wricciot: Moved to `QueryLang`

core/query/delateralize.ml
wricciot: Moved to `QueryLang`

core/query/mixingQuery.ml
wricciot: Factorised into `QueryLang` (and we still need to consider "index" there...)

core/query/evalMixingQuery.ml
wricciot: No.

core/query/evalMixingQuery.ml
wricciot: I believe this is a leftover.

core/query/mixingQuery.ml
wricciot: I'll remove the "will it work" part, as it is tested now. But it makes sense to document the fact that we are using an empty field name for his corner case.

core/query/mixingQuery.ml
wricciot: This should now be removed.

core/query/mixingQuery.ml
wricciot: No, I'm removing this.

core/query/mixingQuery.ml
wricciot: Indeed, much of that debugging code isn't needed.

core/query/mixingQuery.mli
wricciot: After a couple attempts, I have decided to move the common parts to a separate module called `QueryLang`. There's still some overlapping, but I can't do better than this while at the same time guaranteeing that the standard normalizer has the same semantics as that in the previous releases of Links.

core/query/mixingQuery.ml
wricciot: Unifying the two query types by changing the `Var` annotation to `Types.datatype` everywhere seems to have done the trick. It took a lot of work, but all the alternatives were less convincing...

core/query/evalMixingQuery.ml
jamescheney: Actually I don't think a projection from a table would be well-typed - projection in NRC queries takes a record, not a list of records (or a table, which is a different type from a list of records).  So probably safe to leave out this specific error but OK to leave it too.

jamescheney (approved):
Looks good to me, just a couple of small nitpicks, the most substantial being to consider renaming the generic `queryLang` to something like `Nrc`.

core/query/evalMixingQuery.ml
wricciot: Yeah, I guess there was no need to be worried about it in the first place. I probably wasn't thinking straight when I wrote that comment. I'm just a bit hesitant to remove it because of the remote possibility that last year I saw something that I'm not seeing now.

core/query/sql.ml
```diff
@@ -22,10 +22,14 @@ type query =
...
-and select_clause = (base * string) list * from_clause list * base * base list
+and select_clause =
+    multiplicity * select_fields * from_clause list * base * base list
+and select_fields =
+  | Star
+  | Fields    of (base * string) list
 and from_clause =
   | TableRef of table_name * Var.var
-  | Subquery of query * Var.var
+  | Subquery of dependency * query * Var.var  (* bool = LATERAL? *)
```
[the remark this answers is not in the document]
dhil:
```suggestion
  | Subquery of dependency * query * Var.var
```

dhil (approved):

dhil:
It looks to me like the outstanding TODOs are of fairly minor character. We can fix these in a subsequent PR. I am going to merge this PR in now as it blocks a whole string of PRs are from being merged (because we have made these ad-hoc agreements to merge this first, then Simon's PR, and so on).

jamescheney:
Actually never mind, let's just merge this - the two things I suggested modifying are very minor.

Issue #937, opened 2020-11-10
Extensible SQL serialiser


SimonJF:
Currently, the SQL serialiser is a monolithic function in `sql.ml`. This works fine for the most part, but some drivers expect different dialects of SQL. As an example, string concatenation in PSQL and SQLite3 can be written `'x' || 'y'`, but it is `concat('x', 'y')` in MySQL (:roll_eyes:).

The DB drivers do currently allow us to override `string_of_query` for serialisation, but this only works at the top-level of a query:
as soon as we need to change a pretty-printing function, we need to either duplicate the logic, or pass extra ad-hoc functions. To some extent, we already have this issue by needing to pass `quote` around as a separate argument.

Instead, I propose that we take a more object-oriented approach, with all of the pretty printers (e.g., `pr_delete`, `pp_quote`, etc.) defined as methods which can be overridden by the DB drivers as necessary.

PR #941, opened 2020-11-13
Extensible SQL Serialiser

SimonJF:
# Summary and Motivation

This commit aims to implement #937 by refactoring the monolithic SQL serialiser into an object. As a result, different database drivers may now override parts of the SQL serialiser in order to cleanly and modularly implement driver-specific code. You can see this in action in `pg_database.ml`, which now only needs to override `pp_insert`.

The issue with the previous approach is that it was very difficult to override *parts* of the SQL serialisation code, since it was a monolithic function. The only ways to implement specific functionality for each database was to pass in parameters everywhere (which in turn was an artifact of having the code very tightly-coupled before -- see #832). I got stung badly by this in my concurrent PR, #935 .

# Details

* `sql.ml`: The various printing functions have been made into an object, `Sql.printer`, which takes a `quote` function as its parameter. Methods of `Sql.printer` can be overridden on a per-driver basis, while inheritance allows us to use the default behaviour elsewhere.
* `value.ml`: 
  * The `Value.database` class now takes an `Sql.printer` object as a parameter, which can be specialised per-database. This is now the preferred way to implement driver-specific behaviour, rather than overriding `string_of_query`.
  * `quote_field` has been removed and put inside `Sql.printer`, where it belongs. Code (such as the lens code) can still get access to DB-specific quoting functionality using `db#get_printer#quote_field`.
* Drivers have been updated to take the changes into account.

# Warts

We still have `escape_string` in the `Value.database` class, even though it morally belongs in the `Sql.printer` class. This is because the `pg_database` driver implements `escape_string` by calling through to an underlying `connection` instance which is only defined after initialisation, whereas we need this at initialisation time to pass as a parameter to `pg_printer`.

This patch also does not address the existing wart of `insert returning` statements being tied to `Value.database` rather than in the SQL printer (mainly because of the fact that it can create multiple queries). Future work!

### DIFF
--- a/core/query/sql.ml
+++ b/core/query/sql.ml
@@ -68,13 +68,6 @@ let fresh_dummy_var () =
   incr dummy_counter;
   "dummy" ^ string_of_int (!dummy_counter)
 
-let string_of_label label =
-  if Str.string_match (Str.regexp "[0-9]+") label 0 then
-    "\"" ^ label ^ "\""     (* The SQL-standard way to quote an identifier;
-                               works in MySQL and PostgreSQL *)
-  else
-    label
-
 module Arithmetic :
 sig
   val is : string -> bool
@@ -145,7 +138,7 @@ let order_by_clause n =
    element type at all. *)
 
 
-(* TODO: OCaml cannot infer polymorphic record signatures.
+(* TODO: OCaml cannot infer polymorphic method signatures.
  * For now, hoisting these, but eventually we'd want to put
  * these into the object so they can be used by overridden methods. *)
 let pp_comma_separated pp_item =
@@ -155,16 +148,16 @@ let pp_comma_separated pp_item =
 let gen_pp_pair ppf fmt_str fl fr (l, r) =
   Format.fprintf ppf fmt_str fl l fr r
 
-class printer =
+class printer (quote : string -> string) =
   object (self : 'self_type)
 
-  method quote q =
-    "\"" ^ Str.global_replace (Str.regexp "\"") "\"\"" q ^ "\""
+  (* Defer to class parameters *)
+  method quote_field = quote
 
   method private pp_quote : Format.formatter -> string -> unit = fun ppf q ->
     (* Copied from the pg_database.ml, which seems to be the default,
      * I guess? *)
-    Format.pp_print_string ppf (self#quote q)
+    Format.pp_print_string ppf (self#quote_field q)
 
   method private gen_pp_option ppf fmt_str f option =
     OptionUtils.opt_iter (Format.fprintf ppf fmt_str f) option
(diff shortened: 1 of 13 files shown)


core/query/sql.ml
```diff
@@ -144,201 +137,250 @@ let order_by_clause n =
    returned. This allows these operators to take lists that have any
    element type at all. *)
 
-let pp_comma ppf () = Format.pp_print_string ppf ","
-let pp_comma_separated pp_item = Format.pp_print_list ~pp_sep:pp_comma pp_item
+class printer (quote : string -> string) =
```
jamescheney, line 140: It seems like there are two strategies for customization being used by the printer class:

1. passing in a (function) argument (e.g. `quote`)
2. extending/overriding the printer class, or creating an object of this class and then customizing it

Why not, e.g., leave the `quote` method abstract and thus force it to be defined by each DB_specific printer, and only use extension strategy (2)?

core/query/sql.ml
```diff
@@ -357,13 +399,11 @@ let rec inline_outer_with q =
     | q -> q
 
 
-let string_of_base quote one_table b =
-  Format.asprintf "%a" (pr_base quote one_table) b
+(* TODO: Still a lot to do to get this to compile, I'm sure.
```
jamescheney, line 402: Is this comment still relevant?

jamescheney (approved):
This looks reasonable.  I have the tiniest of quibbles regarding the extensibility strategy (quote as a parameter vs. other things as virtual methods).

Regarding the insert-returning thing, I guess this is really a higher-level question than SQL printing, the current workaround seems fine for now.

I am uncertain whether this PR includes the needed extensibility point to allow #935 to go ahead, or more work is needed there.  I don't see anything about printing string concat differently so assume this is deferred to #935 after this is merged.

core/query/sql.ml
SimonJF: Oops!

core/query/sql.ml
SimonJF: Yes, you're correct. I experimented with both options. I decided upon the approach of having `quote` as a parameter since there's no reasonable default, from what I can see: all 3 DBs do something slightly different.

Upon reflection, I think the best thing to do here would be to make the `quote` method virtual, and then have each printer override it (more specifically, the `default_printer` would still take `quote` as a parameter). This would have the dual advantages of a more uniform design while still not requiring an arbitrary default. Thanks for prompting me to think about this more carefully!

SimonJF:
Thanks for the helpful reviews, @jamescheney; I've just pushed a patch which should hopefully resolve them.

As per your question: 

> I am uncertain whether this PR includes the needed extensibility point to allow #935 to go ahead, or more work is needed there. I don't see anything about printing string concat differently so assume this is deferred to #935 after this is merged.

This patch is the extensibility point needed for #935; the point at which I needed concatenation printing is only contained in the #935 (serialisation of `SqlLike`). In fact, I've only just seen the `FIXME` on line 91 of `sql.ml` -- I expect I can work something out in the other patch which should sort this out, too.

PR #943, opened 2020-12-11
Add support for OCaml 4.12

kit-ty-kate:
OCaml 4.12 added `List.compare` which conflicts with the use of `compare` in those modules where the `List` module is opened.


dhil (approved):
Thanks!

PR #946, opened 2020-12-26
WIP: attempt to fix issue 773

### LINKED ISSUE #773: Promoting previously inferred types to signatures vs. behavior of recursive functions
The following properties seem to be at odds with each other:

1. When `TypeSugar` runs again after a previous run, the types stored in binders (put there by the previous run) are treated as if they were type signatures to check against.

2. Polymorphic functions are typed differently based on whether they have a type signature or not. If they do, they may be used polymorphically in their bodies, if they don't, they can only be used monomorphically inside.

3.  The transformation `DesugarInners` runs after type-checking and treats all (on the outside) polymorphic recursive functions polymorphically in their bodies as well. Hence, in their bodies it inserts the necessary type applications around usages of itself/other recursive functions from the same group. This step essentially removes the distinction between inside and outside in terms of typings, as the IR expects.

Together, these three properties mean that a problem can arise when `TypeSugar` is run again, but before `DesugarInners` has run. In this situation, by the first point, the recursive functions have signatures and are treated polymorphically inside. However, `TypeSugar`  should then notice that the necessary type applications are missing, because they are only inserted later on by `DesugarInners`.

The easiest solution seems to be to make sure that `DesugarInners` runs immediately after `TypeSugar`. Currently, however, `DesugarCP` runs between `TypeSugar` and  `DesugarInners`. We should make sure if we can swap `DesugarCP` and  `DesugarInners`.

jamescheney:
Not to be merged, we should have tests exercising `desugarCP` and any interactions with `desugarInners` first.  Would close #773.

### DIFF
--- a/core/frontend.ml
+++ b/core/frontend.ml
@@ -200,8 +200,8 @@ module Typeability_preserving = struct
 
   (* Collection of transformers. *)
   let transformers : transformer array
-    = [| (module DesugarCP)
-       ; (module DesugarInners)
+    = [| (module DesugarInners) (* this must always be done first after type inference *)
+       ; (module DesugarCP)
        ; only_if
            Basicsettings.Sessions.exceptions_enabled
            (module DesugarSessionExceptions)


jstolarek:
> Not to be merged

Github now allows to create draft pull requests:



Sadly, you can't convert existing PR into a draft.

jamescheney:
Actually it does allow to convert to draft too, but for some reason I can never see the option to create as draft, I just have to do it later and forgot in this case.

jamescheney:
After reviewing this I think it is probably safe to merge if say @dhil agrees.  I do not feel like coming up with examples of the CP syntax that would test for interactions between `desugarInners` and `desugarCP`; if someone else thinks this should be done I will create an issue to record this.

dhil (approved):
> I do not feel like coming up with examples of the CP syntax that would test for interactions between desugarInners and desugarCP; if someone else thinks this should be done I will create an issue to record this.

I am happy to dispense entirely of the CP extension, albeit that may never happen -- but a maintainer can dream, right? 

I suggest that we create a "meta-issue" linked to this PR for provenance, in case bad interactions between the interactions between CP, desugarCP, and desugarInners are discovered later. We can close the issue if no such interactions are discovered in, say, the next year.

jamescheney:
Alternatively, what if we make the CP syntax and translation pass conditional (like some other extensions that are not thoroughly tested/maintained) and then whoever wants it to become unconditional can add tests if they want?

That way it will still be "maintained" in the sense of we continue to make sure it typechecks, but we won't have a false sense of security that it works otherwise.

dhil:
Yes we can do something like that, though, it would still be a breaking change, alas perhaps less severe than ripping out the CP component. I think @slindley needs to sign off on such a change as he is one who is gatekeeping CP.

jamescheney:
Yes, that seems sensible.  What if we merge this and create an issue saying "someone should either make CP syntax an option-guarded feature, or write tests that exercise how CP might interact with local functions"

dhil:
I am happy to merge this PR as is.

PR #947, opened 2020-12-26
fix to set response to 500 on error

### LINKED ISSUE #198: Server Error Responses should have correct HTTP Status Codes
When making a RPC that crashes on the server, the response is the HTML doc displaying the `h1` 
> ## Links Fatal Error

This response comes with HTTP Status `200 OK`

Clearly a `500 Internal Server Error` is more appropriate.

jamescheney:
This fixes #198 by mutating the request data to set http response code to 500 whenever there is an error.


dhil (approved):
Looks good to me!

Issue #924, opened 2020-10-08
Experiences with links 0.9.2 installation via opam on MacOS with PostgreSQL


vcgalpin:
As a recent new user of links, I am documenting my experiences with installation on MacOS to record the process and to support development of better documentation for new users. Since my goal is to install the GtoP DB, I'm using PostgreSQL. My focus is on installing links without compiling anything from scratch, but rather using various package managers to install and/or compile.

These installations were done on the latest MacOS version (Catalina - 10.15.7) with latest version of Xcode (12.0.1)<sup>*</sup>.

The basic steps are to
- install a MacOS package manager that can install opam
- install opam via MacOS package manager
- install PostgreSQL, possibly using the MacOS package manager
- install links-postgres using opam
  (this step must happen after the installation of PostgreSQL)
- ensure the PostgreSQL server is running 
  (how much of this is done automatically depends on the specific installation method)
- setup appropriate databases and users in PostgreSQL for use by links
- edit the links config file with user details
- run links 

An important aspect of installing links with database support is the ordering of installation of different packages. In the case of PostgreSQL, PostgreSQL must be installed before opam compiles `links-postgresql`, because `links-postgresql` is dependent on the opam package `postgresql` (which is dependent on `conf-postgresql`) and to install `postgresql`, `pg_config` which is provided by the PostgreSQL installation must be run to provide information about the PostgreSQL installation, such as location of libraries.

For MacOS, a number of ways to download and install PostgreSQL can be found at
https://www.postgresql.org/download/macosx/
I've investigated all possibilities except fink, because opam is only available via MacPorts and homebrew

The following installations were **successful**:
- MacPorts (di), opam (MP), postgres96-server (MP), links-postgresql (opam) 
- MacPorts (di), opam (MP), postgres12-server (MP), links-postgresql (opam)
- homebrew (di), opam (hb), postgres12 (hb), links-postgresql (opam)
- EDB PostgreSQL 11.4 (di), opam,  links-postgresql (opam)
- EDB PostgreSQL 13 (di), opam,  links-postgresql (opam)
- postgres.app (di), opam,  links-postgresql (opam)

**Notation** the item in brackets describe how the software was installed/compiled: (MP) MacPorts, (hb) Homebrew, (di) direct install where an installer or executable is available for download.

The following installation was **unsuccessful**
- EDB PostgreSQL 12 (di), opam,  links-postgresql (opam)
[...]

PR #956, opened 2021-05-12
raise Ubuntu version used by Travis to bionic/18.04

### LINKED ISSUE #955: Travis build fails due to Sphinx requiring Python >= 3.6
Travis currently [fails](https://travis-ci.org/github/links-lang/links/builds/770921976#L848) at `make doc` on master because the `sphinx`  Python library seems to have started using features that are only available in Python >= 3.6.

There are several remedies:
1. Try to hard-code using an older version of `sphinx`.
2. Manually request installation of Python 3.6 (Ubuntu Xenial/16.04, which we currently use in Travis, has a default version of 3.5 when installing "python3", but Python 3.6 can be requested manually).
3. Update our Travis bots to use Ubuntu Bionic/18.04, which installs Python 3.6 as default.

The last option seems favorable to me, as we've had several problems in the past due do using an ancient Ubuntu version. However, this would mean that we never make the requirement for Python >= 3.6 explicit anywhere. I have verified in an experiment that simply raising the Ubuntu version in Travis from Xenial to Bionic fixes the problem, without needing to change anything else in the Travis config.

frank-emrich:
This fixes #955.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 sudo: required
-dist: xenial
+dist: bionic
 language: c
 services:
   - postgresql


dhil:
Can you try `focal` too (that's version 20.04)?

frank-emrich:
I can't get 20.04 to work: the opam package `mysql` only works with mysql 5.7, but Ubuntu 20.04 only comes with packages for mysql 8. I have tried both installing mysql 5.7 manually via third party package repositories and using mariadb 10.3 as a drop-in replacement. Neither worked so far and I don't think it's a good idea to sink more time into this.

Unless someone wants to look further into this, let's stay with 18.04 until there are no more database-related problems with 20.04.

dhil:
It does not sound quite right to me. I am running Ubuntu 20.04 LTS on my own machine, and I have `links-mysql` installed. I think I am using `mariadb` bindings. I can try to tinker with this PR before next meeting.

slindley:
If we're in this kind of situation then sticking with the ancient versions of everything is not the correct solution. I'm not sure if mysql can even support a lot of the interesting language-integrated query work that is provided by Links anyway. If we really can't get the OCaml bindings to work with a modern version of mysql then we should just stop testing them as part of CI and perhaps stop maintaining them. Glancing at the bug reports it sounds like one workaround could be to use mariadb instead anyway.

jamescheney:
MySql < 8 does not fully support the SQL:1999 features needed by shredding. The problem though is that ocaml-mysql does not support MySql 8 - at least last I checked (and raised an issue about it). MySql 8 has a new/different C API.  I think MySql 5.7 will continue to be widely used for a while.

It would be good to get things working with MariaDB too.

What stops us installing mysql-5.7 on the Travis vm, even if it's not the default for Ubuntu 20.x?

jamescheney:
> What stops us installing mysql-5.7 on the Travis vm, even if it's not the default for Ubuntu 20.x?

Sorry, I see Frank already addressed this. It looks like there are various ways of doing this but they seem fragile (e.g. using a special external package repository for MySql 5.7 on Ubuntu 20).

frank-emrich:
> Sorry, I see Frank already addressed this. It looks like there are various ways of doing this but they seem fragile (e.g. using a special external package repository for MySql 5.7 on Ubuntu 20).

Indeed! I tried this with no immediate success. However, I'm sure that it can be done if we choose to invest more time into this.

As a different solution, eagle-eyed @rudihorn spotted the `mysql8` package on opam, which is a fork of `mysql` with support for version 8. Using this seems to work and only required cosmetic changes to the links-mysql package, as seen in my "use mysql8 package" commit in this PR. I suspect that this package, however, *only* supports mysql 8, so we would lose support for 5.7.

Finally, there is now still a problem with postgres 12, which fails to start during the `travis_setup_postgresql` step. Unfortunately, the Travis documentation regarding this is very sparse and this requires further tinkering.

vcgalpin:
Probably not relevant, but on MacOS installing links via opam, it was not possible to get a working installation with EDB Postgres 12, although the MacPort and the homebrew versions of postgres 12 were OK. See note at the end of issue #924

dhil:
I just noticed the following warning on travis-ci.org:
```
Please be aware travis-ci.org will be shutting down by end of May 2021. Please consider migrating to travis-ci.com. 
```
I guess today is the last day we can use travis-ci.org. I am not sure what needs to be done in order to migrate to travis-ci.com.

dhil:
I have resolved the original issues regarding installation of `links-mysql` and `postgres`. A new issue has surfaced: I am not sure why, but the relational lenses test package seems to insist on connecting to postgres using the role `travis`, even though, the config uses the role `links`. CC @rudihorn

rudihorn:
@dhil For unit tests it is this config that needs to be updated https://github.com/frank-emrich/links/blob/travis-use-bionic/tests/unit/config.sample which needs to be updated with the new username/password. The relational lenses integration test configuration files should use this: https://github.com/frank-emrich/links/blob/travis-use-bionic/tests/relational-lenses/config.pgsql. I think the config you changed can actually be deleted.

dhil:
Thanks I will give it a try.

SquidDev:
Apologies if everyone is aware of this already, but it looks like Travis are [significantly reducing their offerings for open source](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing) (afaik, requiring you to apply for more credits once the budget has run out).

I don't know if it's worth switching to an alternative CI provider (GH Actions, Azure, CircleCI) while things are needing to be updated anyway?

dhil:
Thanks Jonathan! During yesterday's meeting I briefly mentioned that we should consider looking into GH actions. We tried Azure in the past, which turned out to be pain as their images changed fairly regularly causing our CI scripts to break (our CI scripts are quite possibly too fragile).

dhil:
I think supporting MySQL 5.7 is dead in the water, unless someone volunteers to build a custom image. In its current state this PR restore Travis CI with PostgreSQL 12, MySQL 8, and SQLite3. 

Some difficulties arise from Travis changes to the virtual environment (new usernames, ports, new packages, etc). However, many problems are also our own fault because our code base is not modular. We duplicate information in many places, e.g. database configs. We also have lots of stale stuff in our repository which makes it much harder to navigate and tell what is actually being loaded.

CC @frank-emrich maybe you can take a look at this PR now.

jamescheney:
As I understand it Mysql 5.7 is not quite EOL but will be soon, so this makes sense.  But the problem with MySQL5 is that it's diffcult to install/configure on ubuntu 20 and travis, not that there is something wrong with the Links code or ocaml-mysql driver, right?  

MySQL 8 should support enough SQL:1999 for shredding so we should make an issue to remind us to enable this.

As far as I can tell the ocaml-mysql8 library is beta/forned from ocaml-mysql so I worry slightly about depending on it.  But presumably the case for incorporating this into ocaml-mysql or maintaining ocaml-mysql8 will strengthen as Mysql8 takes over.

At the moment this PR will break code that currently runs for anyone using links-mysql to connect to a mysql5 database. (The number of people affected may be zero for all I know).  Nevertheless, I would be in favor of keeping the (unsupported, not CI tested) capability to connect to mysql5 (and possibly MariaDB) through links-mysql, and making a new Links driver library links-mysql8 that does things the new way.  From now on we only test against mysql8 in the code and remove the old links-mysql library when mysql5 is fully dead.  (Or if it is possible to support both versions of mysql within links-mysql8 that would be fine too but seems more painful).

dhil:
> As I understand it Mysql 5.7 is not quite EOL but will be soon, so this makes sense. But the problem with MySQL5 is that it's diffcult to install/configure on ubuntu 20 and travis, not that there is something wrong with the Links code or ocaml-mysql driver, right?

Yes.

> Nevertheless, I would be in favor of keeping the (unsupported, not CI tested) capability to connect to mysql5

I suppose we can rename the `links-mysql` package in this PR to `links-mysql8` and then retrieve the old `links-mysql` package from the master branch.

.travis.yml
```diff
@@ -1,35 +1,56 @@
...
-  - ./opam2_local pin add links-mysql . -y
+  - opam pin add links . -y
+  - opam pin add links-postgresql . -y
+  - opam pin add links-sqlite3 . -y
+  - opam pin add links-mysql . -y
   - make doc
   - make all-ci
   - make tests
-  - ./run-database-tests tests/database -d all
+  - ./run-database-tests tests/database -d mysql
```
jamescheney, line 53: What is the reason for this change?  If it's just to avoid spending time running the other database tests while getting this PR to work, it should be reverted.

jamescheney (changes requested):
I am happy with moving to mysql8 for testing and providing a Links mysql8 driver library: this should finally support shredding for mysql and probably has other advantages.  However I think the existing, working driver links-mysql should be preserved as is for now, we just don't commit to testing/maintaining it and drop it when mysql5 is fully dead.  This looks like it would be straightforward just by copy-pasting the mysql-driver library and reverting the old driver code.  It might also be possible to support both from within one library but I'm not sure this is worth the trouble.

jamescheney:
> I suppose we can rename the links-mysql package in this PR to links-mysql8 and then retrieve the old links-mysql package from the master branch.

That would be good. Looking over the changes so far, the diff to the actual links-mysql driver code is minimal, I'm happy to take a crack at this if you want.

.travis.yml
dhil: good catch! That is supposed to read "all"

dhil: I changed it to `mysql` for faster turn around when I was debugging the mysql build.

dhil:
> > I suppose we can rename the links-mysql package in this PR to links-mysql8 and then retrieve the old links-mysql package from the master branch.
> 
> That would be good. Looking over the changes so far, the diff to the actual links-mysql driver code is minimal, I'm happy to take a crack at this if you want.

Yes please, that would be great.

jamescheney:
OK, I have tried.  It builds locally but I have not tried to run the database tests locally (at all...)

For the moment what I have done duplicates, rather than replaces, the mysql configurations and tests; the mysql(5) ones are just disabled.  So there are two drivers, mysql and mysql8, and in the tests we actually run we need to make sure to use the latter.  There are also a handful of places where stringly-typed programming happens (I think mainly in the lenses code) where we now need to match either mysql or mysql8; this is a slightly bad code smell that I felt guilty about exacerbating.

For reasons of sanity preservation I have not tried to change the various identifiers named 'mysql_*' in 'mysql8_database.ml' since these are not visible outside the driver as far as I know.  But there is now some code duplication between the two drivers which is slightly ugly, and makes me reconsider whether it would be better to package them into one driver that can handle either mysql5 or 8.  I suppose that should be a separate issue since it is not necessary just now to get this PR closed and make CI work again.

rudihorn:
Is there actually a reason why we aren't splitting up all the tests we are doing into different CI tasks? Then the `make doc` can run using the latest image. That way there should be less tripping up between opam with the different package versions when installing ocamlformat and then the Links dependencies as well. Thy mysql 5.7 tests could just run on an old image, while others like postgres could use the new version. Some of it could also run in parallel.

rudihorn:
Also for Azure pipelines can't you specify any dockerhub image and thus target a very specific image which should not change?

rudihorn:
Also look at https://github.com/rudihorn/links/blob/azure_hack_serv/azure-pipelines.yml for a reference on using a service container for e.g. postgresql

dhil:
@rudihorn I appreciate the enthusiasm, though, I think the main problem is noone is willing to maintain it. I am definitely not willing, and whilst you may be willing, you will relatively soon be moving onto greener plains, so it makes little sense for you to take on the responsibility.

Another thing to bear in mind is to *to keep it simple*; it is quite possible that we can achieve ever so more by increasing the complexity of our software stack, but it comes at the expense of raising the bar to entry -- and I remind you that this is project where people come and go. I am in favour of generally "dumbing down" the codebase to make it easier to extend with new features (easier in the sense that a newcomer can rapidly get something going). I think a similar argument applies to CI. The CI script should just be a handful of steps: install this dependency, install links, go! This way it is easy for us, and newcomers, to follow the steps of the CI in a standard environment. It shouldn't be necessary to spin up a range of containers -- at the end of day we do not envisage people running Links in containerised form.

rudihorn:
In the current state the CI is complicated and broken because of dependency mismatches between two completely separate functions (documents and databases). If MySQL would be using the 5.7 MySQL service container (https://hub.docker.com/layers/mysql/library/mysql/5.7.34/images/sha256-92ad1d7e3f8eb7e67d35bf251912fb7cd12676a601dc90b6beb1aece7c1f5073?context=explore) it would still be running as it should. Separating out CI tasks is just standard software practice and saves pain and hassle. It also doesn't contradict the "install dependencies then links", it just means taking tasks that the user doesn't need to perform like `make doc` and shoving them somewhere were we don't need to care about them because they probably won't break (e.g. https://hub.docker.com/r/sphinxdoc/sphinx).

dhil:
> Separating out CI tasks is just standard software practice and saves pain and hassle.

You are missing the bigger picture. You are introducing a massive abstraction, someone (preferably *someone* in plural) needs to understand how it works internally.

jamescheney:
> All checks have passed
> 1 successful check
Woo hoo!

jamescheney (approved):
Light is green, trap is clean

dhil (approved):

Issue #963, opened 2021-06-01
Consider mergng links-mysql and links-mysql8


jamescheney:
In #958 we have created a new library `links-mysql8` which uses `ocaml-mysql8` to connect to MySQL 8.x DBMSs.

I recall seeing somewhere that the MySQL8 client library is backwards compatible with older MySQLs but not the reverse.  So, the extra library may not be necessary.  We are not currently able to test this in CI, since it is difficult to set up MySQL5.7 on Ubuntu 20, but it should be easy enough to test it in a VM using an older distribution that comes with MySQL 5.7 standard.

If this works, I'm happy with going back to using only `ocaml-mysql8` and `links-mysql`.  If not, given the current code duplication between the `links-mysql` and `links-mysql8` libraries I'd slightly prefer merging the two libraries so that we adaptively use the right one dpeending on the mysql version of the database we are connecting to.

jamescheney:
On reflection combining into two libraries doesn't seem like a good thing to do, since the single library would depend on both ocaml-mysql and ocaml-mysql8.  So the main question is whether we can connect to legacy MySql 5.x instances from ocaml-mysql8.

dhil:
I am closing this issue as #966 tests `links-mysql8` against a MySQL 5.7 instance.

PR #966, opened 2021-06-07
Migrate CI to GHA.

dhil:
This PR migrates our CI from Travis to GitHub Actions (GHA).

* Removes Travis configuration (Travis will have to be uninstalled manually either before or after merging this PR).
* CI is triggered on pushes to `master` or pull requests opened against `master`. We can change this behaviour if need be.
* Tests `links-mysql` in a Ubuntu 18.04 environment (this environment ought to be available for the next two years or so).
* The database argument strings in database configs are now parameterised by environment variables to make them more portable. The variables and their default values are:
```
# PostgreSQL
LINKS_POSTGRES_HOST=localhost
LINKS_POSTGRES_USER=links
LINKS_POSTGRES_PASSWORD=links
LINKS_POSTGRES_PORT=5433

# MySQL8 variables
LINKS_MYSQL8_HOST=127.0.0.1
LINKS_MYSQL8_USER=links
LINKS_MYSQL8_PASSWORD=links
LINKS_MYSQL8_PORT=3306

# MySQL5.7 variables
LINKS_MYSQL_HOST=127.0.0.1
LINKS_MYSQL_USER=links
LINKS_MYSQL_PASSWORD=links
LINKS_MYSQL_PORT=3306
```
It may be worthwhile to put them in a file, `.dbenv`, under revision control for an easy way to access and modify them, e.g. `eval $(cat .dbenv); ./run-database-tests -d all tests/database`. It may also be useful in reducing some of the redundancy in the workflow definition.

* The OCaml 4.08.0 compiler is cached, which substantially reduces the time taken to run the CI.
* Testsuites are now run in separate parallel jobs, which are
   + Repository rule check
   + Build documentation
   + Interpreter non-database tests
   + PostgreSQL, MySQL8, and SQLite3 tests
   + MySQL5.7 tests
   + `links-mysql8` on MySQL5.7 tests

Future things we can look into:
* Caching Links dependencies.
* Test with multiple OCaml compilers.
* Extend the operating system matrix with macOS.


dhil:
SUCCESS!

dhil:
**Note** I have updated the first comment.

rudihorn:
These are essentially the changes that I suggested so I approve of this PR.

.github/workflows/default.yml
```diff
@@ -0,0 +1,289 @@
...
+        env:
+          POSTGRES_USER: links
+          POSTGRES_PASSWORD: links
+        ports:
+          - 5433:5432
+      mysql:
+        image: mysql:8.0
+        env:
+          MYSQL_USER: links
+          MYSQL_USER_PASSWORD: links
```
rudihorn, line 134: I think if this becomes MYSQL_PASSWORD, the code to create the user shouldn't be necessary.

.github/workflows/default.yml
```diff
@@ -0,0 +1,289 @@
...
+          - 4.08.0
+
+    runs-on: ${{ matrix.os }}
+
+    services:
+      mysql:
+        image: mysql:5.7
+        env:
+          MYSQL_USER: links
+          MYSQL_USER_PASSWORD: links
```
rudihorn, line 224: ditto

.github/workflows/default.yml
```diff
@@ -0,0 +1,289 @@
...
+    steps:
+
+      - name: Configure MySQL
+        run: |
+          mysql --version
+          sudo systemctl enable mysql
+          sudo systemctl start mysql
+          mysql -h 127.0.0.1 -P 3308 -uroot -plinks -e "CREATE USER links IDENTIFIED BY 'links';"
+          mysql -h 127.0.0.1 -P 3308 -uroot -plinks -e "GRANT ALL ON  links.* TO links;"
+          mysql -h 127.0.0.1 -P 3308 -uroot -plinks -e "FLUSH PRIVILEGES;"
```
rudihorn, line 245: ditto

.github/workflows/default.yml
```diff
@@ -0,0 +1,289 @@
...
+      mysql:
+        image: mysql:5.7
+        env:
+          MYSQL_USER: links
+          MYSQL_USER_PASSWORD: links
+          MYSQL_ROOT_PASSWORD: links
+          MYSQL_DATABASE: links
+          MYSQL_ROOT_HOST: '%'
+        ports:
+          - 3308:3306
```
rudihorn, line 229: I would suggest mapping to the same standard MySQL port 3306.

.github/workflows/default.yml
dhil: It sadly is.

.github/workflows/default.yml
```diff
@@ -0,0 +1,289 @@
...
+          --health-interval=10s
+          --health-timeout=5s
+          --health-retries=3
+
+    steps:
+      - name: Configure MySQL
+        run: |
+          mysql --version
+          sudo systemctl enable mysql
+          sudo systemctl start mysql
```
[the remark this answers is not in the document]
dhil: They are sadly necessary.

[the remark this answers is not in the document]
dhil: The `enable` step can probably be dropped.

.github/workflows/default.yml
dhil: These variables appear to be ignored by the container, and can probably be deleted.

.github/workflows/default.yml
dhil: I did that initially, but it doesn't work as the host image already occupies that port.

rudihorn:
The problem is that you are starting an instance of MySQL that is included with the host container, so this blocks you from mapping the port in the mysql service container. I deleted all the MySQL setup code here and it works: https://github.com/rudihorn/links/pull/1

dhil:
Nice!

dhil:
Regardless, the exact choice of port is largely unimportant as everything is now parameterised. Even for PostgreSQL we can't pick a "standard" port reliably, because some have postgresql 9.x, postgresql 10.x, and postgresql 12.x installed on their local system (e.g. me), each instance runs on a separate port (which I think is determined by the order of installation).

.github/workflows/default.yml
rudihorn: You still need to fix this otherwise the user account setup won't work.

dhil:
Something is odd here, @rudihorn your suggestion does not seem to work off the shelf here
```
ERROR 1045 (28000): Access denied for user 'links'@'172.18.0.1' (using password: YES)
```
I am also confused about the address `172.18.0.1`. Not sure where that is coming from.

rudihorn:
Addressing https://github.com/links-lang/links/pull/966#discussion_r649006748 will fix it. The other address is that of the service container, just ignore it as it is fine.

dhil:
I don't understand your comment in the light of the recent changes.

rudihorn:
You need to change `MYSQL_USER_PASSWORD` into `MYSQL_PASSWORD`

dhil:
Aha, thanks

dhil:
Thanks it is slightly slicker now

dhil:
I would still like to learn how to index into the environments of services such that I don't have to duplicate information.

rudihorn:
> 
> 
> I would still like to learn how to index into the environments of services such that I don't have to duplicate information.

I suspect this is not possible. I found https://bloggie.io/@_junrong/using-environment-variables-secrets-in-github-actions which dumps the json output of various variables and the env vars are not included anywhere, see: https://github.com/rudihorn/links/runs/2793276737?check_suite_focus=true


Maybe it is possible to store the information elsewhere (e.g. in the matrix), but I'm not super convinced by this approach...

rudihorn:
Yes I think it is fine.

rudihorn (approved):

frank-emrich (approved):

core/database.ml
```diff
@@ -2,11 +2,13 @@ open CommonTypes
...
 
 let connection_info
-  = Settings.(option "database_args"
-              |> synopsis "Database host, name, user, and password"
+  = let setting =
+      Settings.(option "database_args"
+              |> synopsis "Database host, port, user, and password"
               |> to_string from_string_option
-              |> convert Utility.some
+              |> convert (fun s -> Utility.some (Sys.expand s))
```
jamescheney, line 9: This is a bit ad hoc.  Are there any other places where environment variable expansion would be helpful?  Maybe the fact that environment vars are expanded for this specific option (and the reason why) should be documented in the synopsis or a comment?

jamescheney (approved):
This looks reasonable to me.  I wonder (per #963 ) whether it is possible to test Links with links-mysql8 against a MySQL 5.7 database.  If we could confirm that links-mysql8 still connects fine to at least some older MySQLs then I'd be happier about doing #963.  I suspect this is not straightforward though so won't insist on it.

core/database.ml
dhil: We do the same thing in places where filenames/pathnames come up, e.g. `parse.ml`, `moduleUtils.ml`, etc.

dhil: Actually, I see that past-me has defined a utility function for this purpose called `parse_paths` in `settings.ml:128`. However, it converts colon/comma separated string into a list of strings (each expanded). I could use this. It will require me to patch up the places where we parse the database arguments string, though, it might be better to parse it up front with `parse_paths` (maybe rename this function too).

dhil: I think my suggestion has to be implemented as a separate patch. See #970.

dhil:
> I wonder (per #963 ) whether it is possible to test Links with links-mysql8 against a MySQL 5.7 database. 

@jamescheney The latest commit adds a job that runs a Ubuntu 20.04 instance, which tests `links-mysql8` against MySQL 5.7 database

dhil:
I'll merge this PR tomorrow and uninstall Travis. Subsequently, I will take a look at the other PRs and trigger CI reruns for each of them.

PR #968, opened 2021-06-08
fix #967

### LINKED ISSUE #967: Calling serverTime from client fails
In this example:
```
fun testPage(msg) {
  fun foo(x) client { replaceNode(
      (fun () {<p><b>{stringToXml(x)}</b>:<b>{intToXml(serverTime())}</b></p>})()
    , getNodeById("xyz"))   
  }
  page
  <html>
    <body>
      <div>Message: {stringToXml(msg)}</div>
        <form l:onsubmit="{foo(x)}" method="post">
          <input type="text" l:name="x"/>
        </form>
      <div id="xyz"/>
    </body>
  </html>
}
```
This creates a form that when submitted calls a client function that replaces another node with content provided by a server function that refers to the argument x to the client function. When run, if we submit the form we get a client-side error message:

```
    Uncaught Error: Fatal error: call to server returned an error. Details: ***: Error: "Assert_failure core/evalir.ml:246:18" 
        remoteCallHandler http://localhost:8080/lib/jslib.js:899
```
The assertion failure in evalir line 246 seems to be an unhandled case where it is expected that either we have both free variables and an environment, or neither free variables nor an environment, in a server call. So somewhere something things are getting out of sync.

@kwanghoon has suggested a fix in a comment in #907.

kwanghoon:
This PR fixes #967. 
 - A debug point to the FunctionPtr of apply in evalir.ml to make noise in case 'find_fun f' fails silently
 - parse_remote_call in webif.ml can now generate either FunctionPtr or PrimitiveFunction for each client remote call request.


core/evalir.ml
```diff
@@ -238,7 +238,8 @@ struct
         [], Types.Not_typed)) in
     function
     | `FunctionPtr (f, fvs), ps ->
-      let (_finfo, (xs, body), z, _location) = find_fun f in
+      let (_finfo, (xs, body), z, _location) = (try find_fun f with NotFound s ->
```
jamescheney, line 241: I would suggest putting the try-with on th enext line to make it more obvious there is nontrivial control flow here.

Perhaps also clarify (if I am right) that the change here does not actually contribute to fixing #967, but instead is meant to improve reporting if something goes wrong.  We could even make the debug message a more obvious warning that tells us something is wrong, or translate the NotFound exception to a more informative one that tells the clients of this function that an unsuccessful attempt to look up the function name was made.  (Indeed, a malicious client could try to do this intentionally...)

core/evalir.ml
```diff
@@ -238,7 +238,8 @@ struct
...
+           (Debug.print("NotFound in fun_defs: "^string_of_int f); raise (NotFound s)) ) in
```
jamescheney, line 242: and put the final "raise" on its own line too

core/webif.ml
```diff
@@ -38,7 +38,10 @@ struct
 
     let func =
       match fvs with
-      | `Record [] -> `FunctionPtr (int_of_string fname, None)
+      | `Record [] -> let i_fname = int_of_string fname in
```
jamescheney, line 41: I would add a comment here explaining the reason for the conditional behavior: basically IMO the reason is there is an artificial distinction between primitive functions (builtin functions) and function pointers (functions defined as part of a Links program), and this is the point where we have to do something about it in order for attempts to remotely call primitive functions to work properly.

jamescheney (changes requested):
This looks reasonable provided CI is happy.  I guess the change to `evalir` is just to aid debugging?  Otherwise I have minor formatting requests.

This should probably wait for 0.9.3 to be finalized.

core/evalir.ml
kwanghoon: I want to improve reporting if something is not found here. Your suggestion is a good idea to move the debug message to the NotFound exception after making it more informative. I will change the lines accordingly, and will also make the try-with more explicit having new lines here.

core/evalir.ml
```diff
@@ -238,7 +238,10 @@ struct
         [], Types.Not_typed)) in
     function
     | `FunctionPtr (f, fvs), ps ->
-      let (_finfo, (xs, body), z, _location) = find_fun f in
+      let (_finfo, (xs, body), z, _location) =
+           try find_fun f
+           with NotFound s ->
+             raise (NotFound (s^" attempting to look up the function name")) in
```
jamescheney, line 244: Sorry, I think what I had in mind was to raise an Evalir.internal_error here, with a more informative error message than "NotFound", not to change the payload data of NotFound.  

Have a look at the `run_program` function which catches `NotFound` exceptions raised during program execution.  I think we should do something similar here.  (However, the bug illustrated in #967 doesn't seem to be raising a NotFound exception, rather it is finding something but the z and fvs values don't match the first two cases so `assert false` is being raised.  Perhaps the `assert false` should also be turned into a more informative error message.

jamescheney (approved):
Looks reasonable, but I had something different in mind for how to handle the NotFound.

If that is fixed, I am happy, though this PR should still wait until 0.9.3 is finalized.

core/evalir.ml
kwanghoon: To @jamescheney: Is this more like what you have in your mind? I'll update the previous PR with this. 

try find_fun f
with NotFound s ->
   raise (internal_error s) 

An interesting point is this. 

NotFound exception related with this issue #967 is never propagated to the `with` clause in run_program.
The reason is that `perform_request` (in `webif.ml`) and `callback` (in `webserer.ml`) directly call `Eval.apply` without going through `run_program`.

So the exception does not have any information (like a file name) that `internal_error` appends to it. I think, as a result, it has been a bit hard to find out where this exception is raised.

dhil, line 244:
```suggestion
             raise (internal_error ("Failed to find function name: " ^ (string_of_int f)))
```

dhil (changes requested):

dhil (approved):

kwanghoon:
👍 @dhil

PR #971, opened 2021-06-11
Freshen generalized type variables after turning lambda into named function

### LINKED ISSUE #871: Generalisation breaks the value restriction

frank-emrich:
In `DesugarFuns`, lambdas are translated to named functions as follows:
```
fun (x) {
...
}
```
becomes 
```
{
  fun f(x) {...}
  f[arg1, ... ,argn]
}
```
Here, f is generalized to some type \forall a1, ... an. A, and the `arg1` to `argn` type arguments are inserted based on the lambda's original type.

However, there is an issue where we may generalize type variables that aren't present in the type environment, but nevertheless appear elsewhere outside of `f` as flexible variables. The current generalization approach mutates all generalized type variables directly, turning them into rigid ones. This means that any variable `a` appearing outside of `f` that gets generalized is turned rigid, too, and is therefore unbound, causing IR type errors.

This PR adds `generalise_without_mutation` to `Generlise`, which does not mutate/rigidify the generalized type variables. Instead, it uses the type variable renaming infrastructure to replace the generalized type variables with fresh, rigid variables inside the body of `f`.

Note that this is only used for generalization done by `DesugarFuns`, while `TypeSugar` still mutates type variables as before. We need to investigate further if we need to do the same refreshing there, too.

This fixes #871.


jamescheney:
This looks reasonable, but feels a bit like a local fix where a global redesign might be preferable (e.g. because the generalise_without_mutation leaves work that needs to be done somewhere else later, which we must therefore remember to do).  But at least it fixes the examples that break because of #871.

frank-emrich:
I agree that there are two kinds of issues with this fix:

### Type-checking

It seems that instead of refreshing the quantifiers, we could also ensure that no type flexible type variable b gets generalized if it appears somewhere in the AST outside of the function to be generalized. This is a stronger condition than what we currently check (where we just check that be does not appear in the type environment). I'm not sure if this can be implemented easily/efficiently. Maybe rank tracking could be used for this?

### Implementation
I agree that the `generalise_without_mutation` is somewhat ad-hoc in the sense that it leaves work left to be done. We could introduce something like `generalise_and_refresh` in `Generalise`, which gets handed the function binding itself and does all of the refreshing immediately, rather than relying on the use site to do this.

I shied away from this because it would push additional knowledge about the AST into `Generalise`, which ideally should be independent from the AST, but only require knowledge about the representation of types. 
It seems that currently, `Generalise` does to have `Sugartypes` in its dependency chain. However, `Generalise` already knows about type environments.

core/generalise.ml
```diff
@@ -174,8 +174,14 @@ let mono_type_args : type_arg -> unit =
 (** generalise:
     Universally quantify any free type variables in the expression.
 *)
-let generalise : gen_kind -> ?unwrap:bool -> environment -> datatype -> ((Quantifier.t list * type_arg list) * datatype) =
-  fun kind ?(unwrap=true) env t ->
+let generalise :
+      gen_kind ->
+      rigidify_vars:bool ->
```
dhil, line 179: Not sure what the point of labelling this parameter is. If you insist on it being labelled, then I suggest moving it to the front of the parameter list. You can move `unwrap` too whilst you are at it.

core/generalise.mli
```diff
@@ -4,5 +4,11 @@ val generalise : ?unwrap:bool -> Types.environment -> Types.datatype
               -> ((Quantifier.t list * Types.type_arg list) * Types.datatype)
 val generalise_rigid : ?unwrap:bool -> Types.environment -> Types.datatype
                     -> ((Quantifier.t list * Types.type_arg list) * Types.datatype)
+
+(* Like generalise, but does not mutate the generlized type variables by making them rigid.
```
dhil, line 8:
```suggestion
(* Like generalise, but does not mutate the generalised type variables by making them rigid.
```

core/generalise.mli
```diff
@@ -4,5 +4,11 @@ val generalise : ?unwrap:bool -> Types.environment -> Types.datatype
...
+  This means that the generlized type variables must be post-processed elsewhere! *)
```
dhil, line 9:
```suggestion
  This means that the generalised type variables must be post-processed elsewhere! *)
```

dhil (changes requested):

core/generalise.ml
frank-emrich: I usually prefer naming boolean parameters.
The reason for the unconventional ordering of parameters here was that the definition of `generalise_rigid`, `generalise`, ... further down in the file need this order for currying. I solved this now by making the definitions in lines ~235 slightly more verbose.

tests/typed_ir.tests
```diff
@@ -88,11 +88,10 @@ function pattern matching has right effects (#691)
 sig f : ([a]) {}-> a fun f ([x]) { x }
 stdout : () : ()
 
-Generalisation obeys value restriction (#871)
+Generalisation obeys value restriction / Freshening of generalized type variables when turning lambda into named fun (#871)
```
dhil, line 91:
```suggestion
Generalisation obeys value restriction / Freshening of generalised type variables when turning lambda into named fun (#871)
```

core/desugarFuns.ml
```diff
@@ -101,18 +101,24 @@ object (o : 'self_type)
...
     let o = o#bind_binder bndr in
+    let fun_definition =
+      { fun_binder           = bndr
+      ; fun_linearity        = lin
+      ; fun_definition       = (s_tvs, def)
+      ; fun_location         = loc
+      ; fun_signature        = None
+      ; fun_frozen           = true
+      ; fun_unsafe_signature = false } in
+    (* We replace the generalized quantifiers (instead of mutating the
```
dhil, line 116:
```suggestion
    (* We replace the generalised quantifiers (instead of mutating the
```

dhil (approved):

PR #972, opened 2021-06-12
Sugartoir: Fix setting ambient effect when traversing zero-ary function

### LINKED ISSUE #873: "Incompatible effects" IR typechecking error
Compiling the following function from Prelude:

```
sig freshResource : () ~> ()
fun freshResource() {
  escape esc in {
    redirect("?_k=" ^^ unsafePickleCont(fun () {esc(())})); exit([])
  }
}
```
produces an IR typechecking error:
```
***: IR Type Error: Incompatible effects:
 Allowed:
  wild
 Actual:
  wild|350
occurring in IR tail computation:
(Ir.Apply ((Ir.TApp ((Ir.Variable 172), [{ |wild|%365 }; %359; { |350 }])),
   [(Ir.Variable 228)]))
```
Reproduced on 2b26e2fe with following config:
```
typecheck_ir=true
fail_on_ir_type_error=true
optimise=false
debug=true
print_types_pretty=false
hide_fresh_type_vars=false
show_flavours=true
show_quantifiers=true
print_types_pretty=true
show_raw_type_vars=true
```

See #697 for a similar error but with a different cause.

frank-emrich:
In `eval_bindings` in `Sugartoir`, the ambient effect saved in the environment needs to be updated before traversing the body of a function binding.

Currently, this is achieved by reading the effect row  `eff` from the function's type and passing it to `CompilePatterns.desugar_pattern`, which is used to fold over the function's parameters.

However, as a direct consequence of this being a fold, it has no effect (pun intended) when the function has no parameters: The fold is a no-op, and the unchanged, outer environment is used when traversing the function body, without updating the ambient effect.

This fixes #873.


frank-emrich:
@jamescheney expressed some interested in seeing this fix when we discussed this last week

jamescheney:
Looks reasonable; I remember spending some time banging my head on this last summer and am happy if this means I don't have to try again :)

core/sugartoir.ml
```diff
@@ -1156,13 +1157,14 @@ struct
...
                     let eff = TypeUtils.effect_row ft in
+                    let env' = with_effects env eff in
                     let ps, body_env =
                       List.fold_right
                         (fun p (ps, body_env) ->
                            let p, penv = CompilePatterns.desugar_pattern eff p in
                              p::ps, body_env ++ penv)
                         ps
-                        ([], env) in
+                        ([], env') in
```
dhil, line 1167:
```suggestion
                        ([], with_effects env eff) in
```

core/sugartoir.ml
```diff
@@ -1156,13 +1157,14 @@ struct
                     let f  = Binder.to_name bndr in
                     let ft = Binder.to_type bndr in
                     let eff = TypeUtils.effect_row ft in
+                    let env' = with_effects env eff in
```
dhil, line 1160:
```suggestion
```

core/sugartoir.ml
```diff
@@ -1194,6 +1196,7 @@ struct
                           let f  = Binder.to_name bndr in
                           let ft = Binder.to_type bndr in
                           let eff = TypeUtils.effect_row ft in
+                          let env' = with_effects env eff in
```
dhil, line 1199:
```suggestion
```

core/sugartoir.ml
```diff
@@ -1202,7 +1205,7 @@ struct
                                   let p, penv = CompilePatterns.desugar_pattern eff p in
                                     p::ps, body_env ++ penv)
                                ps
-                               ([], env) in
+                               ([], env') in
```
dhil, line 1208:
```suggestion
                               ([], with_effects env eff) in
```

tests/typed_ir.tests
```diff
@@ -69,11 +69,21 @@ tests/typed_ir/T697b.links
...
+exit : 0
+stdout : () : ()
+
+Wild effect compatibility #4 (#873, 2)
+fun () { escape e in { (fun (esc) { esc(()) })(e) } }
+exit : 0
+stdout : fun : forall a::Row.() ~a~> ()
+
+Wild effect compatibility #5 (#873, 3)
+sig foo : () {a}~> () fun foo () { escape e in { (fun (esc) { esc(()) })(e) }}
```
dhil, line 84: What's the point of `a` (`= a : () {}-> ()`) here?

tests/typed_ir.tests
```diff
@@ -69,11 +69,21 @@ tests/typed_ir/T697b.links
...
+exit : 0
+stdout : () : ()
```
dhil, line 87:
```suggestion

Call/cc
sig callcc : (((a) ~> b) ~e~> a) ~e~> a fun callcc(f) { escape k in f(k) }
exit : 0 
stdout : () : ()
```

dhil (changes requested):

tests/typed_ir.tests
frank-emrich: I'm not sure if there's a deeper reason, I've taken the test cases 2 and 3 in verbatim as things @jamescheney experimented with in issue #873.

dhil: I have my own agenda here, I want to disallow `{name}` as syntactic sugar for `{name:()}` in order to fix #975 -- I essentially want to use the syntax `{name}` to mean presence variable in all types (it already does in rows). We will probably discuss this at the next meeting.

dhil (approved):
LGTM, though, I would like the test case to use `{a:()}` rather than `{a}`. However, this can be patched at a later if this is considered "controversial" at this time.

tests/typed_ir.tests
jamescheney: In the examples I wrote,  I think I meant for `a` to be a (row) type variable, anyway.  So this test case could be changed to something like 

```
 sig foo : () { |e}~> () fun foo () { escape e in { (fun (esc) { esc(()) })(e) }}
```

(assuming that in fact works too)

frank-emrich: done, the test works with this modification

PR #981, opened 2021-06-17
[Experiment] Run interpreter testsuite on ubuntu-latest and macos-latest

dhil:
This patch extends the operating system matrix for the interpreter job with `macos-latest`. In addition it changes `ubuntu-20.04` to `ubuntu-latest` since the interpreter should not be sensitive to system dependencies.

### DIFF
--- a/.github/workflows/default.yml
+++ b/.github/workflows/default.yml
@@ -145,17 +145,11 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v2
 
-      - name: Install system dependencies
-        if: matrix.os == 'ubuntu-latest'
-        run: sudo apt-get install -y libev-dev
-        shell: bash
-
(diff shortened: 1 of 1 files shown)


dhil:
It seems that `macos` does not support containers, so we cannot easily add `macos` to the operating system matrix of the database jobs. We would need to define a separate job for testing databases on `macos`.

jamescheney:
Though testing on MacOs would be nice, their switch to custom hardware seems likely to mean that x86 MacOs will be eol within a few years so I'm not sure how much effort is worth putting into this.

dhil:
Hopefully GitHub would provide access to new hardware eventually. At the moment we get "interpreter testsuite on macOS" for free. So perhaps, we just take what we can get for free, and leave database testing for later. We can discuss at the next meeting.

PR #982, opened 2021-06-17
New round-tripping type pretty printer

### LINKED ISSUE #926: Row types should be printed in round-trippable canonical forms
Type inference can result in row types that get printed out in non-parsable ways, for example:

```
links> fun f(tbl) { for (x <-- tbl) where (x.id == 42) [x] };
f = fun : (TableHandle((|id:Int|a::Base),(|_::Base),(|_::Base))) -> [(|id:Int|a::Base)]
```
If we try to actually ascribe the above signature then this doesn't parse:
```
links> sig f : (TableHandle((|id:Int|a::Base),(|_::Base),(|_::Base))) -> [(|id:Int|a::Base)]
***: Parse error: <stdin>:1

  sig f : (TableHandle((|id:Int|a::Base),(|_::Base),(|_::Base))) -> [(|id:Int|a::Base)]
                            ^ 
```
because of the superfluous `|`, whereas this is fine:
```
links> sig f : (TableHandle((id:Int|a::Base),(|_::Base),(|_::Base))) -> [(id:In
...... fun f(tbl) { for (x <-- tbl) where (x.id == 42) [x] };
f = fun : (TableHandle((id:Int|a::Base),(|_::Base),(|_::Base))) -> [(id:Int|a::Base)]
```

HT @vcgalpin

SamuelNovak:
A new pretty printer for types. Uses a string buffer to build the printed string (quasilinear time), this is hidden away under a functional interface (see module `StringBuffer` in `types.ml`).
The aim is that this printer will fix roundtripping issues of the old printer (hence it will fix #926).

The new printer is now the default. The old printer is still present, and the setting that controls which printer is used is `types_pretty_printer_engine` with allowed values:
- `roundtrip` (also allows entering `new`) - this new printer
- `old` - the original printer
- `both` (also allows entering `all`) - invokes both printers, for comparison, e.g.:
  ```links
  links> @set types_pretty_printer_engine both;

  links> fun (x) {x};
  fun
      : Roundtrip: `(a::Any) -> a::Any'
  Old: `(a::Any) -> a::Any'
  Roundtrip and old printer AGREE. 

  links> 1+1;
  2 : Roundtrip: `Int'
  Old: `Int'
  Roundtrip and old printer AGREE. 
  ```
- `none` - no pretty printing, uses the derived printer instead

This PR includes also a couple of new test cases, specifically for testing the problems that this printer fixes. They are mainly in `tests/roundtrippable_types.tests` and `tests/parser.tests` (because some of them were needed also for testing #993).

Note: there is currently nothing done about sugar for shared effects.


core/types.ml
```diff
@@ -2397,6 +2499,785 @@ struct
...
+          | true -> write "@");
+         buffer
+      | _ -> failwith ("Illformed effect:\n" ^ datatype ctx p r)
+    in
+    
+    (* func starts here *)
+    fun ?(is_lolli=false) ctx p domain effects range ->
+    (* dpr' "func"; *)
+    let { buffer; concat; add_buffer; _ } = create_buffer () in
+    let dom_row =
```
dhil, line 3004: You should use the utility functions from `typeUtils.ml` to extract rows, etc.
```suggestion
    let domain = TypeUtils.extract_row domain in
```

dhil:
Great stuff. I can help you polish this PR when you have fixed the remaining outstanding issues (or at least tried to ;-)).

core/types.ml
```diff
@@ -3494,14 +3495,14 @@ let test_type_roundtrip : datatype (* -> tycon_environment *) -> string -> bool
    *    (\* TODO compare if the parsed type is isomorphic to t *\)
    *    true *)
   (* TODO dependency cycle; wanna use DesugarDatatypes.read or parts of it *)
-  true
+  failwith "Testing for type round trip is not implemented."
```
dhil, line 3498: Not here. You should define any testing logic in a separate module.

test-harness
```diff
@@ -97,8 +97,12 @@ def split_arguments(args = []):
             links_args.append(args[i])
     return (links_args, rest_args)
 
-def evaluate(name, code, config_file, stdout='', stderr='', exit = '0', env = None, filemode='', args='', ignore = None):
+def evaluate(name, code, config_file, old_printer='', stdout='', stderr='', exit = '0', env = None, filemode='', args='', ignore = None):
```
dhil, line 100: These changes are fine whilst you are developing, but they must be reverted before this PR can be merged.

SamuelNovak: I reverted this

core/types.ml
```diff
@@ -1711,12 +1711,51 @@ let extract_tuple = function
 (* Some helper functions from typeUtils.ml, to prevent cyclical dependencies *)
 exception TypeDestructionError of string
 
-let extract_row t = match concrete_type t with
+let typeUtils_concrete_type t =
```
dhil, line 1714: You should make `TypeUtils.concrete_type` an alias of this.

dhil (commented):
A few nit comments:
* rename `roundtrippable-types.tests` to `roundtrippable_types.tests`
* remove the file `examples/alias_row_update.links`. I suggest you add the contents to test file, e.g. `roundtrippable_types.tests`.

core/types.ml
SamuelNovak: I'll remove that entirely for now, maybe in the future it would be nice to have though.

tests/roundtrippable_types.tests
```diff
@@ -12,7 +12,10 @@ filemode : true
...
-filemode : true
+fun w(g) { handle(g()) { case Return(()) -> () case F(h, res) -> w(h) }} w
 args : --enable-handlers
-ignore : not roundtrippable
\ No newline at end of file
+stdout : fun : (() { |(mu a . F:(() { |a}~> ()) {}-> _::Any,wild:()|c)}~> ()) {F{_}|c}~> ()
+
+Recursive variant
+sig f : ([|(mu a. Op:() { |a}-> ())|]) {}-> () fun f(_) { () } f
+stdout : fun : ([| |(mu a . Op:() { |a}-> ())|]) {}-> ()
```
dhil, line 21: I don't think this type is entirely correct. The `|` in front of `(mu ...` should not be printed here, I think.

SamuelNovak: It seems internally the `(mu ...)` is in a row variable (at least from the non-pretty printout)
```ocaml
(Types.Function
   ((Types.Record (Types.Row
                     ({1 => (Types.Present
                               (Types.Variant
                                  (Types.Row
                                     ({}, Point @ (...) (Types.Recursive (6642,...

```

SamuelNovak: This is now resolved, variants have special treatment - no pipe for row variable.

SamuelNovak:
TODO Found a broken thing (simulating OOP, example from dhil) (also this is using the extended parser from #993):
```links
links> sig mno : (String) -> (name:String | (mu r . pn:( ((name:String|a)) ~> String )))
...... fun mno(n) { (name=n, pn = fun(r) { r.name }) };

mno = fun : (String) -> (name:String,pn:((name:String|_)) ~> String)
The old printer would say:
(String) -> (name:String|(mu b . pn:((name:String|_)) ~> String))
New and old pretty types agree:
false

links> var o = mno("Aa");
o = (name = "Aa", pn = fun) : (name:String,pn:((name:String|_::(Unl,Mono))) ~> String)
The old printer would say:
(name:String|(mu a . pn:((name:String|_::(Unl,Mono))) ~> String))
New and old pretty types agree:
false
```

SamuelNovak:
The above problem is fixed

dhil:
> TODO Found a broken thing (simulating OOP, example from dhil) (also this is using the extended parser from #993):

OK, I understand this to mean that we have to merge #993 before this PR.

core/types.ml
```diff
@@ -2401,6 +2470,964 @@ struct
...
+      = fun amb pr v ->
+      Printer ("<wrap>", fun ctx () buf ->
+                         let inner_ctx = Context.set_ambient amb ctx in
+                         apply pr inner_ctx v buf)
+
+    let eval : 'a printer -> Context.t -> 'a -> string
+      = if Settings.get trace_printer
+        then
+          (fun pr ctx v ->
+            let buf = create 10 in (* initial size is an arbitrary value here (TODO) *)
```
dhil, line 2824: Make the initial size much bigger, let's arbitrarily say, the maximum length of a tweet: 280. The initial allocation is cheap, subsequent reallocations are more expensive, so you generally want to avoid too many reallocations.

core/types.ml
```diff
@@ -2401,6 +2470,964 @@ struct
...
+         apply lp ctx x buf
+      | Empty, Printer _ ->
+         apply rp ctx y buf
+      | Empty, Empty -> ()
+
+    let constant : string -> unit printer
+      = fun s ->
+      Printer ("<constant>", fun _ctx () buf -> write buf s)
+
+    let wrap : 'a printer -> 'a -> unit printer
```
dhil, line 2810: I prefer `with_value` to `wrap`

core/types.ml
```diff
@@ -2401,6 +2470,964 @@ struct
...
+      = fun pr v ->
+      Printer ("<wrap>", fun ctx () buf -> apply pr ctx v buf)
+
+    let wrap_ambient : Context.ambient -> 'a printer -> 'a -> unit printer
```
dhil, line 2814: `with_ambient`

core/types.ml
```diff
@@ -2401,6 +2470,964 @@ struct
...
+      = fun amb pr v ->
+      Printer ("<wrap>", fun ctx () buf ->
```
dhil, line 2816: "<with_ambient>"

dhil (changes requested):

core/types.ml
```diff
@@ -2445,78 +3472,159 @@ See Note [Variable names in error messages].
...
+        let t' = t in
+
+        let policy = policy () in
+        let t = if policy.quantifiers then t
+                else NewPrint.strip_quantifiers t in
+        build_tyvar_names ~refresh_tyvar_names free_bound_type_vars [t];
+        let context = NewPrint.Context.setup policy Vars.tyvar_name_map (* (fun o -> o#typ t) TODO for shared effects *) in
+        let new_type = NewPrint.string_of_datatype context t in
+        let old_type = begin
+            if Settings.get print_old_new then
```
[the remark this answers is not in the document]
SamuelNovak: This is done, note that I changed the interface `types.mli`, but it compiles so hopefully it's fine.

core/types.ml
```diff
@@ -2445,78 +3472,159 @@ See Note [Variable names in error messages].
...
-    build_tyvar_names ~refresh_tyvar_names free_bound_type_vars [t];
-    let context = Print.context_with_shared_effect policy (fun o -> o#typ t) in
-    Print.datatype context (policy, Vars.tyvar_name_map) t
+    begin
+      if not (Settings.get use_old_type_pp) then
+        (* preserve the original arguments so old printer can be invoked as well *)
+        let policy' = policy in
+        let t' = t in
+
+        let policy = policy () in
```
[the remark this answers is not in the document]
SamuelNovak: This should be correct now.

SamuelNovak:
There is an issue with printing quantifiers even when they should be hidden currently.

core/types.mli
```diff
@@ -5,7 +5,8 @@ open CommonTypes
 type 'a stringmap = 'a Utility.StringMap.t [@@deriving show]
 type 'a field_env = 'a stringmap [@@deriving show]
 
-val print_types_pretty : bool Settings.setting
+type pretty_printer_engine = PPOld | PPRoundtrip | PPAll
```
dhil, line 8: I don't think you need this, the `pretty_printer_engine` can be a private setting to the module. Thinking about it, I am not even convinced that `print_types_pretty` needs to be exposed. I think you can delete it as well.

core/types.mli
```diff
@@ -37,8 +38,11 @@ module Vars : sig
   type vars_list = (int * (flavour * kind * scope)) list
 end
 
+(* TODO return this to original (~S) *)
+type pp_policy = {quantifiers:bool; flavours:bool; hide_fresh:bool; kinds:string; effect_sugar:bool}
```
dhil, line 42: It would be great if you created a module `Policy` for this datatype, e.g.
```ocaml
module Policy = struct
  type t = {quantifiers: bool; .... }
  let default_policy () = ...
end
```

dhil (changes requested):

core/types.mli
SamuelNovak: This is done now

SamuelNovak:
> (TODO) There is an issue with printing quantifiers even when they should be hidden currently.

The above issue is fixed now.

Note that some tests broke because the type variable names generated are different, not sure why this is. The names agree in both printers. I changed the breaking test cases to reflect this.
To illustrate what the above is about (note the only difference is `a` vs. `c`)
```links
expected `fun : () { |(mu a . Op:(() ~> ()) -> ())}~> ()';
     got `fun : () { |(mu c . Op:(() ~> ()) -> ())}~> ()'
```

tests/parser.tests
```diff
@@ -18,25 +18,25 @@ filemode : true
 
 Recursive row variable in variant [1]
 sig f : ([|(mu a. Op:() { |a}-> ())|]) {}-> () fun f(_) { () } f
-stdout : fun : ([||(mu a . Op:() {|a}-> ())|]) {}-> ()
+stdout : fun : ([|(mu a . Op:() { |a}-> ())|]) {}-> ()
 
 Recursive row variable in variant [2]
 sig tsize : ([|( mu t. Node:([|t|],[|t|])|Leaf)|]) ~> Int fun tsize(t) { switch (t) { case Leaf -> 1 case Node(l,r) -> 1 + tsize(l) + tsize(r) } } tsize
-stdout : fun : ([||(mu a . Leaf|Node:([||a|], [||a|]))|]) ~> Int
+stdout : fun : ([|(mu b . Leaf:(),Node:([|b|], [|b|]))|]) ~> Int
```
dhil, line 25: I think this printing is incorrect. It is a recursive row variable in a **variant** row. Thus the fields should be `vfields`, i.e. the separator is `|`.

dhil:
> > (TODO) There is an issue with printing quantifiers even when they should be hidden currently.
> 
> The above issue is fixed now.
> 
> Note that some tests broke because the type variable names generated are different, not sure why this is. The names agree in both printers. I changed the breaking test cases to reflect this.
> To illustrate what the above is about (note the only difference is `a` vs. `c`)
> 
> ```
> expected `fun : () { |(mu a . Op:(() ~> ()) -> ())}~> ()';
>      got `fun : () { |(mu c . Op:(() ~> ()) -> ())}~> ()'
> ```

What if you run the example on master? I think the most plausible explanation is that this patch has introduced a change.

tests/parser.tests
SamuelNovak: You are right, fixing this.

SamuelNovak:
> What if you run the example on master? I think the most plausible explanation is that this patch has introduced a change.

On master, the result is as the original test cases would expect, so it must be something I did.

SamuelNovak:
This was indeed my doing! The original `string_of_datatype` stripped away toplevel quantifiers if `show_quantifiers = false`, but I forgot about this when I rewrote it, and consequently the type var names were built starting with the first quantifier (which should have been hidden).

dhil:
Great! I will give it a final review later today.

dhil (approved):

PR #985, opened 2021-06-21
Fix test-harness silently failing when config file provided in header

frank-emrich:
Commit bacc5e2e930ab7453f6d7ea6bcac48457e3e22c8 introduced a bug into `test-harness` that made all .tests files with a header containing a config file fail silently.

This PR fixes the concrete bug causing the exception in `test-harness` and adds logic for detecting such exceptions in the future (instead of the ignoring them and considering the test as passed).

### DIFF
--- a/test-harness
+++ b/test-harness
@@ -216,18 +216,17 @@ def main():
         # enqueues execution of
         # evaluate(name, code, config_file, **opts)
         # in the thread pool
-        jobs.append(tp.submit(evaluate, name, code, config_file, **opts))
+        jobs.append((name, tp.submit(evaluate, name, code, config_file, **opts)))
 
     # Wait for all jobs to finish
     tp.shutdown(wait=True)
 
     # Check that none of the jobs raised an exception:
-    exceptions = [ j.exception() for j in jobs ]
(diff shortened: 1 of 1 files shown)


test-harness
```diff
@@ -211,11 +216,19 @@ def main():
...
-        tp.submit(evaluate, name, code, config_file, **opts)
+        jobs.append(tp.submit(evaluate, name, code, config_file, **opts))
 
     # Wait for all jobs to finish
     tp.shutdown(wait=True)
 
+    # Check that none of the jobs raised an exception:
+    exceptions = [ j.exception() for j in jobs ]
+    if any(exceptions):
+      raise SystemExit(("Error: One of the test jobs failed to execute, producing"
```
dhil, line 227: I think you should print the name of the job.

dhil (changes requested):

test-harness
frank-emrich: good point, done

dhil (approved):

dhil:
LGTM. Just merge when the CI has completed.

PR #1000, opened 2021-07-03
Surface syntax for Presence type arguments.

### LINKED ISSUE #975: There is no syntax for supplying a presence variable as a type argument
We have syntax for supplying type variables of kind `Type` and `Row` as type arguments, but we have no syntax for variables of kind `Presence`. There used to be syntax support for this, but it was dropped in 2014 for whatever reasons (c.f https://github.com/links-lang/links/blob/master/core/parser.mly#L1004).

Note this issue should be fixed independently of kind inference, that we have previously discussed implementing.

dhil:
This patch adds surface syntax for Presence type arguments. Presence type arguments use the `{<field_spec>}` syntax. Here are some example of the syntax:

```links
typename T(p::Presence) = (foo{p});

(foo=4200) : T({:Int})         # Present with type Int
()         : T({-})            # Absent
(foo=4200) : T({%})            # Unnamed flexible variable
(foo=true) : T({%p})           # Named flexible variable
fun(r : T({_})) { () }         # Anonymous presence variable
fun(r : T({p})) { r : T({p}) } # Named presence variable
```

To resolve the ambiguity between the presence type argument `{wild}`
from the row type argument `{wild}`, I have elected to remove the
syntax sugar for effect and record fields which lets one omit the
`()`. Note however that it is still possible to omit the `()` for
variant fields. The rationale behind this decision is

* We never use the syntax sugar `wild` to mean `wild:()` since we have
 `~>`.
* The `()` in a record field means that the label carries a
 *unit* payload, in an effect field it means `() -> ()`, i.e. a 
 nullary operation which returns unit, whilst `()` in a 
 variant field it means the label carries a *nullary* payload;
 thus despite the similar syntax the three `()` are conceptually 
 different.

As a result this patch introduces a breaking change.

Resolves #975.


jamescheney (commented):
This looks reasonable to me, but I'm not sure I'm knowledgeable enough about this for my "approval" to be meaningful.  

When you say 
> The () in a record field means that the label carries a
> unit payload, in an effect field it means () -> (), i.e. a
> nullary operation which returns unit, whilst () in a
> variant field it means the label carries a nullary payload;
> thus despite the similar syntax the three () are conceptually
> different.

I appreciate that there is a difference in kind between effect and record/variant, but I'm not sure how a variant case with type `()` being "nullary" contrasts with a record field with payload `()`.  In both cases the data associated with the case/field is a unit.  The rationale for treating them differently (which I agree with) is that conventionally in ML/Haskell-like datatype declarations, we omit the type information associated with variant cases that have no associated data, which from a type theoretic point of view correspond to a case with data "unit", but this is purely a notational convenience to save us having to write "of unit" or "()" in many, many places.  (Though I suppose that implementations could treat such cases differently from variants that are explicitly described as "of unit", we are not doing this in Links as far as I know.)  Whereas for records, it is (and probably should be) unusual to have a field with no associated data, so if you really want that, it is reasonable to have to write "foo:()" to document that you are doing something unusual.  

(The more likely scenario is where foo is part of a record type that's abstracted over some type variable, and its type is the abstracted variable.  If we happen to plug in "()" for the abstracted variable, it would then look strange for this type to print out as "(foo,...)" rather than "(foo:(),...)".  I suppose this irregularity could affect variant types as well, though.)

slindley (approved):
This seems like a reasonable and necessary change to me. We should continue to support the syntactic sugar that allows enumerations as variant types, but we shouldn't extend that syntactic sugar to other row types where it leads to ambiguities and gets in the way of presence type arguments.

PR #1005, opened 2021-07-06
Transfer `benchmarks` to links-lang/links-benchmarks

dhil:
This patch moves the directory `benchmarks` from this repository to the new links-lang/links-benchmarks repository, which is intended to collect programs that we can use to benchmark various components of Links.


jamescheney:
It might have been good to discuss this.  The fact that these have type-rotted over time suggests that they might be good to add to the typechecking tests even if we are not running them as part of the test suite.

dhil:
The repository links-lang/links-benchmarks type checks all programs in the repository using the latest released version of Links.

PR #1006, opened 2021-07-07
Effect Sugar: Act 1: Enter the Printer

### LINKED ISSUE #682: Effect sugar does not handle omitted presence variables
As of #669, we can now omit presence variables from effects when the `effect_sugar` flag is enabled. However, the type printer does not handle this, and always prints them out:

```
(Comp((), {Eff:Int|_})) -> Comp((), {|_})
```

 will be printed as

```
(Comp((), {Eff:Int|_})) -> Comp((), {Eff{_}|_})
```

### LINKED ISSUE #986: Effect sugar: conflation of effect variables
The effect sugar erroneously conflates some effect variables in the presence of higher-order functions.

```links
# effect_sugar=true
links> fun f(g)(k)() { k(fun(x) { g(x) }) };
f = fun : ((a::Any) -b-> c::Any) -> (((a::Any) -b-> c::Any) -> f::Any) -> () -> f::Any
```

The four bare `->` arrows are understood to share the same effect variable by the sugar, however, the implicit effect variable on the second arrow is distinct from the (shared) effect variable on the second and fourth arrows. This is clear without the effect sugar enabled.

```links
# effect_sugar=false
links> fun f(g)(k)() { k(fun(x) { g(x) }) };
f = fun : ((a::Any) -b-> c::Any) -> (((a::Any) -b-> c::Any) -e-> f::Any) -> () -e-> f::Any
```

Here both `->` hides (fresh) effect variables, which are different from `e`.

### LINKED ISSUE #992: Extended effect sugar
In Links a row is closed if it does not end in a row variable. Setting `effect_sugar=true` allows the programmer to elide effect variables in open empty effect rows, e.g.

```links
sig apply : ((a) -> b, a) -> b
fun apply(f,x) { f(x) }
```
really means
```links
sig apply : ((a) -e-> b, a) -e-> b
fun apply(f,x) { f(x) }
```

For nonempty effect rows the shared effect variable is written using "wildcard syntax"

```links
sig apply : ((a) {Op:() |_}-> b, a) {Op:() |_}-> b
fun apply(f, x) { f(x) }
```

Here, `_` denotes the name of shared effect variable, rather, than two distinct names. From the point of view of the programmer writing `|_` rather than, say, `|e` does not really buy much. Instead we should change the default interpretation of effect rows to be *open* rather than closed as it is today such that

```links
sig apply : ((a) {Op:()}-> b, a) {Op:()}-> b
fun apply(f, x) { f(x) }
```

really means

```links
sig apply : ((a) {Op:() |e}-> b, a) {Op:() |e}-> b
fun apply(f, x) { f(x) }
```

for fresh `e`. We would require some new syntax to denote a closed effect row, e.g.
```links
sig apply : ((a) {Op:() |.}-> b, a) {Op:() |.}-> b
fun apply(f, x) { f(x) }
```

Note this change would be for effect rows only, it would not change the behaviour of record, variant, or session rows. I think there may be some intricacies with effect rows in type applications that require some thinking.

SamuelNovak:
This is the printer part of the effect sugar fix/enhancement project. It fixes a few bugs, creates some enhancements and introduces multiple fine-grained options for the Roundtrip printer, so that we can set what we want the effect sugar to be.

The relevant setting is `effect_sugar_policy` and it can take any combination of the options below, separated by commas (`,`) and enclosed by quotes (`"..."`). Example usage:

```
@set effect_sugar_policy "presence_omit,alias_omit,contract_operation_arrows"
```

The documentation of these options is also available by using

```
@help effect_sugar_policy;
```

---

The available options are:

- **`presence_omit`** [shortcut `pres`] - hide presence-polymorphic operations in open effect rows

  ```links
  sig handleE : (() {E:() {}-> ()|e}~> ()) -> (() {E{_}|e}~> ())
                                                  ^^^^^^^^
  fun handleE(f) {
    fun () {
      handle(f()) { case E(resume) -> resume(()) }
    }
  }
  ⇝
  handleE : (() {E:() {}-> ()|e}~> ()) -> () { |e}~> ()
                                             ^^^^^
  ⇝ (this is on an arrow, so it reduces further to)
  handleE : (() {E:() {}-> ()|_}~> ()) -> () ~> ()
                                             ^^
  ```

- **`alias_omit`** [shortcut `alias`] - hide empty (or emptied using `presence_omit` above) effect rows in the last argument of type alias applications

  ```links
  typename Ord = [| Lower | Greater | Equal |];
  typename Order(a, e::Eff) = (a, a) ~e~> Ord;

  sig ord : Order(a, { |_})
                     ^^^^^
  fun ord(x, y) { if (x == y) Equal else if (x < y) Lower else Greater }
  ⇝
  ord : Order(a)
             ^^^
  ```

- **`arrows_show_implicit_effect_variable`** [shortcut `show_implicit`] - display the implicit shared effect on function arrows, and hide fresh effect variables instead. See examples from #986 at the bottom.

  ```links
  # without `arrows_show_implicit_effect_variable`:

  sig map : ((a) ~e~> b, [a]) ~e~> [b]
                 ^^^^         ^^^^
  fun map(f, lst) { switch(lst) { case [] -> [] case (x::xs) -> f(x) :: map(f, xs) }}
  ⇝
  map : ((a) ~> b, [a]) ~> [b]
             ^^         ^^

  sig map' : ((a) ~e~> b) -> ([a]) ~e~> [b]
                  ^^^^             ^^^^
  fun map'(f)(lst) { switch(lst) { case [] -> [] case (x::xs) -> f(x) :: map'(f)(xs) }}
  ⇝
  map' : ((a) ~> b) -_-> ([a]) ~> [b]
              ^^    ^^^^         ^^
                    ||||
                    (fresh variables appearing; note that in this particular case of
                     a curried function, the next setting
                     `arrows_curried_hide_fresh` may take effect,
                     however, in general, fresh vars will appear)

  ⇝ (with `arrows_show_implicit_effect_variable`)
  map : ((a) ~b~> c, [a]) ~b~> [c]
             ^^^^         ^^^^
  map' : ((a) ~b~> c) -> ([a]) ~b~> [c]
              ^^^^    ^^       ^^^^
  ```

- **`arrows_curried_hide_fresh`** [shortcut `chf`] - in curried functions, the "collection" arrows (function being partially applied to arguments, collecting them) are assumed to have fresh effect variable, and these are not shown. See examples from #986 at the bottom.

  ```links
  # see above, without `arrows_show_implicit_effect_variable`

  # without `arrows_curried_hide_fresh`:
  map' : ((a) ~> b) -_-> ([a]) ~> [b]
                    ^^^^
  ⇝ (action of `arrows_curried_hide_fresh`)
  map' : ((a) ~> b) -> ([a]) ~> [b]
                    ^^
  ```

- **`contract_operation_arrows`** [shortcut `contract`] - contracts operation arrows in effect rows in the following cases:

  ```links
  E:(a) {}-> b    ⇝ E:(a) -> b
        ^^^^              ^^
  E:() {}-> a     ⇝ E:a
    ^^                ^


  sig f : () {E:() {}-> a}-> a
  fun f() { do E };
  ⇝
  f : () {E:a}-> a
  ```

- **`open_default`** [shortcut `open`] - **effect** rows are open by default; uses `{fields| .}` to mean a closed row

  ```links
  # without `open_default`
  sig f : () {E:() {}-> a}-> a
             ^^^^^^^^^^^^^
  fun f() { do E }

  sig g : () {E:() {}-> a|_}-> a
             ^^^^^^^^^^^^^^^
  fun g() { do E }

  f : () {E:() {}-> a}-> a
               ^^    ^
  g : () {E:() {}-> a|_}-> a
               ^^    ^^^

  ⇝ (applying `open_default`)
  f : () {E:() { | .}-> a| .}-> a
                 ^^^^    ^^^^
  g : () {E:() { | .}-> a}-> a
                 ^^^^    ^

  # this works well together with arrow contractions
  # (see the options above)

  ⇝ (applying both `open_default` and `contract_operation_arrows`)
  f : () {E:() -> a| .}-> a
                   ^^^^
  g : () {E:() -> a}-> a
                   ^
  ```

Apart from the above, there are meta-options:

- **`all`** - enables all of the options
- **`none`** - disables all of the options
- **`default`** - resets the setting to default, which is currently:
  - `presence_omit`
  - `alias_omit`
  - `contract_operation_arrows`
  - `arrows_curried_hide_fresh`

Note that when setting `default`, you still need to surround it in quotes as below, otherwise Links complains:

```
@set effect_sugar_policy "default";
```

The names are intentionally long and descriptive, but they also have shortcuts for convenient entering, for example:

```
@set effect_sugar_policy "pres,chf,alias,open";
≡
@set effect_sugar_policy "presence_omit,arrows_curried_hide_fresh,alias_omit,open_default";
```

### TODO:
- [x] Breaking tests in `tests/handlers.tests`
- [x] Breaking tests in `tests/parser.tests`
- [x] Breaking tests in `tests/patterns.tests`
- [x] Breaking tests in `tests/recheck_frontend_transformations.tests`
- [x] Breaking tests in `tests/typecheck_examples.tests`
- [x] Introduce a shortcut for `arrows_show_implicit_effect_variable`

---

This PR fixes #682 (with the option `presence_omit` is enabled).

It will also close #992 (option `open_default`, implemented both in printer and parser in this PR).

---

It fixes #986 as well, where the options `arrows_show_implicit_effect_variable` [`show_implicit`] and `arrows_curried_hide_fresh` [`chf`] can be used to specify what we want to see, or what the assumptions are:

Using the example from #986:
```links
fun f(g)(k)() { k(fun(x) { g(x) }) };
```

This will print as:

- Without the above settings:
  ```links
  f = fun : ((a::Any) -b-> c::Any) -_-> (((a::Any) -b-> c::Any) -> f::Any) -_-> () -> f::Any
  ```
  Collector arrows are not assumed fresh, and so fresh variables will be printed explicitly. The implicit shared effect is hidden.

  The difference between the fresh effect variables and the implicit shared effect is visible here.

- `arrows_curried_hide_fresh` [`chf`]:
  ```links
  f = fun : ((a::Any) -b-> c::Any) -> (((a::Any) -b-> c::Any) -> f::Any) -> () -> f::Any
  ```
  Collector arrows are assumed fresh, and hence do not show up (here by assume I mean that the user assumes this, and hence does not need to see the implicit vars). Implicit shared effect still hidden.

  Here, the difference between the fresh and the implicit shared effect variables is hidden by syntax.

- `arrows_show_implicit_effect_variable` [`show_implicit`]:
  ```links
  f = fun : ((a::Any) -b-> c::Any) -> (((a::Any) -b-> c::Any) -e-> f::Any) -> () -e-> f::Any
  ```
  The implicit shared effect is made explicit: all of its occurences are shown. Collector arrows are not assumed fresh; however, since the implicit shared effect is visible, all fresh variables can be safely hidden.

  This case also differentiates between the fresh effects and the implicit (in this case explicit) shared one.

- both: `show_implicit,chf`:
  ```links
  f = fun : ((a::Any) -b-> c::Any) -> (((a::Any) -b-> c::Any) -e-> f::Any) -> () -e-> f::Any
  ```
  This case works the same way as the above (`show_implicit`).


SamuelNovak:
TODO: 
- [ ] fix tests
- [ ] create new tests for the sugar

examples/handlers/sugar_examples.links
```diff
@@ -0,0 +1,27 @@
+# Effect sugar examples:
```
dhil, line 1: These should be added as tests rather than examples.

core/types.mli
```diff
@@ -37,12 +37,28 @@ end
...
+    type t = opt list
+    val default : unit -> t
+
+    val presence_omit              : t -> bool
+    val alias_omit                 : t -> bool
+    val arrows_show_the_one        : t -> bool
+    val arrows_collection_fresh    : t -> bool
+    val contract_operation_arrows  : t -> bool
+    val open_default               : t -> bool
+    val different_operation_arrows : t -> bool
```
dhil, line 52: We don't want this option now as there is no surface syntax to match the printing. I want to maintain the invariant that anything we print is parseable.

core/types.ml
```diff
@@ -1995,28 +1999,144 @@ module Policy = struct
...
+           ; "   argument of aliases"
+           ; " * arrows_show_the_one_effect: display the imlicit shared effect"
+           ; "   on arrows"
+           ; " * arrows_curried_collection_assume_fresh [ccf]: in curried"
+           ; "   functions, collection arrows are assumed to have fresh"
+           ; "   effects and these are hidden"
+           ; " * contract_operation_arrows [contract]: contract operations"
+           ; "   `E:() {}-> a' to `E:a'"
+           ; " * open_default [|.}]: effect rows are open by default,"
+           ; "   closed with syntax { |.}"
```
dhil, line 2086: Have you implemented the closing syntax?

dhil: I don't see it as part of this PR.

dhil (changes requested):

core/types.ml
SamuelNovak: It is implemented on the printer side, but not in the parser. I'll disable it for now (keeping the code, I'll just make the option unaccessible), and then reenable it when parser support is ready.

SamuelNovak: I have added the closing syntax into this PR (patched the Lexer, Parser and the desugaring pass).

core/types.mli
SamuelNovak: Created the syntax here.

core/lexer.mll
```diff
@@ -181,6 +181,7 @@ rule lex ctxt nl = parse
   | '_'                                 { UNDERSCORE }
   | '='                                 { EQ }
   | "->"                                { RARROW }
+  | "->>"                               { RRARROW }
```
dhil, line 184: I don't want to introduce a new arrow. Let's remove this.

core/lexer.mll
```diff
@@ -212,6 +213,7 @@ rule lex ctxt nl = parse
   | "||"                                { BARBAR }
   | "&&"                                { AMPAMP }
   | '|'                                 { VBAR }
+  | "|."                                { VBARDOT }
```
dhil, line 216: I don't want to introduce a new token. Let's remove this. I want `.` to be allowed in `evar` position in the parser, if the effect sugar is on.

core/sugarTraversals.ml
```diff
@@ -104,6 +104,7 @@ class map =
       let open Datatype in
       function
       | Closed -> Closed
+      | DotClosed -> DotClosed
```
dhil, line 107: Should not be necessary.

core/parser.mly
```diff
@@ -1068,15 +1068,27 @@ vfield:
 
 efields:
 | fields_def(efield, row_var, kinded_row_var)                  { $1 }
+| DOT                                                          { ( [] , Datatype.DotClosed) }
```
dhil, line 1071: I don't think this is correct. The `.` should only be allowed to appear in a row variable position, e.g. the productions `row_var` and `kinded_row_var` should allow `.` and interpret it as `Datatype.Closed`.

dhil (changes requested):
Let's not introduce new tokens.

core/parser.mly
SamuelNovak: How about `(a) {.}-> b`, i.e. a function that admits no effects at all?

```links
sig f : (Int, Int) {.}-> Int
fun f(x, y) { x + y }
```
Arguably this is less useful, but if it can happen (is a legal type), I think it should be supported.

We can have this type without the `{.}` syntax, but then it has to be `{ |.}`.

core/lexer.mll
SamuelNovak: Without this token, the parser doesn't recognize the dot syntax without a space:
```links
links> sig f : () {Ask:() -> a| .}-> a
...... fun f() { do Ask };
f = fun : () {Ask:a|.}-> a

links> sig f : () {Ask:() -> a|.}-> a
***: Parse error: <stdin>:1

  sig f : () {Ask:() -> a|.}-> a
                           ^ 
```

I think it's getting confused with something else, but introducing a token creates the correct behaviour. Unless the problem is actually caused by something else?

core/lexer.mll
SamuelNovak: Removed both from printing and parsing.

core/sugarTraversals.ml
SamuelNovak: This is now removed.

core/parser.mly
SamuelNovak: `{.}` syntax removed

core/lexer.mll
SamuelNovak: As discussed, I created a special case for this in the `op` rule, and kept it (at least for now) as a token - this seems to be a clean solution.

dhil:
Following #1000 the following is no longer a valid rewrite
```
E:() {}-> ()    ⇝ E
```

SamuelNovak:
> Following #1000 the following is no longer a valid rewrite
> 
> ```
> E:() {}-> ()    ⇝ E
> ```

You're right! I forgot to change it in the description, but the printer handles it correctly.
@dhil

core/lexer.mll
dhil: I have thought a little bit more about this. I think we should keep things simple for now and disallow `|.` and it instead insist you write `| .` for a closed row. I don't envisage closed rows being used at all when the effect sugar is on. I am sort of concerned about this special case, as I think this needs to be replicated elsewhere, in particular when I land the hygienic names patch. So lets remove the token again, and simply the grammar.

core/types.ml
```diff
@@ -2646,8 +2858,449 @@ module RoundtripPrinter : PRETTY_PRINTER = struct
...
+             let (o, _) = o#effect_row e in
+             let (o, _) = o#typ r in
+             (o, tp)
+          | Alias ((_,kinds,tyargs,_), _)
+            | RecursiveApplication { r_quantifiers = kinds; r_args = tyargs ; _ } ->
+             (o#alias_recapp kinds tyargs, tp)
+          | _ -> super#typ tp
+      end
+
+    type row_omis = Omissible of row
```
dhil, line 3068:
```suggestion
    type row_erasability = Erasable of row
```

core/types.ml
```diff
@@ -2646,8 +2858,449 @@ module RoundtripPrinter : PRETTY_PRINTER = struct
...
+                  | NonOmissible of row
```
dhil, line 3069:
```suggestion
                  | Indelible of row
```

core/types.ml
```diff
@@ -2825,11 +3500,20 @@ module RoundtripPrinter : PRETTY_PRINTER = struct
       Printer (fun _ctxt prim buf ->
           StringBuffer.write buf (Primitive.to_string prim))
 
-  let is_var_anonymous : Context.t -> tid -> bool =
-    fun ctxt vid ->
-    let _, (_, _, count) = Vars.find_spec vid (Context.tyvar_names ctxt) in
-    (count = 1 && (Policy.hide_fresh Context.(policy ctxt))) (* we want to hide it *)
-    && not (Context.is_tyvar_bound vid ctxt) (* and it is not bound (if it is bound, it has to show up *)
+  type var_anonymity = Visible | Anonymous | ImplicitEff
```
dhil, line 3503:
```suggestion
  type var_anonymity = Visible | Anonymous | ImplicitEffectVar
```

core/types.ml
```diff
@@ -3018,20 +3715,27 @@ module RoundtripPrinter : PRETTY_PRINTER = struct
                 | Var _ | Recursive _ | Absent | Present _ -> ()
                 | _ -> StringBuffer.write buf ":"
               in
-              let ctx' =
-                if Context.is_ambient_type_arg ctx then ctx
-                else Context.(set_ambient Presence ctx)
-              in
-              Printer.apply (meta pt) ctx'  () buf
+              (* TODO (merge conflict resolution) check this; I
```
dhil, line 3718: Has this todo been completed?

core/types.ml
```diff
@@ -3018,20 +3715,27 @@ module RoundtripPrinter : PRETTY_PRINTER = struct
...
+                 removed the ambient change to pass the original
+                 ambient through for effect row printing *)
+              Printer.apply (meta ctx pt) ctx () buf
            | _ -> raise tag_expectation_mismatch))
 
-  and meta : typ point -> unit printer
+  (* TODO (merge conflict resolution) check this: has to do with |.} *)
```
dhil, line 3724: Done?

core/types.ml
```diff
@@ -3086,6 +3791,29 @@ module RoundtripPrinter : PRETTY_PRINTER = struct
...
+                | Anonymous -> arrows_show_impl_shared ||
+                                 (arrows_collection_fresh && Context.is_ambient_arrow_curried ctx)
+                | ImplicitEff -> not arrows_show_impl_shared
+              end
+            else match anonymity with
+                 | Anonymous -> true (* skip *)
+                 | Visible   -> false (* no skip *)
+                 | _         ->
+                    (* this is not supposed to happen if there is no shared effect variable *)
+                    raise tag_expectation_mismatch
```
dhil, line 3815: raise `internal_error` here. In general `tag_expectation_mismatch` should only be used for things that match over the structure of types.

core/types.ml
```diff
@@ -3094,44 +3822,65 @@ module RoundtripPrinter : PRETTY_PRINTER = struct
...
+                  StringBuffer.write buf "{";
+                  Printer.apply row_parts (Context.set_ambient Context.Effect ctx) r' buf;
+                  StringBuffer.write buf "}";
                 end;
               (if is_wild then StringBuffer.write buf "~" else StringBuffer.write buf "-");
-              (* add the arrowhead/lollipop *)
-              (if is_lolli then StringBuffer.write buf "@" else StringBuffer.write buf ">")
+              (* add the arrowhead/lollipop/oparrow *)
+              (if is_lolli then StringBuffer.write buf "@"
+               else (* if Policy.EffectSugar.different_operation_arrows (Policy.es_policy (Context.policy ctx))
```
dhil, line 3880: Delete comment.

core/types.ml
```diff
@@ -3278,19 +4043,52 @@ module RoundtripPrinter : PRETTY_PRINTER = struct
...
+                                     (with_tyvar_names names (empty ())))in
+               let no_sugar_tp = Printer.generate_string datatype tctx ty
+               in
+                 SharedEffect.ensugar_datatype
+                   (Policy.es_policy policy') shared_effect ty
+                   no_sugar_tp
+             else ty
+    in
+    let ret = Printer.generate_string datatype ctxt ty in
+    (* (if effect_sugar
```
dhil, line 4067: Delete comment.

core/types.ml
```diff
@@ -227,8 +228,12 @@ let is_field_spec_body = is_type_body
 module Transform : TYPE_VISITOR =
 struct
   class visitor = object ((o : 'self_type))
-    val rec_vars : (meta_type_var) IntMap.t = IntMap.empty
 
+    (* (Samo) need to disable this sometimes; hacky solution *)
```
dhil, line 232: What hack?

dhil (changes requested):
Looks mostly good to me. One nitpick thing: you change terminology between "implicit" and "shared" effect variable. I suggest you pick one of them, and stick to it consistently.

core/types.ml
```diff
@@ -1995,28 +2009,197 @@ module Policy = struct
...
+  module EffectSugar : sig
+    type opt = PresenceOmit
+             | AliasOmit
+             (* vvv TODO name; for now it's the One Effect To Rule Them All *)
+             | ArrowsShowTheOneEffect
+             | ArrowsCurriedCollectionAssumeFresh
+             | ContractOperationArrows
+             | OpenDefault
+             (* | DifferentOperationArrows *)
+             | FinalArrowSharesWithAlias
```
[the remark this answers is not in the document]
SamuelNovak: This is handled by #1032. In your example, this setting would cause the type to desugar into
```links
f = fun : (Comp(a, { |e})) -e-> Comp(a, { |e})
                      ^-----^--------------^
```

The printer will handle each case correctly. I've included the setting in this PR so that it's compatible with #1032.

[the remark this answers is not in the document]
dhil: Hmm. This is not great. It is problematic that the arrow gets endowed with e.  I think we may need to think more carefully about this case.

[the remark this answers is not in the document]
SamuelNovak: I can disable this setting in this PR and then just reenable it in the other one.
The idea behind this is
```links
sig upd_map : (a, b, Map(a,b)) ~> Map(a,b)
# which would mean
sig upd_map : (a, b, Map(a,b,{ |e})) ~e~> Map(a,b,{ |e})
```
where `Map` contains a function that is relevant to its transformation, and so its effect has to be shared with the arrows as well.

core/types.ml
SamuelNovak: I believe that is now fully resolved

core/types.ml
SamuelNovak: The visitor sometimes needs to keep the original tyvars, and sometimes in needs to refresh them (having only one or the either broke either sugar or the unifier (I think due to decycling)).

Since I don't know exactly why the issues happened, I created this value in the class, and I'm just setting it where I need it to be different from default. It felt hacky at the time (mainly because I didn't understand what was going on), though in hindsight this is probably clean enough.

core/types.ml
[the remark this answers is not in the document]
dhil: Hmm. We need to be careful here. It is often incorrect to assume that the last arrow and the alias should share the effect variable, though, sometimes it is correct. We need to think more carefully about a robust design.

core/types.ml
```diff
@@ -1995,28 +2009,197 @@ module Policy = struct
...
-           ; flavours     : bool
-           ; hide_fresh   : bool
-           ; kinds        : kind_policy
-           ; effect_sugar : bool }
+  module EffectSugar : sig
+    type opt = PresenceOmit
+             | AliasOmit
+             (* vvv TODO name; for now it's the One Effect To Rule Them All *)
+             | ArrowsShowTheOneEffect
+             | ArrowsCurriedCollectionAssumeFresh
```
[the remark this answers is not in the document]
SamuelNovak: How about CurriedCollectionFresh? It doesn't imply that a transformation will happen (as this setting is for now just about an assumption to apply when printing). I also wanted to keep it clear that this is about the arrows where arguments are just being collected (this may not be necessary though).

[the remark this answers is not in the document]
dhil: The suggested name means absolutely nothing to me. It is not clear what it's semantics is.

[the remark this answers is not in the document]
SamuelNovak: I'd settle for `ArrowsCurriedHideFresh`, this seems it should be clear enough. I'll update the PR description too.

dhil:
I will look at this again after I have finished reviewing/merging #1028.

SamuelNovak:
I've removed the presets (from meta-options), because they are irrelevant for this PR.
In fact, they also became irrelevant for #1032, which was their original motivation.

dhil (approved):

PR #1009, opened 2021-07-10
Fix error with global let bound anonymous functions

### LINKED ISSUE #353: Global let bound anonymous functions are not compiled correctly to the client
The JavaScript compiler produces meaningless code for global let-bound anonymous functions. The following minimal example illustrates the problem
```
var f = fun() { () };

fun mainPage(_) {
  page
   <#>
    <button l:onclick="{f()}">Click me!</button>
   </#>
}

fun main() {
  addRoute("/", mainPage);
  servePages()
}

main()
```

The global let binding `f` gets compiled as
```
function _fun__g131_2064(__kappa) { _yieldCont(__kappa, {}) }
var f_2065 = {"func":"_2064", "location":"unknown"};
function _fun__g132_2068(_2067, __kappa) { var event_2066 = _2067; _yield(function () { f_2065(__kappa) }) }
```
Clearly, `f_2065` is not a function.

This problem happens irrespective of whether handlers are enabled.

If the binding is localised in for example `mainPage`, then the code generator emits the correct code.

ChefYeum:
Fixes #353


lib/js/jslib.js
```diff
@@ -761,65 +759,60 @@ const LINKS = new (function() {
...
+   * environment:e }, where the environment is optional. If an
+   * environment is specified, we assume that the function denoted by
    * f is actually a wrapper and that f(e) is the desired function.
    * Without an environment, f itself denotes the desired function, a
    * standard CPS compiled Links function. This is recursive, so each
    * object in `obj` also has its functions resolved.
-   *
+   * 
    * @param {any} state
    * @param {any} obj
```
dhil, line 786:
```suggestion
   * @param {any} obj
   * @return {any} obj
```

lib/js/jslib.js
```diff
@@ -761,65 +759,60 @@ const LINKS = new (function() {
...
     }
     return;
   }
 
 
   /**
-   * Resolve function references in the object `obj`, specified as records
-   * { function:f, environment:e }, where the environment is optional. If
-   * an environment is specified, we assume that the function denoted by
+   * Resolve function references in the object `obj` and return the
```
dhil, line 776: You should note here that the resolution process fails for deeply nested structures.

dhil (changes requested):
LGTM. Have you checked that the examples page still works?

lib/js/jslib.js
dhil: BTW you can fix this problem in a follow up PR.

dhil (approved):
LGTM.

lib/js/jslib.js
dhil: I have added the comment.

SimonJF:
FYI, `git bisect` tells me that this breaks Commands in the MVU library (see, e.g., `examples/mvu/time.links`). I get the error "`env_3442_3596.3432` is not a function".

I can look at a fix, but also happy if someone with more knowledge gets there first :)

dhil:
Hi Simon, yes this is known. @ChefYeum is working on the fix.

Issue #1011, opened 2021-07-12
Regression: progress.links causes the webserver to respond with 500 Internal Error


dhil:
Input a positive natural number `n` into the input field and subsequently click `Submit`, then the server responds with `n` "500 Internal Error" messages. Curiously, the progress bar example completes, so the errors seem to only be observable if the debugging console is open. I have only tested with Chrome v91.0.4472.114.

dhil:
The issue occurs in Firefox 89.0.2 too.

dhil:
The problem was either introduced or exposed by patch #947 since after removing it the server responds with OK.

dhil:
It seems the problem occurs because the process on the server-side throws `Aborted`. It seems that every client call invokes abort on its process. I am not sure, whether that is sensible, but we can discuss that design at another time. The way I see it there are two possible corrective measures for this issue here: 1) we can special case `Aborted` in `handle_error`, or 2) push the `set_http_response_code` into the error branches.

PR #1012, opened 2021-07-12
Default HTTP response when process throws `Aborted`.

dhil:
This patch fixes issue #1011. Patch #947 modified the response code
for error messages to be 500. Every client call aborts its server
process, and when a process aborts it throws the exception `Aborted`,
meaning it would inadvertently set the response code to 500. In the
long term we should probably reconsider the design of processes,
remote calls, etc.

### DIFF
--- a/core/webif.ml
+++ b/core/webif.ml
@@ -182,7 +182,9 @@ struct
           Lwt.return (mime_type, error_page (Errors.format_exception_html exc)) in
 
     let handle_exception = function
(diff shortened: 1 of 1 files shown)


jamescheney (approved):
This looks reasonable, but I wonder if it would be clearer to call `handle_error` something like `handle_exception` instead, since `Aborted` is an exception that does not signal an error.

PR #1015, opened 2021-07-14
add tag to distinguish the structure of the serialised JSON

### LINKED ISSUE #996: Tag server-side values for easy client resolution
As pointed out in #353 the client side resolver for server side values is completely broken. A related problem is that "jsonised" server-side values are not easily identifiable. Some inference over JSON object is required to determine the Links value, however, in some instances it is **impossible** to reliably determine what kind of value the server sent, for example some records are indistinguishable from the special object with event handlers. It would be good to be more principled about serialisation of server-side values. A first step would be to include in each serialised object a tag, which describes whether the object at hand is a record, variant, etc.

ChefYeum:
fixes #996

### DIFF
--- a/core/json.ml
+++ b/core/json.ml
@@ -63,9 +63,9 @@ let json_of_lens (db, lens) : Yojson.Basic.t =
 let jsonize_location loc = `String (Location.to_string loc)
 
 let rec cons_listify : Yojson.Basic.t list -> Yojson.Basic.t = function
-  | [] -> `Null
+  | [] -> `Assoc [("_tag", `String "List")]
   | x::xs ->
-      `Assoc [("_head", x); ("_tail", cons_listify xs)]
+      `Assoc [("_head", x); ("_tail", cons_listify xs); ("_tag", `String "List")]
 
 let rec jsonize_value' : Value.t -> Yojson.Basic.t =
   function
@@ -76,6 +76,7 @@ let rec jsonize_value' : Value.t -> Yojson.Basic.t =
   | `Socket _
       as r ->
       raise (Errors.runtime_error ("Can't jsonize " ^ Value.string_of_value r));
+
   | `FunctionPtr (f, fvs) ->
     let (_, _, _, location) = Tables.find Tables.fun_defs f in
     let location = jsonize_location location in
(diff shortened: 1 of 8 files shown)


lib/js/jslib.js
```diff
@@ -800,21 +800,76 @@ const LINKS = new (function() {
...
+          '_tail': resolveServerValue(state, obj._tail)
+        };
+      
+      case "ClientAccessPoint":
+      case "ServerAccessPoint":
+      case "ClientPid":
+      case "ServerPid":
+      case "SessionChannel":
+      case "ServerSpawnLoc":
+      case "ClientupawnLoc":
```
dhil, line 857:
```suggestion
      case "ClientSpawnLoc":
```

lib/js/jslib.js
```diff
@@ -800,21 +800,76 @@ const LINKS = new (function() {
...
+          value[k] = resolveServerValue(state, obj[k]);
+        }
+        return newObj;
+        
+      case "List":
+        if (obj._head === null) {
+          return {} // check
+        }
+        return {
+          '_head': resolveServerValue(state, obj._head),
```
dhil, line 847: Use `LINKEDLIST.Cons(resolveServerValue(state, obj._head), resolveServerValue(state, obj._tail))` here.

lib/js/jslib.js
```diff
@@ -800,21 +800,76 @@ const LINKS = new (function() {
...
+      
+      case "Char":
+        delete obj._tag; 
+        return obj;
+
+      case "FunctionPtr":
+      case "ClientFunction":
+        const f = (!TYPES.isObject(obj.environment)) ?
+          eval(obj.func) :
+          partialApply(eval(obj.func), eval(obj.environment));
```
dhil, line 822: I think `environment` needs to be resolved. It can contain function pointers.

lib/js/jslib.js
```diff
@@ -800,21 +800,76 @@ const LINKS = new (function() {
...
+      case "ServerPid":
+      case "SessionChannel":
+      case "ServerSpawnLoc":
+      case "ClientupawnLoc":
+        // TODO: should they return a new object?
+        delete obj._tag;
+        return obj;
+
+      default:
+        delete obj._tag;
```
dhil, line 863: Let's move the default case to the top, e,g.
```javascript
    function resolveServerValue (state, obj) {
        if (obj._tag == null && obj.key != null) {
            obj.key = _lookupMobileKey(state, resolveServerValue(null, obj.key))
            return obj;
        }

        switch (obj._tag) {
        case "Bool": ....
```

lib/js/jslib.js
```diff
@@ -800,21 +800,76 @@ const LINKS = new (function() {
...
+        if (obj.key) {
+          obj.key = _lookupMobileKey(state, obj.key)
+          return obj;
+        } else {
+          for (let k in obj) {
```
dhil, line 868: This branch should be deleted.

dhil (changes requested):

dhil:
We need to make a few more changes to `jslib.js`, specifically all of the call sites of `resolveServerValue`.

lib/js/jslib.js
```diff
@@ -800,21 +800,76 @@ const LINKS = new (function() {
...
+        var value = {};
+        var newObj = { '_label': obj._label, '_value': value};
+
+        for (let k in obj._value) {
+          value[k] = resolveServerValue(state, obj[k]);
+        }
+        return newObj;
+        
+      case "List":
+        if (obj._head === null) {
```
ChefYeum, line 843: Also needs to cover `undefined`

lib/js/jslib.js
```diff
@@ -800,21 +800,76 @@ const LINKS = new (function() {
...
+      case "ClientPid":
+      case "ServerPid":
+      case "SessionChannel":
+      case "ServerSpawnLoc":
+      case "ClientSpawnLoc":
+        // TODO: should they return a new object?
+        delete obj._tag;
+        return obj;
+
+      default:
```
dhil, line 862: Lets do something like
```javascript
        default:
            throw "Untagged server value \"" + JSON.stringify(obj) + "\"";
```

lib/js/jslib.js
ChefYeum: `_lookupMobileKey` is no longer used in the function. Is that fine?

lib/js/jslib.js
ChefYeum: Added, I went with "Mistagged"

lib/js/jslib.js
```diff
@@ -843,7 +843,7 @@ const LINKS = new (function() {
         return newObj;
 
       case "List":
-        if (obj._head === null) {
+        if (obj._head === null || obj._head === undefined) {
```
ChefYeum, line 846: Explicit condition for `undefined`

core/json.ml
```diff
@@ -63,9 +63,9 @@ let json_of_lens (db, lens) : Yojson.Basic.t =
 let jsonize_location loc = `String (Location.to_string loc)
```
dhil, line 64: On line 57 you need to add `("_tag", "Lens")`

lib/js/jslib.js
```diff
@@ -800,23 +800,70 @@ const LINKS = new (function() {
...
+          resolveServerValue(state, obj._tail)
+        )
+
+      case "ClientAccessPoint":
+      case "ServerAccessPoint":
+      case "ClientPid":
+      case "ServerPid":
+      case "SessionChannel":
+      case "ServerSpawnLoc":
+      case "ClientSpawnLoc":
```
dhil, line 861: Add a `"Lens"` case to this list of cases.

lib/js/jslib.js
```diff
@@ -800,23 +800,70 @@ const LINKS = new (function() {
...
+        // TODO: should they return a new object?
```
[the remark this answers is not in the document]
dhil: Having thought about it, I think it is best we mutate the given object. Ultimately, I would like to return a new object, but some of the code in `jslib.js` may depend on some of the side effects caused by `resolveServerValue`. Therefore I suggest we initially mutate each object until we have investigated each call site of `resolveServerValue`.

lib/js/jslib.js
```diff
@@ -800,23 +800,70 @@ const LINKS = new (function() {
...
-      return obj;
-    } else {
-      for (let key in obj) {
-        obj[key] = resolveServerValue(state, obj[key]);
-      }
-      return obj;
-    }
+    switch (obj._tag) {
+      default:
+        throw new Error("Mistagged server value \"" + JSON.stringify(obj) + "\"");
```
dhil, line 805:
```suggestion
        throw new Error("Unrecognised or missing tag on server value \"" + JSON.stringify(obj) + "\"");
```

dhil (changes requested):

lib/js/jslib.js
```diff
@@ -800,23 +800,70 @@ const LINKS = new (function() {
...
-      obj.key = _lookupMobileKey(state, obj.key)
-      return obj;
-    } else {
-      for (let key in obj) {
-        obj[key] = resolveServerValue(state, obj[key]);
-      }
-      return obj;
-    }
+    switch (obj._tag) {
+      default:
```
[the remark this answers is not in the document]
dhil: We need to investigate whether the `obj.key` needs to be resolved. I suspect it may contain serialised server values.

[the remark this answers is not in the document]
dhil: Judging from 
```javascript
function _lookupMobileKey(state, serverKey) {
  return state.mobileKeys[serverKey];
}
```
it seems that `obj.key` is just an index. So it should probably not be resolved.

dhil:
We also need to add a tag to processes, e.g. ```("_tag", `String "Process")``` https://github.com/links-lang/links/pull/1015/files#diff-951e2296b8afaef998cbf54def5bc5b91bbe143e757d3092222bd07b576ea31bR195. In `resolveServerValue` for an object `obj` with `_tag == "Process"` you will need to recursively resolve `obj.process` and each entry in `obj.messages`.

dhil:
I have updated the patch now. I managed to restore all examples. There were multiple bugs hiding in the client interaction with server values. It is a sort of miracle that anything worked before. I reckon we are due a clean up of the client-server interaction logic in `jslib.js`.

dhil (approved):

dhil:
@SimonJF hopefully this patch restored the functionality of your MVU examples. With this patch I was able to run the examples pages again.

PR #1022, opened 2021-07-16
Resolve arguments inside call packages.

### LINKED ISSUE #1021: Regression: progress.links progress bar no longer renders correctly.
Whilst patch #1015 restored the functionality of the examples page, it broke the rendering of the progress bar in `progress.links`. The progress bar renders as full during the entire duration of the counting.

dhil:
This patch fixes #1021. The problem was that the arguments inside call
packages received from the server were unresolved.

A unified approach to handling of values received from the server is
warranted.


dhil:
@SimonJF, this patch truly restores all the examples, and I hope your MVU examples too. Let me know if you experience any problems following this patch.

PR #1023, opened 2021-07-17
Robust resolution of deeply nested server values.

### LINKED ISSUE #1013: Resolution of deeply nested server values crashes the client
Following #1009 deeply nested server values causes the client to crash, because the resolution algorithm is now recursive.

Here is an example that exhibits the behaviour (the crash is machine dependent, so you may have to increase the length of the list `xs`)
```links
var xs = [0..2085];

fun mainPage(_) {
  page
    <#>
      <button l:onclick="{println(intToString(length(xs)))}">Click me</button>
    </#>
}

fun main() {
  addRoute("/", mainPage);
  servePages()
}

main()
```

The fix is to turn it back into iterative form.

dhil:
This patch converts the function `resolveServerValue` into CPS with a
trampoline such that resolution of deeply nested server values no
longer causes a stack overflow.

Resolves #1013.

### DIFF
--- a/lib/js/jslib.js
+++ b/lib/js/jslib.js
@@ -941,17 +941,30 @@ const LINKS = new (function() {
         // or continuing with a final result.
         // TBD: Would be more elegant to use JS constructors instead of
         // using a signal member like __continuation.
-        const unresolvedServerValue = serverResponse.content.value;
+        const callPackageOrServerValue = serverResponse.content.value;
 
-        if ((unresolvedServerValue instanceof Object) && ('__continuation' in unresolvedServerValue)) {
+        if ((callPackageOrServerValue instanceof Object) && ('__continuation' in callPackageOrServerValue)) {
+          // callPackageOrServerValue is a call package.
           // Bouncing the trampoline
 
-          DEBUG.debug("Client function name, before evaluation, is ", unresolvedServerValue.__name);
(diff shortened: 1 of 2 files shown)


dhil:
@SimonJF do you mind testing this patch with your examples?

lib/js/jslib.js
```diff
@@ -772,6 +772,17 @@ const LINKS = new (function() {
...
+   // Server value resolution
+   class ServerValueResolutionYield extends Error {
+     constructor(k) {
+       super("resolveServerValueCPS yielding");
+       this.k = k;
+     }
+   }
+
+   let _resolveDepth = 0;
+   let _resolveDepthMax = 1000;
```
SimonJF, line 784: Minor, but this can be a `const`, right?

SimonJF, line 784:
```suggestion
   const _resolveDepthMax = 1000;
```

SimonJF (approved):
This seems to work on my GtoPdb examples. There is an issue with the session-typed chatserver, but I think that's separate as I tried it before and after this.

dhil:
I will look at your session type example. I have a feeling it is related to value resolution

dhil:
@SimonJF please test again. It should work now.

SimonJF:
Confirmed working. Happy for this to go in at your convenience.

SimonJF:
And thanks for looking at the broken example!

PR #1026, opened 2021-07-21
Fix nested effect rows printing incorrectly

### LINKED ISSUE #1024: Type pretty printer bug
A minor bug in the new type pretty printer
```links
links> fun() { error("") };
fun : () { |_}~> _
```
It should print `fun : () ~> _`. I guess this went unnoticed because we have few/or no test cases for the `wild` sugar.

SamuelNovak:
Fixes #1024.

The issue was that the original code did not use the already flattened version of the row for printing the row variable in function arrows.

The problem was caused by the internal structure like this (in the particular example from #1024):
```links
fun() { error("") } : () { | {wild:()} }-> _
```
note the nested internal row within the variable of the external row, which reduces to `() {wild:()}-> _` and hence to `() ~> _`.

### DIFF
--- a/core/types.ml
+++ b/core/types.ml
@@ -3065,7 +3065,8 @@ module RoundtripPrinter : PRETTY_PRINTER = struct
           in
           Printer (fun ctx r buf ->
               let is_lolli = Context.is_ambient_linfun ctx in
-              (* flatten here in case there are nested row variables *)
+              (* flatten here in case there are nested row variables,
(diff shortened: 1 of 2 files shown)


dhil (changes requested):
LGTM. Can you please add the example as a test case?

SamuelNovak:
I created a test case in roundrippable types for this.

dhil (approved):

PR #1032, opened 2021-07-28
Effect Sugar: Sharing of Implicit Effect Variables

SamuelNovak:
This patch enhances the desugaring pass responsible for spreading the implicit shared effect variable across a type. It includes two new settings for `effect_sugar_policy` (see #1006, though they can likely be merged in any order), namely:

- **`final_arrow_shares_with_alias`** [shortcut `final_arrow`]:
  *(This will be enabled by default)*

  Where a last arrow is followed by a type alias containing an effect variable as its range, and the arrow has an implicit effect variable, the two may unify = share the effect variable.

  Originally, a possible shared effect variable was assumed to occur only on the last function arrow or the last type alias, and these two excluded each other. This lead to problems in types like the ones below (note that the examples already use the enhanced effect sugar).

  Consider the (minimal) example below: (*for a more elaborate/realistic example, see* [1])

  ```links
  typename S(a) = (v:a, ord:(a,a) ~> Bool);
  # really S(a, e::Eff) = (v:a, ord:(a,a) ~e~> Bool);

  sig upd_s : (a, S(a)) ~> S(a)
  # sig upd_s : (a, S(a, { |e})) ~e~> S(a, { |e})
  #                         ^-----^-----------^
  fun upd_s(v', s) {
    if (s.ord(v', s.v))
      (s with v=v')
    else
      s
  }
  ```

  The above example is a structure which carries around some helper which has an effect variable, and this helper is used when transforming it. Following the original rules, this would desugar into:
  ```links
  sig upd_s : (a, S(a, { |e})) ~f~> S(a, { |e})
                          ^-----------------^
  ```
  where `e` and `f` are different - this would, however be the wrong type.

  Instead the intended and inferred type is:
  ```links
  sig upd_s : (a, S(a, { |e})) ~e~> S(a, { |e})
                          ^-----^-----------^
  ```
  where the effect variables on the arrow and the aliases unified.

  On the other hand, in some cases, we do not want these to unify; consider:
  ```links
  sig handleOp : (Comp(a, {Op:()|_})) -> Comp(a)
  fun handleOp(f)() { handle(f()) { ... case Op(r) -> ... }}

  # we want the above to mean
  # sig handleOp : (Comp(a, {Op:()|e})) -f-> Comp(a, {Op{_}|e})
  #                                ^------------------------^
  ```
  where again `e` and `f` are different, except here this is intended.

  To this end, in this PR, the effect desugaring pass introduces a fresh **flexible** effect variable on the last arrow (if its range is an alias that can have a shared effect), and the type inference will correctly unify/not unify it with the effect variable on alias. [2]

  To illustrate, a signature like `upd_s` above transforms as follows:

  ```links
  sig upd_s : (a, S(a)) ~> S(a)

  ↦ (effect sugar)
  sig upd_s : (a, S(a, { |e})) ~%f~> S(a, { |e})
                          ^------------------^

  ↦ (type inference/unification)
  (if the internal function is used)
  sig upd_s : (a, S(a, { |e})) ~e~> S(a, { |e})
                          ^-----^-----------^
  (otherwise)
  sig upd_s : (a, S(a, { |e})) ~f~> S(a, { |e})
                          ^-----------------^
                               e ≠ f
  ```

  This way, the two possibilities (exemplified by `S`, where the effect needs to be shared on the arrows transforming it; and `Comp`, where this is not the case) are transformed by desugaring and type inference to the intended types.

- **`all_implicit_arrows_share`** [shortcut `all_arrows`]:
  All arrows with implicit effect variables will be unified. This is an experimental setting.

Note that both settings only impact the sharing of the implicit effect variable - the operation propagation is unchanged by this PR.

### TODO:
- [x] Check in more examples if this is no longer buggy (shouldn't be)
- [x] Create tests
- [x] Clean up code
- [x] Fix breaking handlers examples
  - [x] `examples/handlers/coins.links`
  - [x] `examples/handlers/fringe.links`
  - [x] `examples/handlers/shiftreset.links`
  - [x] `examples/handlers/shallow_state.links`
- [x] Check correctness of the big match that I rewrote
- [x] Update/elaborate this PR description (new stuff was added that isn't here yet)

---

[1]: This is an excerpt from a program that uses `Map`s based on trees. The keys are allowed to be arbitrary types, and hence we need to provide an ordering function.
  ```links
  typename Ord = [| Lower | Greater | Equal |];
  typename Order(a) = (a, a) ~> Ord;
  # Order(a, e::Eff) = (a, a) ~e~> Ord;

  typename Tree(a, b) = mu t. [| Node:(left:t, key:a, value:b, right:t) | Leaf |];
  typename Map(a, b) = (ord:Order(a), tree:Tree(a,b));
  # Map(a, b, e::Eff) = (ord:Order(a, { |e}), tree:Tree(a,b));

  [...]

  sig upd_map : (a, b, Map(a,b)) ~> Map(a,b)
  # sig upd_map : (a, b, Map(a,b,{ |e})) ~e~> Map(a,b,{ |e})
  #                                 ^-----^--------------^
  fun upd_map(key, value, mp) {
    var (ord=ord, tree=tree) = mp;
    (mp with tree = upd_tree(ord, key, value, tree))
  }
  ```

  The intended (and inferred) type of `upd_map` is:
  ```links
  upd_map : (a, b, Map(a,b,{ |e})) ~e~> Map(a,b,{ |e})
                              ^-----^--------------^
  ```
  where the effect variable `e` is shared across both the aliases and the arrow - this is because here, to transform a `Map`, we need to use the helper it carries around.

  The type that resulted from desugaring (before this PR) was however:
  ```links
  upd_map : (a, b, Map(a,b,{ |e})) ~f~> Map(a,b,{ |e})
                              ^--------------------^
  ```
  where `e` ≠ `f`.

---

[2]: There currently exist types, where this will not work correctly - this is however a problem with the type inference, see #1041. In the examples where this was a problem, I explicitly typed the impacted functions.


dhil:
I will look at this PR after #1006.

core/types.ml
```diff
@@ -2077,7 +2077,7 @@ module Policy = struct
            ; "   that contain functions"
            ; "   enables " ^ (shortcuts_of_opts preset_struct)
            (* TODO!! change this if default changes *)
-           ; " * default: revert to default value (currently preset_comp)"
+           ; " * default: revert to default value (currently preset_struct)"
```
dhil:
```suggestion
           ; " * default: revert to default configuration"
```

dhil (approved):
LGTM

core/types.ml
SamuelNovak: Done! :)

dhil (approved):

Issue #155, opened 2017-03-30
Headless Web testing framework


jamescheney:

jamescheney:
I was thinking (re: #156 ) that it would be handy to set up a simple script for regression testing of relatively simple web things, along the following lines:

- run short Links program that builds a web page and starts using servePages(), background it
- use wget or similar to check that the http response is what we expect
- shutdown the Links program

Would it be worthwhile to set up a separate script to do this, or is it a trivial special case of the headless web testing frameworks you've looked at?

SimonJF:
Hi James,

We'd need to do step 1 manually (I would imagine using Python's subprocess library or something similar), but steps 2 and 3 would be doable using the headless web testing frameworks.

ChefYeum:
I'm on it

PR #1035, opened 2021-08-02
E2E tests 

ChefYeum:
implements #155 and more.


.github/workflows/default.yml
```diff
@@ -131,7 +132,7 @@ jobs:
           MYSQL_PASSWORD: links
           MYSQL_ROOT_PASSWORD: links
           MYSQL_DATABASE: links
-          MYSQL_ROOT_HOST: '%'
+          MYSQL_ROOT_HOST: "%"
```
dhil, line 135: I don't understand the necessity of this change.

.github/workflows/default.yml
```diff
@@ -342,3 +343,47 @@ jobs:
           LINKS_MYSQL_PASSWORD: links
           LINKS_MYSQL_PORT: ${{ job.services.mysql.ports[3306] }}
         shell: bash
+
+  end-to-end:
```
dhil, line 347: Lets name this job "webbrowser" or something long those lines (maybe client even?)

.github/workflows/default.yml
```diff
@@ -342,3 +343,47 @@ jobs:
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Use OCaml ${{ matrix.ocaml-compiler }}
+        uses: ocaml/setup-ocaml@v2
+        with:
+          ocaml-compiler: ${{ matrix.ocaml-compiler }}
+
+      - run: opam install -y ounit2
```
dhil, line 371: I think you can omit this step.

.github/workflows/default.yml
```diff
@@ -342,3 +343,47 @@ jobs:
...
+        run: |
+          eval $(opam env)
+          make all-ci
+        shell: bash
+
+      - uses: actions/setup-node@v2
+      - run: npm install
+        working-directory: ${{ env.NPM_DIR }}
+      - run: npm test
+        working-directory: ${{ env.NPM_DIR }}
```
dhil, line 389: Remove the submodule "links-database-setup" from revision control. It was added in commit 04337e.

tests/end-to-end/linksServerRunner.js
```diff
@@ -0,0 +1,23 @@
...
+
+    // TODO: Find workaround to wait for the server to start.
+    // The following line produces an uncondtiional timeout.
+    return new Promise(resolve => {
+      setTimeout(() => {
+        resolve(process)
+      }, 5000)
+    });
+  },
+  DEFAULT_BASE_URL: 'http://localhost:8080/',
```
dhil, line 21: Can we take localhost and the port as input parameters, e.g. `HOST` and `PORT`?

tests/end-to-end/package-lock.json
```diff
@@ -0,0 +1,8630 @@
+{
```
dhil, line 1: Is this file intended to be under revision control? Isn't it a local file that NPM uses to acquire sole access to a project?

tests/end-to-end/tests/buttons.test.js
```diff
@@ -0,0 +1,46 @@
...
+const { startServer, DEFAULT_BASE_URL, LINKS_ROOT } = require('../linksServerRunner');
+
+let driver, process;
+
+beforeAll(async () => {
+  // Instantiate browser driver
+  driver = await loadBrowser();
+
+  // Start Links server
+  process = await startServer(`${LINKS_ROOT}/examples/webserver/buttons.links`);
```
dhil, line 12: Can we make the example an input parameter?

dhil (changes requested):

.github/workflows/default.yml
ChefYeum: It must have been my linter

tests/end-to-end/linksServerRunner.js
ChefYeum: From environment variables?

.github/workflows/default.yml
ChefYeum: Done

tests/end-to-end/tests/buttons.test.js
ChefYeum: I think it makes sense to hardcode them under the current setup

tests/end-to-end/linksServerRunner.js
ChefYeum: Where should I get the parameters?

ChefYeum: Ideally, it should read off the Links config file but I don't know if that is necessary

.github/workflows/default.yml
ChefYeum: I can't seem to find any mention of `links-database-setup` nor the commit hash. Please double-check

tests/end-to-end/package-lock.json
ChefYeum: Yeah it's meant to be version controlled

.github/workflows/default.yml
dhil: It is in this patch. Look in "Files Changed" tab, there you will see it.

tests/end-to-end/linksServerRunner.js
dhil: Exactly. It should get them as environment variables such that people can run the tests locally without having to modify the script locally (port 8080 may be in UAE by another service).

SimonJF:
(Just swooping in to say thank you *so much* for doing this!)

tests/end-to-end/linksServerRunner.js
ChefYeum: I added `PORT` to be read from the environment variable `LINKS_PORT`. But I am assuming this to match the port specified in the Links config file.

But I'm not sure if it makes sense for the hostname to be something other than `localhost`. What do you think?

dhil: It does make sense for the host to be different, e.g. `0.0.0.0`, `127.0.0.1`, or something completely different. The name `localhost` may or may not be bound to those values, so it is better to make this a parameter.

dhil: Regarding the port number, you are right that a port number may be supplied via a configuration file. In any case, the web driver should probably also be parameterised by the config. It could potentially be an optional parameter.

dhil:
The `webbrowser` job should install and setup PostgreSQL database too. It will have to populate the database with the data for the examples page.

ChefYeum:
It doesn't seem so trivial to add more than one test under the current setup.

After the first test, the server for the first test needs to be killed so it can start a server for the next test. But Node only seems to keep track of the shell process used to execute the Links server but not the process for Links server itself.

I'm investigating how I could start a process group when starting the server and kill them all at the end.

dhil:
> It doesn't seem so trivial to add more than one test under the current setup.
> 
> After the first test, the server for the first test needs to be killed so it can start a server for the next test. But Node only seems to keep track of the shell process used to execute the Links server but not the process for Links server itself.
> 
> I'm investigating how I could start a process group when starting the server and kill them all at the end.

Did you find a workaround? Otherwise, I have some ideas.

ChefYeum:
> > It doesn't seem so trivial to add more than one test under the current setup.
> > After the first test, the server for the first test needs to be killed so it can start a server for the next test. But Node only seems to keep track of the shell process used to execute the Links server but not the process for Links server itself.
> > I'm investigating how I could start a process group when starting the server and kill them all at the end.
> 
> Did you find a workaround? Otherwise, I have some ideas.

Yeah, that's resolved now. Now I'm working [here](/ChefYeum/links/tree/e2e-tests-test) for populating the DB. I think I'm almost there.

.github/workflows/default.yml
```diff
@@ -342,3 +343,107 @@ jobs:
...
+
+      - run: opam install -y conf-libev
+      - run: opam install -y ounit2
+
+      - name: Install Links dependencies
+        run: >-
+          opam install
+          ./links.opam
+          ./links-postgresql.opam
+          ./links-mysql8.opam
```
dhil, line 412: This shouldn't be necessary. We are only testing against Postgresql

.github/workflows/default.yml
```diff
@@ -342,3 +343,107 @@ jobs:
...
+          ./links-sqlite3.opam
```
dhil, line 413: Remove this too

tests/end-to-end/tests/buttons.test.js
```diff
@@ -0,0 +1,48 @@
...
+  // Wait for the result to load
+  await driver.wait(until.urlIs(DEFAULT_BASE_URL + '/#'));
+
+  var elem = await driver.findElement(By.css('body'));
+  var output = await elem.getText();
+
+  expect(output.trim()).toBe('3');
+});
+
+// test("Google", async () => {
```
dhil, line 41: Delete these comments.

dhil (approved):
LGTM modulo some minor change requests.

.github/workflows/default.yml
dhil: The UI isn't highlighting the right line here. My comment is about line 412.

.github/workflows/default.yml
dhil: Line 413.

ChefYeum (commented):
marked requested changes

PR #1038, opened 2021-08-05
Effect Sugar: Fix operations not propagating from within aliases

### LINKED ISSUE #1036: Effect Sugar does not propagate operations correctly
During desugaring, some operations are not being propagated. So far I've noticed this in cases, where the operations have to be pulled out of a type alias. Examples below:

This works correctly:
```links
sig handleL : (() {L:()|_}~> ()) -> (() ~> ())
fun handleL(f) {
  fun () {
    handle(f()) { case L(resume) -> resume(()) }
  }
}

handleL = fun : (() {L:() {}-> ()|a}~> ()) -> () {L{_}|a}~> ()
```

However, if `L` is hidden inside an alias, it is not discovered:
```links
typename HasL = () {L:()|_}~> ();
# HasL = a::Eff.() {L:() {}-> ()|a::Eff}~> ()

typename AnyFun = () ~> ();
# AnyFun = a::Eff.() ~a~> ()

# Example 1:

sig handleL1 : (HasL) -> (() ~> ())
fun handleL1(f) {
  fun () {
    handle(f()) { case L(resume) -> resume(()) }
  }
}

Type error: The non-recursive function definition has return type
    `() {L{a}|b}~> ()'
but its annotation has return type
    `() ~b~> ()'
In expression: fun handleL1(f) {
  fun () {
    handle(f()) { case L(resume) -> resume(()) }
  }
}.

# Example 2:

sig handleL2 : (HasL) -> (AnyFun)
fun handleL2(f) {
  fun () {
    handle(f()) { case L(resume) -> resume(()) }
  }
}

Type error: The non-recursive function definition has return type
    `() {L{a}|b}~> ()'
but its annotation has return type
    `AnyFun ({ |b})'
In expression: fun handleL2(f) {
       fun () {
         handle(f()) { case L(resume) -> resume(()) }
       }
     }.
```

The desugaring correctly causes the effect variable in `HasL` to unify with the effect on the last arrow (ex. 1) or the alias in codomain (ex. 2), but the operation `L` is not propagated.

I suspect we just need to descend into the alias when collecting operations.

SamuelNovak:
Fix #1036 

This PR introduces changes into the effect desugaring pass. As shown in #1036, when operations appear within a type alias, they are hidden from the effect sugar and do not propagate the same way visible (written explicitly in the signature) operations do.

This PR makes the desugaring pass descend into the type alias/recursive application when collecting operation labels, so that they can be propagated correctly and code such as following will typecheck:

```links
typename HasL = () {L:()|_}~> ();
# HasL = a::Eff.() {L:() {}-> ()|a::Eff}~> ()

sig handleL1 : (HasL) -> (() ~> ())
fun handleL1(f) {
  fun () {
    handle(f()) { case L(resume) -> resume(()) }
  }
}
# handleL1 = fun : (HasL ({ |a})) -> () {L{_}|a}~> ()
```
(example from #1036)

Note that `L` is not made visible in the alias `HasL` in the above type - it is already contained within. Currently it prints as `HasL ({ |a})`, but this is handled by #1006, and will be printed without the `{ |a}` after that is merged (it is an orthogonal issue).

### TODO:
- [x] Code cleanup
- [x] Check out the breaking examples
  - [x] `examples/handlers/fringe.links`
  - [x] `examples/handlers/monadic_reflection.links`
  - [x] `examples/handlers/shiftreset.links`
  - [x] `examples/handlers/shallow_state.links`


core/desugarEffects.ml
```diff
@@ -548,11 +548,11 @@ let collect_operation_of_type tp
       | Some vid -> RowVarMap.add_raw (-1)
                       (RowVarMap.find_raw vid operations) operations
     in
-    (* print_endline "Collected operations:";
-     * print_endline
-     *   (RowVarMap.show (fun ppr x -> Format.fprintf ppr "%s"
-     *                                   (StringSet.show x)) operations);
-     * flush_all (); *)
+    print_endline "Collected operations:";
```
dhil: I don't think these should be printed unless `debug` is enabled (and possibly some other option). Maybe just delete them again.

core/desugarEffects.ml
```diff
@@ -641,12 +641,16 @@ let gather_operations (tycon_env : simple_tycon_env) allow_fresh dt =
                let operations =
                  RowVarMap.fold
                    (fun vid sset acc ->
+                     print_endline "folding";
```
dhil: Same as above.

core/desugarEffects.ml
```diff
@@ -710,6 +714,9 @@ let preprocess_type (dt : Datatype.with_pos) tycon_env allow_fresh shared_effect
     =
   let dt = cleanup_effects tycon_env dt in
   let row_operations = gather_operations tycon_env allow_fresh dt in
+  print_endline "Preprocessing: gathered operation labels. RowVarMap, stringmap";
```
dhil: Same as above.

dhil (changes requested):

core/desugarEffects.ml
dhil: Seems like I was viewing an outdated commit.

core/desugarEffects.ml
```diff
@@ -95,20 +95,20 @@ let unpack_var_id = function
 
 module SEnv = Env.String
 
-type tycon_info = Kind.t list * bool
+type tycon_info = Kind.t list * bool * Types.typ option
```
dhil, line 98: I think you should add a comment about the purpose of each of the three components.

dhil (approved):
LGTM

dhil (approved):

core/desugarEffects.ml
SamuelNovak: Done!

PR #1039, opened 2021-08-05
[WIP] Fix #1037

### LINKED ISSUE #1037: Xml type bug
Output from Links REPL 0.9.3
```
Welcome to Links version 0.9.3 (Burghmuirhead)
links> var z = <a><a1>A1</a1><a2>A2</a2></a>;
Finishing process MAIN
z = <a><a1>A1</a1><a2>A2</a2></a> : Xml
links> var z_children = childNodes(z);
Finishing process MAIN
z_children = <a1>A1</a1><a2>A2</a2> : Xml
links> var z_grandchildren = childNodes(z_children);
***: Internal Error in lib.ml (Please report as a bug): Runtime type error: non-XML given to childNodes.
This should not happen if the type system / checker is correct. Please file a bug report.
```

jamescheney:
A quick and hopefully not too dirty fix for #1037.  Fixes a couple of other weird things about XML along the way.

Let's see if this passes CI, then we should add a test or two.

### DIFF
--- a/prelude.links
+++ b/prelude.links
@@ -502,6 +502,15 @@ fun asList(t) server {
   for (x <-- t) [x]
 }
 
+# Gets children of elements in an XML forest, concatenated in order
+# @param {Xml} xml
+# @returns {Xml}
+sig childNodes : (Xml) ~> Xml
+fun childNodes(xml) {
+  for (x <- xml) elementChildNodes(x)
+}
+
+
 # Appends a XML forest to an XmlItems children
 # @param {XmlItem} item
 # @param {Xml} child
--- a/prelude.links
+++ b/prelude.links
@@ -507,9 +507,16 @@ fun asList(t) server {
 # @returns {Xml}
 sig childNodes : (Xml) ~> Xml
 fun childNodes(xml) {
-  for (x <- xml) elementChildNodes(x)
+  for (x <- xml) itemChildNodes(x)
 }
 
+# Gets text content of an XML forest, concatenated in order
+# @param {Xml} xml
+# @returns {String}
+sig getTextContent : (Xml) ~> String
+fun getTextContent(xml) {
+  fold_left((^^),"",for (x <- xml) [itemTextContent(x)])
+}
 
 # Appends a XML forest to an XmlItems children
 # @param {XmlItem} item
--- a/core/lib.ml
+++ b/core/lib.ml
@@ -577,7 +577,7 @@ let env : (string * (located_primitive * Types.datatype * pure)) list = [
   PURE);
 
   (* XML *)
-  "elementChildNodes",
+  "itemChildNodes",
   (p1 (function
          | `XML (Value.Node (_, children)) ->
            let children = List.filter Value.is_node children in
@@ -586,7 +586,7 @@ let env : (string * (located_primitive * Types.datatype * pure)) list = [
            let children = List.filter Value.is_node children in
            `List (List.map (fun x -> `XML x) children)
          | `XML (_) -> `List []
-         | _ -> raise (runtime_type_error "non-XML given to elementChildNodes")),
+         | _ -> raise (runtime_type_error "non-XML given to itemChildNodes")),
    datatype "(XmlItem) -> Xml",
   IMPURE);
 
@@ -798,8 +798,11 @@ let env : (string * (located_primitive * Types.datatype * pure)) list = [
   IMPURE);
 
 
-  "getTextContent",
-  (`Client, datatype "(Xml) ~> String",
+  "itemTextContent",
+  (p1 (function
+         | `XML (Value.Text str) -> Value.box_string str
+         | _ -> raise (runtime_type_error "non-text node given to textContent")), 
+   datatype "(XmlItem) ~> String",
   IMPURE);
 
   "getAttributes",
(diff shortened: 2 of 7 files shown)


jamescheney:
One possibly undesirable thing about the current implementation of `childNodes` is that it only selects child element or namespaced element nodes, and excludes attribute or text nodes, which are also considered nodes of the DOM tree, though attribute nodes are generally not considered "children" of an element node they're attached to.  However, text nodes could be returned and aren't.  Maybe this is often what we want, but it differs from (say) XPath which I believe can return both text and element nodes when following a child step.  Of course, Links != XPath.

vcgalpin:
A related XML issue I'm seeing is the following:
```
links> var t = <b>name</b>;
Finishing process MAIN
t = <b>name</b> : Xml
links> getTextContent(t);
***: Error: Cannot call client side function 'getTextContent' because of outside of web mode
```
Is there a way around this?

vcgalpin:
It may be better to have distinct functions with meaningful names:
getChildElements (rather than childNodes)
getTextNode (which works outside web mode)
getAttributes (which already exists)
and
getChildren
to return everything?

jamescheney:
Argh, I forgot about the issue of aligning with the javascript versions.  `jslib.js` has this to say about `childNodes`:
```
// childNodes
//   This is badly named
//    - what is its intended semantics?
//    - what should it be called?
//    - should it even be here?
//
// [BUG]: This expects a DomNode, while the server version expects Xml.
function _childNodes(elem) {
  DEBUG.assert(false, "childNodes is not implemented properly");
  var result = [];
  for (var i=0; i<elem[0].childNodes.length; i++) {
    result.push(elem[0].childNodes[i].cloneNode(true));
  }
  return(result);
}
```

whereas \getChildNodes` and `getTextContent` are defined like this:
```
function _getChildNodes(xml) {
  var first = LINKEDLIST.head(xml);
  if (first.type !== "ELEMENT") {
    throw new Error("getChildNodes() applied to non-element node");
  }
  return first.children || [ ];
}

function _getTextContent(xml) {
  var first = LINKEDLIST.head(xml);
  if (first.type !== "TEXT") {
    throw new Error("getTextContent() applied to non-text node");
  }
  return first.text || '';
}
```
which if I understand right, means that they expect a list of xmlitems, the first of which is expected to be an element resp. text node, and return the children resp. text content.

We should be able to provide `getTextContent` on the server side, or emulate it by converting the Xml to a variant.
I'm inclined to fix these to apply to a single xmlitem, like `elementChildNodes`.

jamescheney:
In this PR let's just fix broken parts of the existing design, and collect requirements for a redesign somewhere else.

It looks like `childNodes` in the JS library brokenly expects a dom node rather than an Xml, so won't give equivalent behavior.  If it isn't being used anywhere, maybe it should just be gotten rid of.

jamescheney:
I've tried to rationalize this a bit further, and provide a server-side implementation of `getTextContent`.  But the JS version expects to be applied to a single-element Xml which is a text node.  So after this change
```
getTextContent(<#>asdf</#>);
```
works and returns `"asdf"`, but 
```
getTextContent(<a>asdf</a>);
```
fails (and this should also happen on the client, and would have before my changes).

Moreover you can't get the text node(s) out using `childNodes` since this only returns children that are element nodes.  But fixing this would involve changing the behavior, not just generalizing it.

vcgalpin:
I think I can work around all the current problems by using xmlToVariant and variantToXml, and working with the variant type rather than the Xml type.

jamescheney:
Yes, that should work and is probably the right thing to do if you are doing anything serious with XML.  Longstanding issue/enhancement #46 is about redefining the `Xml`/`XmlItem` types to just be the equivalent variant types you get from xmlToVariant.  This would have lots of advantages, but since in Links we don't have e.g. type classes or other ways of specializing printing to particular types, the nicer ad hoc printing of XML literals would be hard to retain.  By keeping `XmlItem` abstract we keep this ability, and also in principle keep the ability to enforce that an `Xml` value is a well-formed XML document (or fragment), though I'm not sure we currently actually do this.

jamescheney:
In fact a step towards #46 would probably be to use `xmlToVariant` and `variantToXml` to implement the other XML convenience functions natively, though this would probably slow many of them down.

prelude.links
```diff
@@ -502,6 +502,22 @@ fun asList(t) server {
   for (x <-- t) [x]
 }
 
+# Gets children of elements in an XML forest, concatenated in order
+# @param {Xml} xml
+# @returns {Xml}
+sig childNodes : (Xml) ~> Xml
```
dhil, line 508: Changelog: generalised behaviour

prelude.links
```diff
@@ -502,6 +502,22 @@ fun asList(t) server {
...
+fun childNodes(xml) {
+  for (x <- xml) itemChildNodes(x)
+}
+
+# Gets text content of an XML forest, concatenated in order
+# @param {Xml} xml
+# @returns {String}
+sig getTextContent : (Xml) ~> String
```
dhil, line 516: Changelog: now runs on the server too.

prelude.links
dhil: (should we adopt the habit of extending the changelog as we go rather than all at once just before a release?)

dhil (approved):
Looks good to me.

vcgalpin (approved):

prelude.links
jamescheney: Not sure, but for 0.9.4 it seems that there will be plenty of PRs that haven't been doing this so it might be confusing to start halfway through.  Shall we discuss at the next meeting?

Issue #1042, opened 2021-08-30
Update changelog


dhil:
@SamuelNovak @wricciot and myself to update the `CHANGES.md` with a summary of our respective patches.

wricciot:
I've added my summary now.

dhil:
I have added my entries as well.

SamuelNovak:
It seems I cannot push to master directly, so I've created a PR with my entries: #1043.

Issue #1041, opened 2021-08-17
Type inference incorrectly unifies flexible effect rows in recursive functions


SamuelNovak:
For certain recursive functions that have both a flexible effect variable, and an operation label occuring in a different effect row, the two effect variables get unified, even though the flexible one should be instantiated as a fresh rigid var.

Two minimal examples:

```links
sig g : (a) -%f-> Comp((), {Op:(a) {}-> ()|e})
#           ^^^^^
fun g(x)() {
  g(x)()
}

g = fun : (a) {Op:(a) {}-> ()|b}~> Comp ((),{Op:(a) {}-> ()|b})
#             ^^^^^^^^^^^^^^^^^^


sig g' : (a) -%f-> () {Op:(a) {}-> ()|e}~> ()
#            ^^^^^
fun g'(x)() {
  g'(x)()
}

g' = fun : (a) {Op:(a) {}-> ()|b}~> () {Op:(a) {}-> ()|b}~> ()
#              ^^^^^^^^^^^^^^^^^^
```

As you can see above, `%f` incorrectly unified with `e`. However, we would expect `%f` to be instantiated to some fresh variable, so that is it different from `e` - this is also the type that gets inferred if the signature is omitted.

```links
fun g'(x)() {
  do Op(x);
  g'(x)()
}

g' = fun : (a) -> () {Op:(a) {}-> ()|_}~> _::Any
#              ^^                   ^^^ these are different!
```

(Note that in the above example, I changed this definition of `g'` so that the operation `Op` would be inferred).

### More details:

* This happens in cases with and without type aliases containing effects (such as `Comp`).
* The declared operations do not need to be invoked from within the function.
* It seems this only happens with recursive functions (at least I have not yet managed to replicate it with a non-recursive function).
* This occurs on `master`, independent of effect sugar.

---

For a more elaborate example, I first noticed this in [examples/handlers/fringe.links](https://github.com/links-lang/links/blob/master/examples/handlers/fringe.links), when I placed a flexible variable on the arrow in `walkTree`, as follows:

```links
sig walkTree : (Tree(a)) -%f-> Comp((), {Yield:(a) {}-> ()|_})
fun walkTree(tree)() {
  switch (tree) {
    case Leaf(v)   -> yield(v)
    case Node(l,r) -> walkTree(l)(); walkTree(r)()
  }
}
```

(placing flexible variables here is motivated by #1032)

PR #1043, opened 2021-08-30
Update CHANGES.md with recursive rows, roundtrip printer and effect sugar

SamuelNovak:
Update to CHANGES.md, see #1042. It seems I cannot push it directly without a PR.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,19 @@
 # 0.9.4
 
+## Queries mixing set and bag semantics
+Links now provides experimental support for SQL queries mixing set and bag semantics. 
+
+When the `mixing_norm=on` flag is added to the configuration file, or when a query is defined in a `query mixing { ... }` block, Links will use a new query evaluator, allowing the programmer to call deduplication functions (`dedup` and `distinct`) within database queries. These are handled with set-based SQL statements `select distinct / union`, in addition to the usual bag-based `select / union all`.
+
+    # will run on the DB as "select distinct e.dept as dept from employees"
+    query mixing {
+      dedup(for (e <-- employees) [(dept = e.dept)])
+    }
+
+Queries mixing set and bag semantics may, in some cases, require the use of the SQL:1999 keyword `lateral`; Links implements an optional query transformation to produce queries that do not use `lateral` (allowing the use of older DBMSs): this behaviour is enabled by using `query delat` in place of `query mixing`.
+
+Further information on this feature is provided in the [Links GitHub wiki](https://github.com/links-lang/links/wiki/Deduplication-in-database-queries).
+
 ## DateTime type
 Links now includes a primitive type, `DateTime`, for dates and times.
 This is a *breaking change* from previous versions, where `dateToInt`
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -43,6 +43,23 @@ You can also print out the `DateTime` using `show` (which is an alias of
 
 Due to limitations of the underlying library, the minimum timezone granularity is one hour. Unfortunately, this means we can't handle Indian timezones, for example.
 
+## Surface syntax for presence type argument
+
+**Breaking change**: New syntax has been added to support type arguments of kind `Presence`. Here are some example of the syntax:
+
+```links
+typename T(p::Presence) = (foo{p});
+
+(foo=4200) : T({:Int})         # Present with type Int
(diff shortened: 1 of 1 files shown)


CHANGES.md
```diff
@@ -66,6 +66,59 @@ The syntactic sugar for effect and record fields which lets one omit the `()` ha
 
 It is now possible to annotate type variables with `Mono` restriction, e.g. `sig id : (a::(Any,Mono)) -> a::(Any,Mono)`.
 
+### Recursive rows
+
+* We can now enter a recursive row effect type, such as `{ |(mu a.F:(() { |a}-> ()) {}-> b|c)}`.
+* **Breaking change**: recursive rows of different types cannot mix anymore - a variant can only contain fields with variant syntax (i.e. uppercase and separated by `|`) in its recursive row variable, and the same goes for others (e.g. records - lowecase, separated by `,`);
+  * these are correct: `[|(mu a. Foo:()|Bar:()|d)|]`, `(|(mu a. foo:(),bar:()|d))`
+  * these are not: `[|(mu a. foo:()|bar:()|d)|]`, `(|(mu a. foo:()|bar:()|d))`
+* **Breaking change**: Recursive variants with no directly exposed fields no longer require (nor allow) the vertical bar separating fields from the row variable: the following is no longer correct: `[| |(mu a . Foo:()|Bar:()|d)|]` (note the bar before `mu`)
```
dhil, line 75: Hmm. I missed this point. I think we should add that syntax back in.

dhil: I am not sure why I missed this, but I think it is "wrong" to be able to write `[|(mu a. ...)|]`, we should keep some syntactic consistency between record, effect, and variant rows.

SamuelNovak: I think this was intentional at the time, to have a special sugar for variants, because `[|(mu r. φ)|]` is unambiguous, whereas `((mu r. φ))` is not the same as `(|(mu r. φ))`.

Ultimately there is a tradeoff to be made - either we want simplicity and uniformity (I think this is what I'd go for), or we want the syntax to be shorter even if it forces a special case to be made (this variant thing affords us to lose 2 characters: ` |`; there is a middle ground in allowing `[||` but then this may need to be a new lexeme).

dhil: Yes. But we should not remove the ability to write `[| |p|]`, it is the unsugared/"correct" syntax.

SamuelNovak: Ah, I see what you mean. I reckon this should only require duplicating one of the production rules for `vfields`, with the `vbar`.

dhil: I have restored the syntax

CHANGES.md
```diff
@@ -66,6 +66,59 @@ The syntactic sugar for effect and record fields which lets one omit the `()` ha
...
+
+Note that one can select multiple printers at once, for comparison; this is done by separating printer names by commas, e.g.:
+
+```links
+@set types_pretty_printer_engine "roundtrip,old";
+```
+
+## Effect Syntactic Sugar
+
+**Breaking change** (for code using effect sugar): This version implements enhanced syntactic sugar for effects. The changes influence both the Roundtrip printer (see above) and the desugaring passes (between parsing and typechecking).
```
dhil, line 95: I don't think this is a breaking change or need to be. The default preset should do whatever `effect_sugar=true` used to do.

SamuelNovak: I believe the breaking part comes from the implicit sharing inference: the rule `final_arrow`. I think this can just be removed from the default and that will restore the original behaviour. 

Shouldn't cause problems with tests either, but I'm not too sure of this.

I'll have a look

SamuelNovak: I've opened #1045 to return the option of the original sharing rule for arrow + alias.

However, I've realised that the changes listed here are likely not breaking on their own - the new version of sharing between last arrow + type alias is a more general approach that handles old cases as they were expected, and also handles new cases. 

However, the problem is that #1041 (incorrect behaviour in the type inference engine/unifier) interacts with this new general approach (which relies on correct type inference to infer whether the last arrow will share its effect var with a following alias) and can break existing code - which I think is why I originally marked this as breaking.

For reference see #1032 and #1041.

SamuelNovak: I've removed the "breaking" mark, since the effect sugar itself should not actually break stuff - it's just more general.

However I think the above is still a problem

PR #1045, opened 2021-08-31
Reinstated old sharing rules as default

SamuelNovak:
In #1032, the original rules for sharing of implicit effect variables between a last arrow and a following type alias were completely replaced by the new inference-based rule. This PR removes the new sharing rule from the default setting.

See [this review](https://github.com/links-lang/links/pull/1043#pullrequestreview-742323777).

### DIFF
--- a/core/types.ml
+++ b/core/types.ml
@@ -2023,6 +2023,7 @@ module Policy = struct
              | OpenDefault
              | FinalArrowSharesWithAlias
              | AllImplicitArrowsShare
+             | OldSharing
 
     type t = opt list
     val default : unit -> t
@@ -2035,6 +2036,7 @@ module Policy = struct
     val open_default              : t -> bool
     val final_arrow_shares_with_alias : t -> bool
     val all_implicit_arrows_share : t -> bool
+    val old_sharing               : t -> bool
   end = struct
     type opt = PresenceOmit
              | AliasOmit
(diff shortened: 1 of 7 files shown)


core/types.ml
```diff
@@ -2046,7 +2046,7 @@ module Policy = struct
              | AllImplicitArrowsShare
     type t = opt list
 
-    let default_opts = [PresenceOmit ; AliasOmit ; ContractOperationArrows ; ArrowsCurriedHideFresh ; FinalArrowSharesWithAlias ]
+    let default_opts = [PresenceOmit ; AliasOmit ; ContractOperationArrows ; ArrowsCurriedHideFresh ]
```
dhil, line 2049: ... and these are only toggled if `effect_sugar=true`, right?

dhil (approved):

core/types.ml
SamuelNovak: Yes. The setting itself is independent of `effect_sugar`, so the values are there, but they take no effect unless `effect_sugar=true`

SamuelNovak: What I mean is that the `default_opts` will be set even without `effect_sugar=true`, but this will have no effect on anything.

core/types.ml
```diff
@@ -2135,17 +2136,16 @@ module Policy = struct
            ; " * open_default [open]: effect rows are open by default,"
            ; "   closed with syntax { | .}"
            ; " * final_arrow_shares_with_alias [final_arrow]: final arrow and"
-           ; "   a following type alias will be assumed to share implicit effects"
-           ; "   (This is only a desugaring setting)."
+           ; "   a following type alias may share the implicit effect variable"
            ; " * all_implicit_arrows_share [all_arrows]: all arrows with implicit"
-           ; "   effect vars will be unified"
+           ; "   effect vars will be unified (EXPERIMENTAL)"
```
dhil, line 2141:
```suggestion
           ; "   effect vars will be unified (experimental)"
```

dhil (approved):

PR #1061, opened 2021-10-18
Implement System F-style general-purpose type abstractions

frank-emrich:
This PR enables writing expression of the form `/\ [Q]. { E }`, where Q is a list of quantifiers and E is an expression with the quantifiers Q in scope. Assuming that E has type T, the overall type is `forall Q.T`.

 For example, we may write
```
  /\ [a, e :: Row] { sig f : (a) -e-> a fun f(x) {x} ~f }
```
of type `forall a, e::Row. (a) -e-> a`.

One important restriction to ensure that such /\ expressions have most general types is that the type T of E must be unambiguous/unique.
As an example why this is important, consider the following function:
```
fun outer() {
  var inner = /\ [e::Row] { fun f(x) {x + 1} };
  ~inner
}
```
What should `outer`'s type be? If the inner function uses `e` as its effect variable, we have 
```
outer : forall e::Row. (Int) -e-> (Int)
```

However, if the inner function uses some arbitrary effect %f as its effect variable, the binding of `inner` could generalize this independently, and we would get
```
outer : forall f::Row, e::Row. (Int) -f-> (Int)
``` 

We avoid all these problems with a simple rule: After type-checking any type abstraction of the form mentioned initially, we require that the type T of E does not contain any flexible type variable. This is sufficient to ensure that E has a unique type.

My plan is to integrate similar syntax using quantifiers in square brackets into `fun` and `rec` bindings in a follow-up PR.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,10 @@
+# 0.9.6
+
+* Links now supports System F-style explicit type abstractions:
+  For instance, writing `/\ [a, e::Row] { foo }` abstracts the expression `foo`
+  over type variable `a` and row variable `e`. Here, `foo` must have a unique
+  type and must be pure (to satisfy the value restriction).
+
 # 0.9.5
 
 This is a minor hotfix release.
--- a/core/desugarTypeVariables.ml
+++ b/core/desugarTypeVariables.ml
@@ -254,6 +254,8 @@ object (o : 'self)
 
   method set_toplevelness at_toplevel = {< at_toplevel >}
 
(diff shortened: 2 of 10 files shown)


dhil:
Need to implement value restriction check.

dhil:
Ping @frank-emrich

frank-emrich:
Sorry for the delay. I've now made sure that `/\` obeys the value restriction in the same way as other generalising constructs in Links. I've also extended CHANGES.md to reflect the change.

PR #1062, opened 2021-10-29
Deserialise messages in `onMessage`.

dhil:
This commit deserialises received messages in the `MESSAGE_DELIVERY`
branch of `jslib.js`. This fixes the problem that the UG students have
run into.

As the comment in the code says, a refactoring of this code is
warranted.


slindley:
Some tidying will be needed later, but this PR fixes important bugs and doesn't seem to break existing code so I've merged it already.

PR #1072, opened 2021-12-23
Fixes #1067

### LINKED ISSUE #1067: Use of dedup and distinct
I'm not getting the results I expect from the following code. The `numbers` table has one integer field; there are 1000 rows, each containing a value 0 between 99. I'm running version 0.9.4 installed via opam on MacOS 11.6.1 and the database and table are defined as follows:
```
var db = database "links";
var numbers = table "numbers" with (number: Int) from db;
```
This is the output from the REPL (with timing information removed for clarity). I've explicitly used `query mixing` but I also have `mixing_norm=on` in the config file.
`dedup` outside the query gives the expected result.
```
links> length ( dedup (query mixing { for (x <-- numbers) [x] }));
Running query: 
select (t2386."number") as "number"
from "numbers" as t2386
Finishing process MAIN
100 : Int
```
`dedup` inside the query gives an error.
```
links> length (query mixing { dedup(for (x <-- numbers) [x] )});
***: Internal Error in lib.ml (Please report as a bug): Unoptimized table access!!!
```
and `distinct` appears to do nothing as it returns the same list as when it isn't used.
```
links> length (query mixing { for (x <- distinct(numbers)) [x] });
Running query: 
select (t2393."number") as "number"
from "numbers" as t2393
Finishing process MAIN
1000 : Int
```

wricciot:
This PR fixes issue #1067 with queries using deduplication:

The fix consists of three small changes:

- the function `QueryLang.used_database`, which incorrectly ignored deduplicated subexpressions, now performs recursion under deduplication: this corrects the "unoptimized query" problem (which was caused by Links incrrectly determining that deduplicated subexpressions didn't require the database)

- the function `MixingQuery.freshen_for_bindings` was incorrectly throwing away `Q.Prom` and `Q.Dedup` constructors when performing recursion. This was at the basis of the problem that caused certain uses of distinct/dedup to be ineffective and has been fixed

- the function `QueryLang.lookup_fun` does NOT hardcode `distinct` as a `Primitive` operation any more, letting Links unfold it to a composition of `asList` and `dedup` (`dedup`, however, will still be hardcoded as a `Primitive`); this ensures that normalized queries have the expected normal form and fixes a "fatal error: disjunct" problem that was not detected in the issue, but still occurred when evaluating queries in the form `query { distinct(factorials) }`.

### DIFF
--- a/core/query/evalMixingQuery.ml
+++ b/core/query/evalMixingQuery.ml
@@ -36,14 +36,14 @@ and disjunct is_set = function
 | QL.Prom p -> sql_of_query S.Distinct p
 | QL.Singleton _ as j -> S.Select (body is_set [] [] j)
 | QL.For (_, gs, os, j) -> S.Select (body is_set gs os j)
-| _arg -> Debug.print ("error in SimpleSqlGen.disjunct: unexpected arg = " ^ QL.show _arg); failwith "disjunct"
+| _arg -> Debug.print ("error in EvalMixingQuery.disjunct: unexpected arg = " ^ QL.show _arg); failwith "disjunct"
 
 and generator locvars = function
 | (v, QL.Prom p) -> (S.Subquery (dependency_of_contains_free (E.contains_free locvars p), sql_of_query S.Distinct p, v))
 | (v, QL.Table (_, tname, _, _)) -> (S.TableRef (tname, v))
 | (v, QL.Dedup (QL.Table (_, tname, _, _))) ->
     S.Subquery (S.Standard, S.Select (S.Distinct, S.Star, [S.TableRef (tname, v)], S.Constant (Constant.Bool true), []), v)
-| (_, _arg) -> Debug.print ("error in SimpleSqlGen.disjunct: unexpected arg = " ^ QL.show _arg); failwith "generator"
+| (_, _arg) -> Debug.print ("error in EvalMixingQuery.disjunct: unexpected arg = " ^ QL.show _arg); failwith "generator"
 
 and body is_set gs os j =
     let selquery body where =
--- a/core/query/mixingQuery.ml
+++ b/core/query/mixingQuery.ml
@@ -56,8 +56,8 @@ let rec freshen_for_bindings : Var.var Env.Int.t -> Q.t -> Q.t =
     | Q.Singleton v -> Q.Singleton (ffb v)
     | Q.Database db -> Q.Database db
     | Q.Concat vs -> Q.Concat (List.map ffb vs)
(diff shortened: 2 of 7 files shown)


jamescheney:
Looks like the CI is failing due to an error that appears to have nothing to do with any of your changes:

```
File "core/proc.ml", line 180, characters 8-22:
180 |         Lwt_main.yield () >>= pstate
              ^^^^^^^^^^^^^^
Error (alert deprecated): Lwt_main.yield
Use Lwt.pause instead
```

Presumably due to CI installing a newer version of lwt than you are using... 

It would also be good to add a few tests to tests/database that exercise dedup and distinct, possibly the ones given in #1067 that use factorials would suffice for now.

wricciot:
> Looks like the CI is failing due to an error that appears to have nothing to do with any of your changes:

I believe that's supposed to be corrected by Simon's PR 1073.

> It would also be good to add a few tests to tests/database that exercise dedup and distinct, possibly the ones given in #1067 that use factorials would suffice for now.

Ok, tests added.

jamescheney (approved):
OK, looks good to me.  We should eventually add some more elaborate tests...

wricciot:
> OK, looks good to me. We should eventually add some more elaborate tests...

Yes. I'm checking the other issue with mixing normalization so I'll be adding more tests as part of that fix.

wricciot:
I have spotted another problem with `QueryLang.used_database` and deduplication: the function ignores the subexpressions of a `Concat` because in a homogeneous setting they cannot reference databases directly; however, that is not true in the heterogeneous setting. I will commit a fix for that tomorrow morning (after which this PR should be ready for merging).

wricciot:
Ok, as far as I'm concerned this PR is ready for merging, However I'm confused by the web browser check failing: this only appeared after the last commit, which was a harmless removal of whitespace and comments, I am not familiar with this particular check, so I want to ask if anybody knows what the problem could be.

jamescheney:
That's strange.  The failing test looks like one that does exercises database functionality, but not in a way that seems relevant to any of the changes (specifically the last commit).  Might be worth making a dummy change to force CI retry in case the web browser failure is something transient.

dhil:
> That's strange. The failing test looks like one that does exercises database functionality, but not in a way that seems relevant to any of the changes (specifically the last commit). Might be worth making a dummy change to force CI retry in case the web browser failure is something transient.

As discussed on a previous meeting, the webbrowser step is broken. It can be ignored.

dhil:
I have disabled (commented out) the webbrowser workflow. I expect the CI to tick all green now.

PR #1073, opened 2021-12-24
Fix yield

SimonJF:
Fix deprecation warning with latest LWT

### DIFF
--- a/core/proc.ml
+++ b/core/proc.ml
@@ -177,7 +177,7 @@ struct
       begin
         (* Debug.print ("yielding"); *)
         reset_step_counter ();
-        Lwt_main.yield () >>= pstate
+        Lwt_main.pause () >>= pstate
       end
     else
       pstate ()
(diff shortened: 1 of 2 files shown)


SimonJF:
I don't know whether this will conflict with earlier versions of LWT that people may have installed, so I won't merge it yet. Perhaps it would be better to constrain the LWT version too.

dhil:
> I don't know whether this will conflict with earlier versions of LWT that people may have installed, so I won't merge it yet. Perhaps it would be better to constrain the LWT version too.

Yes, we should constrain the lwt version to be at least the version when pause was introduced or conservatively to the latest release.

SimonJF:
Just checked and `Lwt.pause` was added 5 years ago (prior to 3.1.0, which we currently require), so I think we're safe. Regardless though, we should probably constrain to at least 5.0.0 (tested and working) at some point rather than the ancient 3.1.0, since I'm pretty sure 3.1.0 wouldn't work.

dhil (approved):

Issue #565, opened 2019-04-12
Exceptions


jamescheney:
While thinking about #563, I wondered whether instead of changing the syntax to have special syntax for inserting just one thing with `returning`, what we really need is actual exceptions, so that instead of getting an interpreter-killing fatal error if you do the wrong thing in this case, you get a catchable exception.

This might be useful everywhere else exceptions normally are too.

So.  I read this nice paper about sessions and exceptions in Links.  It supports exceptions of type "unit" at the moment only.  (that is, when you `raise` you cannot send any data to the handler of the exception).

For example, I'd like to be able to do something like this:

```
try 
  insert t (a,b) xs returning a
with `InsertReturningError(e) -> print(e)
```
and implement the failure cases of `InsertReturning` in `evalir` by raising a Links-catchable exception `InsertReturningError` instead of an OCaml internal error.

How far away are we from being able to generalize the sessions-exceptions extension, so that we can throw / catch arbitrary Links variants as exceptions?  Is this dependent on (or expected to be subsumed by) a future version of effects-and-handlers?

I'm guessing that one headache here could be if the `try...with` starts out running on the server and then control flow passes to the client, where an exception is raised...

SimonJF:
As effect handlers generalise exception handlers, the way I'd implement this would be to use effect handlers, and implement a similar translation to the one that I did in STWT in order to handle possibly-linear resources correctly. I think that we have all of the theory and complex machinery that we need.

Having exceptions with payloads should be a reasonably straightforward extension to what we currently have (I've done some theory for exception payloads already) -- it'd just be a case of implementing it. Then, raising the exception would invoke something like `DoOperation ("_InsertReturningError")`.

Would you want this treated as a checked, or unchecked exception? For a checked exception, we almost have it already -- I believe all that would need doing is:

  1. An extension of the `cancel` construct to arbitrary terms as opposed to just session endpoints (we do this internally already)
  2. An exception operation, and a modification of the typing for  `insert` to state that it can invoke this exception operation
  3. Wrap all uses of `insert` in a `handle` block, where the handle and success cases are explicit

jamescheney:
I'd be happy with checked exceptions (i.e. types force handling) as a start. Only `insert...returning` would need this, not plain inserts.

slindley:
[...]

PR #1079, opened 2022-01-17
Temporal Language-Integrated Query

SimonJF:
# Temporal Language-Integrated Query

This patch adds support for period-stamped valid- and transaction-time temporal tables to Links.

## Temporal databases?

Here's a whistle-stop tour of temporal database concepts to contextualise the patch. For further information, see http://www2.cs.arizona.edu/~rts/tdbbook.pdf .

### Key concepts

Temporal databases model how data changes over time. There are two main dimensions:

  * Transaction time, which concerns the *state* of the database (e.g., "what was
      the contents of the 'employees' table last week?")

  * Valid time, which concerns the data being modelled (e.g., "when did Simon
      stop being a PhD student?")

      The goal is to add support for both of these dimensions of time to Links' language-integrated query functionality.

### Period stamping

There are a few ways of keeping track of this information, but the one we choose is period-stamping. Here, each temporal table is extended with "from" and "to" fields:

  | name  | position | valid_from          | valid_to            |
  |-------|----------|---------------------|---------------------|
  | alice | teacher  | 2013-03-20 00:00:00 | 2022-01-01 00:00:00 |
  | bob   | teacher  | 2022-01-17 00:00:00 | infinity            |

(Note that the intervals are closed-open: they *include* the `valid_from` field, and *exclude* the `valid_to` field).

Imagine this is a valid time table. We define the table as follows:

```
var employees =
  table "employees" with (name: String, position: String)
  using valid_time(valid_from, valid_to) from db;
```

Each row is then endowed with period-stamping information, so:

```
query nested {
    for (x <-v- employees)
    [ x ]
}
```

has type `ValidTime((name: String, position: String))`.

We can project the data out using the `vtData` function, and the `from` and `to` times using `vtFrom` and `vtTo` respectively.

We can get the current state of the database using the `ttCurrent` and `vtCurrent` functions (these are defined in the Prelude). Example:

```
query nested {
    for (x <- vtCurrent(employees))
    [ x ]
}
```

has type `(name: String, position: String)`.

(We can also timeslice at a particular point in time using `ttAt` and `vtAt`).

Insert, update, deletion operators will then modify these periods. As an example, if we had a valid-time table and wanted to fire Bob, and it was 10AM on the 17th of January, then we'd write

```
delete (x <-v- employees)
  where (x.name ==  "bob")
```

and this will end up with the database state:

  | name  | position | from                | to                  |
  |-------|----------|---------------------|---------------------|
  | alice | teacher  | 2013-03-20 00:00:00 | 2022-01-01 00:00:00 |
  | bob   | teacher  | 2022-01-17 00:00:00 | 2022-01-17 10:00:00 |

Similar things happen with insertions and updates.

### Current, Sequenced, Nonsequenced

The above refers to "current" updates, which mean "alter the current state of the database". There are two other dimensions of DB inserts/updates/deletions:

  * Sequenced: an update over a time period. For example, seconding someone to a different post between January and March (sequenced update) or having them take unpaid leave (sequenced deletion)

  * Nonsequenced: where the time period is explicitly altered (for example, extending someone's contract, or correcting their start date).

We support all of these.

## Feature Overview

This patch adds support for:

  * Transaction-time and valid-time tables using a period-stamped representation (i.e., with explicit "from" and "to" fields)

  * Insert, update, and delete operations on temporal tables

  * Querying temporal tables

  * Homogeneous temporal joins (e.g., joining two transaction time tables)

## Implementation

The core concepts are as follows:

  * Tables / update operations now optionally carry a "temporality" argument (either Current, Transaction, or Update). Tables also carry the names of their fields.

  * The `query/temporalQuery.ml` file contains all of the rewrites on inserts / updates / deletions

  * Internally, the `ValidTime` and `TransactionTime` types are represented as records of the form `(!data: a, !from: DateTime, !to: DateTime)`. The accessors are implemented explicitly rather than via a desugaring pass to ensure that the translation is type-preserving.

  * `query/query.ml` augments query results with the from- and to- times for temporal tables.

  * Temporal joins are achieved as a post-processing pass after query normalisation (the translation is actually quite easy after normalisation)

Naturally all of the usual things (Sugared AST, traversals, typechecker, IR, IR traversals, value representation) need to be changed as well.

You can see all of the features in action in the tests: `tests/shredding/{validtime, transactiontime, vt_join}.links`.

## Breaking Changes

  * New keywords: `vt_insert`, `tt_insert`, `valid_time`, `transaction_time`, `valid`, `to`, `current`

  * The `TableHandle` type now contains the temporality (either `Current`, `Transaction`, or `Valid`). For example: `TableHandle(Current, r, w, n)`

## Limitations / Future Work

  * At the moment, the temporal functionality only works with PostgreSQL. This is because Postgres is the only DBMS we support that allows an `infinity`
      timestamp.

  * We do not currently support bitemporal tables (i.e., tables which have both valid time and transaction time functionality). This is the next big feature we need to design and think about.

  * Valid time sequenced inserts are limited -- they currently do not attempt to do any clever "gap filling", nor do they do any in-SQL checking that they respect primary keys. Therefore, an invalid VT sequenced insert (for example, one where the period of validity overlaps an existing row) will fail at runtime.

  * Temporal queries can only currently be flat (i.e., result types cannot contain any further collections). Shredded temporal queries are entirely nontrivial and a research project of their own.

  * We only currently support a period-stamped representation. It would be interesting to also support other temporal partitioning strategies (preferably in a uniform way).


SimonJF:
@vcgalpin I'd be interested to see whether this works with your larger C-19 curation examples before we decide to merge. Let me know if you need any help migrating.

vcgalpin:
> @vcgalpin I'd be interested to see whether this works with your larger C-19 curation examples before we decide to merge. Let me know if you need any help migrating (although I think most of the breaking changes are outlined in my earlier email).

After some syntax changes, the C-19 curation app appears to be working. I've not done any comprehensive testing.

[I have one problem still to fix relating to the use of parseData in a sequenced update (I'm getting a 'wild' type error now, but I didn't get that error previously) but the code needs improving anyway since with the older version of temporal Links, the DateTime type didn't play nicely with MVU and the work around was to transform dates to strings and back again. This is no longer necessary with the current version of temporal Links.]

I'll now work with the XML curation interface and see if I can get that working.

vcgalpin:
> I'll now work with the XML curation interface and see if I can get that working.

Two syntax changes were required to get it up and running, neither of which had to do with temporal syntax.

SimonJF:
Hurrah! That's good news on both counts. Thanks for testing it out!

> [I have one problem still to fix relating to the use of parseData in a sequenced update (I'm getting a 'wild' type error now, but I didn't get that error previously) 

Hrm -- I think it's right that `parseDate` is wild since it can fail (for example given an incorrect date), and at present I don't think it's runnable on the DB (although that could be implemented, probably). Open to tweaking it, but I think having it tame is the best option for now.

dhil:
I will look at this in detail, but I don't think we should merge this PR before our next meeting on January 31.

An immediate red flag for me is the number of new keywords being introduced. I wager we can do it with less, or at least smarter than reserving fairly typical words.

core/query/sql.ml
```diff
@@ -293,11 +318,13 @@ class virtual printer =
...
           self#pp_insert ppf ins_table ins_fields ins_records
-      | With (z, q, q') ->
+      | With (z, q, qs) ->
           Format.fprintf ppf "with %s as (@[<v>%a@])\n%a"
             z
             pr_q q
-            pr_q q'
+            pp_qs qs
+      | Transaction qs ->
+          Format.fprintf ppf "BEGIN; %a COMMIT;" pp_qs qs
```
jamescheney, line 327: Here of course is a place where in general a DB might return an exception/error because of concurrent write activity that committed first.  I guess currently if this ever happens we will just get a hard run time failure. Of course, standalone update operations can fail too for example by violating integrity constraints.  Sounds like a great use case for effect / exception handlers.

core/query/temporalQuery.ml
```diff
@@ -0,0 +1,857 @@
...
+    ((Ir.var * string) * Q.t option * Q.t) ->
+    string ->
+    string ->
+    Sql.query =
+  fun table_types ((tbl_var, table), where, body) tt_from tt_to ->
+
+    let now_const = Constant.DateTime.now () in
+    let sql_now = Sql.Constant now_const in
+    let forever_const = Constant.DateTime.forever in
+    let sql_forever = Sql.Constant forever_const in
```
jamescheney, line 52: Could be hoisted since similar constants are used elsewhere

core/query/temporalQuery.ml
```diff
@@ -0,0 +1,857 @@
...
+    ])
+
+
+  let delete :
+    ((Ir.var * string) * Q.t option) ->
+    string -> (* tt_to field *)
+    Sql.query =
+    fun ((tbl_var, table), where) tt_to ->
+      let now = Sql.Constant (Constant.DateTime.now ()) in
+      let forever = Sql.Constant (Constant.DateTime.forever) in
```
jamescheney, line 152: Could be hoisted

core/query/temporalQuery.ml
```diff
@@ -0,0 +1,857 @@
...
+      Sql.query =
+      fun table_types ((tbl_var, table), where, body) from_field to_field ->
+        (* Valid time current updates are similar to transaction-time updates
+         * in structure: fetch the current rows, but with updated fields;
+         * insert the updated rows; and close off the previous ones.
+         * There's an extra step, though: ensure any future rows are also
+         * updated. *)
+        let open Sql in
+        let now_const = Constant.DateTime.now () in
+        let sql_now = Constant now_const in
```
jamescheney, line 282: Hoist

core/query/temporalQuery.ml
```diff
@@ -0,0 +1,857 @@
...
+        let sql_proj field = Project (tbl_var, field) in
+        let sql_binop op x1 x2 = Apply (op, [x1; x2]) in
```
jamescheney, line 284: This is OpHelpers.binop

core/query/temporalQuery.ml
```diff
@@ -0,0 +1,857 @@
...
+        let current_at =
+          sql_binop "&&"
```
jamescheney, line 286: This is `OpHelpers.and` etc.

core/query/temporalQuery.ml
```diff
@@ -0,0 +1,857 @@
...
+            OptionUtils.opt_map (base []) (StringMap.lookup k fields_with_time)
+            |> OptionUtils.from_option (Project (tbl_var, k))) table_types
+          |> StringMap.to_alist
+          (* Need to swap (col, val) pairs to (val, col) to fit select_clause AST,
+           * which mirrors "SELECT V as K" form in SQL *)
+          |> List.map (fun (k, v) -> (v, k)) in
+        (* Add "current at" clause to predicate *)
+        let sel_where =
+          match where with
+            | Some where -> Sql.Apply ("&&", [base [] where; current_at])
```
jamescheney, line 318: `OpHelpers.and`

core/query/temporalQuery.ml
```diff
@@ -0,0 +1,857 @@
...
+        (* Next, we need to insert the results *)
+        let insert =
+          Insert { ins_table = table; ins_fields = field_names;
+            ins_records = Query sel_var } in
+
+        (* Next: close off the old rows *)
+        let upd_current =
+          let pred =
+            let starts_before_now =
+              sql_binop "<" (sql_proj from_field) sql_now in
```
jamescheney, line 336: `OpHelpers.lt`

core/query/temporalQuery.ml
```diff
@@ -0,0 +1,857 @@
...
+            let var = Var.fresh_raw_var () in
+            let ins =
+              Sql.Insert {
+                ins_table = table;
+                ins_fields = field_names;
+                ins_records = Query var
+              } in
+            Sql.With (Sql.string_of_table_var var, sel, [ins]) in
+
+          (*  - Selection / insert #1: Old values at beginning of PA *)
```
jamescheney, line 446: What is PA?

core/query/temporalQuery.ml
```diff
@@ -0,0 +1,857 @@
...
+  module Delete = struct
+    let current :
+      ((Ir.var * string) * Q.t option) ->
+      string -> (* valid from field *)
+      string -> (* valid to field *)
+      Sql.query =
+      fun ((tbl_var, table), where) from_field to_field ->
+        let open Sql in
+        let now_const = Constant.DateTime.now () in
+        let sql_now = Constant now_const in
```
jamescheney, line 533: Hoist

core/query/temporalQuery.ml
```diff
@@ -0,0 +1,857 @@
...
+        let sql_binop op x1 x2 = Apply (op, [x1; x2]) in
+        let sql_proj field = Project (tbl_var, field) in
+
+        let current_at =
+          sql_binop "&&"
```
jamescheney, line 538: `OpHelpers`

core/query/temporalQuery.ml
```diff
@@ -0,0 +1,857 @@
...
+         * in structure: fetch the current rows, but with updated fields;
+         * insert the updated rows; and close off the previous ones.
+         * There's an extra step, though: ensure any future rows are also
+         * updated. *)
+        let open Sql in
+        let now_const = Constant.DateTime.now () in
+        let sql_now = Constant now_const in
+        let sql_proj field = Project (tbl_var, field) in
+        let sql_binop op x1 x2 = Apply (op, [x1; x2]) in
+        let current_at =
```
jamescheney, line 285: Are all the definitons of `current_at` the same?  Hoist if so

jamescheney (changes requested):
This looks good.  The changes are very broad though.  I have a few nitcpicks and some larger requests:

* I think it would be helpful to somewhere document that the intention is that behavior of existing Links programs (that don't use temporal tables) is preserved, and that if we ever want to get rid of the temporal functionality, it's just a matter of specializing all the places where Table types/constructors occur to the "current" case, and eliminating dead code.  

  As it is, lots of existing Links programs will need changes though since the Table type now has an additional parameter.  Perhaps one way to make this slicker could be to introduce a primitive TemporalTable type (i.e. renaming the current Table type internally) and define the legacy Table type as an alias for TemporalTable(Current,...).  That way existing code doesn't have to change (and shouldn't have to change back, if we we adopt a different design in the future).

* A second, longer-term thought is that this design makes changes to both the frontend (sugar/typechecking) and backend (IR and evaluation). It would be nice (and I think seems possible in principle) to desugar the temporal constructs to the existing IR, thus avoiding the need for changes to the backend.  However, that is definitely not something I'm suggesting trying to do now.

* Finally, it seems like there's a fair amount of room for cleanup/code deduplication in `temporalQuery.ml`, see the comments there.

SimonJF:
Thank you both for having a look and for the detailed suggestions. I'm entirely in agreement that we shouldn't merge this prior to a dev meeting, and that the changes are a lot wider-ranging than I'd hoped (really for something that's actually not conceptually huge).

Regarding keywords: I have the same concern (esp. regarding `current`, `valid`, and `to`). There were places I tried to do the usual `VARIABLE` trick but ran into `reduce/reduce` errors, but granted I didn't try too hard and there's certainly scope for reducing the number of new keywords.

I'll have a go at doing these changes toward the end of the week.

core/unify.ml
```diff
@@ -388,6 +388,15 @@ let rec unify' : unify_env -> (datatype * datatype) -> unit =
 
   let ut = unify' rec_env in
   let ur = unify_rows' rec_env in
+  let utmp (t1, t2) =
```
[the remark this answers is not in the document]
SimonJF: Yes, that's right. In practice, I don't see many places where we would want to quantify over the temporality due to typing (in short I don't think the additional complexity is worth it) but I can rethink if you like.

core/types.mli
```diff
@@ -374,8 +376,10 @@ val make_list_type : datatype -> datatype
 val make_process_type : row -> datatype
 val make_record_type  : datatype field_env -> datatype
 val make_variant_type : datatype field_env -> datatype
-val make_table_type : datatype * datatype * datatype -> datatype
+val make_table_type : Temporality.t * datatype * datatype * datatype -> datatype
```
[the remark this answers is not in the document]
SimonJF: Almost certainly!

core/evalir.ml
```diff
@@ -585,6 +585,25 @@ struct
              | `Bool true     -> t
              | `Bool false    -> e
              | _              -> eval_error "Conditional was not a boolean")
+
+  and eval_nested_query env cont (db: Value.database) pkg =
```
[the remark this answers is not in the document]
SimonJF: Yes, it's entirely because of nested records: the normal forms are nested (since the results are of the form `(!data = ..., !from = ..., !to)` (a la the Provenance translation Stefan did). The actual querying can probably be done with the flat policy; I think I insisted on nested-ness for typechecking.

We'll need the full generality of shredding when we think of shredded temporal queries, but I think that's along the line a bit.

core/query/queryLang.ml
```diff
@@ -173,8 +173,18 @@ let field_types_of_row r =
...
+     * the period-stamping fields included in the temporal metadata. *)
+    let dt x = (x, Types.Primitive Primitive.DateTime) in
+    let metadata_fields =
+        OptionUtils.opt_app (fun (x, y) -> [dt x; dt y]) [] temporal_fields
+    in
+    let declared_fields = field_types_of_spec_map fields in
+    (* Add metadata fields *)
+    List.fold_left
+      (fun acc (k, v) -> StringMap.add k v acc)
+      declared_fields metadata_fields
```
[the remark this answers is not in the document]
SimonJF: Love it!

core/query/temporalQuery.ml
SimonJF: Agreed on `forever`, but given that `now` is side-effecting, hoisting that has little benefit.

core/query/queryLang.ml
```diff
@@ -525,260 +535,379 @@ let check_policies_compatible env_policy block_policy =
     raise (Errors.runtime_error error)
 
 (* convert a regexp to a like if possible *)
-  let rec likeify v =
```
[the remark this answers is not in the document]
SimonJF: It's all whitespace apart from the change in the debug message. I've put back the indentation but annoyingly, GH still doesn't seem to want to show the proper diff.

core/types.mli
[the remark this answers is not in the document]
SimonJF: Upon trying -- it actually has the opposite effect of requiring even more edits!

core/query/sql.ml
```diff
@@ -13,15 +13,21 @@ type query =
...
     }
   | Update    of {
       upd_table: table_name;
       upd_fields: (string * base) list;
       upd_where: base option
     }
   | Delete    of { del_table: table_name; del_where: base option }
-  | With      of table_name * query * query
+  | With      of table_name * query * query list
+  | Transaction of query list (* SQL Transaction: Complete atomically*)
```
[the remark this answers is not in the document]
SimonJF: Yes, at present there's no support for transactions other than in the SQL DSL.

It would definitely be good to have some support for this. We'd probably need a better idea of exceptions, for a start, and I think after that we can do a design for purely update-based transactions. I wouldn't know the best design for then intertwining with regular and shredded queries, especially when relying things like `insert returning`, but I think it could be fun to look at.

core/query/sql.ml
SimonJF: Ah, you read my mind! Yes, indeed, currently this will net you a runtime failure (like with the lens code). 
Maybe we should think of working out a proper design for #565.  I think it's outside of the scope of this patch though.

SimonJF:
OK, so made a dent.

With some lexer hacking we're down to `vt_insert`, `tt_insert`, `valid`, `to` as the new keywords. Ideally `valid`, `to` wouldn't be reserved either so there's still room for improvement.

Still also need to do the TableHandle backwards compatibility but I'll probably tackle that next week.

core/evalir.ml
```diff
@@ -601,7 +620,11 @@ struct
           let sort = Type.sort t in
           value env table >>= fun table ->
           match table with
-            | `Table (((db,cstr), table, _, _) as tinfo) ->
+            | `Table tinfo ->
+              (* I can't get this to work without turning the warning off, even
+                 with "Value" in scope. *)
+              let [@warning "-40"] (db, cstr) = tinfo.database in
```
dhil, line 626: Disabling this warning is unacceptable! The correctness of this code now depends on the order of type inference. It is really bad and it **will** come back to bite us later. It can be difficult to understand the origin of these warnings, however, this particular case is because of heavy re-use of module names and record labels. The correct thing to do here is to make type disambiguation possible. To make it possible you must introduce namespaces.

I will fix these instances for you. Next time please ask for help.

SimonJF: This was a temporary workaround for issues encountered while merging, which I didn't intend to be in the PR. Apologies. Thank you for sorting it.

dhil (approved):
I will merge this PR shortly. 

This PR introduces a lot of breaking changes due to the array of new keywords. You need to update the file `CHANGES.md` accordingly.

SimonJF:
Thanks @dhil. Can you hold off until noon tomorrow? There are a few last things I'd like to do (including the `CHANGES.md` change).

core/query/sql.ml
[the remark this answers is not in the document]
jamescheney: Tthe SQL queries resulting from a single shredded query ought to be wrapped in a transaction anyway, and aren't; we could do this now, using the new transaction case in Sql.query.  Having a design for source-level transactions seems like a separate (future) issue.

core/query/sql.ml
jamescheney: Agreed.

jamescheney (approved):
I've made one tiny nitpicky suggestion but otherwise looks good overall.

core/types.ml
```diff
@@ -1543,8 +1557,8 @@ and normalise_datatype rec_names t =
   | Record row              -> Record (nr row)
   | Variant row             -> Variant (nr row)
   | Effect row              -> Effect (nr row)
-  | Table (r, w, n)         ->
-     Table (nt r, nt w, nt n)
+  | Table (t, r, w, n)         ->
+     Table (t, nt r, nt w, nt n)
```
jamescheney, line 1561: A tiny nitpick: 
```suggestion
  | Table (t, r, w, n)      -> Table (t, nt r, nt w, nt n)
```

SimonJF:
Thanks for the reviews. I followed through with @jamescheney's suggestion to rename the general table type to `TemporalTable` and make `TableHandle(r, w, n)` an alias for `TemporalTable(Current, r, w, n)`.

Happy for this to go in when you are.

dhil:
Merging this next!

PR #1082, opened 2022-01-31
Fixes #1075

### LINKED ISSUE #1075: Mixing normalization and concatenation
This problem was reported by @jamescheney in #1067. It's unrelated to deduplication, but caused by a bug in the mixing normaliser:

```
links> query {for (x <-- factorials) [(a=x.i),(a=x.f)]};
[(a = 1), (a = 1), (a = 2), (a = 2), (a = 3), (a = 6)] : [(a:Int)]
links> query mixing {for (x <-- factorials) [(a=x.i),(a=x.f)]};
***: Fatal error : body 
links> query nested {for (x <-- factorials) [(a=x.i),(a=x.f)]};
[(a = 1), (a = 2), (a = 3), (a = 1), (a = 2), (a = 6)] : [(a:Int)]
links> query delat {for (x <-- factorials) [(a=x.i),(a=x.f)]};
***: Fatal error : body
```

The mixing normaliser does not split comprehensions along concatenation correctly. It then returns an expression in the shape `U{ {a}++{b} | x <-- factorials }` which is not in normal form, so the conversion to SQL fails. The culprit is one between `MixingQuery.reduce_for_source/reduce_for_body` (or perhaps their combination).

wricciot:
This is straightforward: the mixing normaliser (`mixingQuery.ml`) has an issue with `For` when its body is a `Concat`: the comprehension would normally be split along concatenation, but this does not happen. 

The issue is caused by the changes made to normalisation to allow it to consider normal forms where comprehension generators can be complex queries: in order for these changes to work, we must add a recursive call to `MixingQuery.reduce_for_body` (which until now was non-recursive).

The old normaliser doesn't show this bug, for reasons not entirely clear (but related to the simpler structure of homogeneous bag normal forms). However it's now advisable that the code responsible for normalising comprehensions should be refactored to make its rationale more understandable (without a clear rationale, every further change is at risk of creating new bugs).

### DIFF
--- a/core/query/mixingQuery.ml
+++ b/core/query/mixingQuery.ml
@@ -211,7 +211,6 @@ let rec reduce_where_then (c, t) =
 let rec reduce_for_body (gs, os, body) =
   let rb = reduce_for_body in
   match body with
-    (* | Concat []                 -> body *)
     | Q.For (_, gs', os', body') -> rb (gs @ gs', os @ os', body')
     (* | Prom _ as u               ->
           let z = Var.fresh_raw_var () in
--- a/core/query/mixingQuery.ml
+++ b/core/query/mixingQuery.ml
@@ -208,10 +208,11 @@ let rec reduce_where_then (c, t) =
   | _ ->
     Q.If (c, t, Q.Concat [])
 
-let reduce_for_body (gs, os, body) =
+let rec reduce_for_body (gs, os, body) =
+  let rb = reduce_for_body in
   match body with
     (* | Concat []                 -> body *)
-    | Q.For (_, gs', os', body') -> Q.For (None, gs @ gs', os @ os', body')
+    | Q.For (_, gs', os', body') -> rb (gs @ gs', os @ os', body')
     (* | Prom _ as u               ->
           let z = Var.fresh_raw_var () in
           let tyz = type_of_expression u in
@@ -220,6 +221,7 @@ let reduce_for_body (gs, os, body) =
           For (None, gs @ [(z, u)], [] (* os *), (Singleton vz)) *)
     (* make sure when we reach this place, gs can NEVER be empty
       | _ when gs = [] (* && _os = [] *) -> body *)
+    | Q.Concat vs -> reduce_concat (List.map (fun v -> rb (gs, os, v)) vs)
     | _                         -> Q.For (None, gs, os, body)
 
 let rec reduce_for_source : Q.env -> var * Q.t * Types.datatype -> (Q.env -> (Q.t list -> Q.t list) -> Q.t) -> Q.t =
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,8 @@
   For instance, writing `/\ [a, e::Row] { foo }` abstracts the expression `foo`
   over type variable `a` and row variable `e`. Here, `foo` must have a unique
   type and must be pure (to satisfy the value restriction).
+* Fixed a bug in "mixing" query normalisation, which prevented certain queries using 
+  concatenation inside `for` statements from being correctly converted to SQL.
 
 # 0.9.5
 
--- a/tests/database/mixing.links
+++ b/tests/database/mixing.links
@@ -4,7 +4,8 @@ var bagtable = table "bagtable" with (i : Int) from db;
 fun test() {
     assertEq(query mixing { distinct(bagtable) }, [(i=0)]);
     assertEq(query mixing { dedup(for (x <-- bagtable) [x]) }, [(i=0)]);
-    assertEq(query mixing { for (x <- distinct(bagtable)) [x] }, [(i=0)])
+    assertEq(query mixing { for (x <- distinct(bagtable)) [x] }, [(i=0)]);
+    assertEq(query mixing { for (x <-- bagtable) [(i=0,j=x.i)] }, [(i=0,j=0),(i=0,j=0)])
 }
 
 test()
--- a/tests/database/mixing.links
+++ b/tests/database/mixing.links
@@ -5,7 +5,7 @@ fun test() {
     assertEq(query mixing { distinct(bagtable) }, [(i=0)]);
     assertEq(query mixing { dedup(for (x <-- bagtable) [x]) }, [(i=0)]);
     assertEq(query mixing { for (x <- distinct(bagtable)) [x] }, [(i=0)]);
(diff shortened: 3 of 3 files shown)


core/query/mixingQuery.ml
```diff
@@ -208,10 +208,11 @@ let rec reduce_where_then (c, t) =
   | _ ->
     Q.If (c, t, Q.Concat [])
 
-let reduce_for_body (gs, os, body) =
+let rec reduce_for_body (gs, os, body) =
+  let rb = reduce_for_body in
   match body with
     (* | Concat []                 -> body *)
```
jamescheney, line 214: Is it worth removing this commented-out line now that Concat is handled uniformly by a new case?

jamescheney:
If it helps, here is the result of manually running the queries from `mixing.links` with sqlite3:

```
Finishing process MAIN
 _     _ __   _ _  __  ___
/ |   | |  \ | | |/ / / ._\
| |   | | , \| |   /  \  \
| |___| | |\ \ | |\ \ _\  \
|_____|_|_| \__|_| \_|____/
Welcome to Links version 0.9.6 (Burghmuirhead)
links> var db = database "links";
debug: found loadable database driver
- /home/jcheney/src/tmp/links-issue1075/_build/install/default/lib/links-sqlite3/links_sqlite3.cmxs
- /home/jcheney/src/tmp/links-issue1075/_build/install/default/share/links-sqlite3/links_sqlite3_dependencies.json

registering driver for sqlite3
Finishing process MAIN
db = (database sqlite3:links : Database
links> var bagtable = table "bagtable" with (i : Int) from db;
Finishing process MAIN
bagtable = (table bagtable) : TableHandle((i:Int),(i:Int),(i:Int))
links> query mixing { distinct(bagtable) }
...... ;
Query.eval time: 44
Running query: 
select distinct (t2340."i") as "i"
from (select distinct *
from "bagtable" as t2340) as t2340
select distinct (t2340."i") as "i"
from (select distinct *
from "bagtable" as t2340) as t2340
Finishing process MAIN
[(i = 0)] : [(i:Int)]
links> query mixing { dedup(for (x <-- bagtable) [x]) };
Query.eval time: 35
Running query: 
select distinct (t2347."i") as "i"
from (select distinct *
from "bagtable" as t2347) as t2347
select distinct (t2347."i") as "i"
from (select distinct *
from "bagtable" as t2347) as t2347
Finishing process MAIN
[(i = 0)] : [(i:Int)]
links> query mixing { for (x <- distinct(bagtable)) [x] };
Query.eval time: 87
Running query: 
select (t2355."i") as "i"
from (select distinct (t2354."i") as "i"
from (select distinct *
from "bagtable" as t2354) as t2354) as t2355
select (t2355."i") as "i"
from (select distinct (t2354."i") as "i"
from (select distinct *
from "bagtable" as t2354) as t2354) as t2355
Finishing process MAIN
[(i = 0)] : [(i:Int)]
links> query mixing { for (x <-- bagtable) [(a=x.i),(a=x.i)] };
Query.eval time: 39
Running query: 
(select (t2361."i") as "a"
from "bagtable" as t2361)
union all
(select (t2361."i") as "a"
from "bagtable" as t2361)
(select (t2361."i") as "a"
from "bagtable" as t2361)
union all
(select (t2361."i") as "a"
from "bagtable" as t2361)
***: Error: Sqlite3.Error("Sqlite3.prepare: near \"(\": syntax error") 
links> 
```
Trying the query in sqlite3 directly, it seems that 
```
select (t2361."i") as "a" from "bagtable" as t2361
```
works but 
```
(select (t2361."i") as "a" from "bagtable" as t2361)
```
is a syntax error, likewise
```
select (t2361."i") as "a" from "bagtable" as t2361 union all select (t2361."i") as "a" from "bagtable" as t2361
```
works but
```
(select (t2361."i") as "a" from "bagtable" as t2361) union all (select (t2361."i") as "a" from "bagtable" as t2361)
```
is a syntax error.  I guess the parentheses are not strictly necessary in SQL (at least as long as all of the operations are "union all") since  Postgres and Mysql must allow a more relaxed syntax.  How strange!

wricciot:
> ```
> select (t2361."i") as "a" from "bagtable" as t2361 union all select (t2361."i") as "a" from "bagtable" as t2361
> ```
> 
> works but
> 
> ```
> (select (t2361."i") as "a" from "bagtable" as t2361) union all (select (t2361."i") as "a" from "bagtable" as t2361)
> ```
> 
> is a syntax error. I guess the parentheses are not strictly necessary in SQL (at least as long as all of the operations are "union all") since Postgres and Mysql must allow a more relaxed syntax. How strange!

So much for "one extra parenthesis never hurt anyone"!

Thanks!

jamescheney:
It seems the SQL standard does not require that parentheses are allowed here, see: https://stackoverflow.com/questions/4653124/what-does-the-sql-standard-say-about-parentheses-in-sql-union-except-intersect-s

Perhaps interestingly if you mix UNION and UNION ALL at the same level then Q UNION Q UNION ALL Q and Q UNION ALL Q UNION Q yield different results.  It seems the first one gets evaluated as (Q UNION Q) UNION ALL Q (i.e. you get dedup(Q) UNION ALL Q) while the second gets evaluated as (Q UNION ALL Q) UNION Q (you get dedup(Q) since the deduplicating union happens last).  

One can emulate the parentheses using subqueries however, e.g. `Q1 UNION (Q2 UNION ALL Q3)` can be written as `Q1 UNION SELECT * FROM (Q1 UNION Q2)`.  So if there are places where we might generate mixes of union and union all (or other set operators) then it might make sense to program defensively here to avoid relying on syntax extensions not required by SQL.

wricciot:
Sigh, welcome to the bureau of complication of simple affairs. Thanks!

jamescheney:
I suppose one way to fix this could be to add a method `parenthesize` to the database class/interface that abstracts over a safe way to parenthesize SQL queries; MySQL and Postgres could implement this by just adding parentheses while sqlite3 could do it the more conservative/defensive way.  This seems a little ugly though, especially if we want pretty-printing to play nicely with it.  So maybe better to just do the horrible `select * from (...)` thing everywhere.  Now that I think of it though, this only seems necessary for arguments to a `Union` that are not `Select`s, since our `Union` construct is homogeneous (that is, with a single `Union` you can only do a bunch of "union all"s, or a bunch of "union"s, not a mix).

jamescheney:
Merging in the fix to #1090 should hopefully make CI happy now.

dhil:
Is this one good to go in now?

wricciot:
It is ready as far as I'm concerned.

PR #1084, opened 2022-02-01
Secure connections

dhil:
This patch makes it possible for the built-in webserver to serve
requests via https.

To enable secure connections, you must first obtain an adequate
certificate and key, e.g. via Let's Encrypt or a self-signed
certificate. The latter can be useful for testing, e.g. the following
command starts an interactive process to create a self-signed
certificate (that uses 4096 bits RSA encryption and is valid for 365 days):

```shell
openssl req -x509 -newkey rsa:4096 -keyout server.key -out server.crt -days 365 -nodes
```

After obtaining a valid certificate, you must tell Links to run in SSL
mode and you must also tell it how to locate the `key` and `crt` file. This can be done via a configuration file, e.g.

```
# ssl.config
ssl=true
ssl_cert_file=server.crt
ssl_key_file=server.key
```

Then running `./links --config=ssl.config <file.links>` will cause the
built-in webserver to only serve requests via https.

When a webpage is served via https, then the websocket layer will
automatically communicate via the wss protocol.

Note the error handling around certificates is fairly minimal. For
example, if a certificate file or key file is not found it dumps a
stack trace.

### DIFF
--- a/core/errors.ml
+++ b/core/errors.ml
@@ -54,6 +54,7 @@ exception DisabledExtension of Position.t option * (string * bool) option * stri
 exception PrimeAlien of Position.t
 exception ForbiddenClientCall of string * string
 exception MissingBuiltinType of string
+exception MissingSSLCertificate
 
 exception LocateFailure of string
 let driver_locate_failure driver = LocateFailure driver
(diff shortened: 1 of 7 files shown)


dhil:
I have successfully tested this patch with the chat server example. Secure chat FTW.

PR #1087, opened 2022-02-02
[Draft] JavaScript compiler / runtime refactor

dhil:
This patch refactors the JavaScript compiler and runtime to collect some of the technical debt that has amassed over the years. This refactoring has unveiled several bugs in the Links JavaScript runtime.

### DIFF
--- a/core/irtojs.ml
+++ b/core/irtojs.ml
@@ -79,18 +79,18 @@ module Code = struct
 
   module Runtime = struct
 
-    module LINKS = struct
-      let project = Var "LINKS.project"
-      let erase   = Var "LINKS.erase"
-      let union   = Var "LINKS.union"
-      let remoteCall = Var "LINKS.remoteCall"
-      let xml = Var "LINKS.XML"
-      let eq = Var "LINKS.eq"
-      let neq = Var "LINKS.neq"
-      let gt = Var "LINKS.gt"
-      let lt = Var "LINKS.lt"
-      let gte = Var "LINKS.gte"
-      let lte = Var "LINKS.lte"
+    module Links = struct
+      let project = Var "_$Links.project"
+      let erase   = Var "_$Links.erase"
+      let union   = Var "_$Links.union"
+      let remoteCall = Var "_$Links.remoteCall"
+      let xml = Var "_$Links.XML"
+      let eq = Var "_$Links.eq"
+      let neq = Var "_$Links.neq"
+      let gt = Var "_$Links.gt"
+      let lt = Var "_$Links.lt"
+      let gte = Var "_$Links.gte"
+      let lte = Var "_$Links.lte"
     end
 
     module CONSTANTS = struct
@@ -98,13 +98,13 @@ module Code = struct
     end
 
     module List = struct
-      let nil = Var "LINKEDLIST.Nil"
-      let cons = Var "LINKEDLIST.Cons"
+      let nil = Var "_$List.Nil"
+      let cons = Var "_$List.Cons"
 
-      let hd = Var "LINKEDLIST.head"
-      let tl = Var "LINKEDLIST.tail"
+      let hd = Var "_$List.head"
+      let tl = Var "_$List.tail"
 
-      let append = Var "LINKEDLIST.append"
+      let append = Var "_$List.append"
     end
 
     module JSON = struct
@@ -119,6 +119,9 @@ module Code = struct
 
   module Aux = struct
     let call f args = Call (f, args)
+
+    let project record label =
+      call Runtime.Links.project [record; label]
 
     module List = struct
       let cons x xs =
@@ -191,22 +194,22 @@ let __kappa = Code.ObjectContinuation.__kappa
 (**
   Required runtime support (documenting any JavaScript functions used):
 
-  LINKS.concat(a, b)
+  _$Links.concat(a, b)
      concatenate two sequences: either strings or lists
-  LINKS.accum(f, i)
+  _$Links.accum(f, i)
     concatMap: apply f to every element of the sequence `i' and concatenate the results.
   _plus, _minus, etc.
     curried function versions of the standard arithmetic operators
-  LINKS.XML(tag, attrs, children)
+  _$Links.XML(tag, attrs, children)
     create a DOM node with name `tag'
                        and attributes `attrs' (a dictionary)
                        and children `children' (a sequence of DOM nodes and strings)
-  LINKS.union(r, s)
+  _$Links.union(r, s)
     return the union of the records r and s
     precondition: r and s have disjoint labels
-  LINKS.project(record, label)
+  _$Links.project(record, label)
     project a field of a record
-  LINKS.erase(record, label)
+  _$Links.erase(record, label)
     return a record like "record" but without the field labeled "label"
 
   _start(tree)
@@ -276,7 +279,7 @@ module Js_CodeGen : JS_CODEGEN = struct
            PP.vsep (punctuate " " (List.map (fun (name, vars, body, _loc) -> show_func name (Fn (vars, body))) defs)) ^^
              break ^^ show rest
         | Fn _ as f -> show_func "" f
-        | Call (Var "LINKS.project", [record; label]) ->
+        | Call (Var "_$Links.project", [record; label]) ->
            maybe_parenise record ^^ (brackets (show label))
         | Call (Var "hd", [list;kappa]) ->
            (maybe_parenise kappa) ^^ PP.text "hd" ^^ (parens (  maybe_parenise list))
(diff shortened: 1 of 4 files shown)


lib/js/jslib.js
```diff
@@ -4217,27 +4345,29 @@ var debugChromiumGC = _debugChromiumGC;
 function _stringEq(l, r) {
   return l === r;
 }
-var stringEq = _stringEq;
+const stringEq = _stringEq; // TODO(dhil): these definitions appear
```
djedr: Hello,
I was the intern who wrote this nearly 8 years ago.

I am still subscribed to this repo after 8 years, partly out of sentiment -- my account was originally created to work on Links. I've been hanging around GitHub a lot recently, I saw the title of your PR and clicked out of curiosity.

I think I can provide some context.

All the functions starting here:
https://github.com/links-lang/links/blob/master/core/lib.ml#L1420

`(* LINKS GAME LIBRARY *)`

up to here:

https://github.com/links-lang/links/blob/master/core/lib.ml#L1555

`(* END OF EQUALITY FUNCTIONS *)`

were written for the purpose of implementing these example games:

https://github.com/links-lang/links/tree/master/examples/games

I suspect they haven't been used elsewhere, as I don't think they were documented.

The idea was to make the games run fast enough to be playable. Optimizations were necessary to do that.

They were achieved with by any means I could come up with, not all of which were necessarily pretty or sensible as I might've not known any better -- I was far from competent on anything I was doing. For example in this definition of `_objectEq`:

```js
function _objectEq(l, r) {
  return l._label === r._label && (l._value === r._value || (l._value === {} && r._value === {}));
}
```

I'm not sure what I was thinking when I wrote `(l._value === {} && r._value === {})`. It seems to be useless nonsense (fortunately harmless).

Anycase I think these functions can be safely refactored as long as it is verified that the game examples run sensibly fast.

Hope this helps.

Cheers,
Dariusz

PS I'd create a PR, but I couldn't get past the setup. I hope this is better than nothing.

dhil: Hi Dariusz! 

Thanks for the clarification. In this patch, I have updated and implemented all the infix operators on the client side. The performance difference between my definitions and those used by the games should be negligible. The JS engines have become really apt at optimising code. 

I think my comment about the brokenness of those functions refer to the fact that they drop the current continuation, meaning they cannot be used in an effectful or higher-order context. Though, they were probably OK because they were marked `PURE` in `lib.ml`, so the JavaScript compiler should eta-expand them accordingly.

Ideally, I think we want the JavaScript compiler to be type-directed such that it can insert the efficient native versions of equality, relational, string operators.

dhil:
I will merge this patch later today. In a follow-up patch I will further trim down the runtime and split it into non-web and web dependent modules.

core/irtojs.ml
```diff
@@ -604,54 +801,60 @@ module CPS_Compiler: functor (K : CONTINUATION) -> sig
...
-                   ("_value", Lit (UnixTimestamp.of_calendar ts |> string_of_float))]
-         | Constant.DateTime Timestamp.Infinity ->
-             Dict [("_type", strlit "infinity")]
-         | Constant.DateTime Timestamp.MinusInfinity ->
+    | Ir.Constant c ->
+      begin
+        let open Constant in
+        match c with
+        | Constant.Int v   -> Lit (string_of_int v)
+        | Constant.Float v -> Lit (string_of_float' v)
```
slindley, line 828: Get rid of `Constant.`

core/irtojs.ml
```diff
@@ -830,83 +1037,91 @@ end = functor (K : CONTINUATION) -> struct
...
+      let f = strip_poly f in
+      begin
+        match f with
+        | Ir.Variable f ->
+          let f_name = VEnv.find f env in
+          begin
+            match vs with
+            | [l; r] when StringOp.is f_name ->
+              (* TODO(dhil): It'd be safer and more robust to hoist
+                 and let bind `gv l` and `gv r`. *)
```
slindley, line 1058: good point

core/irtojs.ml
```diff
@@ -921,54 +1136,52 @@ end = functor (K : CONTINUATION) -> struct
...
+      | Ir.DoOperation (name, args, _) ->
          let maybe_box = function
            | [v] -> gv v
            | vs -> Dict (List.mapi (fun i v -> (string_of_int @@ i + 1, gv v)) vs)
          in
-         let cons k ks =
-           Call (Var "_Cons", [k;ks])
-         in
-         let nil = Var "Nil" in
+         let nil = Runtime.List.nil in
```
slindley, line 1184: Consider inlining

core/irtojs.ml
```diff
@@ -992,45 +1205,24 @@ end = functor (K : CONTINUATION) -> struct
...
-         | Shallow -> raise (Errors.runtime_error "CPS compilation of shallow handlers is currently not supported")
-         | Deep params ->
+         | Ir.Shallow -> raise (Errors.runtime_error "CPS compilation of shallow handlers is currently not supported")
+         | Ir.Deep params ->
             let translate_parameters params =
               let is_parameterised = List.length params > 0 in
               let param_ptr_binder =
-                Var.fresh_binder
-                  (Var.make_local_info (Types.Not_typed, "_param_ptr"))
+                Var'.fresh_binder
```
slindley, line 1224: Rename `Var'` to `Ir.Var`

core/irtojs.mli
```diff
@@ -8,39 +8,58 @@ module VEnv = Env.Int
...
+    type t = string
+    [@@deriving show]
+  end
+
+  module Label: sig
+    type t = string
+    [@@deriving show]
+  end
+
+  type t = Var    of Var.t
```
slindley, line 22: document this data type

core/irtojs.mli
```diff
@@ -8,39 +8,58 @@ module VEnv = Env.Int
...
+         | LetFun of (Var.t * Var.t list * t * Ir.location) * t
+         | LetRec of (Var.t * Var.t list * t * Ir.location) list * t
+         | Call   of t * t list
+         | Unop   of Var.t * t
+         | Binop  of t * Var.t * t
+         | If     of t * t * t
+         | Switch of t * t stringmap * t option
+         | Dict   of (Label.t * t) list
+         | Arr    of t list
+         | Select of t * Label.t
```
slindley, line 35: Consider renaming - `Select` clashes with session typing constructs

core/irtojs.mli
```diff
@@ -8,39 +8,58 @@ module VEnv = Env.Int
...
+
+         | LetFun of (Var.t * Var.t list * t * Ir.location) * t
+         | LetRec of (Var.t * Var.t list * t * Ir.location) list * t
+         | Call   of t * t list
+         | Unop   of Var.t * t
+         | Binop  of t * Var.t * t
+         | If     of t * t * t
+         | Switch of t * t stringmap * t option
+         | Dict   of (Label.t * t) list
+         | Arr    of t list
```
slindley, line 34: consider removing `Arr`

lib/js/jslib.js
```diff
@@ -1029,7 +1444,7 @@ const LINKS = new (function() {
   /**
    * Perform a client call as specified in _callPackage, then re-invoke
    * the server using _remoteContinue
-   * NOTE: variables defined within this function could shadow
+   * NOTE: letiables defined within this function could shadow
```
slindley: letiables!

lib/js/jslib.js
```diff
@@ -2547,112 +2800,67 @@ function _replaceDocument(tree) {
...
-        } else {
-          _handlingEvent = false;
-          _current_pid = active_pid;
-          throw e;
-        }
-      }
-    }
-  }
-}
+let _eventHandlers = {};
```
slindley: should this be a global variable?

lib/js/jslib.js
```diff
@@ -3006,97 +3142,76 @@ function get_process_id(pid) {
...
 //   continuation; hence there is no _recv form (direct-style).
 function recv(kappa) {
-  DEBUG.assert(
+  _$Debug.assert(
     arguments.length == 1,
     'recv received ' + arguments.length + ' arguments, expected 1'
   );
-  //DEBUG.assert(_current_pid != _mainPid,
-  //           "Cannot call recv() in main process.");
-  DEBUG.assert(
```
slindley: why has this assertion been deleted?

dhil: I remember now. It's because `mainPid` is a hidden property now. I could expose it again to keep this assertion alive.

dhil:
I ran jslib.js through jshint (a JavaScript linter). It gave me 559 warnings/errors. My latest commit addresses the severe warnings/errors (some warnings/errors are false positive as the linter does not have knowledge of the glue code generated by the compiler). I have brought the number down to 147 warnings/errors. I believe all but one of these warnings/errors are harmless.

dhil:
I will merge this when the CI says OK. 

@SimonJF I have tested the session exception extension to the best of my ability. However, as this is a 1000+ refactoring I am sure there are new bugs in the runtime, in particular in the communication stuff as it was rather difficult to refactor. Ping me if you stumble upon some unexpected behaviour following the merge of this patch.

PR #1091, opened 2022-02-03
Fix #1090

### LINKED ISSUE #1090: SQL-standard generation for sqlite3
As found in #1082, SQLite3 doesn't accept queries with parentheses in UNION or UNION ALL (and presumably other set operations).  This manifests in a query like:
```
query  { for (x <-- factorials) [(a=x.i),(a=x.i)] }
```
whose normal form will have a UNION ALL at the top level (the problem originally manifested in a query  from `mixing.links` in #1082 but the problem has nothing at all to do with mixing normalization).

Fixing this should make #1082 pass CI.

Intended fix: avoid generating parentheses in parts of SQL queries where their use is nonstandard and use `select * from (...)` instead to group subexpressions where necessary.

jamescheney:
This uses SQL-safe parentheses conditionally: if a query appearing under a UNION/UNION ALL is a SELECT then we just print it, otherwise we wrap the subquery in a silly "SELECT * FROM (...)" that seems to be the only fully SQL-standardized way to do this.

Let's see what the CI thinks of this.

### DIFF
--- a/core/query/sql.ml
+++ b/core/query/sql.ml
@@ -274,13 +274,14 @@ class virtual printer =
             Format.pp_print_string (order_by_clause n)
       | Union (mult, qs, n) ->
         let pp_sep_union ppf () = Format.fprintf ppf "\nunion %a\n" pp_all mult in
-        let pp_value ppf x =
+        let pp_union_term ppf x =
           match x with  (* parenthesize safely wrt SQL standard *)
           | Select _ -> pr_q ppf x
           | _ -> Format.fprintf ppf "select * from (%a)" pr_q x
-        in
-        Format.fprintf ppf "%a%a"
-          (Format.pp_print_list ~pp_sep:pp_sep_union pp_value) qs
+        in 
+          (* defensively add a select in case there is an order by clause *)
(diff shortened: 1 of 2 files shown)


core/query/sql.ml
```diff
@@ -274,7 +274,11 @@ class virtual printer =
             Format.pp_print_string (order_by_clause n)
       | Union (mult, qs, n) ->
         let pp_sep_union ppf () = Format.fprintf ppf "\nunion %a\n" pp_all mult in
-        let pp_value ppf x = Format.fprintf ppf "(%a)" pr_q x in
+        let pp_value ppf x =
+          match q with  (* parenthesize safely wrt SQL standard *)
+          | Select _ -> Format.fprintf ppf "%a" pr_q x
```
wricciot, line 279: I might be confused, but isn't this equivalent to `pr_q ppf x`?

jamescheney: Sorry, yes, there was a silly typo there

jamescheney: I was just adapting what was there already, happy to simplify this if I can

PR #1098, opened 2022-02-09
Allow direct dispatch of messages to MVU event loop

SimonJF:
# Overview
This fairly simple patch makes it possible to dispatch an MVU message from outside of the event loop. This is particularly useful, for example, when dealing with a persistent, stateful thread which is receiving messages from a server.

# Implementation
* Introduce type alias `MvuHandle(msg)`
* Introduce variants of the `run` functions: `runHandle`, `runCmdHandle`, `runSimpleHandle` which return an `MvuHandle(msg)` rather than the unit value
* Add `Mvu.dispatch : (msg, MvuHandle(msg)) ~> ()`, which directly dispatches a message to the MVU loop

The patch is structured in such a way that the change is backwards-compatible.
See `examples/mvu/dispatch.links` for an example.

### DIFF
--- a/examples/mvu/dispatch.links
+++ b/examples/mvu/dispatch.links
@@ -1,0 +1,51 @@
+open import Mvu;
+import MvuHTML;
+import MvuAttrs;
+
+typename Msg = [| Incr |];
+
+typename Model = Int;
+
+fun incrModel(hndl) {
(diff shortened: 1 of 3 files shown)


dhil (approved):

SimonJF:
@dhil if you're happy then I'm happy; feel free to merge at your leisure

PR #1105, opened 2022-02-18
Add missing access point functions on client

SimonJF:
The Links standard library implements four functions for creating access points: `new`, `newServerAP`, `newClientAP`, and `newAP`. Of these, only `new` is implemented on the client.

This simple patch implements (at least stubs for) the other three. We use the same assumptions as `spawn`, and only support creating APs on the local client rather than the server or another client. `newClientAP` is simply wired to `new`.

This allows us to run, e.g.,

```
fun main() {
    var _ = spawnClient {
        var ap = newClientAP();
        var _ = spawn { var s = accept(ap); close(send((), s)) };
        var t = request(ap);
        var ((), t) = receive(t);
        close(t);
        print("done")
    };
    page <html></html>
}

serveThis(main)
```

### DIFF
--- a/lib/js/jslib.js
+++ b/lib/js/jslib.js
@@ -378,8 +378,8 @@ const WEBSOCKET = {
         path = "";
       }
     }
-
-    return "ws://" + loc.host + "/" + path + "/" + _client_id;
+    let protocol = location.protocol === "https:" ? "wss:" : "ws:";
+    return protocol + "//" + loc.host + "/" + path + "/" + _client_id;
   },
 
   connect_if_required: function (state) {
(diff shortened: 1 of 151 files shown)


lib/js/jslib.js
```diff
@@ -2922,6 +2923,20 @@ function spawnAt(loc, f, kappa) {
...
+    check_location(loc);
+    return _new();
+}
+
+function _newClientAP() { return _new(); }
+function _newServerAP() {
+    throw new Error('Cannot create a server access point from a client');
+}
+
+const newClientAP = LINKS.kify(_newClientAP);
```
slindley, line 2936: Is there a good reason why `const new = LINKS.kify(_new);` doesn't appear somewhere in `jslib.js`?

SimonJF: I wouldn't be surprised if it's because `new` is a keyword.

slindley: That's probably it... but we should really have a way with dealing with that kind of name clash. Something to think once #1087 gets merged.

PR #1112, opened 2022-03-24
Fix regression in vdom.js

dhil:
This patch fixes a regression in `lib/js/vdom.js`, as since patch #1087 there is no longer any `_makeCont` function, instead client code has to call `_$K.make`.


dhil:
Supposedly already fixed by @slindley, since this PR is empty since syncing with master...

PR #1118, opened 2022-03-26
Fix remote calls involving client closures.

### LINKED ISSUE #1050: Broken remote calls
I am somewhat surprised to discover that remote calls seem to be thoroughly broken. The following program causes a runtime error on the server when you click the button.
```links
fun runOnServer(f) server { f() }

fun mainPage(_) {
  page
    <#><button l:onclick="{runOnServer(fun() {()})}">Click me!</button></#>
}

fun main() {
  addRoute("/", mainPage);
  servePages()
}

main()
```
Error message:
```
jslib.js:1392 POST http://localhost:10088/ 500 (Internal Server Error)
(anonymous) @ jslib.js:1392
runOnServer_2344 @ (index):3295
(anonymous) @ (index):3297
_yield @ jslib.js:3667
_fun__g139_2349 @ (index):3297
_2349 @ (index):418
_cont @ jslib.js:2609
(anonymous) @ jslib.js:2618
(anonymous) @ jslib.js:1503
jslib.js:1085 Uncaught Error: Fatal error: call to server returned an error. Details: ***: Error: Links_core.Evalir.Exceptions.EvaluationError("Attempting to apply (_closureTable = 0) (need function instead)") 
    at XMLHttpRequest.<anonymous> (jslib.js:1085)
```
This (mis)behaviour persists even if I force the anonymous function `fun() { () }` to be located on the server. However, if I run it as a first-order function, then it seems to work as expected.
```links
fun g() server { () }

fun mainPage(_) {
  page
    <#><button l:onclick="{g()}">Click me!</button></#>
}

fun main() {
  addRoute("/", mainPage);
  servePages()
}

main()
```

At first I thought the problem was related to higher-order remote calls in event handlers, however, it is not as the following program illustrates:
```links
fun runOnServer(f) server { f() }

fun mainPage(_) {
  var pid = spawnClient { runOnServer(fun() { () }) };
  page
    <#></#>
}

fun main() {
  addRoute("/", mainPage);
  servePages()
}

main()
```
The above program crashes after start up with a similar error message as the first program.

dhil:
This patch fixes a problem with calling remote server functions that
will call a client side function. Previously, client side closures
were stored in a global table `_closureTable`. Such a closure would
get serialised as an index into this table, however, the server
evaluator is unaware of how to invoke a closure contained within this
table, thus resulting in a 500 Internal Error being thrown.

This patch adds a new value form on the server side `ClientClosure
of int` to represent a client side closure, where the integer payload
is the index into the global closure table. In addition the evaluator
is extended with a clause for `ClientClosure` which produces a
client side call to a special function that knows how to access and
invoke the targetted closure in the global table.

Resolves #1050

Partial fix for #219. This patch unveils another problem with the example program in #219. It produces a "forbidden" client call, e.g.
```
Links Error
***: Error: Cannot call client side function '_$ClosureTable.apply' because of before server page is ready
```

### DIFF
--- a/core/webif.ml
+++ b/core/webif.ml
@@ -135,7 +135,6 @@ struct
       | RemoteCall (func, env, args) ->
         Debug.print("Doing RemoteCall for function " ^ Value.string_of_value func
           ^ ", client ID: " ^ client_id_str);
-        Debug.print ("func: " ^ Value.show func);
         Proc.resolve_external_processes func;
         List.iter Proc.resolve_external_processes args;
         List.iter (Proc.resolve_external_processes -<- fst -<- snd)
--- a/core/webif.ml
+++ b/core/webif.ml
@@ -136,7 +136,6 @@ struct
         Debug.print("Doing RemoteCall for function " ^ Value.string_of_value func
           ^ ", client ID: " ^ client_id_str);
         Debug.print ("func: " ^ Value.show func);
-        Debug.print ("args: " ^ mapstrcat ", " Value.show args);
         Proc.resolve_external_processes func;
         List.iter Proc.resolve_external_processes args;
         List.iter (Proc.resolve_external_processes -<- fst -<- snd)
--- a/core/webif.ml
+++ b/core/webif.ml
@@ -31,7 +31,6 @@ struct
   let parse_remote_call (valenv, _, _) cgi_args =
     let fname = Utility.base64decode (assoc "__name" cgi_args) in
     let args = Utility.base64decode (assoc "__args" cgi_args) in
-    (* Debug.print ("args: " ^ Value.show (Json.parse_json args)); *)
     let args = Value.untuple (U.Value.load (Yojson.Basic.from_string args)) in
 
     let fvs = U.Value.load (Yojson.Basic.from_string (Utility.base64decode (assoc "__env" cgi_args))) in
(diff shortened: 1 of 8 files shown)


core/webif.ml
```diff
@@ -124,19 +124,19 @@ struct
...
         let result_json =
           Json.jsonize_value_with_state result json_state |> Json.json_to_string in
         Lwt.return ("text/plain", Utility.base64encode result_json)
-      | RemoteCall(func, env, args) ->
+      | RemoteCall (func, env, args) ->
         Debug.print("Doing RemoteCall for function " ^ Value.string_of_value func
           ^ ", client ID: " ^ client_id_str);
-        (* Debug.print ("func: " ^ Value.show func); *)
-        (* Debug.print ("args: " ^ mapstrcat ", " Value.show args); *)
+        Debug.print ("func: " ^ Value.show func);
```
jamescheney, line 138: Do you want these debug messages to stay uncommented?

jamescheney (approved):
This looks good to (non expert) me.  Regarding #219 I'm pretty sure that this is a constructive step, I think resolving #219 may require generalizing client/server annotations so that they can occur inside local function definitions (and for the right thing to happen if so), and I suspect the lack of a sensible treatment of client-side closure calls would have prevented that working.  So I'd be in favor of merging sooner rather than later.

core/webif.ml
dhil: I'll axe them! Thanks!

dhil, line 138:
```suggestion
```

PR #1128, opened 2022-04-01
Fix comparison operations for DateTime

### LINKED ISSUE #1126: Comparison with infinity/forever
Some comparisons with infinity give errors:
```
links> var t1 = parseDate("2022-08-04 12:03:12");
t1 = 2022-08-04 12:03:12.000000+1 : DateTime

links> var t2 = forever;
t2 = infinity : DateTime

links> t2 == t1;
***: Runtime error: Comparing infinity with 2022-08-04 12:03:12.000000+1 which either does not make sense or isn't implemented. 

links> t2 <> t1;
***: Runtime error: Comparing infinity with 2022-08-04 12:03:12.000000+1 which either does not make sense or isn't implemented.

links> t2 <= t1;
***: Runtime error: Comparing infinity with 2022-08-04 12:03:12.000000+1 which either does not make sense or isn't implemented. 
```
Other  comparisons are OK:
```
links> t2 >= t1;
true : Bool

links> t1 < t2;
true : Bool

links> t2 < t1;
false : Bool
```

SimonJF:
Equality was not implemented on the server, and comparisons weren't
correctly implemented on the client.

Fixes #1126

### DIFF
--- a/lib/js/jslib.js
+++ b/lib/js/jslib.js
@@ -806,6 +806,14 @@ const _$Links = (function() {
       return true;
     }
 
+    if (_$Types.isDateTime(l) && _$Types.isDateTime(r)) {
+        if (l._type === "timestamp" && r._type === "timestamp") {
+            return l._value === r._value;
+        } else {
+            return l._type === r._type;
+        }
+    }
+
     throw ("Comparing " + l + " with " + r + " which either does not make sense or isn't implemented.");
   }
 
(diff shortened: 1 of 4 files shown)


lib/js/jslib.js
```diff
@@ -806,6 +806,14 @@ const _$Links = (function() {
       return true;
     }
 
+    if (_$Types.isDateTime(l) && _$Types.isDateTime(r)) {
```
dhil, line 809: I think this code block needs to be moved up before the generic object comparison code. Otherwise I don't think this code will get executed properly.

dhil (changes requested):

lib/js/jslib.js
SimonJF: Whoops, good catch.

SimonJF: (Funnily enough I noticed and fixed this for a few other cases -- this one slipped through the net! I'm grateful for you spotting it.)

dhil (approved):

PR #1129, opened 2022-04-01
Allow record extension in presence of temporal projections

### LINKED ISSUE #1124: Record extension in the context of temporal tables
The following works for non-temporal tables:
```
var test_table = table "test_table" with (
                   id: Int,
                   integer: Int)
                 where id readonly from db;
query { for (x <-- test_table)  [( a=x.id | ( b=x.integer | x ))]};
```
but trying this on temporal tables when using the functions that extract the information from the temporal tables gives an error:
```
links> query nested {  for (x <-t- reference2ligand_temporal) [(tt_from=ttFrom(x) | ttData(x) ) ]};)
***: Error: Links_core.QueryLang.DbEvaluationError("Error adding fields: non-record") 
```

SimonJF:
This allows record extension to play nicely with temporal projections.
I was forgetting that all arguments to reduce_artifacts are already
eta-expanded, so we can work with record literals.

Fixes #1124.

### DIFF
--- a/core/query/query.ml
+++ b/core/query/query.ml
@@ -259,20 +259,33 @@ struct
   | Q.Apply (Q.Primitive "AsListT", [xs])
   | Q.Apply (Q.Primitive "AsListV", [xs]) -> xs
   (* Temporal projection operations *)
-  (* Note: These should already all be eta-expanded, so we can
-     treat them as record literals. *)
   | Q.Apply (Q.Primitive "ttData", [x])
   | Q.Apply (Q.Primitive "vtData", [x]) ->
-        Q.unbox_record x
-        |> StringMap.find TemporalField.data_field
+    begin
+        match x with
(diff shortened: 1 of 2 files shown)


SimonJF:
Looking at it, this is a bit more complex than I'd thought (was testing on a restricted case -- the cases I was encountering all happened to have record literals, but this is not the case in general since `xlate` does not in itself normalise terms, merely translate them).

The issue is that `Extend` needs to have access to a record literal during translation, whereas it doesn't necessarily on an un-normalised term.

SimonJF:
I've reopened this with the view to partially sorting #1130. Extension may not work in the general case, but we can at least make it work for temporal accessors who operate on a record literal.

SimonJF:
Can this go in? I don't think it's too controversial.

dhil (approved):
LGTM.

Issue #1116, opened 2022-03-26
Option or Maybe?


dhil:
The prelude and runtime make inconsistent use of optional data types. In some place we use a `Maybe(a) = [|Just:a|Nothing|]` type, whereas in others we use `Option(a) = [|Some:a|None|]`. We should pick either and stick to it throughout Links. I suggest we have a brief discussion about this matter and make a swift decision on Monday.

This will be a breaking change.

ezrakilty:
Historical note: at the beginning, Jeremy really preferred the Haskell style, Just/Nothing, saying that "Nothing" was more natural English. I preferred the ML style, because Some/None have the same number of letters. I suppose we put our own preferences into the prelude in different places. In other ways, I used to advocate the ML point of view while Jeremy did the Haskell.

Ironically, he now works on OCaml, and I do all my hacking in Haskell.

dhil:
I had a conversation with @slindley the other day about this "issue". Sam asked me who introduced either; my guess off the cuff was you and @yallop, though, apparently I had it wrong way around -- I thought Jeremy had introduced `Option` and you had introduced `Maybe`. Thanks for the correction and historical note, @ezrakilty!

dhil:
`Maybe` won.

PR #1131, opened 2022-04-06
Hand refactor Option -> Maybe, fixing #1116

jamescheney:
Just a quick search and replace, no attempt has been made to check the changed examples still work yet...

### DIFF
--- a/examples/TicTacToe-Console.links
+++ b/examples/TicTacToe-Console.links
@@ -2,7 +2,9 @@
 
 typename Player = [|X|O|];
 
-var initialState = [Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing] : [Maybe(Player)];
+var initialState = [Nothing, Nothing, Nothing,
+                    Nothing, Nothing, Nothing,
+		    Nothing, Nothing, Nothing] : [Maybe(Player)];
 
 sig playerToInt : (Player) -> Int
 fun playerToInt(player) {
@@ -20,9 +22,9 @@ fun playerToString(player) {
     }
 }
 
-sig optionToString : ((a) -e-> String) -> (Maybe(a)) -e-> String
-fun optionToString(xToString)(option){
-    switch(option){
+sig maybeToString : ((a) -e-> String) -> (Maybe(a)) -e-> String
+fun maybeToString(xToString)(ma){
+    switch(ma){
         case Nothing -> "_"
         case Just(player) -> xToString(player)
     }
--- a/examples/TicTacToe-Console.links
+++ b/examples/TicTacToe-Console.links
@@ -54,7 +54,7 @@ fun listToString(list, xToString){
 
 sig printState : ([Maybe(Player)]) ~> ()
 fun printState(state){
-    print(listToString(state, optionToString(playerToString)))
+    print(listToString(state, maybeToString(playerToString)))
 }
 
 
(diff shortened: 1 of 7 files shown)


examples/TicTacToe-Console.links
```diff
@@ -1,9 +1,8 @@
 #tictactoe console edition
 
 typename Player = [|X|O|];
-typename Option(a) = [|None|Some:a|];
 
-var initialState = [None, None, None, None, None, None, None, None, None] : [Option(Player)];
+var initialState = [Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing] : [Maybe(Player)];
```
dhil, line 5: I think this change is fine, but it worth noting that it is not really necessary. There might be reasons to use an `Option` type over `Maybe` type in examples, e.g. an ML crowd will be familiar with `Option`, whereas a Haskell crowd would be mostly familiar with `Maybe`.

examples/TicTacToe-Console.links
```diff
@@ -21,11 +20,11 @@ fun playerToString(player) {
     }
 }
 
-sig optionToString : ((a) -e-> String) -> (Option(a)) -e-> String
+sig optionToString : ((a) -e-> String) -> (Maybe(a)) -e-> String
```
dhil, line 23: `maybeToString`?

dhil (commented):
The big changes will be in `prelude.links`, `core/lib.ml`, and `lib/js/jslib.js`.

jamescheney:
> The big changes will be in prelude.links, core/lib.ml, and lib/js/jslib.js.

Good point.  I looked over prelude.links and didn't see any occurrence of `Option`, `None`, or `Some`.  There are a few in lib.ml and jslib.js (in the latter, `Some` and `None` are used for the JS version of the `attribute` library function, but in lib.ml the `Just`/`Nothing` constructors are used...)

examples/TicTacToe-Console.links
jamescheney: I think it is worthwhile for consistency with the prelude and to avoid having to re-declare Option whenever needed(i.e. keep things DRY).

examples/TicTacToe-Console.links
jamescheney: OK, though possibly this could be a library function, it is a special case of the "fold" (or "case") operation for maybe:
```
foldMaybe :: (Maybe(a),b,a -e-> b) -e-> b
```

slindley: It certainly shouldn't be called `optionToString` any more!

jamescheney:
Ready for review, though the interpreter check seems to have failed for a random reason (something to do with opam credentials)

examples/TicTacToe-Console.links
jamescheney: https://github.com/links-lang/links/pull/1131/commits/6cffb2d68915e733fa243963b12d73312823f5b4

lib/js/jslib.js
```diff
@@ -3811,8 +3811,8 @@ function _attribute(xml, attr) {
 
   const val = obj[attr];
 
-  if (val == undefined) return({_label:"None", "_value":(_$Constants.UNIT)});
-  else return({"_label":"Some", "_value":val});
+  if (val == undefined) return({_label:"Nothing", "_value":(_$Constants.UNIT)});
```
dhil, line 3814:
```suggestion
  if (val == undefined) return {_label: "Nothing", "_value": _$Constants.UNIT};
```

lib/js/jslib.js
```diff
@@ -3811,8 +3811,8 @@ function _attribute(xml, attr) {
...
+  else return({"_label":"Just", "_value":val});
```
dhil, line 3815:
```suggestion
  else return {"_label": "Just", "_value": val};
```

dhil (approved):
LGTM

Issue #907, opened 2020-08-01
client/server annotations on local functions


jamescheney:
In concrete syntax, Links only allows toplevel functions to carry client or server annotations.  Inner and anonymous functions are not allowed to have locations in Links source code, and the location component of the abstract syntax is initialized to `loc_unknown` for these.  

Internally, there are a few places where functions with explicit locations are created by desugaring (e.g. in `desugarLAttributes`) and it seems that these do not cause problems.

This issue proposes that the concrete syntax be extended to allow client/server locations on any function.  Relatedly, we could also support `client` and `server` blocks, which would simply be syntactic sugar for:
```
client {...} --> (fun () client {...})()
server {...} --> (fun () server {...})()
```
The implications of allowing client and server annotations anywhere are not clear.  There may be places where we implicitly assume only top-level definitions have such annotations, which may result in wrong behavior if this is not the case.  An initial experiment with this discussed in #906 resulted in an assertion failure in `evalir`, suggesting that in at least one place the existing code is not dealing with located local functions correctly.  The non-working example is as follows:
```
mutual {
fun testPage(msg) {
  page
  <html>
    <body>
	<div>Message: {stringToXml(msg)}</div>
      <div id="t">
        <h4>HAHAH</h4>
      </div>
      <div id="p">
        <p>LALA</p>
      </div>
      <div id="x">
        <p>GAGA</p>
      </div>
      <a href="" l:onclick="{changeContent()}">Click me</a>
    </body>
  </html>
}


fun changeContent() client {
  var time = clientTime();
  replaceNode(
      <form l:onsubmit="{error("asdf")}" method="post">
        onSubmit handled OK
	<button type="submit">Submit!</button>
      </form>
    ,
    getNodeById("t")
  );
  fun snippet1 () {
      <form l:action="{testPage(intToString(time))}" method="post">
        action handled OK at <b>{intToXml(time)}</b>
	<button type="submit">Submit!</button>
      </form>
  }
  replaceNode(snippet1(),getNodeById("p"));
  fun snippet2() server {
    <a l:href="{testPage(intToString(time))}">href handled OK at <b>{intToXml(time)}</b></a>
  }
  replaceNode(snippet2(),getNodeById("x"))
}
}

fun main () {
  addRoute("",testPage);
  servePages()
}

main()

```

(Manually lambda-lifting this example and hoisting the server-functions to the toplevel resulted in correct behavior.)

jamescheney:
[...]

PR #1133, opened 2022-04-16
initial attempt to enable client/server annotations anywhere

jamescheney:
work in progress on #907


jamescheney:
It sounded like people were happy with merging, perhaps a quick review would be helpful.  In particular we are currently not using the new features in any tests whatsoever (even just syntax/typechecking) so suggestions what tests should be added are welcome.

dhil:
Yes! I will have a look tomorrow!

core/evalir.ml
```diff
@@ -245,11 +245,12 @@ struct
...
-      let env =
+      let env, xs =
         match z, fvs with
-        | Some z, Some fvs -> Value.Env.bind z (fvs, Scope.Local) env
-        | None, None -> env
-        | _, _ -> assert false
+        | Some z, Some fvs -> Value.Env.bind z (fvs, Scope.Local) env, xs
+        | None, None -> env, xs
+        | Some z, None -> env, z::xs (* kludge to deal with possible mismatch *)
+        | None, Some _ -> assert false
```
jamescheney, line 253: This was an attempt to work around an apparent bug in closure conversion which surfaces if you have local client functions with non-empty closures.  However I'm, sure it is not the right way to fix this problem, so I will undo it.  The right thing to do is merge the rest of this PR and then make a new issue describing the bug that needs to be worked around.

dhil (approved):
LGTM.

Issue #1136, opened 2022-04-28
Error when calling non-closed local server functions


jamescheney:
Thanks to #1133 it is now possible to write local / anonymous functions that have client or server annotations.  Such functions need not be closed, for example:
```
fun foo(x) client { replaceNode(
    (fun (y) server  {<div id="xyz"><p><b>{stringToXml(x)}</b>:<b>{stringToXml(y)}</b></p></div>})("xyzzy")
    , getNodeById("xyz"))
}

fun testPage(msg) {
  page
  <html>
    <body>
      <div>Message: {stringToXml(msg)}</div>
        <form l:onsubmit="{foo(x)}" method="post">
          <input type="text" l:name="x"/>
  	<button type="submit">Submit!</button>
        </form>
      <div id="xyz">This should be replaced</div>
    </body>
  </html>
}



fun main () {
  addRoute("",testPage);
  servePages()
}

main()
```
However, running this and clicking the Submit button yields (I think) the same error @kwanghoon reported in a comment on #907, adjusting for changes:
```
Uncaught Error: Fatal error: call to server returned an error. Details: ***: Error: File "core/evalir.ml", line 252, characters 18-24: Assertion failed
```
This error appears in the browser console, but relays an error raised on the server.  

Removing the `server` label from the anonymous function yields correct behavior, presumably because the function is called on the client and it is closure converted correctly there.

Replacing `foo` with the following:
```
fun foo(x) client { replaceNode(
    (fun (x,y) server {<div id="xyz"><p><b>{stringToXml(x)}</b>:<b>{stringToXml(y)}</b></p></div>})(x,"xyzzy")
    , getNodeById("xyz"))
}
```
so that the anonymous server function is lambda-lifted in place, yields correct behavior.

The error is being raised here:
https://github.com/links-lang/links/blob/474137f31a43d7589d2711968d8a564bb14c173f/core/evalir.ml#L252

which is an "unreachable" case of a pattern match that tries to match up the closure variable and environment.  Evidently these are getting out of sync somehow, possibly due to an unhandled case or no-longer-guaranteed invariant in closure conversion.  Through random hacking I found that the change in commit [8cee258](https://github.com/links-lang/links/pull/1133/commits/8cee258c0419172e314bda2d4b3fb42a1f315287) "fixes" this problem.  But this is almost certainly not the right way to fix it, the right way is to make sure that the function info and environment/argument list do not get out of sync in the first place.

kwanghoon:
[...]

PR #1138, opened 2022-05-02
Attempted workaround for #1136

jamescheney:
This is an attempt to work around #1136 in a principled, but ugly, way.
A shorter fix is possible ([8cee258](https://github.com/links-lang/links/pull/1133/commits/8cee258c0419172e314bda2d4b3fb42a1f315287)) but this approach delays fixing the mismatch between closure variable and environment until the point FunctionPtr is called.  Since we are already looking up the function info there and checking various things, that's no real problem, but I don't like this way of doing things because it delays the point where we make things right to the point where something is about to go wrong, rather than the point where we create the FunctionPtr value representing a server-side call request received from the client.

An approach that is both clean and principled would be nicer, but might take a lot more work.  For example, fully closure-converting the IR functions so that the environment is passed as an ordinary variable, all function bodies are closed, and IR code does the bureaucracy of passing around environment variables would potentially fix this, but it woudl also be a lot more work (possibly including adding existentials to the IR if we want to be able to typecheck after closure conversion ala Minamide et al.)

### DIFF
--- a/core/webif.ml
+++ b/core/webif.ml
@@ -41,13 +41,23 @@ struct
        have to do something about it in order for attempts to remotely
        call primitive functions to work properly. *)
 
-    let func =
-      match fvs with
-      | `Record [] -> let i_fname = int_of_string fname in
-          if Lib.is_primitive_var i_fname
-          then `PrimitiveFunction (Lib.primitive_name i_fname, Some i_fname)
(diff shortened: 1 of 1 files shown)


dhil (approved):
LGTM

Issue #44, opened 2016-07-28
There is no way to define row aliases


slindley:

jamescheney:
More generally, for any kind it would be nice to be able to define (parameterized) aliases of things of that kind, uniformly.  Something like this syntax:

```
typename Foo(a,b,c)::Kind = ...
```
where the RHS of `...` would need to allow several different syntactic classes (which might in turn be tricky to parse).

We can currently already put kind specifiers on the type parameters but they are not handled in a hygienic way, and no kind inference is done (#491, #526, #301).

dhil:
Since #1141 we can define aliases for effect rows. This mechanism should be generalised along the lines outlined by @jamescheney above.

PR #1141, opened 2022-05-24
Effect aliasing

Orbion-J:
I added the possibility to write effect aliases, similarly as what already exists for types.
- there is a new keyword `effectname`
- to write an alias for an effect row : `effectname MyEffectRow(a, ... ,e::Eff, ...) = { Op1 : type, ... | e }` the arguments being type variable of any kind (kinds other than type must be explicit)
- we can have, as above, an open row (the row variable is a parameter of the effect alias) or a closed one (just do not write the `| e`
- To use it in a signature or in another type or effect alias just apply it with the right arguments as for `typename` things.
- In arrows, we can use them as row variables: `() -MyEffectRow(args)-> ()` (idem with `~>`)
- However, due to lack of kind inference, row variables and aliases have to be used carefully so that links does not think they are of kind type. We need to write them most of the time between braces ` { | ... }`. For instance, if you have `effectname E(a::Eff) = {X : ... | a }` and a row variable `e::Eff`, you will have to write `E({ |e})`. (Idem for another effect alias instead of the variable). This makes the usage of several nested aliases a bit messy, it would be nice if we could avoid it.
- We cannot write recursive effect aliases for now. In the branch `visitor`, I added another transformer that makes possible simple recursion by inlining a mu type in one pass.
- For now the aliases are replaced by the row they correspond to : we do not keep aliases. 
- In the repl, effect alias definitions are printed but without the braces ! Rows are in general printed without braces and the alias body is a row. => this might need to be enhanced

About implementation, I copied and then merged most of the time what existed for `typename`.


jamescheney:
I don't know the backstory of this, and haven't had a chance to look at this carefully, but instead of adding a keyword and duplicating a lot of code for `typename`, would it be possible (and wouldn't it be nicer) to augment `typename` to allow an optional kind, and generalize it to allow any kind of type alias (with Type being the default used if no explicit kind is given)?

e.g. instead of 
```
effectname MyEffectRow(a, ... ,e::Eff, ...) = { Op1 : type, ... | e }
```
could we write
```
typename MyAlias(...) :: K = ty
```
for any kind K, with Eff being a special case, and just generalize the existing code for aliases to allow this?

The description and comments suggest that this may not be straightforward, because of existing idiosyncrasies, but the above design seems cleaner and something we should work toward (see also #44 and comments about row aliases).  Perhaps a constructive compromise could be to adopt this PR with a small change: allowing (in addition or only)
```
typename  MyEffectRow(a, ... ,e::Eff, ...):: Eff =  { Op1 : type, ... | e }
```
as a special case without (yet) trying to generalize to allow aliases of any kind.

Orbion-J:
@jamescheney I can surely do something that looks like that. At least syntacticly, I can check during parsing which kind do we have and use after the same things I ve already done.
I will still try see how complicated it would be to extend to presence, in order to be able to at least use it for every primary kind.

bin/repl.ml
```diff
@@ -155,14 +155,14 @@ let rec directives : (string * ((Context.t -> string list -> Context.t) * string
...
-            tenv.Types.tycon_env
+            tenv.Types.alias_env
           in
           StringSet.iter (fun k ->
               let s = Env.String.find k tycon_env in
               Printf.fprintf stderr " %s = %s\n"
                 (Module_hacks.Name.prettify k)
                 (Types.string_of_tycon_spec s))
-            (StringSet.diff (Env.String.domain tycon_env) (Env.String.domain Lib.typing_env.Types.tycon_env));
+            (StringSet.diff (Env.String.domain tycon_env) (Env.String.domain Lib.typing_env.Types.alias_env));
```
dhil, line 165:
```suggestion
            (StringSet.diff (Env.String.domain tycon_env) (Env.String.domain Lib.typing_env.Types.tycon_env));
```

bin/repl.ml
```diff
@@ -155,14 +155,14 @@ let rec directives : (string * ((Context.t -> string list -> Context.t) * string
         ((fun context _ ->
           let tycon_env =
             let tenv = Context.typing_environment context in
-            tenv.Types.tycon_env
+            tenv.Types.alias_env
```
dhil, line 158:
```suggestion
            tenv.Types.tycon_env
```

bin/repl.ml
```diff
@@ -221,7 +221,7 @@ let rec directives : (string * ((Context.t -> string list -> Context.t) * string
         ((fun context args ->
           let tenv, aliases =
             let tyenv = Context.typing_environment context in
-            tyenv.Types.var_env, tyenv.Types.tycon_env
+            tyenv.Types.var_env, tyenv.Types.alias_env
```
dhil, line 224:
```suggestion
            tyenv.Types.var_env, tyenv.Types.tycon_env
```

bin/repl.ml
```diff
@@ -268,8 +268,8 @@ let handle previous_context current_context = function
      let tycon_env' =
        let tenv  = Context.typing_environment previous_context in
        let tenv' = Context.typing_environment current_context in
-       let tycon_env, tycon_env' =
-         Types.(tenv.tycon_env, tenv'.tycon_env)
+       let tycon_env, tycon_env'=
```
dhil, line 271:
```suggestion
       let tycon_env, tycon_env' =
```

bin/repl.ml
```diff
@@ -268,8 +268,8 @@ let handle previous_context current_context = function
...
+         Types.(tenv.alias_env, tenv'.alias_env)
```
dhil, line 272:
```suggestion
         Types.(tenv.tycon_env, tenv'.tycon_env)
```

core/defaultAliases.ml
```diff
@@ -3,7 +3,7 @@ open CommonTypes
 (* Alias environment *)
 module AliasEnv = Env.String
 
-let alias_env : Types.tycon_environment =
+let alias_env : Types.alias_environment =
```
dhil, line 6:
```suggestion
let alias_env : Types.tycon_environment =
```

core/desugarDatatypes.ml
```diff
@@ -68,7 +72,7 @@ module Desugar = struct
   let desugar_quantifiers  (sqs: SugarQuantifier.t list) :  Quantifier.t list =
     List.map SugarQuantifier.get_resolved_exn sqs
 
-  let rec datatype (alias_env : Types.tycon_environment) t' =
+  let rec datatype (alias_env : Types.alias_environment) t' =
```
dhil, line 75:
```suggestion
  let rec datatype (alias_env : Types.tycon_environment) t' =
```

core/desugarDatatypes.ml
```diff
@@ -209,6 +213,72 @@ module Desugar = struct
...
+                else t
+              in
+              let type_args qs ts =
+                List.combine qs ts
+                |> List.mapi
+                     (fun i (q,t) ->
+                       let  t = match_kinds i (q, t) in
+                       type_arg alias_env t node)
+              in
+
```
dhil, line 244:
```suggestion
```

core/desugarDatatypes.ml
```diff
@@ -209,6 +213,72 @@ module Desugar = struct
...
+                let primary_kind_of_type_arg : Datatype.type_arg -> PrimaryKind.t = function
+                  | Type _ -> PrimaryKind.Type
+                  | Row _ -> PrimaryKind.Row
+                  | Presence _ -> PrimaryKind.Presence
+                in
+                let q_kind, _ = proj q in
+                let t_kind = primary_kind_of_type_arg t in
+                if q_kind <> t_kind then
+                  raise
+                    (TypeApplicationKindMismatch
```
dhil, line 228: `TypeApplicationKindMismatch` should probably be defined in `errors.ml`.

core/desugarDatatypes.ml
```diff
@@ -209,6 +213,72 @@ module Desugar = struct
...
+              in
+
+              let qn = List.length qs and tn = List.length ts in
+              if qn = tn then
+                type_args qs ts
+              else
+                raise (TypeApplicationArityMismatch { pos = node.pos; name = name; expected = qn; provided = tn })
+            in
+            begin match SEnv.find_opt name alias_env with
+              | None -> raise (UnboundTyCon (node.pos, name))
```
dhil, line 252: `UnboundTyCon` should probably also be defined in `errors.ml`

core/desugarDatatypes.ml
```diff
@@ -236,11 +306,16 @@ module Desugar = struct
     | Row r      -> Row, row alias_env r node
     | Presence f -> Presence, fieldspec alias_env f node
 
-
-
   let datatype' alias_env ((dt, _) : datatype') =
     (dt, Some (datatype alias_env dt))
 
+  let row' alias_env ((r, _) :row') =
+    (r, Some (row alias_env r (WithPos.make (Datatype.Effect r)))) (* TODO(rj) should keep the pos *)
```
dhil, line 313: I think this is fine for now. Ideally, `row` would be part of the datatype definition -- we should work towards that at a later point.

core/desugarDatatypes.ml
```diff
@@ -340,36 +417,43 @@ object (self)
 
         (* Add all type declarations in the group to the alias
          * environment, as mutuals. Quantifiers need to be desugared. *)
-        let ((mutual_env : tycon_spec SEnv.t), ts) =
-          List.fold_left (fun (alias_env, ts) {node=(t, args, (d, _)); pos} ->
+        let ((mutual_env : Types.alias_environment), ts) =
```
dhil, line 420:
```suggestion
        let ((mutual_env : Types.tycon_environment), ts) =
```

core/desugarDatatypes.ml
```diff
@@ -340,36 +417,43 @@ object (self)
...
             (* Check if the datatype has actually been desugared *)
-            let (t, dt) =
-              match dt' with
-               | (t, Some dt) -> (t, dt)
-               | _ -> assert false in
-            WithPos.make ~pos (name, args, (t, Some dt))
+            let b' = match Desugar.aliasbody mutual_env b with
+                | Typename   (_, Some _) as b' -> b'
+                | Effectname (_, Some _) as b' -> b'
+                | _ -> assert false
```
dhil, line 440: You should use `internal_error` here.

core/desugarDatatypes.ml
```diff
@@ -468,17 +557,17 @@ let toplevel_bindings alias_env bs =
...
+  let alias_env = typing_env.alias_env in
   let alias_env, bindings =
     toplevel_bindings alias_env bindings in
   (* let typing_env = { typing_env with tycon_env = alias_env } in *)
   (bindings, opt_map ((phrase alias_env) ->- snd) p)
 
 let sentence typing_env = function
   | Definitions bs ->
-      let _alias_env, bs' = toplevel_bindings typing_env.tycon_env bs in
+      let _alias_env, bs' = toplevel_bindings typing_env.alias_env bs in
```
dhil, line 568: `tycon_env`

core/desugarDatatypes.ml
```diff
@@ -468,17 +557,17 @@ let toplevel_bindings alias_env bs =
...
         Definitions bs'
-  | Expression  p  -> let _o, p = phrase typing_env.tycon_env p in
+  | Expression  p  -> let _o, p = phrase typing_env.alias_env p in
```
dhil, line 570: `tycon_env`

core/desugarDatatypes.ml
```diff
@@ -468,17 +557,17 @@ let toplevel_bindings alias_env bs =
 
 let program typing_env (bindings, p : Sugartypes.program) :
     Sugartypes.program =
-  let alias_env = typing_env.tycon_env in
+  let alias_env = typing_env.alias_env in
```
dhil, line 560: `tycon_env`

core/desugarDatatypes.mli
```diff
@@ -1,4 +1,4 @@
-val read : aliases:Types.tycon_environment -> string -> Types.datatype
+val read : aliases:Types.alias_environment -> string -> Types.datatype
```
dhil, line 1:
```suggestion
val read : aliases:Types.tycon_environment -> string -> Types.datatype
```

core/desugarEffects.ml
```diff
@@ -1089,16 +1091,28 @@ class main_traversal simple_tycon_env =
...
+                    let d = cleanup_effects tycon_env d in
+                    let eff = gather_mutual_info tycon_env d in
+                    let has_imp = eff#has_implicit in
+                    let implicits = StringMap.add t has_imp implicits in
+                    let used_mutuals = StringSet.inter eff#used_types tycons in
+                    let dep_graph =
+                        StringMap.add t (StringSet.elements used_mutuals) dep_graph
+                    in
+                    (implicits, dep_graph)
+                  | Effectname (r,_) -> (* is this the right thing to do ? *)
```
dhil, line 1106: We have to be careful here. I am not sure this is the right thing to do. We need to check what the effect desugaring procedure is doing under the hood. The arrows on labels in effect rows means something else than function arrows.

core/parser.mly
```diff
@@ -1130,7 +1136,7 @@ type_arg_list:
 type_arg:
 | datatype                                                     { Datatype.Type $1     }
 | braced_fieldspec                                             { Datatype.Presence $1 }
-| LBRACE row RBRACE                                            { Datatype.Row $2      }
+| LBRACE trow RBRACE                                           { Datatype.Row $2      }
```
dhil, line 1139: I think this should just be `row`.

core/parser.mly
```diff
@@ -1185,21 +1199,32 @@ vfield:
 | CONSTRUCTOR                                                  { ($1, present) }
 | CONSTRUCTOR fieldspec                                        { ($1, $2)      }
 
+tfields:
```
dhil, line 1202: The `fields_def` macro is supposed to eliminate some of the redundancy w.r.t. rows. We should use that rather than duplicating the parsing rules for rows.

dhil:
> I don't know the backstory of this, and haven't had a chance to look at this carefully, but instead of adding a keyword and duplicating a lot of code for `typename`, would it be possible (and wouldn't it be nicer) to augment `typename` to allow an optional kind, and generalize it to allow any kind of type alias (with Type being the default used if no explicit kind is given)?

Ultimately, we want to do something like that. But we before we do that we should have a design in mind. The point of this change is to introduce the minimal thing that let us tweak the programming experience / presentation with / of effects and handlers.

In its current state this patch is more invasive than I had originally intended. However, it is probably fine, but we need to be a bit more careful about mixing kind type and row in one alias environment -- I am sure this approach opens up an avenue for a lot of new bugs.

core/desugarDatatypes.ml
Orbion-J: yes it is, what should I do there ?

core/desugarDatatypes.ml
Orbion-J: idem

core/parser.mly
Orbion-J: I just merged `trow` and `erow`.
Initially, I did this because `efields` are used in arrows so it seemed to me the best thing to use. Do you really think we should add the effect application to the `row` ? And what about the dot thing ? I think that, when it is activated, we should be able to use the dot in effect aliases, isn't it ?

core/parser.mly
Orbion-J: cf supra

Orbion-J:
I fixed the simple things.
I tried to add what james was suggesting, I had to do a a bit of a hack in the parser to make it work but it seems to be ok. I added also a few things for presence but this is neither complete nor working.
Anyway, you are right mixing everything could cause some unexpected behavior as :
```ocaml
links> typename A = Int ;
A = Int
links> effectname B = { Op:() -A-> () } ;
***: Error: File "core/types.ml", line 1435, characters 11-17: Assertion failed 
```
this should for instance cause a kind mismatch. I think checking in `instantiate.ml` would help but there could be other issues.

core/parser.mly
```diff
@@ -413,6 +430,7 @@ arg:
 | UFLOAT                                                       { string_of_float' $1 }
 | TRUE                                                         { "true"  }
 | FALSE                                                        { "false" }
+| DEFAULT                                                      { "default" }
```
Orbion-J, line 433: this was missing, making impossible to reset settings to default since
```ocaml
links> @set something default ;
```
caused a parsing error

dhil:
> I fixed the simple things. I tried to add what james was suggesting, I had to do a a bit of a hack in the parser to make it work but it seems to be ok. I added also a few things for presence but this is neither complete nor working. Anyway, you are right mixing everything could cause some unexpected behavior as :
> 
> ```ocaml
> links> typename A = Int ;
> A = Int
> links> effectname B = { Op:() -A-> () } ;
> ***: Error: File "core/types.ml", line 1435, characters 11-17: Assertion failed 
> ```
> 
> this should for instance cause a kind mismatch. I think checking in `instantiate.ml` would help but there could be other issues.

Yes. Do not implement James' suggestion. That's for later when we have worked out a proper design. It is too much to bundle it all up.

core/parser.mly
dhil: You should submit this in a separate PR, do not include it in this one.

Orbion-J: ok, i ll do that

jamescheney:
> Do not implement James' suggestion. That's for later when we have worked out a proper design. It is too much to bundle it all up.

Of course, I was just asking if this was a "good enough for now" change or intended as a long-term proper design.

core/instantiate.ml
```diff
@@ -169,6 +169,7 @@ let instantiates : instantiation_maps -> (datatype -> datatype) * (row -> row) *
       match t with
       | Row _ -> t
       | Meta row_var -> Row (StringMap.empty, row_var, false)
+      | Alias (_,row) -> row
```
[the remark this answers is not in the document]
Orbion-J: cf https://github.com/links-lang/links/pull/1141#issuecomment-1141321170

core/desugarEffects.ml
Orbion-J: yes, I am not really sure of what this is really doing. I tried a bit to understand what happens but I did not really manage to do it...
There is something else we may want to do there. For now, effectname does not support a useful syntactic sugar. We have this :
```ocaml
links> effectname E = { Get:Int } ;
E = Get:Int
```
But we might want this ;
```ocaml
links> effectname E = { Get:Int } ;
E = Get:() {}-> Int
```
It should happens in this file but I did not figured out what to do to make this work yet.

core/desugarEffects.ml
```diff
@@ -978,6 +982,7 @@ class main_traversal simple_tycon_env =
       let module D = Datatype in
       let o, rv =
         match rv with
+        | D.EffectApplication _ -> (o, rv) (* TODO(rj) do i need to do something there ? *)
```
dhil, line 985: You need to make sure the alias constructor is desugared and then make sure the arguments are desugared.

core/desugarEffects.ml
dhil: The latter is the correct behaviour, I believe.

core/types.ml
```diff
@@ -1429,7 +1428,12 @@ and flatten_row : row -> row = fun row ->
     | Row _ -> row
     (* HACK: this probably shouldn't happen! *)
     | Meta row_var -> Row (StringMap.empty, row_var, false)
-    | _ -> assert false in
+    (* | Alias (PrimaryKind.Row, _, row) -> row *)
```
dhil, line 1431: Maybe remove this?

core/types.ml
```diff
@@ -1429,7 +1428,12 @@ and flatten_row : row -> row = fun row ->
...
+    (* | RecursiveApplication { r_dual ; r_args ; r_unwind ; _ } -> *)
+    (* (\* TODO(rj) what should this function do ? r_unwind like this provokes a stack overflow *\) *)
+    (*       r_unwind r_args r_dual *)
+    | _ ->  Debug.print ("row: " ^ show_row row) ; assert false
```
dhil, line 1435: Maybe we should raise `internal_error` here?

dhil (changes requested):
LGTM modulo the requested changes.

core/desugarEffects.ml
Orbion-J: it works now, with a bit of a hack
I added something also we might want (or not) :
```ocaml
typename F = () {}-> Int ;
effectname E = {Get:F} ;
effectname E' = {Get':Int} ;
```
now, we get `E' = Get': () {}-> Int` and also `E = Get:() {}-> Int`. (before we would have had `E = Get:() {}-> () {}-> Int`) 
If we detect a function type application for an operation, it is not desugared into a new function type with unit input. I think this is what we might want

dhil (approved):
LGTM. I will merge this tomorrow.

PR #1148, opened 2022-06-27
Fresh label

Orbion-J:
This grants the possibility to define local labels, ie labels that can only be used in a specific scope.

The user can define one or more local labels that will be bound in the scope with
```ocaml
fresh `A, `B {
   [ scope where labels `A and `B are bound ]
}
```
He can use the local labels so defined in the scope as any other effect label and he cannot use them outside that scope.
A function, a value or a typename that has such a label in its type cannot escape the scope. When we leave the scope, we try to erase the label from the type and if it it possible (ie if the label as a polymorphic or absent presence) the function/value is available in the rest of the program, it's type being cleaned from any local label. Otherwise, the function/value/typename is removed from the context and trying to use it will cause an unknown variable or `Unbound type constructor` error.
There's something that is still not perfect. The effectnames being inlined before typechecking, we can get a label going out of scope. This will be identified by the typechecker and will raise a `local label ... is not bound` error. It seems that the effectname has escaped the scope.
Example
```ocaml
fresh `A {
      effectname E = {`A:Int} ;
}
typename T = () -E-> () ;
```
```ocaml
Type error: The local label `A<1> is not bound
In expression: typename T = () -E-> ().
```

Briefly, here's how the code works :
- All labels in the code are now of type `Label.t`. They can be either global or local, have a name and the local ones have an id.
- labels with a backtick are parsed into free local label and normal label are parsed are global ones
- during desugaring, I added a pass that binds the local labels
- during typechecking, when we encounter a `fresh ... { ... }` binding, we add the labels into the context and typecheck each declaration in the scope. 
- before that we clean the context from any label that will be shadowed by the fresh declaration (labels that have the same name)
- after that we clean the context from the labels to avoid them to escape the scope
- we check during typechecking if types do not contain unbound or shadowed labels.

**Remark**
for now if all labels are internally of the same type and then can be local or global, the parser only accepts local labels in effect related things


dhil:
Superceded by #1169.

Issue #1132, opened 2022-04-16
Make default paths more uniform


jamescheney:
It seems that by default, if some recent version of Links is both installed using OPAM and also built using a checked-out repository clone, the various paths assume values like the following:
```
  db_driver_path                   /home/jcheney/src/links/_build/install/default/share/links-mysql,/home/jcheney/src/links/_build/install/default/share/links-postgresql,/home/jcheney/src/links/_build/install/default/share/links-sqlite3,/home/jcheney/src/links/_build/install/default/share/links-mysql8
  jslibdir                         /home/jcheney/.opam/4.10.0/lib/links/js
  prelude                          /home/jcheney/src/links/_build/default/lib/prelude.links
  stdlib_path                      /home/jcheney/.opam/4.10.0/lib/links/stdlib
  config                           /home/jcheney/.opam/4.10.0/etc/links/config
```
Thus, some (like `db_driver_path` and `prelude`) look in the current checked-out repository while the others look in the currently installed OPAM package for links.  This means that changes to jslib.js, stdlib, or the default configuration won't be visible until they are published in a new OPAM package which is then installed (unless one manually copies things over which is not ideal since it could break the current links repository install).  This is noticeable for example with the significant changes to jslib done recently in #1110.

 It might be a good idea to rationalize these choices.  For example I guess the jslibdir and stdlib_path variables are getting set by the config file in the OPAM install, but why are we using that by default? On the other hand, if I make my own config file and just don't set the jslibdir and stdlib_path settings, then these default to:
```
  stdlib_path                      /home/jcheney/src/links/_build/default/lib/stdlib
  jslibdir                                

```
i.e. stdlib_path gets set to something sensible but jslibdir is empty, and I have to manually set it to the right thing.

It would be good to rationalize this somehow, for example either

1. have a "local" default config file that lives in the repository and points to the right things, rather than inheriting the opam one; or 
2. make `jslibdir` default to the right thing if it is not set by whatever config file is used

I don't have strong feelings, option 2 seems like less work.

PR #1150, opened 2022-06-30
New syntax for handlers

Orbion-J:
This PR changes the syntax for handlers.
Instead of
```ocaml
handle (...) {
  case Op(params, resumption) -> ...
  case Return(x) -> x
}
```
we write
```ocaml
handle (...) {
  case <Op(params) => resumption> -> ...
  case x -> x
}
```
We can also write `case <Op(params) -> resumption> -> ...`. For now, the semantics for `->` and `=>` are the same. This is in prevision for a future step : we will not have `shallowhandler` anymore and we will indicate if a resumption is shallow with `->` and deep with `=>`.

### DIFF
--- a/core/compilePatterns.ml
+++ b/core/compilePatterns.ml
@@ -1173,24 +1173,25 @@ module Handlers = struct
     | Pattern.CNOperation names ->
        let cases =
          StringMap.fold
-           (fun name annotated_clauses cases ->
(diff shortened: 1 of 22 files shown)


dhil:
Thanks. I will merge this after we have fixed #1132 and shipped the next release.

PR #1154, opened 2022-07-12
Fix database unit test suite sample config

dhil:
The sample config for the database unit test suite was using hardcoded constants rather than the `$LINKS_POSTGRES_*` family of environment variables like the rest of the testing infrastructure.

### DIFF
--- a/run-tests
+++ b/run-tests
@@ -32,6 +32,19 @@ prepare_database_config () {
     fi
 }
 
+prepare_no_database_config () {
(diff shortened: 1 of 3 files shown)


dhil:
Something is wrong in the configuration of no database unit tests. The description says "NO DATABASE" but the test suite is evidently trying to load whatever database driver is described by the config file.

dhil:
I figured out what is happening. The test runners always generate a config with database information, and Links currently eagerly loads a default config if noone is given.

PR #1155, opened 2022-07-15
Add h0 and a0 to MvuHTML and MvuAttrs libraries

SimonJF:
The monoidal units of MvuHTML.HTML and MvuAttrs.Attr are accessed by `MvuHTML.empty` and `MvuAttrs.empty`.
When creating a tree, these are quite verbose to use, so the pattern I've ended up doing is:

```
var a0 = MvuAttrs.empty;
var h0 = MvuHTML.empty;
div(a0, h0)
```

I do this so much that I think it would actually be better to have these shortcuts in the actual libraries. It's handy for me and I don't see any harm in it.

### DIFF
--- a/lib/stdlib/mvuHTML.links
+++ b/lib/stdlib/mvuHTML.links
@@ -11,7 +11,7 @@ typename HTML(a :: Type(Any, Any)) =
    |];
 
 var empty = HTMLEmpty;
-var h0 = HTMLEmpty;
+var h0 = empty;
 
 fun textNode(str) {
   HTMLText(str)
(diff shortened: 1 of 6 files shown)


lib/stdlib/mvuHTML.links
```diff
@@ -11,6 +11,7 @@ typename HTML(a :: Type(Any, Any)) =
    |];
 
 var empty = HTMLEmpty;
+var h0 = HTMLEmpty;
```
dhil, line 14:
```suggestion
var h0 = empty;
```

dhil (commented):
LGTM.

SimonJF:
Thanks!

PR #1167, opened 2023-01-28
Hide session fail

slindley:
Session exceptions are compiled into effects and handlers. A practical difficulty with this is that code can become polluted with the `SessionFail` effect. Moreover, if session exceptions are switched on then the typing rules for all of the spawn operators change to take into the `SessionFail` effect (see core/lib.ml). This PR adds an experimental setting `expose_session_fail` which defaults to `true` (giving the old behaviour), but when set to `false` will hide the `SessionFail` effect as part of the built-in `wild` effect.

The PR is technically incomplete as it does not include special cases for treating `SessionFail` in the IR type checker when `expose_session_fail` is disabled, but it is safely hidden behind an experimental setting and is already sufficient to use for practical examples - and in particular makes it possible to implement functionality for detecting when a browser window closes in MVU applications which are not compatible with the `SessionFail` effect appearing explicitly in types.

### DIFF
--- a/core/basicsettings.ml
+++ b/core/basicsettings.ml
@@ -41,6 +41,16 @@ module Sessions = struct
               |> convert parse_bool
               |> CLI.(add (long "session-exceptions"))
               |> sync)
+
+  let expose_session_fail =
+    Settings.(flag "expose_session_fail" ~default:true
+              |> synopsis "Exposes the SessionFail effect"
+              |> depends Handlers.enabled
+              |> depends exceptions_enabled
+              |> privilege `System
+              |> convert parse_bool
+              |> CLI.(add (long "session-fail-exposed"))
+              |> sync)
 end
 
 module System = struct
--- a/core/basicsettings.ml
+++ b/core/basicsettings.ml
@@ -49,7 +49,7 @@ module Sessions = struct
               |> depends exceptions_enabled
               |> privilege `System
               |> convert parse_bool
-              |> CLI.(add (long "session-fail-exposed"))
+              |> CLI.(add (long "expose-session-fail"))
               |> sync)
 end
 
--- a/core/compilePatterns.ml
+++ b/core/compilePatterns.ml
@@ -646,7 +646,15 @@ and match_variant
                    if StringSet.mem name names then
                      (cases, cs)
                    else
-                     let case_type = TypeUtils.variant_at name t in
+                     let case_type =
+                       if Settings.get Basicsettings.Sessions.exceptions_enabled &&
+                          not (Settings.get Basicsettings.Sessions.expose_session_fail) &&
+                          String.compare name Value.session_exception_operation = 0
+                       then
+                         Types.empty_type
+                       else
+                         TypeUtils.variant_at name t
+                     in
 (*                     let inject_type = TypeUtils.inject_type name case_type in *)
                      let (case_binder, case_variable) = Var.fresh_var_of_type case_type in
                      let match_env = bind_type case_variable case_type env in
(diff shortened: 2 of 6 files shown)


core/compilePatterns.ml
```diff
@@ -646,7 +646,15 @@ and match_variant
                    if StringSet.mem name names then
                      (cases, cs)
                    else
-                     let case_type = TypeUtils.variant_at name t in
+                     let case_type =
+                       if Settings.get Basicsettings.Sessions.exceptions_enabled &&
+                          not (Settings.get Basicsettings.Sessions.expose_session_fail) &&
+                          String.compare name Value.session_exception_operation = 0
```
dhil, line 652:
```suggestion
                          String.equal name Value.session_exception_operation
```

core/compilePatterns.ml
```diff
@@ -668,7 +676,13 @@ and match_variant
             let default_type =
               StringSet.fold
                 (fun name t ->
-                   let _, t = TypeUtils.split_variant_type name t in t) cs t in
+                   if Settings.get Basicsettings.Sessions.exceptions_enabled &&
+                      not (Settings.get Basicsettings.Sessions.expose_session_fail) &&
+                      String.compare name Value.session_exception_operation = 0
```
dhil, line 681:
```suggestion
                      String.equal name Value.session_exception_operation
```

core/typeSugar.ml
```diff
@@ -4053,7 +4055,14 @@ let rec type_check : context -> phrase -> phrase * Types.datatype * Usage.t =
                      | Pattern.Operation (name, _, k) -> name, k
                      | _ -> assert false
                    in
-                   let effrow = Types.Effect (Types.make_singleton_open_row (effname, Types.Present efftyp) (lin_any, res_any)) in
+                   let effrow =
+                     if Settings.get Basicsettings.Sessions.exceptions_enabled &&
+                        not (Settings.get Basicsettings.Sessions.expose_session_fail) &&
+                        String.compare effname Value.session_exception_operation = 0
```
dhil, line 4061:
```suggestion
                        String.equal effname Value.session_exception_operation
```

core/typeSugar.ml
```diff
@@ -4287,7 +4296,14 @@ let rec type_check : context -> phrase -> phrase * Types.datatype * Usage.t =
           let opname = find_opname (erase op) in
           let infer_opt = no_pos (Types.make_operation_type (List.map typ ps) rettyp) in
           let term = (exp_pos op, opt) in
-          let row = Types.make_singleton_open_row (opname, T.Present (typ op)) (lin_unl, res_effect) in
+          let row =
+            if Settings.get Basicsettings.Sessions.exceptions_enabled &&
+               not (Settings.get Basicsettings.Sessions.expose_session_fail) &&
+               String.compare opname Value.session_exception_operation = 0
```
dhil, line 4302:
```suggestion
               String.equal opname Value.session_exception_operation
```

core/typeSugar.ml
```diff
@@ -4297,10 +4313,14 @@ let rec type_check : context -> phrase -> phrase * Types.datatype * Usage.t =
           in
             doop, rettyp, usage
         | Operation name ->
-           if String.compare name Value.session_exception_operation = 0 && not context.desugared then
-             Gripers.die pos "The session failure effect SessionFail is not directly invocable (use `raise` instead)"
+           if String.compare name Value.session_exception_operation = 0 then
```
dhil, line 4316:
```suggestion
           if String.equal name Value.session_exception_operation then
```

dhil (approved):
LGTM

core/basicsettings.ml
```diff
@@ -41,6 +41,16 @@ module Sessions = struct
               |> convert parse_bool
               |> CLI.(add (long "session-exceptions"))
               |> sync)
+
+  let expose_session_fail =
+    Settings.(flag "expose_session_fail" ~default:true
```
SimonJF, line 46: I think this should be an option rather than a flag?

SimonJF (commented):
Superficially this looks fine to me. I'm going to try and port over my linear MVU stuff and I'll let you know what I find.

core/basicsettings.ml
dhil: A flag is basically a convenience for an option over values of type bool. So this instance looks right to me.

SimonJF: Oh, I see. I think my hesitation was that it ended up exposed at the command line (i.e., we have the option for `--expose_session_fail`) rather than being config only.

PR #1170, opened 2023-01-31
Refactoring of lens tests and benchmarking

rudihorn:
These are just various changes I made during my PhD write-up which clean up the lens tests.


frank-emrich (approved):
LGTM

Issue #1103, opened 2022-02-10
Scale back distributed delegation?


SimonJF:
Distributed delegation (sending channel endpoints over channel endpoints in a distributed setting) is currently implemented using a distributed algorithm loosely based on https://www.doc.ic.ac.uk/~rhu/sessionj/hyh08session-based.pdf. (The issue is that sending a receive endpoint requires care to ensure that messages are relocated, and that messages aren't lost during migration.)
This is only a problem when migrating a client endpoint to the server, and a client endpoint to another client.

The code for this is a mess and massively complicates `proc.ml` in particular. To be quite honest, I did some limited testing during my PhD but the testing was nowhere near exhaustive. In practice, the only time I've ever used distributed delegation is sending an endpoint from the server to a client, which does not require any special treatment.

I don't think the currently-unused expressive power is worth the complexity in `proc.ml` and `jslib.js`. I propose the following:

  * Default behaviour: remove the client-to-client and client-to-server delegation code, and have these cause a runtime error
 
  * To support the rare case that such functionality is needed, add a setting which keeps all buffers associated with the server; this will be less performant since each client buffer read will require a call to the server, but then the problem is eliminated since we only need to think about name passing

Feedback appreciated; thanks!

dhil:
@SimonJF has this issue been fully or partially resolved by #1172?

SimonJF:
Thanks @dhil , completely addressed

PR #1172, opened 2023-03-30
Remove distributed session delegation in order to clean up proc.ml

SimonJF:
Channels are not innately suited to distributed computation because of the 'distributed delegation' problem which arises when sending channels over channels: if we send endpoint A over endpoint B, then we don't know where to route any messages to endpoint A.

Before, there was a ton of complex code which may or may not have worked, which attempted to implement a version of the distributed delegation algorithm described in the original Session Java paper. This significantly complicated the distributed sessions runtime in proc.ml and jslib.js.

In practice, we only really delegate from server to client, which does not require any special treatment. None of our present code sends session endpoints from client to other client, or client to server. Therefore, all of this spaghetti code is actually redundant. This patch removes all the distributed delegation code and therefore vastly streamlines the distributed session runtime. Attempting to do distributed delegation from client to server, or client to other client, will now result in a runtime error.

In future we may want to re-enable a form of this. I think the easiest way of doing this would be to explicitly allow the creation of "remote channels" whose buffers reside on the server with sends / receives requiring a remote call -- we would then only need to do name passing.


SimonJF:
This should address #1103 (got round to it eventually!)

SimonJF:
@slindley Unless you have any objections, can this go in? I don't think it's too controversial.

PR #1175, opened 2023-04-28
[WIP] Update test using mysql 5.7 driver to use ubuntu 20.04

frank-emrich:
See my first comment below.

### DIFF
--- a/.github/workflows/default.yml
+++ b/.github/workflows/default.yml
@@ -308,7 +308,7 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Install system dependencies
-        run: sudo apt-get install -y libev-dev
+        run: sudo apt-get install -y libev-dev libmariadb-dev-compat
         shell: bash
 
       - name: Use OCaml ${{ matrix.ocaml-compiler }}
--- a/.github/workflows/default.yml
+++ b/.github/workflows/default.yml
@@ -279,7 +279,7 @@ jobs:
       fail-fast: false
       matrix:
         os:
-          - ubuntu-20.04
+          - ubuntu-22.04
         ocaml-compiler:
           - 4.14.1
           - 5.0.0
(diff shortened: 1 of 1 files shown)


jamescheney:
So is OCaml 5.7 not supported as of Ubuntu >=20?

Using mariadb should be fine, in fact @SimonJF had had in mind to try to use it for comparing temporal features, but I made an attempt or two to try it using hte MySQL driver and didn't get anywhere (possibly because I didn't know what I was doing).

frank-emrich:
Sorry, I created this as a playground to see how the CI would react to certain changes, but hadn't added a description to the PR yet.

The problem isn't using MySQL 5.7 per se, but only the client packages used on the OCaml side.

In Links, we currently have two "drivers" for talking to mysql servers. These are basically thin wrappers around client libraries for talking to mysql servers.

1. The first one is provided by our own `links-mysql` package and in turn relies on the [mysql](https://opam.ocaml.org/packages/mysql/)  opam package.
2. The second one is provided by our `links-mysql8` package and in turn relies on the [mysql8](https://opam.ocaml.org/packages/mysql8/)  opam package.

Here, the `mysql8` opam package is a fork of the `mysql` opam package to support the mysql *client* version 8.0. Note that the mysql 8 clients can still talk to *servers* using older versions, in particular 5.7. Our test suite currently covers this in the tests called "Links main workflow / links-mysql8_mysql57". These use `links-mysql8`, which in turn uses the `mysql8` opam package to talk to a mysql 5.7 server.

What broke is the `links-mysql`/`mysql` package (without 8 in the name): It fails to build against client version 8, which is the one provided on newer Ubuntu versions. We kept this driver around in addition to `links-mysql8`/`mysql8` hoping that the original `mysql` package would eventually gain support for version 8 so that there would only be a single, canonical `mysql` package again in the Ocaml ecosystem. However it seems like this hasn't happened so far.

What I've tried in this PR are some workarounds to make the old `mysql` opam package compile against and use newer client versions, but this seems to have failed.

We should probably abandon our own `links-mysql` package for the time being (until it may get patched some day) and only support`links-mysql8` going forward. Given that this client can still talk to 5.7 servers, we do not really lose any functionality this way.

frank-emrich:
Subsumed by #1178

PR #1178, opened 2023-06-20
Remove links-mysql database driver

frank-emrich:
Links currently has two database drivers for MySQL databases, provided by the packages `links-mysql` and `links-mysql8`. The former uses the opam package `mysql`, while the latter uses `mysql8`.

Unfortunately, the `mysql` opam package cannot be built on recent versions of Ubuntu anymore and has not been updated in a while. `mysql8` is a fork of `mysql` that does work on more recent versions of Ubuntu, and supports MySQL version 8 (in addition to earlier versions).

Since `mysql8` is backwards compatible with earlier versions of MySQL, its functionality subsumes that of `mysql`. However, since it's a fork of `mysql` we kept the `links-mysql` driver around in case the original `mysql` package gets updated eventually. 

This PR is the result of giving up on this hope for the time being: It removes the `links-mysql` package and associated code within Links.

Since the CI is currently blocked by this, this PR also applies a temporary fix for #1179, by imposing a bound on the version of `ppx_sexp_conv`


dhil (commented):
Excellent. Following this patch we should do a minor release to bump the version of all links packages except for `links-mysql`.

frank-emrich:
Yes, but let's wait with the release until #1179 is fixed. This PR hotfixes it, which involves a version constraint in `links.opam`, which I'd like to avoid in the released package.

PR #1183, opened 2023-07-10
Simple implementation of SPARQL queries

jamescheney:
This is for discussion/testing fo rthe moment.  Some problems to potentially resolve before merging:

- [ ] `rdf` and some other libraries are added as hard Links dependencies.  It would be good to have a separate "driver" for the SPARQL querying interface, like for the databases.
- [ ] Currently a single function is exposed which takes three arguments: a "base IRI" (needed only if the result is a graph, AFAICT), the URI of the SPARQL endpoint, and the SPARQL query to actually run.  More functions for example to cater to the common case where the base isn't needed, and to allow for different flavors of SPARQL queries, such as ASK (returns boolean) or CONSTRUCT (returns graph), might be nice.  Currently only SELECT (return list of variable-value bindings) is supported.
- [ ] The code currently throws away type/structure information about the query results and just converts everything to strings.  WE could retain the structure of the returned RDF values/terms in Links instead.


.github/workflows/default.yml
```diff
@@ -17,7 +17,7 @@ jobs:
         os:
           - ubuntu-20.04
         ocaml-compiler:
-          - 4.08.1
+          - 4.14.1
```
jamescheney, line 20:
```suggestion
          - 4.13.1
```

jamescheney:
In today's meeting I think we agreed that this can be merged, with the todos above split into a separate issue along with a reminder to avoid/revisit the dependency on LWT when LWT goes away/is subsumed by other capabilities of OCaml 5.

jamescheney:
I think @vcgalpin agreed to try this branch out with some larger Links programs and @wricciot agreed to look over the code.

core/lib.ml
```diff
@@ -360,26 +362,26 @@ let env : (string * (located_primitive * Types.datatype * pure)) list = [
...
     let client_id = RequestData.get_client_id req_data in
-    `SpawnLocation (`ClientSpawnLoc client_id)),
+    Lwt.return (`SpawnLocation (`ClientSpawnLoc client_id))),
     datatype "() ~> Location",
     IMPURE
   );
 
   "haveMail",
-  (`PFun(fun _ ->
+  (`PFun(fun _ _ ->
```
wricciot, line 384: Is this extra argument to the function needed?

core/lib.ml
```diff
@@ -1693,10 +1695,25 @@ let env : (string * (located_primitive * Types.datatype * pure)) list = [
...
-       (`PFun (fun _ _ ->
+       (pure_pfun (fun _ ->
             Value.(box_list (List.map box_string (Settings.get_rest_arguments ()))))),
      datatype "() ~> [String]",
+     IMPURE);
+
+    (* SPARQL *)
+    "sparql",
+    (`Server
+       (p3D (fun base uri query _ ->
```
wricciot, line 1706: I think you can use p3 instead of p3D here.

core/query/sparql.ml
```diff
@@ -0,0 +1,19 @@
...
+
+let select ~base uri query =
+  get ~base ~accept:"application/json" uri {in_query = query; in_dataset = empty_dataset} >>= fun result ->
+  match result with
+  | Result(Rdf.Sparql.Solutions ss) ->
+      Lwt.return (List.map (fun s -> Rdf.Sparql.solution_fold (fun x t l -> (x, Rdf.Term.string_of_term t)::l) s []) ss)
+  | Result(Rdf.Sparql.Bool _) -> raise (internal_error("expected SELECT query, but result is a boolean"))
+  | Result(Rdf.Sparql.Graph _) -> raise (internal_error("expected SELECT query, but result is a graph"))
+  | Ok -> raise (internal_error("expected SELECT query, but no result returned"))
+  | Error(msg) -> raise (internal_error(Rdf.Sparql_protocol.string_of_error msg))
```
wricciot, line 19: Maybe the error message would be clear if it specified that it was returned by SPARQL and not by Links.

wricciot (approved):
Hi,
The code is straightforward and ready to be merged. I left some minor comments for places that could use some clarification.

core/query/sparql.ml
jamescheney:
```suggestion
  | Error(msg) -> raise (internal_error("SPARQL error: " ^ (Rdf.Sparql_protocol.string_of_error msg)))
```

core/lib.ml
```diff
@@ -63,7 +66,7 @@ let conversion_op' ~unbox ~conv ~(box :'a->Value.t): Value.t list -> Value.t = f
 
 let conversion_op ~from ~unbox ~conv ~(box :'a->Value.t) ~into pure : located_primitive * Types.datatype * pure =
   let open Types in
-  ((`PFun (fun _ x -> conversion_op' ~unbox:unbox ~conv:conv ~box:box x) : located_primitive),
+  ((pure_pfun (fun x -> conversion_op' ~unbox:unbox ~conv:conv ~box:box x) : located_primitive),
```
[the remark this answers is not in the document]
jamescheney:
```suggestion
  ((pure_pfun (fun x -> conversion_op' ~unbox ~conv:conv ~box:box x) : located_primitive),
```

core/lib.ml
jamescheney:
```suggestion
       (p3 (fun base uri query ->
```

jamescheney: Nope, because p3 does not do Lwt.return.

core/lib.ml
```diff
@@ -1719,10 +1721,25 @@ let env : (string * (located_primitive * Types.datatype * pure)) list = [
...
-       (`PFun (fun _ _ ->
+       (pure_pfun (fun _ ->
             Value.(box_list (List.map box_string (Settings.get_rest_arguments ()))))),
      datatype "() ~> [String]",
+     IMPURE);
+
+    (* SPARQL *)
+    "sparql",
+    (`Server
+       (p3 (fun base uri query ->
```
jamescheney, line 1732:
```suggestion
       (p3D (fun base uri query _ ->
```

jamescheney:
Light is green, trap is clean.

PR #1185, opened 2023-09-26
Robust build infrastructure

dhil:
This patch makes the build infrastructure for Links more robust. Previously, the build infrastructure would assume that it is running in an environment populated by OPAM. This assumption is too optimistic; for example a Docker container will give you a blank environment by default. To make it easier to install Links in such environments we now prefix dune build commands with `opam exec`. This command resolves the installation directory of dune and makes sure it is invoked with a suitable environment.

Furthermore, the `gen_linkspath.exe` install helper would only test whether Links is being built by OPAM, and if that was not the case then it would assume Links was being built from sources in a local checkout of the git repository. This assumption is a tad optimistic too. So now, if we cannot detect either an OPAM installation or the git repository, then we fall back to a best effort guess based off the current working directory.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,7 +43,7 @@ ROOT:=$(shell dirname $(firstword $(MAKEFILE_LIST)))
 BUILD_DIR:=$(ROOT)/_build
 
 # The build command and some standard build system flags
-BUILD=dune build
+BUILD=opam exec dune -- build
 MAIN=links
 SOURCES=$(MAIN)
 DB_STABLE=links-postgresql,links-sqlite3,links-mysql8
@@ -148,7 +148,7 @@ database-tests: links
 # Cleans the project directory.
 .PHONY: clean
 clean:
-	dune clean
+	opam exec dune -- clean
 	rm -rf *.install
 	rm -rf links linx
 	rm -rf doc/_build
--- a/Makefile
+++ b/Makefile
@@ -125,7 +125,7 @@ include-db-sources:
 # Auxiliary rule used by `opam install` to build Links
 .PHONY: opam-build-links.opam
 opam-build-links.opam: links.opam
-	$(shell LINKS_BUILT_BY_OPAM=1 dune build -p links)
+	$(shell LINKS_BUILT_BY_OPAM=1 opam exec dune -- build -p links)
 
 # Runs the test suite. We reset the variables CAMLRUNPARAM and
 # OCAMLRUNPARAM, because some of the tests are sensitive to the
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -175,8 +175,8 @@ open-doc: doc/_build/html
 
 # Release Links
 opam-release:
-	dune-release tag
-	dune-release distrib -p $(RELEASE_PKGS)
-	dune-release publish -p $(RELEASE_PKGS) distrib
-	dune-release opam -p $(RELEASE_PKGS) pkg
-	dune-release opam -p $(RELEASE_PKGS) submit
+	opam exec dune-release -- tag
+	opam exec dune-release -- distrib -p $(RELEASE_PKGS)
```
thwfhk, line 179: I got the following error message:
```
Makefile:179: *** missing separator.  Stop.

```

thwfhk: It works now. Seems like some copy-paste problem. Never mind.

thwfhk (approved):
LGTM. Tested successfully in a docker container.

frank-emrich:
I don't quite understand the need for `opam exec`. To me, it seems like the only situation where `dune <foo>` fails but `opam exec dune -- <foo>` succeeds is when the user didn't run `eval $(opam env)`. Is that correct? I'm not opposed to using `opam exec`, but I'd like to understand the situation better.

dhil:
> I don't quite understand the need for `opam exec`. To me, it seems like the only situation where `dune <foo>` fails but `opam exec dune -- <foo>` succeeds is when the user didn't run `eval $(opam env)`. Is that correct? I'm not opposed to using `opam exec`, but I'd like to understand the situation better.

That's correct. It is quite annoying having to do this all the time in, say, a Docker container.

frank-emrich (approved):

dhil:
@thwfhk you can merge this PR when you have a minute (remember to choose the strategy "squash and merge").

dhil:
> > I don't quite understand the need for `opam exec`. To me, it seems like the only situation where `dune <foo>` fails but `opam exec dune -- <foo>` succeeds is when the user didn't run `eval $(opam env)`. Is that correct? I'm not opposed to using `opam exec`, but I'd like to understand the situation better.
> 
> That's correct. It is quite annoying having to do this all the time in, say, a Docker container.

I can expand a little upon what I said. This PR ought to remove some of the friction that you would otherwise experience when wanting to package Links as an artifact for a artifact evaluation submission. The build infrastructure is more portable as a result of this PR.

dhil:
I think @thwfhk is correct. This PR solved the issue with Docker containers, I had simply forgotten about it!

Issue #1182, opened 2023-07-03
Rename links-mysql8 driver to links-mysql?


jamescheney:
I thought we decided at the meeting last week to, in effect, discard the current `links-mysql` driver and rename `links-mysql8` to `links-mysql` (while still using `ocaml-mysql8` under the hood), but just-merged #1178 just removes `links-mysql` without renaming the `links-mysql8` driver.

I think we should do this rename, thus ensuring that the `links-mysql` package continues to be the right thing to use to connect to MySQL independent of what links version one is using.

jamescheney:
@frank-emrich ahs there been any further discussion of this?  I think if we do this it should be done before the next release so that we don't wind up with the extra OPAM package.

frank-emrich:
Yes, I'm planning to implement what we discussed regarding this for the upcoming release: Making the `links-mysql` package rely on the `mysql8` package instead of `mysql`. This way, running opam upgrade should seamlessly upgrade existing installations.



________________________________

PR #1189, opened 2023-10-17
Rename links-mysql8 driver to links-mysql

frank-emrich:
This PR renames  the `links-mysql8` driver to `links-mysql`. 

The idea is that for end users who upgrade from Links 0.9.7 to 0.9.8, the composition of #1178 (which removed the existing `links-mysql` driver) and this one has the effect of the `links-mysql` driver changing from using the `mysql8` package under the hood in the future, with no further changes from users required.


frank-emrich:
This resolves #1182

frank-emrich:
Note that parts of the CI now state "Waiting for status to be reported". That's because I've renamed some of the CI steps, all tests are actually passing as intended. Once this PR lands I will perform the same renaming on which CI steps Github expects to have passed.

dhil:
> Note that parts of the CI now state "Waiting for status to be reported". That's because I've renamed some of the CI steps, all tests are actually passing as intended. Once this PR lands I will perform the same renaming on which CI steps Github expects to have passed.

Just delete the stale steps in the settings.

jamescheney (changes requested):
This looks fine generally, I have a couple of  minor questions.

1.  Looking at tests/database, there are currently two (identical) files `factorials.mysql8` and `factorials.mysql`, containing MySQL-specific SQL code to initialize the database for the test `factorials.links`.  My guess is that after making this change, only the `factorials.mysql` file will ever be used by the test scripts, so the `factorials.mysql8` file can be deleted.

2.  Looking at the changes to tests/relational-lenses, I was initially surprised to see that `config.mysql8` is deleted and no `config.mysql` is created.  The reason seems to be that `config.mysql` exists and has the correct contents already, presumably because it was not deleted earlier.  So that's fine.  However, looking at the rest of this directory, it appears that there are again identical files `serial.mysql` and `serial.mysql8` containing MySQL-specific database setup code where the second one will be dead code after this cange and so can be deleted now.

I mark "request changes" now because these bits of cleanup should probably happen now.

frank-emrich:
Thanks, those file were indeed meant to be deleted, good catch!

jamescheney (approved):

jamescheney:
Sorry, I didn't realize this was still waiting on me, LGTM.

PR #1191, opened 2023-11-15
Use Dune to generate OPAM package definitions

dhil:
This patch moves the definitions of our OPAM packages into the file `dune-project`. The OPAM package definitions are automatically generated from this file. This approach offers some advantages:

* Less redundancy: package descriptions are now shared.
* Dune integration: OPAM package definitions are kept in sync with dune.
* Versioning: `dune-release` automatically versions the OPAM package definitions.

Our build rules for the `links` package are a little more complicated than `dune` supports out-of-the-box. The steps required to build this package are defined in the `links.opam.template` file, which `dune` automatically pastes into the generated `links.opam` file.


dhil:
I fixed the minimum required version of OCaml to `4.14.0`, as that seems to be the minimum version with which Links compiles at the moment.

PR #1192, opened 2023-11-27
Depend on OCaml 5.1.1+

dhil:
This patch raises the lower bound of the dependency `ocaml` to 5.1 and
onwards.

We discussed this during the monthly Links Developer Meeting; and it
was agreed that we would raise the lower bound to OCaml 5 to encourage
the use of effect handlers to extend and simplify the code base.

### DIFF
--- a/.github/workflows/default.yml
+++ b/.github/workflows/default.yml
@@ -17,7 +17,7 @@ jobs:
         os:
           - ubuntu-20.04
         ocaml-compiler:
-          - 5.1.0
+          - 5.1.1
 
     runs-on: ${{ matrix.os }}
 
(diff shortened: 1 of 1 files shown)


jamescheney:
There are several places in documentation where we refer to OCaml 4.08 being the smallest usable version.  These should all be updated to 4.14.0 or (after this change) 5.1.0.

dhil:
I've updated the install instructions for OCaml 5.1.1. It seems to be the only place we mention OCaml 4.08 (as well as predecessors and successors) in the source tree. I guess other mentions may be on the wiki, which should be updated to OCaml 5.1.1 following this PR.

PR #1193, opened 2024-02-16
Redirect command

SimonJF:
It's often the case that we want to change page as a result of, say, clicking a button in an MVU application. Currently we just use `redirect` directly as Links is impure, but this isn't a particularly idiomatic way of doing things and has led to a weird race condition in a student project.

This patch introduces a `Redirect` command, which can be accessed through `MvuCommands.changePage(url)`. The redirect is then processed in the MVU event loop as would be expected.

### DIFF
--- a/examples/mvu/changePage.links
+++ b/examples/mvu/changePage.links
@@ -1,0 +1,46 @@
+import Mvu;
+import MvuHTML;
+import MvuAttrs;
+open import MvuCommands;
+typename Message = [| Navigate |];
+
+fun updt(msg, _) {
+  switch (msg) {
+    case Navigate -> ((), changePage("/2"))
+  }
+}
+
+fun view(_) {
+  open MvuHTML;
+  open MvuAttrs;
+  button(onClick(fun() { Navigate }), textNode("Navigate to second page"))
+}
+
+fun page1(_) {
+  Mvu.runCmd("placeholder", (),
+    view, updt, MvuCommands.empty);
+  page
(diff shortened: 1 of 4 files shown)


dhil (approved):
LGTM.

I looked up the `redirect` function. It occurs to me that it is quite possibly not doing what you really want, but it probably works because of how forgiving web browsers are. The implementation of `redirect`:
```ocaml
  (p1D (fun url req_data ->
         let url = Value.unbox_string url in
           (* This is all quite hackish, just testing an idea. --ez *)
           let resp_headers = RequestData.get_http_response_headers req_data in
           RequestData.set_http_response_headers req_data (("Location", url) :: resp_headers);
           RequestData.set_http_response_code req_data 302;
           Lwt.return (`Record [])
      )
```

It sets the header 302 which is supposed to be interpreted as the requested resource has been temporarily moved (c.f. https://www.rfc-editor.org/rfc/rfc9110#status.302).

I understand your example as navigation to the original placement of the resource, not some temporary placement of the resource. The appropriate status ought to be 200 in this case (c.f. https://www.rfc-editor.org/rfc/rfc9110#name-200-ok).

I suppose in the future we can consider adding HTTP-semantics compatible functions for requesting resources (maybe a student project?).

SimonJF:
Thanks @dhil. Agreed in general — although since this is in MVU it will run on the client and therefore use the jslib implementation of redirect (which I think uses document.url to perform the redirect).

PR #1198, opened 2024-07-28
User-definable abstract types

dhil:
This patch makes it possible to declare abstract types in Links, e.g.

```links
typename MyAbstractType;
```

The primary motivation for adding them to be able to give types to alien data objects.

The representation of user-definable abstract types is distinct from the built-in abstract types. In a future patch we ought to reconcile these two representations.

### DIFF
--- a/core/irCheck.ml
+++ b/core/irCheck.ml
@@ -344,6 +344,11 @@ let eq_types occurrence : type_eq_context -> (Types.datatype * Types.datatype) -
             else false
          | _ -> false
          end
+      | Abstract abs ->
+         begin match t2 with
+         | Abstract abs' -> abs == abs' (* pointer equality *)
+         | _ -> false
+         end
       (* Effect *)
       | Effect l ->
          begin match t2 with
--- a/core/types.ml
+++ b/core/types.ml
@@ -136,6 +136,8 @@ module RecIdMap = Map.Make(RecId)
 module type RECIDSET = Utility.Set with type elt = rec_id
 module RecIdSet : RECIDSET = Set.Make(RecId)
 
+let pp_exn fmt _ = Format.fprintf fmt "exn"
+
 type tygroup = {
   id: int;
   type_map: ((Quantifier.t list * typ) Utility.StringMap.t);
@@ -171,6 +173,7 @@ and typ =
   | Table of (Temporality.t * typ * typ * typ)
   | Lens of Lens.Type.t
   | ForAll of (Quantifier.t list * typ)
+  | Abstract of exn
   (* Effect *)
   | Effect of row
   | Operation of (typ * typ * DeclaredLinearity.t)
@@ -400,6 +403,7 @@ struct
          let (o, names') = o#list (fun o -> o#quantifier) names in
          let (o, body') = o#typ body in
          (o, ForAll (names', body'))
+      | Abstract abs -> (o, Abstract abs)
       (* Effect *)
       | Effect row ->
          let (o, row') = o#row row in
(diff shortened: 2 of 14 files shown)


jamescheney:
I've had a very quick look and am uncertain why the payload type of `Abstract` is `exn`.  It seems that the only way in which the type of `x` in `Abstract(x)` matters is that we use locally declared (generative) exceptions as a form of "gensym", and we use pointer equality to compare these in a couple of places.  But these values are never raised or caught as far as I can see.  

Is there a reason to make this specifically exceptions per se or would any other gensym-able abstract type also be OK?  If so perhaps we could introduce a tiny interface for such types (providing gensym and equality) and make this easy to change later if something else is preferred.

dhil:
You are right. I am piggybacking on the generativity of `exn` in OCaml to ensure that identities are globally unique. We can use any other generative type. The reason I didn't use a simple integer counter (as we do for other things) is that it doesn't scale beyond whole-program compilation; it also doesn't work very well if we at some point want to parallelise compilation. 

I agree with your point, that it would be good to have a structured facility in the code base (e.g. in utils) for generating guaranteed globally unique identities, e.g.

```ocaml
module Gensym: sig
  type t
  val gensym : unit -> t (* generates a unique identity *)
  val to_string : t -> string (* string representation; used in codegen *)
  val eq : t -> t -> bool (* decides whether two identities are the same *)
end = struct
  type t = exn
  let gensym () = 
    let exception E in E
  let to_string exn = Printf.sprintf "%d" (Hashtbl.hash exn)
  let instance i = i
  let eq exn exn' = exn == exn'
end
```

Of course it may be beneficial to include an integer count too for generation of human readable names (would be helpful in codegen and error messages, e.g. to distinguish two distinct entities named `Foo`), but this can also be viewed as something orthogonal to the generative identity, e.g. something that should be handled by the `Binder` or `Name` modules.

core/irCheck.ml
```diff
@@ -344,6 +344,11 @@ let eq_types occurrence : type_eq_context -> (Types.datatype * Types.datatype) -
             else false
          | _ -> false
          end
+      | Abstract abs ->
+         begin match t2 with
+         | Abstract abs' -> abs == abs' (* pointer equality *)
```
jamescheney, line 349: Should this be Gensym.equal?

core/types.mli
```diff
@@ -453,6 +454,7 @@ val pp_row' : Format.formatter -> row' -> unit
 val pp_type_arg : Format.formatter -> type_arg -> unit
 val pp_tycon_spec: Format.formatter -> tycon_spec -> unit
 val pp_field_spec: Format.formatter -> field_spec -> unit
+val pp_exn : Format.formatter -> exn -> unit
```
jamescheney, line 457: Is this still needed?

jamescheney (changes requested):
Looks OK generally, a couple of minor things and one more substantial one:

* consistent use of Gensym.equal
* get rid of now unnecessary pp_exn
* I think piggybacking on alias to preserve type arguments to parameterized abstract types is going to be too fragile, and suggest basically making Abstract take Gensym.t plus a list of type parameter arguments.

core/irCheck.ml
dhil: Yes!

core/types.ml
```diff
@@ -136,6 +136,8 @@ module RecIdMap = Map.Make(RecId)
 module type RECIDSET = Utility.Set with type elt = rec_id
 module RecIdSet : RECIDSET = Set.Make(RecId)
 
+let pp_exn fmt _ = Format.fprintf fmt "exn"
```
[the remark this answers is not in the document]
dhil: No, thanks!

core/unify.ml
```diff
@@ -83,8 +83,18 @@ let rec eq_types : (datatype * datatype) -> bool =
...
       | Alias (_, _, x) -> unalias x
-      | x            -> x in
+      | x            -> x
+    in
     match unalias t1 with
+      | Abstract abs ->
+         begin match unalias t2 with
+         | Abstract abs' ->
+            Gensym.equal abs abs'
+            && (let [@ocaml.warning "-8"]Alias (_, (_, _, tyargs, _), _) = t1 in
```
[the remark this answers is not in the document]
dhil: OK, I will have a look at this.

PR #1211, opened 2025-03-24
Sorting record members first on the length of the key

### LINKED ISSUE #1210: Incorrect arguments printing in `Roundtrip` printer with at least 10 arguments
The `RoundtripPrinter` (which is the default printer) displays the list of arguments of a function type in the wrong order: in the following code,
```links
sig f : ((), (), (), (), (), (), (), (), (), Int) ~> Int
fun f(a, b, c, d, e, f, g, h, i, j) { j }
f((), (), (), (), (), (), (), (), (), 1)
```
`f` has the displayed type `((), Int, (), (), (), (), (), (), (), ()) ~> Int`. This is because the string `10` is sorted before the string `2`, even though the integer 2 is less than the integer 10.

Using the old `Printer` (`--set types_pretty_printer_engine=old`), `f` has the (correct) displayed type `((), (), (), (), (), (), (), (), (), Int) ~> Int` thanks to [the conversion to an IntMap](https://github.com/links-lang/links/blob/c7a899ca7d267f276a2153778dc970e27d03fa92/core/types.ml#L2461). The `DerivedPrinter` does not have this issue as it displays the keys as well.

This issue is due to the incorrect ordering in [the `printers` variable of the `RoundtripPrinter.row_parts` function](https://github.com/links-lang/links/blob/c7a899ca7d267f276a2153778dc970e27d03fa92/core/types.ml#L3709).

rajdakin:
This PR fixes #1210 and the order of the `fn_params` member of the function descriptor.

The implementation of this fix adds a string map which has a different `compare` function: it first sorts keys on length, then on the content. This way, the string `10` is sorted after the string `2`. This map is then used in `Row` types.

### DIFF
--- a/core/buildTables.ml
+++ b/core/buildTables.ml
@@ -255,7 +255,7 @@ struct
 
   let bindings tyenv bound_vars cont_vars bs =
     let o = new visitor tyenv bound_vars cont_vars in
-    let _ = o#computation (bs, Return (Extend (StringMap.empty, None))) in ()
+    let _ = o#computation (bs, Return (Extend (Types.FieldEnv.empty, None))) in ()
 
   let program tyenv bound_vars cont_vars e =
     let _ = (new visitor tyenv bound_vars cont_vars)#computation e in ()
--- a/core/channelVarUtils.ml
+++ b/core/channelVarUtils.ml
@@ -14,6 +14,9 @@ let variables_in_computation comp =
   let rec traverse_stringmap : 'a . ('a -> unit) -> 'a stringmap -> unit =
     fun proj_fn smap -> (* (proj_fn: 'a . 'a -> 'b) (smap: 'a stringmap) : unit = *)
       StringMap.fold (fun _ v _ -> proj_fn v) smap ()
+  and traverse_ststringmap : 'a . ('a -> unit) -> 'a st_name_map -> unit =
+    fun proj_fn smap -> (* (proj_fn: 'a . 'a -> 'b) (smap: 'a st_name_map) : unit = *)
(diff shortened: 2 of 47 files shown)


dhil (commented):
> The implementation of this fix adds a string map which has a different compare function: it first sorts keys on length, then on the content.

I am not sure we want this. I'd think it should be sufficient to localise this change to type printing only.

rajdakin:
The root issue is that `StringMap.fold` iterates the map in the wrong order (`"10"` before `"2"`). I see three ways of fixing this issue, one of which is the current one:
- Change the `StringMap` so that comparison yields the correct ordering -- this is the one implemented in this PR;
- Change every `StringMap.fold` usage which gives a list (there are *at least* two such places, and every subsequent edits *must* keep this caveat in mind, which I don't believe is the correct solution) -- this is the solution you are proposing;
- Change the type of functions to have a list of `typ` instead of a `typ` usually containing a `Record` containing a `Row` as the argument.

I have also tried to look into implementing the third solution, but I believe there exists/existed some use case for types other than `Record`+`Row` as argument types, so I quickly abandoned the idea. I can still try to make it work if you believe this would be a better solution.

Note that this PR not only fixes #1210, but also the issue that the `fn_params` member list is not ordered properly in the backend IR (this is not an issue for now as only the types in that list are incorrectly ordered, an information currently unused in all backends, but I will need this fixed for the WasmFX backend).

dhil:
Thanks for the clarification. I see. I think we need to discuss which kind of design to adopt here. Because the change is pervasive.

Issue #57, opened 2016-08-24
Unified pattern matching representation


dhil:
Links has three different pattern matching constructs (four if you include `receive` which is desugared into `switch`). These are

* `switch` which destructs value patterns
* `choice` which destructs choice patterns
* `handle` which destructs effect patterns

Each pattern-matching construct has its own slightly distinct internal representation. Moreover, each construct has its own entry point to the pattern-matching compiler (`compilePatterns.ml`). The `handle` and `choice` constructs employ various hacks to make use of pattern matching compilation infrastructure for `switch`. This infrastructure is rather rigid, making it difficult to reuse and adapt for new experiments. It would be beneficial to have a common representation of pattern-matching internally. It would require a rewrite of the pattern-matching compiler. In addition, it would reduce the amount of  boilerplate code and code duplication that we currently have in `sugarTraversals.ml` and `transformSugar.ml`. 

I would like a design which partitions clauses into value, effect, and choice clauses.

PR #1214, opened 2025-04-11
Moved dependency chasing to the loader and fixed some tests

rajdakin:
This PR:
- removes some `ignore` lines in the tests file for tests that are now passing;
- fixes some tests;
- moves the dependency chasing to the loader (as the comment suggests) so that we can search for files next to the current one (i.e. while compiling `example/foo.links`, the module `Bar` will be searched in `example/bar.links` instead of `bar.links`);
- adds the `case otherwise ->` construct as an alias to the `case _ ->` construct;
- adds a message to the testing output in case a test that was skipped is now passing.

### DIFF
--- a/bin/driver.ml
+++ b/bin/driver.ml
@@ -30,6 +30,7 @@ module Phases = struct
           Parse.program
           source_code
       in
+      let program = Chaser.add_dependencies "" program in
       let context' = Context.({ context with source_code = pos_context }) in
       Loader.({ program_ = program; context = context' })
 
@@ -38,6 +39,7 @@ module Phases = struct
       let program, pos_context =
         Parse.Readline.parse ps1
       in
+      let program = Chaser.add_dependencies_sentence program in
       let context' = Context.({ context with source_code = pos_context }) in
       Loader.({ program_ = program; context = context' })
   end
--- a/core/chaser.ml
+++ b/core/chaser.ml
@@ -85,7 +85,7 @@ let rec add_module_bindings deps dep_map =
           ~message:"Impossible pattern in add_module_bindings")
 
 
-let rec add_dependencies_inner module_name module_prog visited deps dep_map =
+let rec add_dependencies_inner root module_name module_prog visited deps dep_map =
   if StringSet.mem module_name visited then (visited, [], dep_map) else
   let visited1 = StringSet.add module_name visited in
   let dep_map1 = StringMap.add module_name module_prog dep_map in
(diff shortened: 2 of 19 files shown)


dhil:
I know the context of this patch. What's the motivation? It seems to bundle a bunch of unrelated things together. Specifically, what's the motivation for the `case` change?

> * adds the `case otherwise ->` construct as an alias to the `case _ ->` construct;

rajdakin:
There are three tests that use this construct (`examples/toggle.links`, `examples/handlers/toggle.links`, and `examples/handlers/toggle_web.links`). As a side note, `otherwise` is tokenized to the `OTHERWISE` token, which only appears once in the entire parser in a location unrelated to this, so the change is safe.

The motivation of the entire PR is to make sure the tests are up-to-date, i.e. that tests which should pass do actually pass, that tests that pass are not ignored, and to fix minor issues (the `case otherwise`).

jamescheney:
> The motivation of the entire PR is to make sure the tests are up-to-date, i.e. that tests which should pass do actually pass, that tests that pass are not ignored, and to fix minor issues (the case otherwise).

I think Daniel's point is that unlike fixing tests, this is a language change, and deserves to be pulled out and discussed/agreed sepatrately, as it does not appear to be an already recognized issue that no one had gotten around to yet.

I'm not sure what is wrong with `case _`, or what is less wrong with `case otherwise`.   Could you please factor this out and make a case for it?

rajdakin:
Got it, I have removed this change.
(Also, when the test was created in 1b72bf9d80fa57fb3712e6917c7f712d4734678d, the `OTHERWISE` token did not exist, but it did when the typechecking test script was added in 4d6687e076ab35cdae6d91dfb1a65e10076d041e.)

dhil:
> > The motivation of the entire PR is to make sure the tests are up-to-date, i.e. that tests which should pass do actually pass, that tests that pass are not ignored, and to fix minor issues (the case otherwise).
> 
> I think Daniel's point is that unlike fixing tests, this is a language change, and deserves to be pulled out and discussed/agreed sepatrately

Yes indeed. I think the problem is that `otherwise` became a keyword, when @SimonJF implemented the session exception handling syntax. We have discussed this "issue" previously. I believe the right solution is to unify our various handling-syntaxes into `switch (...) { ... }` (issue #57 discusses some thoughts on a better internal representation). We currently have special syntax for effect patterns. We'd need special syntax for catching session exceptions too.

jamescheney:
Ah - sorry @rajdakin, I misinterpreted your change about `case otherwise` as meaning that you had introduced this keyword and rather than that you were adjusting things to make ordinary `case _`  a desugared form of `case` using `otherwise`.  It does still sound like this is something that should be discussed before further changes, though.

The issue is that session exception syntax is of the form `try` exp `as` pattern `otherwise` exp - it wasn't ever being used as a synonym for `_` before.  So even if this is a non breaking change, it is a nontrivial one.  I feel it is conventional enough to write `case _` or similar in functional languages, rather than have a separate keyword for catch-all or default cases; but if we do want to have such a separate keyword, for me it would be more idiomatic to write `default =>` or (to avoid proliferation of keywords) `otherwise =>` rather than `case otherwise =>`,  but this does feel like bikeshedding in action, now.

rajdakin:
Is the current version good to go then? I have removed everything concerning `case otherwise`.

PR #1216, opened 2025-04-23
Effects typechecking fixes

rajdakin:
This PR fixes two issues:
- If one kind of unification issue related to effects occurs, the old `links` tries to generate an error message but the content of that error message generates an internal error;
- If linearity is not tracked, `lindo` should be parsed exactly the same as `do` and `case <Foo =@ k>` like `case <Foo => k>`.

MWE:
```
sig f : () {A:() =@ Bool|_}~> Bool
fun f() { lindo A() }
```
This MWE generates the error in the old `links` when it is supposed to be correct.
Note that removing the signature of `f` makes it work regardless of the fix, as the type will be automatically inferred with a linear effect before linearity is dropped.

```
sig f : (s) {A:(s) {}-> () |_}-> ()
fun f(st) {do A(st)}
```
This MWE generates an internal error in the old `links`. The proposed fix transforms this internal error into the expected normal error:
```
test.links:2: Type error: Invocation of the operation
    `A(st)'
requires an effect context
    `() => (a) {}-> ()'
but, the currently allowed effect is
    `(a) => (a) {}-> ()'
In expression: do A(st).
```

### DIFF
--- a/core/parser.mly
+++ b/core/parser.mly
@@ -1400,7 +1400,8 @@ resumable_operation_pattern:
 | operation_pattern FATRARROW pattern
     { with_pos $loc (Pattern.Operation (fst $1, snd $1, $3, DeclaredLinearity.Unl)) }
 | operation_pattern FATLOLLI pattern
-    { with_pos $loc (Pattern.Operation (fst $1, snd $1, $3, DeclaredLinearity.Lin)) }
+    { with_pos $loc (Pattern.Operation (fst $1, snd $1, $3,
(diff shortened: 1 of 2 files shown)


thwfhk:
This PR looks good to me. I'm not entirely sure about the second MVE. I think `(s) {}-> ()` is an old-fashioned way of writing effect signatures in Links, which used to be understood as an operation that takes an argument of type `s` and returns unit. @dhil is this feature completely discarded now?

thwfhk:
That makes sense to me. I agree with you that these are two orthogonal issues.

PR #1217, opened 2025-06-20
Fix higher-order CPS

### LINKED ISSUE #1097: CPS compilation is broken for inlined handled computations
The following program does not produce any output
```
fun run() {
  handle({
    println(intToString(do Foo(40,2)))
  }) {
    case <Foo(x,y) => resume> -> resume(x+y)
  }
}

fun mainPage(_) {
  page
   <#><button l:onclick="{run()}">Click me!</button></#>
}

fun main() {
  addRoute("/", mainPage);
  servePages()
}

main()
```
after hoisting the inlined computation to a thunk, the program produces the expected output
```
fun f() {
  println(intToString(do Foo(40,2)))
}

fun run() {
  handle(f()) {
    case <Foo(x,y) => resume> -> resume(x+y)
  }
}

fun mainPage(_) {
  page
   <#><button l:onclick="{run()}">Click me!</button></#>
}

fun main() {
  addRoute("/", mainPage);
  servePages()
}

main()
```

The problem seems to be that the inlined computation is being compiled with the wrong continuation.

dhil:
This patch fixes a bug in the JavaScript backend wrt. the name generation logic in the higher-order continuation module. Previously, name binding the continuation could inadvertently drop a prefix of the static continuation.

I have manually tested the few web examples that we have, and verified that the test case in #1097 works as expected now.

Resolves #1097

### DIFF
--- a/core/irtojs.ml
+++ b/core/irtojs.ml
@@ -706,12 +706,12 @@ module Higher_Order_Continuation : CONTINUATION = struct
            let k = gensym ~prefix:"_kappa" () in
            (fun code -> bs (Bind (k, v, code))), ks, Var k
         | Cons ((Var _) as v, kappas) ->
-           bind bs (fun kappas -> Cons (v, kappas)) kappas
+           bind bs (fun kappas -> ks (Cons (v, kappas))) kappas
         | Cons (v, kappas) ->
            let k = gensym ~prefix:"_kappa" () in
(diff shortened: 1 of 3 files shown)


rajdakin (approved):
Looks good to me, though lines 715-717 could be indented once (right now they are aligned with the `let` below).

